PR #57, opened 2021-09-17
Database Migrations 

knolleary:
*I made a mess of the `migration` branch whilst trying to merge commits - I pulled when I should have pushed. Trying against with a clean branch*

This is a first pass of a database migration framework.

Sequelize provides a migration tool as part of its cli. However, from what I have seen, it is very opinionated as to how you structure your models. It assumes you've bootstrapped your project using their cli tool - unfortunately their getting started docs make no mention of this.

So instead, I've gone for something custom, albeit influenced by their tooling.

 - The models in `forge/db/models` define the latest database structure. They can create the database structure if it is missing, however if a model requires a database change, something is needed to modify the database structure. Sequelize has some support for adding columns automatically - but not changing types/constraints etc. We do not make use of this.
- `forge/db/migrations` contains a set of migration files, using the naming pattern `XXXXXX-name.js`. They export an `up` and `down` function that can be used to apply or reverse the migration. They also provide a set of signatures for how the database should look post-migration - see below.
- A new db table has been added `MetaVersion` that records what migrations have been applied to the database.
- When the app starts it compares the name of the most recently applied migration with the list of migration files. If there are unapplied migrations, the app will refuse to start.
- If it finds the MetaVersion table empty, it automatically applies all migrations as it is a clean start.
- A script is provided to apply any outstanding migrations - `forge/scripts/apply-pending-migrations.js`

### Migration signatures

*This part is experimental*.

A potential problem is a model is changed without a corresponding migration created. This is where the migration signatures come in.

For each table modified by the migration, it can include a sha256 hash of the resulting table schema (when created in an in-memory sqlite db).

The script `forge/scripts/verify-migration-signature.js` loads the Models into an in-memory sqlite database then generates the hash of each table schema. It then loads all migrations in order (without applying them) to get a list of the expected table hashes. If there are any mismatches, it warns.

For now, that can be a manual check, but it can become part of the CI build.

There's a bit of a chicken/egg issue of course - you need to create the model change, create the migration, then run the verify script to be told what hash it should now have.

### DIFF
--- a/forge/app.js
+++ b/forge/app.js
@@ -16,37 +16,39 @@ const containers = require('./containers');
   * @namespace forge
   */
 
-const server = fastify()
+(async function() {
 
-server.addHook('onError', async (request, reply, error) => {
-    // Useful for debugging when a route goes wrong
-    console.log(error.stack);
-})
+    const server = fastify()
 
-// Config : loads environment configuration
-server.register(config);
-// License
-server.register(license);
-// DB : the database connection/models/views/controllers
-server.register(db);
-
-process.env.PORT = process.env.PORT || 3000;
-process.env.BASE_URL = `http://localhost:${process.env.PORT}`;
-
-// Routes : the HTTP routes
-server.register(routes)
-// Containers:
-server.register(containers);
-
-// Wait until everything is loaded so PORT can be set via .env config
-server.ready().then(() => {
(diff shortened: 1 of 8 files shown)


knolleary:
Replaced by #263

PR #201, opened 2022-01-21
Fix install instructions

hardillb:
fix for https://github.com/flowforge/installer/issues/11 and https://github.com/flowforge/installer/issues/12


hardillb:
See comments on https://github.com/flowforge/installer/pull/13

hardillb:
Fixes #202

Issue #242, opened 2022-01-27
Log version number at startup


hardillb:
### Description

To make reporting issues easier we should log the FlowForge Version number at startup. Also make it visible in the UI.

PR #243, opened 2022-01-27
Check NodeJS version and log FF version

### LINKED ISSUE #241: Add explicit check for compatible Node version on startup
### Description

Whilst we document Node 16 as our required version, we don't check it when we start.... which means users could unintentionally run with an older version and hit unexpected issues.

We should have an explicit check on startup and refuse to start if it isn't 16+

### Epic/Story

_No response_

hardillb:
Fixes #241 and #242

### DIFF
--- a/forge/config/index.js
+++ b/forge/config/index.js
@@ -18,7 +18,7 @@ const fs = require("fs");
 const path = require("path");
 const fp = require("fastify-plugin");
 const YAML = require("yaml");
-const semver = require('semver')
+
 // const FastifySecrets = require('fastify-secrets-env')
 
 module.exports = fp(async function(app, _opts, next) {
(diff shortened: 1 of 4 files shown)


hardillb:
Should we hard exit or just log NodeJS version?

forge/config/index.js
```diff
@@ -35,6 +36,21 @@ module.exports = fp(async function(app, _opts, next) {
...
+        //npm start
+        app.log.info(`FlowForge v${process.env.npm_package_version}`)
+    } else {
+        //everything else
+        let {version} = require(path.join(module.parent.path, "..", "package.json"))
+        app.log.info(`FlowForge v${version}`)
+    }
+
+    if (!semver.satisfies(process.version,">=16.0.0")) {
+        app.log.warn(`FlowForge requires at least NodeJS v16, ${process.version} found`)
```
knolleary, line 49: I'd be tempted to move this logic right to start of the main app.js, before we try doing any setup - and console.error and exit if node isn't suitable.

hardillb:
The original task #241 says to exit so have added that.

forge/config/index.js
hardillb: OK, makes sense. I was looking for the first place I could get hold of a `app.log` object. We can still log it here so it's clean for a good version

PR #254, opened 2022-02-01
First pass at SES support

hardillb:
uses the following config in flowforge.yml

```yaml
email:
  ses:
    region: eu-west-1
```


knolleary (approved):

hardillb:
Tested locally, but AWS SES account stuck in "Sandbox" mode until I can convince them we are not going to spam the world.

Need to test in K8s on AWS to get make sure the credentials get passed, but I'm going to merge this as it works.

PR #265, opened 2022-02-07
Docker docs

hardillb:
Add docs on how to install and configure with Docker Compose

### DIFF
--- a/docs/install/docker/README.md
+++ b/docs/install/docker/README.md
@@ -24,7 +24,7 @@ The FlowForge Application will be hosted on `http://forge.example.com`
 
 ## Installing FlowForge
 
-### Building Contianers
+### Building Containers
 
 To build the 2 required containers simply run `./build-containers.sh`.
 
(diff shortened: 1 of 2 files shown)


docs/install/docker/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+To make this work you will need to configure a DNS server to map a wildcard domain entry to the IP address of the host running Docker. e.g `*.example.com`.
+
+The FlowForge Application will be hosted on `http://forge.example.com`
+
+**Note** When testing locally you can add entries for each project to your `/etc/hosts` file but you must use the external IP address of the host machine, not the loopback address (`127.0.0.1`).
+
+## Installing FlowForge
+
+### Building Contianers
```
knolleary, line 27:
```suggestion
### Building Containers
```

docs/install/docker/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+To build the 2 required containers simply run `./build-containers.sh`.
```
knolleary, line 29:
```suggestion
To build the 2 required containers run `./build-containers.sh`.
```

`simple/simply` are words to avoid (which I don't always succeed in doing myself) as it makes the read feel like it should be obvious when it may not be to them.

knolleary (commented):
Looks good - couple of nits to squash

PR #274, opened 2022-02-09
Add telemetry metrics component

knolleary:
This adds the `monitor` component that will be used to periodically gather stats about the platform and ping them back to us.

With this PR, the payload of the ping looks like this:

```
{
  instanceId: '5db51f99-c6fb-4340-9c19-78adce58cc1b',
  os: { type: 'Darwin', release: '20.5.0', arch: 'x64' },
  env: { nodejs: 'v16.13.1', flowforge: '0.1.1' },
  platform: {
    counts: { users: 0, teams: 0, projects: 0 },
    config: { driver: 'stub' }
  }
}
```

There is a framework in place for adding additional metrics - something that has to be done in coordination with the flowforge/usage-ping-collector repo to ensure the values are captured in the database.

With all that said, this PR doesn't do any pings. We need to figure out our strategy for when the pings should run.

They are not intended to be event driven - they are a point in time snapshot of the current state of the system that will get reported back periodically. The question is what that period is. As we expect it to be a long running process, it would be reasonable to have a long interval of once a week.

My starting proposal is:

1. Ping back within the first few minutes of startup.
2. Ping once a week at a randomly chosen time between 00:00 and 01:00 UTC each Monday

(using a random time in that interval to avoid the thundering herd problem of our collector being hammered as all instances report at the same time)

A couple more open questions:

1. Should we automatically disable the ping if running using `npm run serve` - ie developer mode. That could generate a lot of noise as devs are starting/stopping the platform repeatedly
2. Should we have a different granularity of ping back?
3. Are there any other metrics we should be capturing for MVP? I'm aware the current counts are a point-in-time snapshot. and won't necessarily give the whole picture. Would it be useful to `platform.count.projectsCreated` and `platform.count.projectsDeleted` - so we can see churn over time? (The answer is of course yes, but that will require a means to store counts of those events as they happen so they can be reported back when the ping runs).

### DIFF
--- a/forge/monitor/metrics/004-platform.js
+++ b/forge/monitor/metrics/004-platform.js
@@ -2,6 +2,7 @@ module.exports = async (app) => {
     return {
         'platform.counts.users': await app.db.models.User.count(),
         'platform.counts.teams': await app.db.models.Team.count(),
-        'platform.counts.projects': await app.db.models.Project.count()
+        'platform.counts.projects': await app.db.models.Project.count(),
+        'platform.config.driver': app.config.driver.type
     }
 }


knolleary:
Have added `platform.config.driver` to the metric list to capture which driver type is being used.

PR #284, opened 2022-02-11
fix slug issue

### LINKED ISSUE #280: Changing Team Slug doesn't update /account/teams
### Current Behavior

If I change the slug for my team in `/team/[teamname]/settings/general` 

The link for the team on `/account/teams` does not reflect the new slug it still has the original default slug for when the team was created. 
So clicking it returns a 404

### Expected Behavior

The link in the account page should have the correct slug/url

### Steps To Reproduce

Create a team with a name eg `Test Team`
See that the slug is set to `test-team`
Change the slug to `testteam` 
Goto `/account/teams` and click on the team name.
 

### Environment

-FlowForge version: 0.1.1
-Node.js version: 16.x
-npm version: 8.x
-Platform/OS: Alpine/Docker
-Browser: Chrome

sammachin:
this should fix #280

### DIFF
--- a/frontend/src/api/teams.js
+++ b/frontend/src/api/teams.js
@@ -1,5 +1,4 @@
 import client from './client'
-import slugify from '@/utils/slugify'
 import paginateUrl from '@/utils/paginateUrl'
 
 const getTeams = async (cursor, limit) => {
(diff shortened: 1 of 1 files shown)


knolleary (commented):
Need to remove the `import slugify` line to pass the eslint check

PR #304, opened 2022-02-15
Update telemetry option wording & add std forge-link class

### LINKED ISSUE #291: Update Telemetry wording in UI to say we now collect data
### Description

The current wording for the Telemetry option (in both Setup screens and Admin Settings) says:

> This release does not collect any information
> A future release will collect anonymous statistics about how FlowForge is used. This allows us to improve how it works and make a better platform.
> We will clearly communicate when this feature is implemented and exactly what information is being gathered.

That needs to be updated now that we are able to collect data.

### Epic/Story

#142

joepavitt:
Closes #291

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -20,6 +20,12 @@
         @apply py-6;
     }
 
+    .forge-link {
+        @apply underline;
+        @apply hover:cursor-pointer;
+        @apply hover:text-blue-600;
+    }
+
     .forge-button {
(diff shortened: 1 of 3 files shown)


knolleary:
There's a second place this option is exposed - as part of the general Admin Settings:

https://github.com/flowforge/flowforge/blob/main/frontend/src/pages/admin/Settings/General.vue#L36

knolleary (approved):

PR #308, opened 2022-02-16
Bump to v0.2.0

hardillb:


hardillb:
Fails to build because can't install dependencies as they are not currently published to npm

PR #311, opened 2022-02-16
Add host to config

### LINKED ISSUE #309: Add 'host' config option to set what address to listen on
### Description

We currently listen on `0.0.0.0` which means "all available interfaces" - with no option to configure that.

We should have `host` option alongside `port` in the config yaml that allows the user to change it (plus allow IPv6 support when that becomes popular in the far future....)

We should default to `127.0.01` as a more secure default.

### Epic/Story

_No response_

knolleary:
Fixes #309

### DIFF
--- a/docs/install/configuration.md
+++ b/docs/install/configuration.md
@@ -12,8 +12,9 @@ the options available in the configuration file.
 
 Option | Description
 -------|------------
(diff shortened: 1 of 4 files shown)


hardillb (approved):
Looks OK, 

But I need to PR the `flowforge.yml` for docker and k8s to default to `0.0.0.0` so the behaviour doesn't change

PR #326, opened 2022-02-18
Add initial Project Stacks

knolleary:
This PR adds the initial implementation of Stacks.

This covers:
 - the database model
 - the api endpoints for CRUD actions
 - UI updates to allow Admins to manage stacks
 - Create Project form has a drop-down select to pick the stack to use

A stack consists of:
 - `id`
 - `name` - must be unique and is what the end user sees when picking a stack
 - `active` (boolean) - allows a stack to be disabled (cannot be used to create new projects) without deleting
 - `properties` - an object of driver-defined key/value pairs

The `localfs` driver has been updated via https://github.com/flowforge/flowforge-driver-localfs/pull/30 to support stacks - mainly how it provides a list of the properties a stack should contain.

### DIFF
--- a/forge/db/migrations/_20220214-02-add-templates.js
+++ b/forge/db/migrations/_20220214-02-add-templates.js
@@ -11,7 +11,7 @@ module.exports = {
                 primaryKey: true,
                 autoIncrement: true
             },
-            name: { type: DataTypes.STRING },
+            name: { type: DataTypes.STRING, unique: true },
             settings: { type: DataTypes.TEXT },
             createdAt: { type: DataTypes.DATE },
             updatedAt: { type: DataTypes.DATE },
--- a/forge/db/models/ProjectTemplate.js
+++ b/forge/db/models/ProjectTemplate.js
@@ -1,0 +1,27 @@
+/**
+ * A Project Stack definition
+ * @namespace forge.db.models.ProjectTemplate
+ */
+const { DataTypes } = require('sequelize')
+
+module.exports = {
+    name: 'ProjectTemplate',
+    schema: {
+        name: { type: DataTypes.STRING, allowNull: false },
+        settings: {
+            type: DataTypes.TEXT,
+            set (value) {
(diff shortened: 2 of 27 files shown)


hardillb:
First quick read looks good. Will give it another look a little later and complete the revice

forge/db/migrations/_20220214-02-add-templates.js
```diff
@@ -0,0 +1,34 @@
+/**
```
knolleary, line 1: Note the file has been named with a `_` at the start - this means it is ignored and not applied - so the contents can be ignored. We'll review it when we do Project Template next week.

I could have deleted it from the branch entirely, but I didn't.

forge/db/models/ProjectTemplate.js
```diff
@@ -0,0 +1,27 @@
+/**
```
knolleary, line 1: Ignore this - it doesn't get applied in this PR

frontend/src/pages/project/create.vue
```diff
@@ -14,6 +14,8 @@
                     </template>
                 </FormRow>
 
+                <FormRow :options="stacks" v-model="input.stack" id="stack">Stack</FormRow>
+
```
hardillb, line 18: Not sure where to put this, but if you try to create a project with no stacks defined it just silently fails (400 error in the browser console)

forge/routes/api/project.js
```diff
@@ -84,45 +88,60 @@ module.exports = async function (app) {
...
+        //     projectToken: projectToken.token,
+        //     ...request.body.options,
+        //     ...authClient
+        // }
+
+        await team.addProject(project)
+        await project.setProjectStack(stack)
+        await project.reload({
+            include: [
+                { model: app.db.model.ProjectStacks }
```
hardillb, line 124:
```suggestion
                { model: app.db.models.ProjectStack }
```

Sorry, typo when I made the suggestion

PR #335, opened 2022-03-04
Do proper check for valid invite on user registration

### LINKED ISSUE #330: /account/register endpoint availble when self registration is disabled
### Current Behavior

The `/account/register` endpoint doesn't validate that the user signing up has been invited. 

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- FlowForge version: 0.2

knolleary:
Fixes #330

This is a stop-gap fix. We need to improve the user invite flow in general.


hardillb (approved):
I'm 99.9% sure I've worked out what this is doing and it looks right

PR #346, opened 2022-03-09
Frontend Billing

joepavitt:
Best to ensure that `backend-billing` is merged first in order to make the review/diff easier to follow. All of `backend-billing` (at the point of opening this PR) is merged into this branch.


knolleary:
An observation for future reference: wide-spread linting changes ought to be done as a separate piece of work to make reviewing changes easier. But I do appreciate we're in a slightly awkward transition at the moment and we have to rip off the plaster at some point.

hardillb (approved):

hardillb:
Shall we open a new issue to do a full Vue lint pass after 0.3.0 is cut on Thursday as all the open PRs should be closed by then

PR #350, opened 2022-03-10
Add Validation to project names

### LINKED ISSUE #331: Create list of banned project names
### Description

To prevent people creating projects with names like:

 - www
 - forge
 - support
 - help

### Epic/Story

_No response_

### LINKED ISSUE #345: Currently possible to create 2 projects with the same name
### Current Behavior

While looking into how to implement #331 I tested creating 2 projects with the same name (using the localfs driver). This succeeded without any problem

### Expected Behavior

If the name is already in use the second project should not be created

### Steps To Reproduce

- Click Create Project
- Change project name to `foo`
- Click Create Project
- Navigate to Team Overview
- Repeat first 3 steps

### Environment

- FlowForge version: 0.2.0
- Node.js version: 16.x
- npm version: 8.x
- Platform/OS: Linux
- Browser: Chrome

hardillb:
Since project names are used as host names they need to be validated and unique.

Fixes #345 
Fixes #331

Still need to work out the db migration script to not drop the table on SQLite and how to check for existing duplicate names (as will not cause a problem on localfs driver)


forge/db/models/Project.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     name: 'Project',
     schema: {
         id: { type: DataTypes.UUID, primaryKey: true, defaultValue: DataTypes.UUIDV4 },
-        name: { type: DataTypes.STRING, allowNull: false },
+        name: { type: DataTypes.STRING, allowNull: false, unique: true },
```
knolleary, line 13:
```suggestion
        name: { type: DataTypes.STRING, allowNull: false },
```

I think it will be cleaner to *not* add this constraint with this PR - given we're taking a code-first approach to policing it at the moment.

Once we have figured out the migration issue with sqlite, we can add it to the model and have the corresponding migration ready to go - rather than have some databases with it and some databases without it.

hardillb: I've removed it and the migration from the PR

frontend/src/pages/project/create.vue
```diff
@@ -129,7 +140,11 @@ export default {
                 this.$router.push({ name: 'Project', params: { id: result.id } })
             }).catch(err => {
                 console.log(err)
-            })
+                if (err.response.status === 409) {
+                    console.log(err.response)
```
knolleary, line 144:
```suggestion
```

PR #352, opened 2022-03-11
Introduce Project Templates

knolleary:
Initial Project Template implementation

### API Endpoints

Method | Endpoint||Permission|Description
-- | -- | -- | -- | --
GET | /api/v1/templates |   | admin | list of all templates
GET | /api/v1/templates/:id |   | member | get template details
POST | /api/v1/templates |   | template:create | create a template definition
DELETE | /api/v1/templates/:id |   | template:delete | delete a template definition
PUT | /api/v1/templates/:id |   | template:edit | update a template definition

_TIL: copy and paste from a google spreadsheet gives you markdown formatted tables for free_

### Database changes

 - Adds `ProjectTemplate` table
 - Adds associate to the `Project` table

### How it works...

A Template object contains two key properties:
 - `settings` - a JSON object of structured key/value pairs that define the runtime settings the template sets
 - `policy` - a JSON object with an identical structure to the `settings` field, but the values are `true/false` - and indicate whether the value is `locked (false)` or `editable (true)` - this determines whether a project is allowed to modify the corresponding value.

When editing a Project's settings, it loads the template to get the default values, and the policy to determine which fields should be editable. When saving the settings for a project, they are compiled back into a single `settings` object - but it only contains values that are different to the template. This blob is saved in the `ProjectSettings` table under the key name `settings`.

The `/projects/:id/settings` end point used by the launcher to get the runtime settings now includes the `settings` property - this is the settings of the template with any project-specific modifications applied.

A PR on the launcher (https://github.com/flowforge/flowforge-nr-launcher/pull/23) does the work to then populate the settings file.

This PR only exposes the following options:

 - `httpAdminRoot`
 - `disableEditor`
 - `codeEditor`
 - `palette.allowInstall`
 - `nodesExcludes` (provided as `palette.nodesExcludes` over the api)
 - `modules.allowInstall`

### UI

The UI code has got a bit complicated in my efforts to reduce code duplication between editing a template and editing a project.

There are views under `frontend/erc/pages/admin/Template/sections` that provide edit templates for different parts of the Template. They have been written to be used in two different modes (toggled by the `editTemplate` property):

 - `editTemplate = true` : Editing the template itself, including a select box to pick the corresponding policy (editable/locked)
 - `editTemplate = false` : Editing the settings in the context of a project, with only those properties with a `policy=true/editable` able to be changed.

The code under `frontend/erc/pages/admin/Template/utils.js` provides some utility functions for manipulating the template structure and setting up the various shadow properties used to track changes and errors.

Even with that in place, there is still a lot more boilerplate code in there that could be componentised.. but not for today.

### Migration

 - The 'create project' page will block the user if there are no templates, with a prompt to ask an Admin to create a template
 - Existing projects without a template will get a default set of properties applied (consistent with existing behaviour) - but the policy is all locked. In other words, *existing* projects cannot modify any of their settings. We will need the story around changing a project's template to enable them to become editable.
 

### Future work

This PR does not address the following:

 - **Versioning of templates**
    If an admin edits a template, then any projects using that template will get those changes the next time it restarts

 - **Scoping of templates**
   Only admins can create templates. Users see *all* available templates (those with `active=true`)

 - **Copying templates**
    Admin has to create each template from a blank start.

 - **Updating running projects**
    If any settings are changed, the project will need to be restarted to pick them up. Right now there is no prompt to do so.


forge/db/controllers/ProjectTemplate.js
```diff
@@ -0,0 +1,40 @@
...
+            let httpAdminRoot = settings.httpAdminRoot
+            delete settings.httpAdminRoot
+            if (typeof httpAdminRoot === 'string') {
+                httpAdminRoot = httpAdminRoot.trim()
+                if (httpAdminRoot.length > 0) {
+                    if (httpAdminRoot[0] !== '/') {
+                        httpAdminRoot = `/${httpAdminRoot}`
+                    }
+                    if (httpAdminRoot[httpAdminRoot.length - 1] === '/') {
+                        httpAdminRoot = httpAdminRoot.substring(httpAdminRoot.length - 2)
```
hardillb, line 13: Is this right?
```shell
$ node
Welcome to Node.js v16.14.0.
Type ".help" for more information.
> let httpAdmin = "/foo/"
undefined
> httpAdmin.substring(httpAdmin.length -2)
'o/'
> 
```
Should it be `httpAdminRoot = httpAdminRoot.substring(0, httpAdminRoot.length - 1)`

forge/db/models/Project.js
```diff
@@ -12,7 +12,22 @@ module.exports = {
...
         type: { type: DataTypes.STRING, allowNull: false },
-        url: { type: DataTypes.STRING, allowNull: false },
+        url: {
+            type: DataTypes.STRING,
+            allowNull: false,
+            get () {
+                const originalUrl = this.getDataValue('url').replace(/\/$/, '')
+                let httpAdminRoot = this.ProjectSettings?.[0]?.value.httpAdminRoot || this.ProjectTemplate?.settings?.httpAdminRoot
+                if (httpAdminRoot) {
+                    if (httpAdminRoot[0] !== '/') {
```
hardillb, line 22: Do we need this as well as the ProjectSettings controller validation earlier?

hardillb, line 22: Do we need this as well as the ProjectSettings controller validation earlier?

hardillb (commented):
I've looked at all the backend code and as much of the frontend as I currently understand

forge/db/models/Project.js
knolleary: I decided it safer to validate it in multiple places whilst I was figuring out who should 'own' the validation and cleaning.

forge/db/controllers/ProjectTemplate.js
knolleary:
```suggestion
                        httpAdminRoot = httpAdminRoot.substring(0, httpAdminRoot.length - 1)
```

Hmm, not sure what I was thinking there...

hardillb (approved):

Issue #125, opened 2022-01-04
Get proper license workflow in place


knolleary:
We currently have a very limited license mechanism in place. It needs a bit more work for MVP.

Current state:

1. A development-only pair of license signing keys are stored in the repository. These can be used to generate valid licenses - but are *not* what we'll use long term. 
2. The license component checks the database to see if a license has been applied as part of the startup wizard.
3. If the license is present, it is verified and the terms it contains are stored
4. The Admin view in the UI will display the current license details but does not allow it to be modified

Required changes:

 - [x] Generate the real license key files. Store private key in 1Password vault. Store public key in this repo
 - [x] Remove placeholder key files from repository 
 - [ ] Generate a new development-only license using the real key files
 - [x] #128 - Add Admin API end point to update license (`PUT /api/v1/admin/license`) 
 - [x] #128 - Allow Admin UI view to change the license
 - [ ] Include license details (as needed by UI) in `GET /api/v1/settings` response
 - [ ] Add license checks on API actions that create Users/Teams/Projects
 - [ ] Add UI prompts where an action is restricted by license

Some outstanding design decisions that should be considered but could be put-off post-MVP
 - Current model only allows for a single license. We may want to have them in their own DB table so we can have a history of previous licenses available.
 - The current model also means if a user gets a new license, there must be overlap between the 'valid from' date of new license and 'expires at' of the old - so they can apply the new license without having to worry too much about timing. Whereas a DB table of licenses could allow for 'automatic' roll-over onto the whichever is the active license.
 - Need to consider license expiry and what happens if a platform finds itself exceeding the current license (or missing license)
 - Should we require all users to have a license of some type? Even in the completely free tier - a license we provide that reflects the free tier entitlement. When we get to adding usage telemetry, that would allow us to relate usage back to a license. Hard to retrofit. Downside is that could inhibit experimentation and long-tail adoption.

knolleary:
Adding to 0.3 milestone as we have a licensable feature in this release and need to validate the license workflow

knolleary:
With 0.3 we have proper license key files generated and stored securely - #364 

We still need to define how features get licensed - feature-flags for individual features, quantities for certain things (users/teams/projects).

Taking this issue out of 0.3 plan as enough has been done for the requirements of that release. We can either close this issue and then raise new issues as more specific requirements are identified, or keep it around to reference in future discussions.

sammachin:
Need to develop stories around licensing

PR #364, opened 2022-03-15
Add production license keys

knolleary:
Part of #125

With this PR, we have production-ready license handling in place.

1. I have added a Production-ready public key for verifying licenses. The corresponding private key and passphrase are in the FlowForge 1Password vault.
2. The license content has been updated to reflect the current state of things:
   - removed the `tier`, `users`, `projects` and `teams` claims - they did nothing so were confusing to have in place. We can add them back when they mean something.
   - added an optional `dev` claim - more on that in a moment
3. The `license-generator.js` tool now asks a series of questions when generating the license.
    - It first asks if this is a development-mode license. If so, it will use the development-only private key that is in this repo to sign it. It also adds the `dev` flag to the license, set to `true`.
    - If it is not dev-mode, it asks for the path to the production private key (which you need to have downloaded from 1password locally), and then for its passphrase (also in 1password)
    - It asks for the license holder's name and an optional note (in dev mode, the note is hardcoded to say it is a dev-only license... really hammering home the point)
    - It asks when the license should be valid from (default today) - and automatically sets the expiry to +1year.
    - It then prints a summary of the above, and the license content.
4. The UI has been updated where it displays the license details to not include the unused 'tier' claim, and also to add a clear indication if it is a dev license or not

This work also discovered a bug in the `elapsedTime` helper in the front end - that takes a period in seconds and turns it into a human readable string (eg `1 year, 2 months`).... which led to a reword to make it behave properly.


knolleary:
One more thing... when the forge app tries to verify a license, it first tries with the production public key. If that fails, it falls back to the dev-only public key. This will make our lives easier.

hardillb (approved):

PR #365, opened 2022-03-15
Implement password reset

### LINKED ISSUE #325: Forgot  Password
### Epic

_No response_

### Description

**As a:** user

**I want to:** be able to reset my password if I have forgotten

**So that:** I can login

### Acceptance Criteria

- [ ] "Forgot Password" link/button on login screen
- [ ]  Reset password link sent to registered email

hardillb:
This adds the following

**API**

- `/accounts/lostpassword` POST with `{ email: 'alice@example.com' }` always returns 200 even if unknown email address
- `/accounts/passwordreset/:token` GET authenticates user and redirects to password reset form
- `/accounts/resetpassword` POST with `{ password: `aaPassword` }` changes password and signs user out so they have to authenticate.

Reuses the `AccessToken` model, needs the association checking

Could do with limiting the scope of the session to only the password reset page and the password reset page needs the home button removing, the signed in user drop down in the right hand corner is useful to confirm the user changing password.

fixes #325

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ docs/jsdoc
 # Users should have a system based .gitignore that works globally, however, this
 # list is a failsafe when they don't.
 .DS_Store
+
+*.env
+*.env.local
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,4 @@
+var/
 docs/
 config/
 .github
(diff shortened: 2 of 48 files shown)


knolleary:
I don't think leaving them logged in just by having following the link we send to the email address is suitable - even for a short-term solution.

The link should take them to a view where they are not logged in and all they can do is reset their password. I'll propose some concrete changes via a PR.

knolleary:
Given the pending release, I've proposed some changes via #366 that builds on top of this PR.

knolleary (approved):

PR #367, opened 2022-03-16
Add id column (styled subtly as to not distract)

joepavitt:
Closes 323

### DIFF
--- a/frontend/src/pages/admin/Teams.vue
+++ b/frontend/src/pages/admin/Teams.vue
@@ -30,6 +30,7 @@ export default {
             nextCursor: null,
             columns: [
                 {name: "Team", class: ['flex-grow'], component: { is: markRaw(TeamCell) }, link: true},
+                { name: '', class: ['font-mono', 'text-xs', 'text-gray-500'], property: 'id' },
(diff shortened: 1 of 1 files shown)


knolleary (approved):

joepavitt:
Screenshot 2022-03-16 at 11 11 31

sammachin:
Should the column have a heading like 'ID'

joepavitt:
> Should the column have a heading like 'ID'

Yeah, glad you brought that up. I actually went through 3 variants before settling on this. Will include a screenshot of each and then I can just push the preferred:

1. Not overcomplicating it, just adding it in as a standard column:

Screenshot 2022-03-16 at 11 23 26

2. Adding some styling to the column. This required an additional piece in the `ItemTable` to differentiate styling applied to a heading/cell for a given column:

Screenshot 2022-03-16 at 11 18 46

3. This, visually I preferred, but then couldn't click/highlight the id as it was part of the clickable region of the row to navigate to that Team:

Screenshot 2022-03-16 at 11 20 17

knolleary:
Happy to move forward with what is in this PR. All of these tables should get a proper UX overhaul - not going to sweat individual pixels at this point.

PR #386, opened 2022-03-18
362 vue linting

joepavitt:
Most linting changes came into one of these:
- Changing `"` to `'`
- Line spacing
- Removing unnecessary imports & unused components (particularly `FormRow` which was in a lot of places
- Renaming of components that only used 1 word in their name, e.g. `Home` to `HomePage`


frontend/src/components/LogoSquare.vue
```diff
@@ -1,8 +1,8 @@
...
 </template>
 <script>
 export default {
-    name: "LogoSquare"
+    name: 'FlowForgeLogo'
```
knolleary, line 6: This component is no longer used and can be removed. Alternatively it needs a different name as this one is already used by `Logo.vue`.

knolleary:
Changes look good - one minor unused component that can either be removed or renamed.

PR #388, opened 2022-03-18
Set up local development

### LINKED ISSUE #387: Build local development script
### Description

A script to install all the @flowforge dependencies from local git repos rather than npm.

This will allow local development from main/HEAD without manual intervention

should be `npm run dev:local`

Possibly add task to revert changes to package.json as well.

### Epic/Story

_No response_

hardillb:
This adds the `dev:local` script to the package.json.

This will convert the current checked out version to local mode. This means that it will load @flowforge/localfs from a locally checked out version (and it's @flowforge/nr-* dependencies)

It expects the following directory tree

devleopment root
  - flowforge
  - flowforge-driver-localfs
  - flowforge-nr-storage
  - flowforge-nr-auth
  - flowforge-nr-audit-logger

Fixes #387


knolleary:
This'll do for now. It may be more maintainable as an actual script - so we can add some safety checks before it starts doing things.

knolleary:
We need to add details of this to `docs/contribute/README.md`

PR #389, opened 2022-03-18
Install stack script

hardillb:
To make it easier to boot strap a new clone.

`npm run install-stack --vers=2.2.2` will install Node-RED v2.2.2 in var/stacks/2.2.2


knolleary:
As with #388 - I think this may want to be an actual script that is part of the end user docs - that can do proper help and arg parsing to validate what is being passed in.

PR #396, opened 2022-03-18
Update eslint rules to include .vue files in build.

joepavitt:
- Contains additional fixes for issues flagged
- Contains two outstanding issues that are fixed in ui_components


joepavitt:
Build expected to fail due to the 2 outstanding issues in the eslint. These weren't trivial fixes, but were fixed in #395

PR #398, opened 2022-03-19
send correct credentials object

sammachin:
Fix for https://github.com/flowforge/flowforge/issues/397

IMO we should consider a `0.3.1` patch for this rather than waiting till 0.4, as it results in users loosing their credential data from flows on a project restart and there is no log warning or explanation.

### DIFF
--- a/forge/routes/storage/index.js
+++ b/forge/routes/storage/index.js
@@ -75,7 +75,7 @@ module.exports = async function (app) {
         if (project) {
             const creds = await app.db.models.StorageCredentials.byProject(id)
             if (creds) {
-                response.send(creds)
+                response.send(creds.credentials)
             } else {
                 response.send({})
             }


hardillb:
Thanks, I had 90% decided to save this for Sean on Monday as an excuse to get him to commit and release 0.3.1 on day one (and then to push to prod)

sammachin:
Sorry I wondered if the "first issue" tag was aimed at me, happy to pull my PR

hardillb:
We still need the fix against the 0.3.0 tag so we can keep this and still do the other bit

knolleary:
We need a proper strategy for fix releases as I mentioned on Friday. I will document in handbook, but here is the approach we should take whilst we are pre 1.0:

1. when we create a new `0.x` release, create a branch `0.3-maint`
2. when an issue is identified in the released code:
    1. PR the fix to `main` - it should include tests.
    2. Create a second PR to port the fix to `0.3-maint` - if the two branches haven't diverged, it will be usually be a simple `git cherry-pick` to copy the commit to a clean branch for the PR

Future `0.3.x` releases are made from the `0.3-maint` branch as needed.

hardillb:
Closing as this is wrapped up in #401

PR #401, opened 2022-03-21
Add tests for the Storage API

### LINKED ISSUE #397: Project credentials not getting loaded
### Current Behavior

If a project is restarted it losses it's credentials

https://github.com/flowforge/flowforge/blob/7f6f644a1294bb362327caf66022414440b6fc7f/forge/routes/storage/index.js#L78

Should be

                response.send(creds.credentials)

### Expected Behavior

Credentials should be persisted

### Steps To Reproduce

- Create project
- Add nodes with credentials
- restart project
- check for credentials

### Environment

- FlowForge version: 0.3.0
- Node.js version: 16.x
- npm version: 8.x
- Platform/OS: Linux
- Browser: NA

hardillb:
And fix credentials and sessions end points so they now pass the tests

This obsoletes PR #398

Fixes #397

### DIFF
--- a/test/unit/forge/routes/storage/index_spec.js
+++ b/test/unit/forge/routes/storage/index_spec.js
@@ -47,7 +47,7 @@ describe('Storage API', function () {
     it('Save Flow', async function () {
         this.timeout(10000)
         const newFlow = [{ id: '1', type: 'tab', label: 'tab1', disabled: false, info: '' }]
-        const flowURL =  `/storage/${project.id}/flows`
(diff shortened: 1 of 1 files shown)


knolleary:
Yup - fastify.inject is the way to go. We use it in the system tests - https://github.com/flowforge/flowforge/blob/main/test/system/001-setup_spec.js#L53

knolleary:
I think the tests would be more suited as a system level test as it cuts across api and db models... but this is a good step forward we can go with for now.

PR #403, opened 2022-03-21
Fix dev:local task to setup nr-launcher symlinks

knolleary:
The task wasn't creating the right symlinks which left projects unable to start.

This does the right set of npm installs to ensure all the necessary symlinks are created.


hardillb (approved):
yeah, that makes sense.

PR #408, opened 2022-03-22
Fix cancel/create button alignment on Create Template page

joepavitt:


hardillb (approved):
Checked out and tested locally

PR #418, opened 2022-03-25
Fix verbose logging from auth routes

hardillb:
Force these routes to `warn` for requests

These routes log at Info currently, had to add `logLevel: 'warn'` to each route.

### DIFF
--- a/forge/routes/auth/index.js
+++ b/forge/routes/auth/index.js
@@ -170,7 +170,8 @@ module.exports = fp(async function (app, opts, done) {
                     remember: { type: 'boolean', default: false }
                 }
             }
-        }
+        },
+        logLevel: 'warn'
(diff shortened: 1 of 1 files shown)


knolleary:
I think there is a general tidy up needed with how we log routes consistently without having to set log levels on each individual one.

There is some interaction with Fastify plugins that is making this hard. We need to learn more from Fastify on how logging should be approached when splitting routes across plugins.

knolleary:
Previous comment not intended as a 'please go away and re do this'. This PR is good to merge to solve the immediate issue.

knolleary (approved):

hardillb:
I agree, long term review is needed. I was just getting fed up with my logs getting spammed.

PR #425, opened 2022-03-29
Add project meta data to the Subscription

### LINKED ISSUE #423: Stripe doesn't show 2nd project in metadata
### Current Behavior

In the Stripe Admin Dashboard the subscription metadata only contains the project ID for the first project, even though there are 2 in the qty on the subscription item 

### Expected Behavior

Metadata should contain ID's for all projects

### Steps To Reproduce

Setup billing on an account
Create 2 projects
Check stripe dashboard

### Environment

- FlowForge version: 0.3.3
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

hardillb:
Previously the project id was added to the Project line item on the subscription not the subscription for all but the fist project

While stored it's not visible in the stripe dashboard (We could build our own dashboard for this)

Added code to add all projects to the subscriptions metadata.

We might want to revert this when/if we have projects at different price points.

Fixes #423

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -2,7 +2,7 @@ module.exports.init = function (app) {
     const stripe = require('stripe')(app.config.billing.stripe.key)
 
     return {
-        createSubscriptionSession: async (team, user) => {
+        createSubscriptionSession: async (team) => {
             const session = await stripe.checkout.sessions.create({
                 mode: 'subscription',
                 line_items: [{
@@ -20,12 +20,6 @@ module.exports.init = function (app) {
                 success_url: `${app.config.base_url}/team/${team.slug}/overview?billing_session={CHECKOUT_SESSION_ID}`,
                 cancel_url: `${app.config.base_url}/team/${team.slug}/overview`
             })
-            // app.db.controllers.AuditLog.teamLog({
-            //     team.id,
-            //     user.id,
-            //     'billing.session.created',
-            //     { session: session.id }
-            // })
             app.log.info(`Creating Subscription for team ${team.hashid}`)
             return session
         },
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -76,7 +76,8 @@ module.exports.init = function (app) {
                 try {
                     await stripe.subscriptions.update(subscription.subscription, update)
                 } catch (error) {
-                    app.log.info(`Problem adding first project to subscription\n${error.message}`)
+                    app.log.warn(`Problem adding first project to subscription\n${error.message}`)
+                    throw error
                 }
             }
         },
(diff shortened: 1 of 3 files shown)


sammachin (approved):

sammachin:
Fine for now but if/when we have different price point projects I think we would update the value of the key to reflect the type of project instead of just being `True`

forge/ee/lib/billing/index.js
```diff
@@ -47,17 +45,23 @@ module.exports.init = function (app) {
...
                 }
                 // TODO update meta data?
-                stripe.subscriptionItems.update(projectItem.id, update)
+                try {
+                    await stripe.subscriptionItems.update(projectItem.id, update)
+                    await stripe.subscriptions.update(subscription.id, {
+                        metadata: metadata
+                    })
+                } catch (error) {
+                    app.log.info(`Problem adding project to subscription\n${error.message}`)
```
knolleary, line 63: This looks like we swallow any failures to update the subscription and let things continue? Should this fail back and be reported to the user?

Also, not sure `info` is the right level of logging failures like this - unless it is completely benign.

forge/ee/lib/billing/index.js
```diff
@@ -88,18 +96,20 @@ module.exports.init = function (app) {
...
                 }
 
                 try {
-                    stripe.subscriptionItems.update(projectItem.id, update)
+                    await stripe.subscriptionItems.update(projectItem.id, update)
+                    await stripe.subscriptions.update(subscription.id, {
+                        metadata: metadata
+                    })
                 } catch (err) {
                     console.log(err)
```
knolleary, line 114: `console.log` should be swapped for proper app logging - and again, what is the consequence of this failing?

knolleary (commented):
Added some comments on the error handling - we seem to be swallowing errors and it isn't clear what state that will leave things in.

forge/ee/lib/billing/index.js
hardillb: Moving over to `warn` level. I can propagate the error up, would just need to work out what the UI should show. 
At the moment a failure in the first instance means we don't bill but the project exists consuming resources.
For the second the project is not removed from a bill but is shutdown.

I think we should probably change the order on the remove path, so a project is only removed from billing if it is shutdown.

forge/ee/lib/billing/index.js
hardillb: Fixed to use `app.log.warn()`

hardillb:
Add in audit log events for team creation and removal.

forge/ee/lib/billing/index.js
hardillb: Errors now propagate up.

knolleary (approved):

Issue #225, opened 2022-01-24
Project Environment Variables


knolleary:
### Description

**As a:** User

**I want to:** set environment variables for a project

**So that:** I can deploy a common set of flows to different projects, but use env vars to customise each instance

---

Need to consider how env vars are stored - particularly ones that could hold credentials.

sammachin:
Environment variables should be part of the template, and then editable on a per project level

knolleary:
## UI Appearance

This is the rough UI I put together a few months ago when I was experimenting with various things.

<img width="913" alt="image" src="https://user-images.githubusercontent.com/51083/159508653-fc0f245b-932d-4c11-a365-e835e432167d.png">


The table is a list of env var name/value pairs. The user can edit any of the names/values of normal env vars. If the user ticks 'encrypt' then the value is encrypted and not visible to the UI anymore - nor can the name be changed. In this instance, the user can only delete the entry and re-add it.

There will be a 'Save settings' button at the bottom to apply the changes - consistent with the other settings tabs.

## Storing env var values

The current Template/Project Settings model works like this:
 - A `ProjectTemplate` stores the template values in a text field called `settings` as a JSON blob
 - A Project has a `ProjectSettings` object with a key name of `settings` and a text value that is a JSON blob
 
The true value of a particular project setting is taken from the `ProjectSettings` object, falling back to the `ProjectTemplate` if the project hasn't overridden the value.

The simple approach here will be to add `env` to the JSON blob in both places - as we would any other project setting that can be provided by a template.

<details><summary>Side note on thread safety of this blob approach - expand me for details</summary>possibly not an immediate concern, but having just thought of it, I want to write it down. Having a single field in the database as a JSON blob may be fine when a requests to update it happen one at a time. But there will be a race condition if two users edit the project settings at the same time - even if they are editing *different* values in settings. If they save their changes at the same time, there is a real risk of one user's changes overwriting the other user's.</details>

### Encryption of secrets

Given env vars will often be used to provide credentials, it is important we have a sensible strategy for storing their values.

They key requirement is that once a credential type env var has been set, the only thing that should be able to read its true value is the `nr-launcher`. The question is how to best to achieve that.

One approach would be to delegate storage of credentials to the container driver layer. K8S could plug that into K8S Secrets (and rely on whatever administrative level encryption has been setup). The downside is we would still have to implement something sensible for localfs and docker (unless there's a standard dockerised Secrets manager we could plug in).

[...]

PR #426, opened 2022-03-30
Add support for Environment Variables in Project Templates

knolleary:
Delivers #225 

Walk-through video: https://youtu.be/vaIbxbe59ek

 - Adds `env` as a property of both Templates and Project Settings
 - Allows Template to predefine env vars - and whether a project can modify the value of any of them.
 - Allows Projects to define their own env vars (and modify any the template allows to be modified)

 - Refactored the ProjectTemplate controller and added a Project controller to centralise handling of project setting updates
 - Added unit tests for the controllers

This PR does *not* cover:
 - support for encrypted env vars. That will be a separate story.

### DIFF
--- a/forge/db/controllers/ProjectTemplate.js
+++ b/forge/db/controllers/ProjectTemplate.js
@@ -1,8 +1,84 @@
+const validSettings = [
+    'disableEditor',
+    'httpAdminRoot',
+    'codeEditor',
+    'palette_allowInstall',
+    'palette_nodesExcludes',
+    'modules_allowInstall'
+    // 'env' // Handled separately
+]
+
+function getTemplateValue (template, path) {
+    const parts = path.split('_')
+    let p = template
+    while (parts.length > 0) {
+        const part = parts.shift()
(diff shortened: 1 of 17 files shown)


forge/db/controllers/ProjectTemplate.js
```diff
@@ -12,29 +88,56 @@ module.exports = {
                     if (httpAdminRoot[httpAdminRoot.length - 1] === '/') {
                         httpAdminRoot = httpAdminRoot.substring(0, httpAdminRoot.length - 1)
                     }
```
hardillb, line 90: What happens if `httpAdminRoot` is entered as `/`? The substring call makes it an empty string

knolleary: Then an empty string gets passed down which is the same as not setting it - I'm 99% sure NR does the right thing with that.

hardillb: yeah, it would evaluate to false

hardillb (approved):

PR #432, opened 2022-03-31
Update LocalFS install instructions

hardillb:
This updates the instructions to match those in the installer project.

These instructions cover how to install from the zip file so no need to use a tar file.

Closes https://github.com/flowforge/installer/issues/25

### DIFF
--- a/docs/install/local/README.md
+++ b/docs/install/local/README.md
@@ -53,7 +53,7 @@ xcode-select --install
 
 ### Installing FlowForge
 
-1. Create a directory to be the home of your FlowForge install. For example: `/opt/flowforge`
+1. Create a directory to be the base of your FlowForge install. For example: `/opt/flowforge` or `c:\flowforge`
 
    For Linux/MacOS:
     ```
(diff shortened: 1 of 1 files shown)


docs/install/local/README.md
```diff
@@ -56,30 +56,49 @@ xcode-select --install
 1. Create a directory to be the home of your FlowForge install. For example: `/opt/flowforge`
```
sammachin, line 56:
```suggestion
1. Create a `flowforge` directory to be the base of your FlowForge install. 
```

docs/install/local/README.md
```diff
@@ -56,30 +56,49 @@ xcode-select --install
...
 
+   For Windows:
+    ```console
+    cd c:\flowforge
+    install.bat
+    ```
+
 #### Installing as a service
 
 On Linux, the installer will ask if you want to run FlowForge as a service.
```
sammachin, line 104: How does the user run as a service on Windows?

hardillb: You don't easily, you need extra things like pm2 or https://www.hardill.me.uk/wordpress/2014/05/30/running-node-red-as-a-windows-or-osx-service/.

docs/install/local/README.md
hardillb: I've updated this to match the installer/README.md with examples for both (which I want to keep because it adds context to the exampels) and changed `home` to `base`

knolleary (approved):

Issue #392, opened 2022-03-18
Project Lifecycle Review


knolleary:
### Description

We need to review the overall lifecycle of a Project to cover - creation, stopping, restarting etc.

This directly feeds into the following stories:

- #348 
- #370 
- #377 
- #288

hardillb:
State diagram as currently implemented (0.3.0)
```mermaid
stateDiagram-v2
SM: Safe Mode
Created
Starting
Running
Stopped
Crashed
state Crashed_Choice <<choice>>
%% Suspended
%% note right of Suspended
%%     Container stopped, but project 
%%     still in database
%% end note
Deleted
[*] --> Created
Created --> Starting
Starting --> Running
Running --> Stopped
Running --> Crashed
Crashed --> Crashed_Choice
Crashed_Choice --> SM: Loop detected
Crashed_Choice --> Starting: One off
SM --> Running
Stopped --> Starting
%% Stopped --> Suspended
Stopped --> Deleted
%% Suspended --> Deleted
%% Suspended --> Starting
Deleted --> [*]
```

A starting point to implement support for https://github.com/flowforge/flowforge/issues/377 and adding support to allow user to restart in safe mode

```mermaid
stateDiagram-v2
SM: Safe Mode
Created
Starting
Running
Stopped
Crashed
state Crashed_Choice <<choice>>
Suspended
note right of Suspended
    Container stopped, but project 
    still in database
end note
Deleted
[*] --> Created
Created --> Starting
Starting --> Running
Running --> Stopped
Running --> Crashed
Running --> SM
Crashed --> Crashed_Choice
Crashed_Choice --> SM: Loop detected
Crashed_Choice --> Starting: One off
SM --> Running
Stopped --> Starting
Stopped --> SM
Stopped --> Suspended
Stopped --> Deleted
Suspended --> Deleted
Suspended --> Starting
Deleted --> [*]
```

I think this would require a new container api endpoint of `suspend` and a requirement to pass arguments to `start` to allow safe mode and restarting from suspended state. 

The suspended state would also be the intermittent state when a stack is changed for a project, allowing the current container to be stopped and then restarted.

knolleary:
Here's where we are headed. This visualisation separates the Container state (orange) and the Node-RED runtime state (red). Each transition is labelled with the API call that triggers a transition. It also shows the Project model's `state` value (grey) at each point.

This will drive some breaking changes to the existing container driver API - but will give us the level of control needed for the various stories outlined in this issue's description.

[...]

Issue #377, opened 2022-03-17
Allow user to suspend project so it doesnt consume resources


sammachin:
### Epic

#787

### Description

**As a:** Owner

**I want to:** have the ability to stop a container for a project

**So that:** I am not charged for consuming resources 

### Which customers would this be availble to

- [X] All Users, (CE)
- [ ] Specific Project Types (EE)
- [ ] Specific Team Tiers (EE)
- [ ] Other (See comments)

### Acceptance Criteria

- [ ] Container is stopped/removed
- [ ] Project is removed from billing

sammachin:
This needs some discussion about engineering design and UX, specific questions are:

Should this be a separate state from the current stopped project which stops Node-RED from handling any requests but doesn't remove the container?
Is that a useful state to keep for example if you need to just pause whatever your flow is doing.

Should that be renamed to pause?

What do we call this, halted?, terminated?

What is the billing experience, eg if a user stops and starts containers in a short timeframe do we refund and then charge again when its started?

Do we have any kind of charge/limit on the number of projects without a running container that a user might have?

sammachin:
Also how does this experience relate and impact  localfs where there is no container as such, does the UI need to hide the option in localfs mode

knolleary:
The work done in #392 and #437 has introduced the 'suspended' state to a Project - a necessary state when considering how stacks get changed.

Whilst developing and testing #437 I had added a 'suspend' option in the project action menu to help me test different state combinations.

I have *removed* the menu option from the final PR to not confuse matters. For future reference, this is the comit that did hid it - so you can see what front-end api is there to use in the future - https://github.com/flowforge/flowforge/pull/437/commits/cf3606ad4ce85cdd7b431f7c388d7cb2de2d27c6

But the main take-away is that the questions around what a suspended project means have been largely addressed through the lifecycle review work.

The remain piece of dev work is on how to expose it properly in the UI.

Whilst I had put it in the action menu (the drop-down menu in the RH side of the project screen) as a super convenient place whilst I was testing, it is more suited for the 'Danger' settings panel.

The text that goes on screen also needs to reference the impact on billing - but only if billing is enabled.

As such, I am going to add a small sizing for this item to reflect the relatively small piece of UI work needed.

sammachin:
TODO Look at billing experience of stopping & starting with credits

hardillb:
OK, basic UI done for suspending, might have a chat with @joepavitt tomorrow about if we want to change things for restarting.

[...]

Issue #288, opened 2022-02-12
Story: Modify the stack a project is configured to use


knolleary:
### Epic

#285 

### Description

**As a:** User

**I want to:** Modify the stack a project is configured to use

**So that:** I can upgrade Node.js/Node-RED version

### Acceptance Criteria

 - [ ] ...

sammachin:
Shouldn't this be as an *owner*

knolleary:
- [ ] #437 covers the core changes
 - [ ] flowforge/flowforge-driver-localfs#37 - covers the localfs driver changes
 - [ ] flowforge/flowforge-driver-docker#17 - covers the docker driver changes

ToDo:

 - [ ] k8s driver changes

PR #437, opened 2022-04-04
Allow a project stack to be changed

knolleary:
Part of #288

This PR is a significant refactor of the driver API to support the ability to suspend projects. This in turn gives the ability to change the stack a project is using.

There will be PRs in the localfs/docker/k8s projects to match these API changes.

### Driver api changes

 - `init` - unchanged - initialise the driver
 - `start` - was previously called `create`. This should start a project. The project might have previously been run, so could have settings already allocated to it.
 - `stop` - stops the project running. This will put the project into `suspended` state. It will be removed from billing. The driver should stop the launcher/container so it is not consuming any resources. The project could get resumed in the future - so any reserved resources should be kept (ie allocated port number in localfs)
 - `remove` - same as `stop`, but the project isn't coming back. So all resources can be released.
 - `startFlows` - was previously called `start`. This should tell the launcher to run node-red.
 - `stopFlows` - was previously called `stop`. This should tell the launcher to stop node-red (but keep launcher running)
 - `restartFlows` - Tell the launcher to restart
 - All other functions unchanged

### Project States

See also #392 

The `Project.state` field is used to save the desired state of the project. It can be `running`, `stopped`, `suspended`.

We also track, in memory, where there are transitions in states happening (in `forge/app/db/controllers/Project`). This allows us to set an 'in-flight' states of `starting`, `stopping` etc - meaning the UI can give better feedback when things are happening.

### Changing a project stack

A6fIQruwES

### Other related changes

 - The 'editor' button on the project info page is now only shown if the project is running.

### TODO

 - For my testing, I added a 'suspend' action to the UI in the project dropdown. Suspending projects is a different story (#377), although all of the backend work was needed to support Changing Stacks so has been done.
     - As suspending projects is not in scope for 0.4, I will remove it from the menu option before this PR is merged. (Even if we keep it, it should be on the 'Danger' settings tab, not the top-level actions menu).
 - Better test coverage
   -  Having been testing this extensively with localfs, but need to write automated tests for the state management parts. I have updated the stub driver to the new API, and started to add some test utilities to it (by creating a project with a certain name, it will simulate different types of errors/timeouts etc). Need to write the test cases to make use of that.


knolleary:
Having done the localfs and docker implementations, I am happy with these core changes.

They now also have reasonable unit/system test coverage for the core paths around creating/starting/stopping projects and switching stacks - using the stub driver. This includes where the container wrapper interfaces with Billing - checking the logic around adding/remove projects from billing as they are started/stopped/removed.

knolleary:
Actually - I've missed one todo - I need to either hide or move the 'suspend' UI element. Update coming shortly.

knolleary:
I have hidden the suspend option. It has proven super useful in testing project lifecycle and we should look at adding it back properly soon as part of #377.

PR #443, opened 2022-04-05
Add in T&Cs options in Admin Panel

### LINKED ISSUE #363: Agree T&Cs on Signup
### Epic

_No response_

### Description

**As a:** user

**I want to:** view and agree to terms and conditions when signing up

**So that:** I fully understand the contract I am entering into.

### Acceptance Criteria

- [ ] Signup page at `/account/create` has a link to view T&Cs 
- [ ] Signup page at `/account/create` has a checkbox that the user agrees to T&Cs (signup is blocked until they do so)
- [ ] Admin can toggle the requirement for T&Cs in config file 
- [ ] Admin can set the URL for the think to T&Cs in config file
- ...

joepavitt:
closes #363

### DIFF
--- a/forge/routes/api/settings.js
+++ b/forge/routes/api/settings.js
@@ -23,6 +23,9 @@ module.exports = async function (app) {
                 response['telemetry:enabled'] = app.settings.get('telemetry:enabled')
                 response['user:signup'] = app.settings.get('user:signup')
                 response['user:reset-password'] = app.settings.get('user:reset-password')
+                response['user:tcs-required'] = app.settings.get('user:tcs-required')
+                response['user:tcs-url'] = app.settings.get('user:tcs-url')
+                response['user:reset-password'] = app.settings.get('user:reset-password')
                 response['user:team:auto-create'] = app.settings.get('user:team:auto-create')
                 response.email = app.postoffice.exportSettings(true)
             }
@@ -30,7 +33,9 @@ module.exports = async function (app) {
         } else {
             reply.send({
                 'user:signup': app.settings.get('user:signup') && app.postoffice.enabled(),
-                'user:reset-password': app.settings.get('user:reset-password') && app.postoffice.enabled()
+                'user:reset-password': app.settings.get('user:reset-password') && app.postoffice.enabled(),
+                'user:tcs-required': app.settings.get('user:tcs-required') && app.postoffice.enabled(),
+                'user:tcs-url': app.settings.get('user:tcs-url')
             })
         }
     })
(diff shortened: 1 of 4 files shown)


knolleary (changes requested):

forge/routes/api/settings.js
```diff
@@ -23,14 +23,19 @@ module.exports = async function (app) {
...
                 response['user:team:auto-create'] = app.settings.get('user:team:auto-create')
                 response.email = app.postoffice.exportSettings(true)
             }
             reply.send(response)
         } else {
             reply.send({
                 'user:signup': app.settings.get('user:signup') && app.postoffice.enabled(),
-                'user:reset-password': app.settings.get('user:reset-password') && app.postoffice.enabled()
+                'user:reset-password': app.settings.get('user:reset-password') && app.postoffice.enabled(),
+                'user:tcs-required': app.settings.get('user:tcs-required') && app.postoffice.enabled(),
```
hardillb, line 37: Why does email need to be enabled to require T&Cs?

knolleary:
There was a related requirement which didn't get captured in the story about what happens if/when the T&Cs are updated and we need users to acknowledge they have seen the updated terms.

From a high level design that would mean:

 - Adding a 'accepted_tcs' field to the User table that is a Date
 - Include a flag in the user registration API to signal they have been accepted (reject the request if not), and to set the field to the current Date.
 - Have a way for the admin to indicate there are new T&Cs (which a date associated)
 - When a user logs in, if the current TCs are dated newer than their own accepted_tcs flag, then prompt them to review the TCs.

We certainly don't need all of that today.

I will raise a separate story that covers this scenario.

forge/routes/api/settings.js
knolleary: > Why does email need to be enabled to require T&Cs?

Because `user:signup` needs to be enabled - which in turn hinges on email being enabled.

It is possibly a little redundant, but no harm.

knolleary:
Having said all of that, I suggest the signup HTTP API should require a flag to be set (only if `user:tcs-required` is set) as positive acknowledgement that tcs have been accepted at the API level - rather than a pure UI check.

joepavitt:
@knolleary Have amended the comments you've flagged (apologies for missing those). The comments you have above seem to exceed the scope of the issue tackled here, and will belong in the new issue you've mentioned, so I believe everything should be resolved here.

knolleary (approved):

knolleary:
Listing out the future stories related to this:

 - [ ] Signup API should require a T&C flag to be set
 - [ ] User model should record if user accepted T&Cs

Issue #424, opened 2022-03-29
Delete project confirmation dialog has broken layout


knolleary:
### Current Behavior

In the delete project confirmation dialog, the styling of the project name overlaps the surrounding text:

<img width="413" alt="image" src="https://user-images.githubusercontent.com/51083/160612704-1f7021dd-0f9d-4fdd-8b8b-9724a84cacfb.png">

### Expected Behavior

The text should not overlap itself

### Steps To Reproduce

1. create a project
2. Go to project settings -> Danger section -> Click Delete Project

### Environment

- FlowForge version: 0.3

knolleary:
@fakoredeDamilola you are certainly welcome to. I would suggest the simple fix is to ensure the project name is on a newline of its own, rather than trying to have it inline with the rest of the text.

knolleary:
Found another instance of this on the Create Team page.

<img width="198" alt="image" src="https://user-images.githubusercontent.com/51083/161251669-5990349d-f48b-48bb-bfbc-de4d49f727c0.png">

fakoredeDamilola:
Hello can you help me
1) I signup the last time, but I dont think I can remember my password, I cannot find a signup page or a flow for forget password etc
![image](https://user-images.githubusercontent.com/47531232/161485102-c861362d-9cdf-4000-8d8c-a60c38586179.png)
2) @joepavitt I tried what you said for the long term, just for this error, and doing this
`'<pre><code block="inline">{{team.name}}</code></pre>'`
actually removes that code sort of default styling, the background and all and I dont think you will want that
A solution I came up with was to make it a block text based on the length of the string i.e greater than 14

knolleary:
1. We don't have a password reset workflow in place today. If you've forgotten the administrator password, the only way back in is to edit the database directly - not easy to do as you need to generate a proper password hash.

> A solution I came up with was to make it a block text based on the length of the string i.e greater than 14

This doesn't need any logic. The dialog should place the project name on a line of its own - that is all.

joepavitt:
Thanks @fakoredeDamilola - worth noting the long term solution with the `<pre><code block="inline">{{team.name}}</code></pre>` will require quite a lot of other work too, in particular with css changes. We have bigger plans for this, so sticking with the short term solution as Nick as suggested is the way to go at the moment. Thanks for helping out on this.

fakoredeDamilola:
Hello @joepavitt @knolleary 
I made a PR https://github.com/flowforge/flowforge/pull/444

[...]

PR #444, opened 2022-04-06
made the team name block

fakoredeDamilola:
Added a class of block to make the text a block text

### DIFF
--- a/frontend/src/pages/team/dialogs/ConfirmTeamDeleteDialog.vue
+++ b/frontend/src/pages/team/dialogs/ConfirmTeamDeleteDialog.vue
@@ -24,7 +24,7 @@
                                         Are you sure you want to delete this team? Once deleted, there is no going back.
                                     </p>
                                     <p class="text-sm text-gray-500">
-                                        Enter the team name '<code class="block">{{team.name}}</code>' to continue.
+                                        Enter the team name <code class="block">{{team.name}}</code> to continue.
                                     </p>
                                 </div>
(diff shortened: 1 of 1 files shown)


joepavitt:
This work is resolving https://github.com/flowforge/flowforge/issues/424

frontend/src/pages/team/dialogs/ConfirmTeamDeleteDialog.vue
```diff
@@ -24,7 +24,7 @@
                                         Are you sure you want to delete this team? Once deleted, there is no going back.
                                     </p>
                                     <p class="text-sm text-gray-500">
-                                        Enter the team name '<code>{{team.name}}</code>' to continue.
+                                        Enter the team name '<code class="block">{{team.name}}</code>' to continue.
```
joepavitt, line 27: I think we can also remove the `'` either side of the `<code>` block as it looks a little strange with them still in there:

Screenshot 2022-04-06 at 12 11 24

joepavitt (changes requested):

frontend/src/pages/team/dialogs/ConfirmTeamDeleteDialog.vue
fakoredeDamilola: resolved

joepavitt: Thanks! We had some other instances of this styling needing to be applied on a couple of other dialogs, mentioned back in the comment thread on #424 - if you can get those in for this PR then I think we'll be good to go after that

Issue #239, opened 2022-01-26
Set Timezone for a Project


sammachin:
### Epic

#209

### Description

As a team owner 

I want to set the timezone for my project

So that when I schedule events in my flows I can use my local time.

Container accepts timezone as an env. variable but we should expose this in the UI

PR #452, opened 2022-04-07
PLCMercenary/issue239

PLCMercenary:
added tz dropdown via data folder in the frontend/src/  included a json with all JS timezones as the source for the dropdown


knolleary:
Thanks @PLCMercenary. Some high level feedback before getting into the specific details of the code.

We provide an .eslintrc file in the repo which sets our standard code style, including indent requirements. Please use those rules as it avoids lots of reformatting changes. (see https://github.com/flowforge/handbook/blob/main/development/packaging.md#linting)

You've included package.json - but the changes in that file aren't needed here and shouldn't be included. That is a down side of the `npm run dev:local` task - it modifies the package.json file for local work, but you have to be careful not to commit those changes. Easy to get caught out by that.

The PR title should be a short description of the feature the PR adds - as we generate the CHANGELOG files from the PR titles.

Those comments aside, I'll review the specific code changes next.

sammachin:
Linking to issue #239

package.json
```diff
@@ -39,8 +39,8 @@
     "dependencies": {
         "@aws-sdk/client-ses": "^3.49.0",
         "@aws-sdk/credential-provider-node": "^3.49.0",
-        "@flowforge/forge-ui-components": "^0.1.1",
-        "@flowforge/localfs": "^0.3.0",
+        "@flowforge/forge-ui-components": "file:../forge-ui-components",
+        "@flowforge/localfs": "file:../flowforge-driver-localfs",
```
hardillb, line 43:
```suggestion
        "@flowforge/forge-ui-components": "^0.1.2",
        "@flowforge/localfs": "^0.3.3",
```

package.json
```diff
@@ -39,8 +39,8 @@
     "dependencies": {
         "@aws-sdk/client-ses": "^3.49.0",
         "@aws-sdk/credential-provider-node": "^3.49.0",
-        "@flowforge/forge-ui-components": "^0.1.1",
-        "@flowforge/localfs": "^0.3.0",
+        "@flowforge/forge-ui-components": "^0.1.2",
+        "@flowforge/localfs": "^0.3.3",
```
hardillb, line 43:
```suggestion
        "@flowforge/localfs": "^0.3.0",
```

joepavitt:
@PLCMercenary the list of timezones seems rather long? I was going to comment asking why we can't simply have all the UTC Offsets, as I would have expected to see, but found these discussions:

https://stackoverflow.com/questions/38399465/how-to-get-list-of-all-timezones-in-javascript
https://stackoverflow.com/questions/54459791/why-so-many-iana-time-zones-names

Which verified that there is in fact a difference between all of those options.

joepavitt:
I'm also very conscious of the scale of the auto-formatting changes in the `.vue` file. Want to make sure that we aren't pushing excessive formatting changes, and that the code abides by the `.eslintrc` provided. We have guides on how to set this up in VS Code [here in the handbook](https://flowforge.com/handbook/development/contributing/).

joepavitt:
@PLCMercenary can we get the linting updates we discussed yesterday pushed to this branch please so that we can get this merged in please?

PLCMercenary:
Joe,  just got them done.. had a difficult time determining if the `<script> </script>` were poorly linted or if those were OK.

frontend/src/pages/admin/Template/sections/Editor.vue
```diff
@@ -49,19 +57,27 @@
 </template>
 
 <script>
-
-import FormRow from '@/components/FormRow'
+import FormRow from '@/components/ForumRow'
```
joepavitt: @PLCMercenary have you tested these latest changes work? This typo is failing the builds, and should have been picked up when testing the frontend still works after these latest changes

joepavitt: You have a typo here, should be `FormRow`, not `ForumRow`. Please make sure that you're checking the UI runs/works before committing the changes.

joepavitt (changes requested):
`frontend/src/pages/admin/Template/sections/Editor-old.vue` should not have been included in the commit and pushed. Please make sure you're only pushing content that is required to solve the issue/task at hand.

joepavitt (approved):
Need to be aware of any JS autoformatting changes on your next commit @PLCMercenary, the changes you've made here do improve readability, just want to make sure it's not going to happen with every file you touch in the future.

From our own 1:1's, I know we're all good on the `.vue` side.

PR #456, opened 2022-04-07
Fix 404 in Stripe Billing callback for charge.failed

### LINKED ISSUE #454: Stripe Billing Dashboard showing 404 errors for `charge.failed` callback
### Current Behavior

When a charge fails Stripe makes a callback with a `charge.failed` event. This results in a 404 error because the event does not contain the metadata pointing to the associated team.

### Expected Behavior

Notice it's a `charge.failed` event and not try and find the team via metadata

### Steps To Reproduce

Sign up with a credit card that will get rejected when a project is added

### Environment

- FlowForge version: 0.3.5
- Node.js version: 16.x
- npm version: 8.x
- Platform/OS: K8s/Linux
- Browser: NA

hardillb:
This uses the customer id to look up the session to find the team.

This will need further work to actually do something when a charge fails (suspend all projects under that team? probably after multiple fails)

Also adds start of billing unit tests

fixes #454

### DIFF
--- a/forge/ee/routes/billing/index.js
+++ b/forge/ee/routes/billing/index.js
@@ -82,8 +82,14 @@ module.exports = async function (app) {
                     return
                 }
             } else {
-                response.status(404).type('text/html').send('Not Found')
-                return
+                if (event.type === 'charge.failed') {
+                    const sub = await app.db.models.Subscription.byCustomer(customer)
+                    team = sub?.Team
+                }
+                if (!team) {
+                    response.status(404).type('text/html').send('Not Found')
+                    return
+                }
             }
 
             switch (event.type) {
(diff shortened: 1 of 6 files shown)


forge/ee/routes/billing/index.js
```diff
@@ -100,8 +106,8 @@ module.exports = async function (app) {
                 break
             case 'checkout.session.expired':
                 // should remove the team here
-                console.log('checkout.session.expired')
-                console.log(event)
+                app.log.info('checkout.session.expired')
```
knolleary, line 109: Can we add some more context in the message so we can relate it back to a user/team/stripe something?

knolleary (commented):
Looks good. Well done on getting the tests figured out.

One small log message that I think could do with some more info added so we'd have some hope of relating back to an entity should it occur.

knolleary (approved):

PR #458, opened 2022-04-07
Add requirement for price acknowledgement when creating project

joepavitt:
- Removed `float-right` and `inline` from the error messaging as caused awkward overlap with elements and strange width issues with the checkbox FormRow proceeding it.
- May want to make the price retrieval more dynamic in the future.

### DIFF
--- a/frontend/src/pages/project/create.vue
+++ b/frontend/src/pages/project/create.vue
@@ -24,7 +24,7 @@
                 <FormRow type="checkbox" v-model="input.billingConfirmation" id="billing-confirmation">
                     Confirm additional charges
                     <template v-slot:description>
-                        You will be charged $15/month for this project.
+                        You will be charged US$15/month for this project.
                     </template>
                 </FormRow>
 


frontend/src/components/FormRow.vue
```diff
@@ -66,7 +66,7 @@
                     <div class="block sm:inline sm:absolute sm:left-full sm:ml-4 mt-2 sm:mt-0"><slot name="append"></slot></div>
                 </template>
             </div>
-            <div v-if="error" class="float-right ml-4 text-red-400 inline text-xs">{{error}}</div>
```
knolleary, line 69: The intention for the float was so to stop the content below this row getting pushed down when the error is shown - not a fan of moving form elements.

Now you mention it, I do see the issue with the Project name row on the create page - with the recently added text about not being able to change name:

Cursor_and_FlowForge

In theory *think* we could apply `after:clear-both` to the top FormRow containing div - so it stops the float from overflowing the container... but I can't get that to work locally. Not going to sweat it - just wanted to highlight the reason for the float.

frontend/src/pages/project/create.vue
```diff
@@ -21,13 +21,17 @@
 
                 <FormRow :options="templates" :error="errors.template" v-model="input.template" id="template">Template</FormRow>
 
+                <FormRow v-if="features.billing" type="checkbox" v-model="input.billingConfirmation" id="billing-confirmation">Please acknowledge that you will be charged (15 USD) for creating this project.</FormRow>
```
knolleary, line 24:
```suggestion
                <FormRow type="checkbox" v-model="input.billingConfirmation" id="billing-confirmation">
                    Confirm additional charges
                    <template v-slot:description>
                        You will be charged $15/month for this project.
                    </template>
                </FormRow>
```

Counter proposal - @sammachin please lmk your preference.

image

frontend/src/components/FormRow.vue
joepavitt: There was another strange issue whereby the `float:right` was limiting the width of the `FormRow` directly after, if that `FormRow` was of `type="checkbox"`, was really strange as they had no reason to interact.

frontend/src/pages/project/create.vue
sammachin: I'm happy with the design, only thing is we use `US$` where displaying currency just to be clear to other countries that may use $ eg CAD

frontend/src/pages/project/create.vue
```diff
@@ -21,13 +21,22 @@
 
                 <FormRow :options="templates" :error="errors.template" v-model="input.template" id="template">Template</FormRow>
 
+                <FormRow type="checkbox" v-model="input.billingConfirmation" id="billing-confirmation">
+                    Confirm additional charges
+                    <template v-slot:description>
+                        You will be charged $15/month for this project.
```
joepavitt, line 27:
```suggestion
                        You will be charged US$15/month for this project.
```

PR #469, opened 2022-04-11
frontend: Admin created users default to no team

ZJvandeWeg:
Users created by the admin might get a personal team. This is the
default before this change. The default is now false after this commit,
as there's a possibility of billing being enabled. That creates a
requirement for payment details for first-time users and a sub optimal
onboarding experience.


hardillb (approved):
I think we should make this default configurable by the admin at some point, but this change works for now.

PR #472, opened 2022-04-11
Don't try and send negative project count to Stripe

hardillb:
Fix the lint errors so it builds clean and we can review just the changes.

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -93,8 +93,9 @@ module.exports.init = function (app) {
             if (projectItem) {
                 const metadata = existingSub.metadata ? existingSub.metadata : {}
                 metadata[project.id] = ''
+                const newQuantity = projectItem.quantity === 0 ? 0 : projectItem.quantity - 1
                 const update = {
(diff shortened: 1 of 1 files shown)


forge/ee/lib/billing/index.js
```diff
@@ -93,8 +93,9 @@ module.exports.init = function (app) {
             if (projectItem) {
                 const metadata = existingSub.metadata ? existingSub.metadata : {}
                 metadata[project.id] = ''
+                const newQuantity = projectItem.quantity > 0 ? 0 : projectItem.quantity - 1
```
joepavitt: isn't this saying that `newQuantity` should be `0` if `quantity > 0`.

so:
`newQuantity = -1` if `projectItem.quantity === 0`
`newQuantity = 0` if `projectItem.quantity === 1`
`newQuantity = 0` if `projectItem.quantity === 5`

joepavitt: surely, it should be `projectItem.quantity < 0 ? 0 : projectItem.quantity - 1`

although, not sure I follow why the `-1` is needed?

joepavitt (approved):

PR #476, opened 2022-04-12
Add a $1 charge when setting up team billing

hardillb:
This means that billing fails for bad cards up front. The $1 is added to the account as credit as soon as the charge completes.

Uses the config item billing.stripe.actiavation_price as a feature flag
```yaml
billing:
  stripe:
    activation_price: price_......
```

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -3,7 +3,7 @@ module.exports.init = function (app) {
 
     return {
         createSubscriptionSession: async (team) => {
-            const session = await stripe.checkout.sessions.create({
+            const sub = {
                 mode: 'subscription',
                 line_items: [{
                     price: app.config.billing.stripe.team_price,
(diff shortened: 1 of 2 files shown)


sammachin:
Documenting my understanding:
- If there is a activation_price value set in the yaml config it adds that to a checkout session
- on the charge.succeeded event again if activation_price is set in config we get the invoice, check if the price_id matches the activation_price
- If true we then create a credit note against the invoice for 100 - I assume stripe uses integers for all currency amounts so that is $1.00

Would it be cleaner to make the credit amount match the activation price in the invoice? Looks like thats part of the price object
Should we specify the currency in the credit note? again taken from the invoice price object

hardillb:
Will update to use value/currency from invoice line item

hardillb:
@sammachin updated, can't add currency to a credit note, but it will use what ever the invoice line item was.

sammachin (approved):

PR #477, opened 2022-04-13
Add option to use existing stack as a base to fill out the form

### LINKED ISSUE #390: Create a new stack from an existing one
An Admin should be able to create a new stack using an existing one's values as the starting point.

joepavitt:
Closes #390.

(Screen recording in #390)


sammachin:
@joepavitt does this allow you to use an inactive stack as the base to create the new one from?

hardillb:
@sammachin It looks like it uses the full list of stacks, not filtering on active, so yes

hardillb (approved):

PR #485, opened 2022-04-19
Rebrand Login

joepavitt:
- Rebrand as per latest Figma designs
- Removed the option upfront for "Login with username" and "Sign Up" (no need when there is only one option there (login with username) as the sign up option also shows on that age too.

Screenshot 2022-04-19 at 13 38 49


joepavitt:
- Doesn't include mobile styling
- Requires latest from https://github.com/flowforge/forge-ui-components/tree/v1.0 as `forge-ui-components`

knolleary:
> Removed the option upfront for "Login with username" and "Sign Up"

The original rationale was that we will one day have SSO support, so I pictured having an initial choice of 'login with username' and 'login with sso' type thing. But as we don't have SSO today, happy to make it more streamlined.


Design is looking 🔥

knolleary (approved):

joepavitt:
> The original rationale was that we will one day have SSO support, so I pictured having an initial choice of 'login with username' and 'login with sso' type thing

Yep, totally understood why it was there, just doesn't make sense right now, and easy to add when it's available.

joepavitt:
Closes #498

PR #495, opened 2022-04-20
Limit access to Project Settings

hardillb:
This limits access to the /project/:id/settings endpoint to tokens with the type `project`

fixes https://github.com/flowforge/security/issues/1


hardillb:
Working on tests
Trying to work out how to auth as a user

hardillb:
Move and existing test from storage tests  (as it's not really the storage API) to new file and then added a new failure test path when using a users credentials.

knolleary (approved):

PR #512, opened 2022-04-25
Admin-can-delete-a-user-PLCMercenary/issue338

PLCMercenary:
user deletion story


forge/db/models/User.js
```diff
@@ -170,6 +192,15 @@ module.exports = {
                 // TODO: standardize on using hashids externally
                 getTeamMembership: async function (teamId, includeTeam) {
                     return M.TeamMember.getTeamMembership(this.id, teamId, includeTeam)
+                },
+                getTeamsRole: async function () {
+                    const teams = await M.Team.forUser(this)
+                    const teamRole = await teams[0].dataValues.role
```
knolleary, line 198: This will return the role a user has in the *first* team they are part of. That isn't what we need here.

As discussed, the idea was to implement `getTeamsOwned` - that will return an array of the teams the user is an owner of.

You are getting `teams` as the list of all teams the user is a part of. The task is to filter that list to only return the ones where the `role` property is `Owner`.

PR #521, opened 2022-04-28
Rebrand - Stage 1

joepavitt:
Stage 1 as defined in : https://github.com/flowforge/flowforge/issues/430

- Includes restructure of layout and signup/sign-in workflow.


knolleary:
When on the User Settings or Admin Settings pages, the sidebar still shows the Team navigation options.

I think we should hide the sidebar when in those sections to make it clear they aren't tied to the Team.

image

joepavitt:
@knolleary good feedback, how would you envision the user going back to the "Team" view? Clicking the logo in top-right?

joepavitt:
Screenshot 2022-04-28 at 17 38 43
Screenshot 2022-04-28 at 17 39 04
Screenshot 2022-04-28 at 17 38 53

joepavitt:
Alternative "Back" button location:

Screenshot 2022-04-29 at 09 01 45

knolleary:
Agreed - back button better at the top

joepavitt:
@knolleary done

knolleary:
- Login Password field is not a password type input:

image

knolleary:
A couple pages of User Settings still use the old side nav - need moving to the top nav
 - User Settings -> Teams
 - User Settings -> Security

joepavitt:
Good spot - should have caught that. Recently merged in @ZJvandeWeg's latest work on `forge-ui-components` which changed that. Also fixed in `FormRow`

joepavitt:
> A couple pages of User Settings still use the old side nav - need moving to the top nav
> 
> * User Settings -> Teams
> * User Settings -> Security

It's also found in:
* Project -> Settings

The mobile styling for the Project Settings though looks pretty bad with the horizontal tabs as there are so many options. I also quite like the side tabs for all of these pages, but can't justify why they should be different to the others.

knolleary:
Create Project and Team pages - should there be a white background container?

image

joepavitt:
> Create Project and Team pages - should there be a white background container?

ah-ha! Finally worked out the difference between your lowercase and uppercase lettering in the filenames! Uppercase feed through `index.vue` (where I fixed this), and the lowercase does not (`create.vue`)

knolleary:
Subtle one...

 - Click Team Settings in LH nav bar - it is highlighted as the current page
 - Click the 'Danger' internal tab - Team Settings loses its highlighting

I think I've checked all the other pages with internal tabs and this is the only one where this happens.

image

Issue #271, opened 2022-02-09
Duplicate Project


sammachin:
### Epic

#268

### Description

**As a:** Owner

**I want to:** Export an existing project to create a new one

**So that:** I can deploy from Dev to Prod the first time.

All values that are in the export should be set and anything not included should use the platform/policy defaults

knolleary:
The 'export' terminology works when you want to leave the system boundary, but when operating inside the system we can be more direct in the terminology and call this 'Copy project' or 'Duplicate project'


### UX

 - The 'Danger' settings tab is where we put administrative actions on a project - this will add a 'Duplicate Project' option on that tab.
     - Aside: the name 'Danger' made sense when it contained the destructive delete action. It now has the 'Change Stack' option, which is something done with caution, so the name still works. 'Duplicate Project' doesn't quite fit the pattern. We should consider renaming that tab to something more appropriate, with a 'Danger' section.
 - Clicking the button will open a dialog to walk through the process.
    - The dialog should look like the 'create project' view where they set the name etc.
        - the Template/Stack should be pre-filled with the right values. I do not think the Template field should be changeable - see #270 for my thoughts on that (we can iterate on that if needed)
     - The dialog also has the options for which project components to copy over - see #270 for details
  - Confirming the dialog kicks off the create and redirects the user to the new project's page - same behaviour as the existing create project flow.

### Runtime API

As this is about creating a new project - with some pre-existing content - the existing `POST /api/v1/project` end point should be extended to support this functionality.

I propose the payload gains an option additional property:

```
...
sourceProject: {
   id: <source project id>
   options: <copy options>
}
...
```

The `<copy options>` object provides the additional information the user provided as to what gets copied over (#270)

sammachin:
Just to confirm, the stack will be changeable when duplicating a project though? but the default would be the same stack as the source project?
Given that it will (soon) be possible to change the stack on a project anyway I can't see why we would want to force a duplicate to be the same stack? That would just require the user to duplicate and then switch the stack.

knolleary:
> Just to confirm, the stack will be changeable when duplicating a project though? but the default would be the same stack as the source project?

Yes - I called out that Template wouldn't be, but didn't say explicitly that stack should be changeable. Have edited above to be explicit.

knolleary:
Delivered in 0.5

Issue #273, opened 2022-02-09
Export to Zip file


sammachin:
### Epic

#268

### Description

**As a:** Owner

**I want to:** Export components of a project to a ZIP file

**So that:** I can store a backup or move to another flowforge system or node-red instance

sammachin:
We should look at the Node-RED project structure and see if we can be compatible with that, possibly via a small transformation script

sammachin:
We need to consider corresponding stories for what the user would do with the Zip file, is there an 'import from zip file' so that a user can restore from their own external backup, Is there a way to run a project locally from a Zip file with local Node-RED? Export without Import doesn't really provide any value.

hardillb:
If the export was done right....

The zip could contain a userDir, so would contain

- settings.js file with all the settings from the template included
- package.json with any extra nodes from the template/stack
- flow.json and flow_creds.json

When unzipped could Node-RED could be pointed at the directory (would need `npm install` running to import the nodes)

Just a thought

sammachin:
Yes this is the sort of thing I was thinking along the lines of, while this seems like a good story, exporting to a self managed Node-RED doesn't create a lot of value for FlowForge it just encourages you to take projects off the platform.

knolleary:
@sammachin you are right to say we need to consider the import side of this as well.

Being able to import back into FlowForge and being able to run it in a local Node-RED instance are two quite different use cases in terms of what the zip can/should contain.

For importing it back into FlowForge (either as migration to another instance, or to restore a manual backup), we cannot take the raw settings.js file - we need the Project Settings JSON. And we have to make sure it isn't a way to import configurations that the platform disallows based on the available templates.

For exporting it as a self-contained standalone Node-RED project, as Ben says, that is much more understood in terms of required files contents.

Are the three uses cases - backing up, move to another FF and move to a standalone NR instance - actually stories in their own right that each represents a different piece of customer value and should be split out? I wonder if the Zip file nature of this story is more an implementation detail?

Otherwise, to deliver this story as described, we need to handle all of those scenarios in one go.

PR #528, opened 2022-05-03
Add export and project duplication

hardillb:
Contains a Vue component for selecting which parts of a Project should be exported. This will be reused for Project duplicate workflow.

Work in progress towards https://github.com/flowforge/flowforge/issues/270 and #271

### DIFF
--- a/frontend/src/pages/project/Settings/Danger.vue
+++ b/frontend/src/pages/project/Settings/Danger.vue
@@ -30,6 +30,14 @@
             </div>
         </div>
 
+        <div class="flex">
+            <div class="max-w-sm pr-2">Copy a project's state to another existing project</div>
+            <div class="pl-2">
+                <ff-button kind="secondary" @click="showExportToProjectDialog()">Export State</ff-button>
+                <ExportToProjectDialog @exportToProject="exportToProject" ref="exportToProjectDialog"/>
+            </div>
+        </div>
+
         <FormHeading class="text-red-700">Delete Project</FormHeading>
         <div class="flex">
             <div class="max-w-sm pr-2">Once deleted, your project is gone. This cannot be undone.</div>
@@ -47,6 +55,7 @@ import FormHeading from '@/components/FormHeading'
 import ConfirmProjectDeleteDialog from './dialogs/ConfirmProjectDeleteDialog'
 import ChangeStackDialog from './dialogs/ChangeStackDialog'
 import ExportProjectDialog from './dialogs/ExportProjectDialog'
+import ExportToProjectDialog from './dialogs/ExportToProjectDialog'
 import DuplicateProjectDialog from './dialogs/DuplicateProjectDialog'
 
 export default {
@@ -65,6 +74,9 @@ export default {
         showDuplicateProjectDialog () {
             this.$refs.duplicateProjectDialog.show(this.project)
         },
+        showExportToProjectDialog () {
+            this.$refs.exportToProjectDialog.show(this.project)
+        },
         exportProject (parts) {
             // call projectApi to generate zipped json
             projectApi.exportProject(this.project.id, parts)
@@ -79,6 +91,14 @@ export default {
             }).catch(err => {
                 console.log(err)
             })
+        },
+        exportToProject (parts) {
+            console.log('Ben')
+            console.log(parts)
+            const options = {
+                sourceProject: parts.sourceProject
+            }
+            projectApi.updateProject(parts.target, options)
         },
         deleteProject () {
             projectApi.deleteProject(this.project.id).then(() => {
@@ -103,6 +123,7 @@ export default {
         ConfirmProjectDeleteDialog,
         ChangeStackDialog,
         ExportProjectDialog,
+        ExportToProjectDialog,
         DuplicateProjectDialog
     }
 }
--- a/frontend/src/pages/project/Settings/Danger.vue
+++ b/frontend/src/pages/project/Settings/Danger.vue
@@ -1,50 +1,50 @@
 <template>
     <form class="space-y-6">
         <FormHeading>Change Project Stack</FormHeading>
-        <div class="flex">
-            <div class="max-w-sm pr-2">Changing the project stack requires the
-                project to be restarted. The flows will not be running whilst
-                this happens.</div>
-            <div class="pl-2">
+        <div>
+            <div>
                 <ff-button kind="secondary" @click="showChangeStackDialog()">Change Project Stack</ff-button>
                 <ChangeStackDialog @changeStack="changeStack" ref="changeStackDialog"/>
             </div>
+            <div class="max-w-sm pt-2 pl-5">Changing the project stack requires the
+                project to be restarted. The flows will not be running whilst
+                this happens.</div>
         </div>
 
         <FormHeading>Export Project</FormHeading>
-        <div class="flex">
-            <div class="max-w-sm pr-2">Allows you to export a snapshot of the
-                project's current state.</div>
-            <div class="pl-2">
+        <div>
+            <div>
                 <ff-button kind="secondary" @click="showExportProjectDialog()">Export Project</ff-button>
                 <ExportProjectDialog @exportProject="exportProject" ref="exportProjectDialog"/>
             </div>
+            <div class="max-w-sm pt-2 pl-5">Allows you to export a snapshot of the
+                project's current state.</div>
         </div>
 
-        <div class="flex">
-            <div class="max-w-sm pr-2">Allows you to use a snapshot of the
-                project's current state to create a new project.</div>
-            <div class="pl-2">
+        <div>
+            <div>
                 <ff-button kind="secondary" @click="showDuplicateProjectDialog()">Duplicate Project</ff-button>
                 <DuplicateProjectDialog @duplicateProject="duplicateProject" ref="duplicateProjectDialog"/>
             </div>
+            <div class="max-w-sm pt-2 pl-5">Allows you to use a snapshot of the
+                project's current state to create a new project.</div>
         </div>
 
-        <div class="flex">
-            <div class="max-w-sm pr-2">Copy a project's state to another existing project</div>
-            <div class="pl-2">
+        <div>
+            <div>
                 <ff-button kind="secondary" @click="showExportToProjectDialog()">Export State</ff-button>
                 <ExportToProjectDialog @exportToProject="exportToProject" ref="exportToProjectDialog"/>
             </div>
+            <div class="max-w-sm pt-2 pl-5">Copy a project's state to another existing project.</div>
         </div>
 
         <FormHeading class="text-red-700">Delete Project</FormHeading>
-        <div class="flex">
-            <div class="max-w-sm pr-2">Once deleted, your project is gone. This cannot be undone.</div>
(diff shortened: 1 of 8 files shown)


hardillb:
@joepavitt can you review the vue in this please. I'm sure it's ugly as hell but it does work...

hardillb:
I'm starting to think I should disable the export (to JSON blob as that is covered in a different story not in the 0.5 release plan and there is nothing you can do with it yet)

frontend/src/pages/project/Settings/Danger.vue
```diff
@@ -38,6 +68,38 @@ export default {
...
+        },
+        showExportToProjectDialog () {
+            this.$refs.exportToProjectDialog.show(this.project)
+        },
+        exportProject (parts) {
+            // call projectApi to generate zipped json
+            projectApi.exportProject(this.project.id, parts)
+        },
+        duplicateProject (parts) {
+            console.log('copy project', parts)
```
joepavitt, line 85:
```suggestion
```

frontend/src/pages/project/Settings/Danger.vue
```diff
@@ -38,6 +68,38 @@ export default {
...
+            projectApi.create(parts).then(result => {
+                console.log(result)
```
joepavitt, line 87:
```suggestion
```

frontend/src/pages/project/Settings/Danger.vue
```diff
@@ -38,6 +68,38 @@ export default {
...
+                console.log(result.id)
```
joepavitt, line 88:
```suggestion
```

frontend/src/pages/project/Settings/Danger.vue
```diff
@@ -38,6 +68,38 @@ export default {
...
+                console.log(result.name)
```
joepavitt, line 89:
```suggestion
```

frontend/src/pages/project/Settings/Danger.vue
```diff
@@ -38,6 +68,38 @@ export default {
...
+                this.$router.push({ name: 'Project', params: { id: result.id } })
+            }).catch(err => {
+                console.log(err)
+            })
+        },
+        exportToProject (parts) {
+            console.log('Ben')
```
joepavitt, line 96:
```suggestion
```

frontend/src/pages/project/Settings/Danger.vue
```diff
@@ -38,6 +68,38 @@ export default {
...
+            console.log(parts)
```
joepavitt, line 97:
```suggestion
```

frontend/src/pages/project/Settings/dialogs/ExportToProjectDialog.vue
```diff
@@ -0,0 +1,190 @@
...
+                                    <p class="text-sm text-gray-500">
+                                        Select project components to copy to project:
+                                    </p>
+                                </div>
+                                <div>
+                                    <FormRow :error="errors.target" v-model="input.target" :options="projects">
+                                        <template v-slot:default>Target Project</template>
+                                    </FormRow>
+                                </div>
+<!--
```
joepavitt: Remove commented out code

frontend/src/pages/project/Settings/dialogs/ExportToProjectDialog.vue
```diff
@@ -0,0 +1,190 @@
...
+                }
+                isOpen.value = true
+                this.input.stack = this.project.stack.id
+                this.input.template = this.project.template.id
+                this.input.team = this.project.team.id
+                const projectList = await teamApi.getTeamProjects(this.project.team.id)
+                this.projects = []
+                for (let i = 0; i < projectList.count; i++) {
+                    if (this.project.id !== projectList.projects[i].id) {
+                        console.log(projectList.projects[i].id, projectList.projects[i].name)
```
joepavitt, line 174:
```suggestion
```

frontend/src/pages/project/components/ExportProjectComponents.vue
```diff
@@ -0,0 +1,54 @@
...
+        </FormRow>
+        <FormRow class="ml-9" v-if="parts.envVars" type="radio" v-model="parts.envVarsKo" :value="false">
+            Keys and Values
+            <!-- <template v-slot:description>Keys and Values</template> -->
+        </FormRow>
+        <FormRow class="ml-9" v-if="parts.envVars" type="radio" v-model="parts.envVarsKo" :value="true">
+            Keys only
+            <!-- <template v-slot:description>Keys only</template> -->
+        </FormRow>
+        <!-- <FormRow type="checkbox" v-model="parts.settings">Settings</FormRow> -->
```
joepavitt, line 27: Remove commented out code

frontend/src/pages/project/components/ExportProjectComponents.vue
```diff
@@ -0,0 +1,54 @@
...
+            Environment Variables
+            <template #description>Export Environment Variables</template>
+        </FormRow>
+        <FormRow class="ml-9" v-if="parts.envVars" type="radio" v-model="parts.envVarsKo" :value="false">
+            Keys and Values
+            <!-- <template v-slot:description>Keys and Values</template> -->
+        </FormRow>
+        <FormRow class="ml-9" v-if="parts.envVars" type="radio" v-model="parts.envVarsKo" :value="true">
+            Keys only
+            <!-- <template v-slot:description>Keys only</template> -->
```
joepavitt, line 25: Remove commented out code

joepavitt (changes requested):
I will likely modify a few of the dialogs shortly, once I proceed with the update to `forge-ui-components` dialog boxes across the platform.

Mostly looking good, but a lot of commented code and console logs that need clearing.

joepavitt (approved):

hardillb:
Still trying to work out how to clean up the overflow here:

Screenshot from 2022-05-05 16-06-21

And how to line up:

Screenshot from 2022-05-05 15-41-06

joepavitt:
> Still trying to work out how to clean up the overflow here

This is an issue with the existing FormRow. I'm honestly okay with pushing as is, with the knowledge that this will be cleaned by as part of the [Dialog Box rework](https://github.com/flowforge/flowforge/issues/520) I'm due to do in 0.5

knolleary:
#273 (export to zip) is not in plan for 0.5. We haven't done any design work around how the zip is structured. Given this PR implements something, can we get a description of what the zip contents is?

hardillb:
The JSON output is as follows:

```json
{
   "flows": [],
   "credentials": { "$": "b5db5a7f0aca0467b928136ff4c01e0flfGwBx7CcOrJHef+xp5StVrny+QwoLhn48aLck+gtIGXp012ujlkmNaTW2vJ0I8gyiAY06iZpbDb" },
  "envVars": {
    "FOO": "BAR",
    "TESTING": "ONE"
  },
  "nodes": {
    "node-red": {...},
    "node-red-node-base64": { ... }
  }
}
```

Where `flows` is the array of tabs/nodes and the `nodes` is the nodes section from the settings object. This was a bare minimum I thought we needed. Open to adding something about stacks and templates, but would need to decide what from each.

hardillb:
We can always remove the UI component for the export to JSON option if needed which will hide this option until we can iterate on it.

knolleary:
I would prefer we didn't expose this option today.

It will come in useful under the covers for the Device work - but that will also help us validate the format and the contents.

knolleary:
I find the reworked UI (buttons above text) to be top heavy and imbalanced. I think the buttons are too big.

I would prefer to get back to the original intent of having the button in a right-hand column - but know full well why that's a little awkward without a bigger refactor of FormRow.

@joepavitt would setting `size="small"` be inoffensive?

joepavitt:
> @joepavitt would setting `size="small"` be inoffensive?

No objection at all from me

frontend/src/pages/project/Settings/Danger.vue
```diff
@@ -1,23 +1,51 @@
...
+                project's current state.</div>
+        </div> -->
+
+        <div>
+            <div>
+                <ff-button kind="secondary" @click="showDuplicateProjectDialog()">Duplicate Project</ff-button>
+                <DuplicateProjectDialog @duplicateProject="duplicateProject" ref="duplicateProjectDialog"/>
+            </div>
+            <div class="max-w-sm pt-2 pl-5">Allows you to use a snapshot of the
+                project's current state to create a new project.</div>
```
knolleary, line 31:
```suggestion
            <div class="max-w-sm pt-2 pl-5">Create a copy of this project.</div>
```

frontend/src/pages/project/Settings/Danger.vue
```diff
@@ -1,23 +1,51 @@
...
+        </div>
+
+        <div>
+            <div>
+                <ff-button kind="secondary" @click="showExportToProjectDialog()">Export State</ff-button>
+                <ExportToProjectDialog @exportToProject="exportToProject" ref="exportToProjectDialog"/>
+            </div>
+            <div class="max-w-sm pt-2 pl-5">Copy a project's state to another existing project.</div>
```
knolleary, line 39:
```suggestion
                <ff-button kind="secondary" @click="showExportToProjectDialog()">Copy to existing project</ff-button>
                <ExportToProjectDialog @exportToProject="exportToProject" ref="exportToProjectDialog"/>
            </div>
            <div class="max-w-sm pt-2 pl-5">Copy the project's state to an existing project.</div>
```

knolleary (commented):
Some suggested tweaks to the labels/descriptions

knolleary:
Just tried the 'Duplicate project' - kept all the default options - and got this error in the runtime:
```
 req: {
      "method": "POST",
      "url": "/api/v1/projects",
      "hostname": "localhost:3000",
      "remoteAddress": "127.0.0.1",
      "remotePort": 55949
    }
    res: {
      "statusCode": 500
    }
    reqId: "req-s"
    err: {
      "type": "TypeError",
      "message": "Cannot read properties of null (reading 'flow')",
      "stack":
          TypeError: Cannot read properties of null (reading 'flow')
              at Object.<anonymous> (/Users/nol/code/flowforge/flowforge/forge/routes/api/project.js:208:39)
    }
```

Note - this is a NR Project that doesn't actually have any flows as I'm using the stub driver.

---

For the Duplicate project UI, I wonder if we'd be better off taking the user to the existing 'create project' page - but with an added 'copy content from...' section - rather than doing this all via a small dialog box.

Having just one view that handles creating projects would remove duplication in UI code. It would also, for example, ensure the 'agree to extra charges' checkbox is there when billing is enabled.

I am happy to tackle that rework if we think its worthwhile.

hardillb:
OK, I will ladd tests for missing flows/creds/settings to the project endpoints (e.g when using the stub driver, but also for a project that hasn't deployed yet)

No problem with moving to the create view, but how will it handle the choices of what to duplicate? Or will you pull in the new component into the create vue?

knolleary:
> Or will you pull in the new component into the create vue?

Yes - reduce, reuse, recycle.

hardillb:
Sounds good, just working out what to with the export to project, if the source project doesn't have the things it should be exporting. Should it just replace the existing with empty values?

PR #537, opened 2022-05-04
523  platform usage tracking

joepavitt:
Tracking now available via Plausible, and can be configured through the `flowforge.yml`:

Production example:

```
#################################################
# Telemetry Configuration                       #
#################################################
telemetry:
  enabled: true
  plausible:
    domain: <data-domain-here>
```

Testing locally:

```
#################################################
# Telemetry Configuration                       #
#################################################
telemetry:
  enabled: true
  plausible:
    domain: <data-domain-here>
    extension: local
```

### DIFF
--- a/docs/admin/telemetry.md
+++ b/docs/admin/telemetry.md
@@ -70,7 +70,11 @@ For the core tracking, the platform will send the telemetry data:
 
 The data is sent via an HTTP Post to `https://ping.flowforge.com`.
 
-## Plausible
+## Frontend Telemetry
+
+The FlowForge UI can be configured to track usage to help understand how users are navigating the pages.
+
+It currently supports using [Plausible Analytics](https://plausible.io/) to gather the information.
 
 ### Configuring Telemetry
 


hardillb:
Is the domain still required (by Plausible) when testing locally?

joepavitt:
Yes it is - as it'll report to the Plausible dashboard still, but flag as "local" stats.

docs/admin/README.md
```diff
@@ -8,7 +8,7 @@
  - [`flowforge.yml` configuration](../install/configuration.md) - base platform configuration, done before you run.
  - [First Run Setup](../install/first-run.md) - create your admin user
  - [FlowForge Concepts](../user/concepts.md)
- - [Usage Telemetry](./telemetry.md)
+ - [Usage Telemetry](/docs/admin/telemetry.md)
```
hardillb, line 11: Any reason to go to a fully qualified path here?

knolleary:
I think it'll be worth registering a separate domain in Plausible just for our local testing - using a subdomain of flowforge.com that we don't publicise. Whilst the `local` mode is useful, I wouldn't want anyone to accidentally start polluting the production stats through a misconfiguration.

docs/admin/telemetry.md
```diff
@@ -59,10 +61,31 @@ Property | Description
...
+### Configuring Telemetry
+
+Further details about how users are navigating around your instance of FlowForge can be enabled via [Plausible](https://plausible.io/). This can be enabled with:
+
+```
+telemetry:
+  enabled: true
+  plausible:
+    domain: <data-domain>
+```
```
hardillb, line 84: No mention of the `extension: local` but it is mentioned in the configuration.md

docs/install/configuration.md
```diff
@@ -60,8 +60,12 @@ See [here](./email_providers.md) for example configuration with common email pro
...
 
-For more information about this feature, see [here](../admin/telemetry.md)
+Additionally, you can configure your own instance of FlowForge to report back to you on how users are using FlowForge. FlowForge is designed to work with [Plausible](https://plausible.io/). You can setup your own account, and pass the releavnt domain to the `yml` in the telemetry configuration
+
+For more information about this feature, see [here](/docs/admin/telemetry.md)
 
 Option        | Description
 --------------|------------
 `telemetry.enabled` | Enables the anonymous usage telemetry of the platform. Default: `true`
+`telemetry.plausible.domain` | The `data-domain` of your site (see [Plausible docs](https://plausible.io/docs/plausible-script)). Default: `null`
```
hardillb, line 70: Can we take the domain value from the `base_url` value to save the user having to enter it twice?

hardillb (commented):
Also I think we should update the `etc/flowforge.yaml` file with an example in the comments

docs/admin/README.md
joepavitt: There shouldn't be - but if you checkout the "Usage Telemetry" link on the live site (https://flowforge.com/docs/admin/) - it doesn't work. Locally, this was the only way I could get it to run. It was very strange

docs/install/configuration.md
joepavitt: yes, although not sure I see why they're entering twice? Isn't it just the once in this config?

knolleary: @hardillb possibly - it depends what using `localhost:3000` for your base_url would do with Plausible - given they use the page referrer to identify where the data should be sent.

It may be we need to be able to override it

joepavitt:
> I think it'll be worth registering a separate domain in Plausible just for our local testing - using a subdomain of flowforge.com that we don't publicise. Whilst the `local` mode is useful, I wouldn't want anyone to accidentally start polluting the production stats through a misconfiguration.

This is fine by me, although (and this is my issue with Plausible). There is nothing stopping anybody using our actual domain with the `local` extension enabled, and polluting the stats. As could be done with any site using Plausible.

docs/install/configuration.md
hardillb: @joepavitt I meant that the user already has to enter a `base_url` value that is the main entry point for the site, we extract the hostname/domain from that rather than having to type it in a second time to setup to Plausible

knolleary:
The telemetry config now covers two different things - frontend and backend - and the config should make it clear they are different things.

I suggest adding a `frontend` (or another better named property) to make it clear the config only applies to the frontend telemetry.

```
telemetry:
  enabled: true
  frontend:
    plausible:
      domain: <data-domain-here>
      extension: local
```

docs/install/configuration.md
joepavitt: @hardillb that seems sensible. I do want to differentiate in the yml though the Plausible being enabled, separate from the general telemetry, so what would be the alternative config structure?

joepavitt:
> I suggest adding a `frontend` (or another better named property) to make it clear the config only applies to the frontend telemetry.

This is already reflected in the docs changes I've made:

Screenshot 2022-05-05 at 09 57 23

I can change the "Plausible" heading to say "Front End" instead.

docs/install/configuration.md
hardillb:
```yaml
telemetry:
  enabled: true
  frontend:
    plausible:
      enabled: true
      domain: <data-domain-here>
      extension: local
```

So that `plausible.domain` can be optional to override the one extracted from the `base_url` value (say the site gets moved, but wants to keep the data history)

hardillb: OK, that makes sense. I was just trying to reduce duplication in the config file.

forge/routes/ui/index.js
```diff
@@ -15,24 +15,42 @@
...
+     * Inject additional script imports for index.html that are dependant upon flowforge.yml
+     */
+    async function injectPlausible (domain, extension) {
+        const filepath = path.join(frontendAssetsDir, 'index.html')
+        const data = await fsp.readFile(filepath, 'utf8')
+        const injected = data.replace(/<!-- forge injected scripts go here -->/g,
+            `<script defer data-domain="${domain}" src="https://plausible.io/js/plausible${extension ? '.' + extension : ''}.js"></script>`)
+        return injected
+    }
+
```
knolleary, line 34: We only need to do the substitution once as the `flowforge.yml` config cannot change without a full restart. This proposed change caches the index.html in memory with the change applied.

```suggestion
    /**
     * Inject additional script imports for index.html that are dependant upon flowforge.yml
     */
    async function injectPlausible (domain, extension) {
        if (!cachedIndex) {
            const filepath = path.join(frontendAssetsDir, 'index.html')
            const data = await fsp.readFile(filepath, 'utf8')
            cachedIndex = data.replace(/<!-- forge injected scripts go here -->/g,
                `<script defer data-domain="${domain}" src="https://plausible.io/js/plausible${extension ? '.' + extension : ''}.js"></script>`)
        }
        return cachedIndex
    }

    let cachedIndex = null

```

docs/admin/telemetry.md
```diff
@@ -59,10 +61,44 @@ Property | Description
...
+### Schedule
+
+For the core tracking, the platform will send the telemetry data:
  - 30 seconds after the platform starts up (but only if the platform has already been initialised)
  - Once every 24 hours at a time randomly picked when the platform starts
 
 The data is sent via an HTTP Post to `https://ping.flowforge.com`.
+
+## Plausible
+
```
knolleary, line 74:
```suggestion
## Frontend Telemetry

The FlowForge UI can be configured to track usage to help understand how users are navigating the pages.

It currently supports using [Plausible Analytics](https://plausible.io/) to gather the information.

```

forge/routes/ui/index.js
knolleary: Should we host the plausible script ourselves rather than hardcode their CDN?

joepavitt: That is a _plausible_ option

hardillb (approved):

PR #547, opened 2022-05-09
Add project export and duplicate options

knolleary:
This PR includes the work from #528 and does a bit of UX rework.

### Create Project

The Create Project page could be reached via `/create` and `/team/:teamId/projects/create`. The former isn't used directly anywhere. This PR removes that path to access the view - it is now always done in the context of the selected Project.

As such, I've updated the page to include the current team sidebar:

image

### Duplicate Project

The Create Project page can take a query parameter - `sourceProject` - that identifies the id of the project that should be copied. When present (and validated), the UI includes the 'select components to export' ui:

image

### Export to project

The Export to existing project UI is still a dialog. Have included a checkbox confirmation as you don't want to accidentally do this to the wrong target project.

image

### Project Settings page

I've fixed up the Settings page back to the column approach, with proper alignment of the buttons:

image

### API

I have commented out the 'export' endpoint as it is unused currently.

Added unit tests for the Project Create end point that covers a lot of the duplicate project cases.

We don't yet have coverage of the Update Project end point.


frontend/src/pages/project/components/ExportProjectComponents.vue
```diff
@@ -1,44 +1,82 @@
...
             Secret
             <template #description>Provide a Secret to encrypt the exported Credentials</template>
         </FormRow>
-        <FormRow type="checkbox" v-model="parts.envVars">
-            Environment Variables
-            <template #description>Export Environment Variables</template>
-        </FormRow>
-        <FormRow class="ml-9" v-if="parts.envVars" type="radio" v-model="parts.envVarsKo" :value="false">
-            Keys and Values
+        <FormRow v-if="showSettings" type="checkbox" v-model="parts.settings">
```
hardillb, line 18: You've added "Project Settings" to this component, but there is no code in the api to do anything with `options.settings` that I can see

knolleary: Good point. Will fix that.

frontend/src/pages/project/Settings/dialogs/DuplicateProjectDialog.vue
```diff
@@ -0,0 +1,179 @@
+<template>
```
hardillb, line 1: Do we still need this Dialog now it's all in the CreateProject Vue?

knolleary:
Have removed the unused dialog.

PR #555, opened 2022-05-10
Allow admin to delete users

### LINKED ISSUE #338: Admin can delete a User
### Epic

_No response_

### Description

**As an admin**

**I want to delete a user account**

**So that they are unable to login to flowforge and all their node-red sessions are invalidated**

### Acceptance Criteria

- [ ] User is Removed from list of users
- [ ] Email address *can* be re-used on a new account
- [ ] User is logged out of any Node-RED Editor Sessions
- [ ] Teams & Projects belonging to the user are *not* removed

knolleary:
Closes #338 

This PR implements the runtime API updates to allow an admin user to delete another user.

The UI changes to support this option will be added shortly - until then, this is a draft PR.

Rules:
 - Only an admin can delete users
 - A user cannot be deleted if:
    - they own a team
    - they are an admin

This also deletes any pending team invitations for the user (either sent by, or sent to).

It does *not* delete any active Node-RED editor sessions - that requires some changes in Node-RED to support remove invalidation of sessions.

### DIFF
--- a/forge/db/models/TeamMember.js
+++ b/forge/db/models/TeamMember.js
@@ -33,6 +33,20 @@ module.exports = {
     finders: function (M) {
         return {
             static: {
+                getTeamsOwnedBy: async (userId) => {
+                    if (typeof userId === 'string') {
+                        userId = M.User.decodeHashid(userId)
+                    }
+                    return this.findAll({
+                        where: {
+                            UserId: userId,
+                            role: Roles.Owner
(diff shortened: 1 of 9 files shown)


knolleary:
Added the UI side to this.

To provide some slip-of-the-mouse protection, I've added locks to the danger buttons in the edit user dialog...

f0XmbJNHmi

PR #558, opened 2022-05-10
Allow Admin to limit what nodes can be installed

hardillb:
Part of https://github.com/flowforge/flowforge/issues/216

### DIFF
--- a/forge/routes/api/template.js
+++ b/forge/routes/api/template.js
@@ -28,6 +28,7 @@ module.exports = async function (app) {
      */
     app.get('/:templateId', async (request, reply) => {
         const template = await app.db.models.ProjectTemplate.byId(request.params.templateId)
+        console.log(template)
         if (template) {
             reply.send(app.db.views.ProjectTemplate.template(template))
         } else {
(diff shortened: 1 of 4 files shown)


forge/routes/api/template.js
```diff
@@ -28,6 +28,7 @@ module.exports = async function (app) {
      */
     app.get('/:templateId', async (request, reply) => {
         const template = await app.db.models.ProjectTemplate.byId(request.params.templateId)
+        console.log(template)
```
knolleary, line 31:
```suggestion
```

knolleary (approved):

forge/routes/api/template.js
hardillb: Doh, missed one

knolleary:
urg. I spot the console.log, I comment on the console.log, I propose the removal of the console.log..... I click merge anyway.

forge/routes/api/template.js
hardillb: Err, you've merged this before this was fixed...

knolleary: I am as disappointed about it as you are.

PR #561, opened 2022-05-10
change install-stack to node script for cross-dev

### LINKED ISSUE #560: Not possible to run `install-stack` npm script on windows
### Current Behavior

npm run install-stack --vers=2.2.2 results in ...

```bash
@flowforge/flowforge@0.4.0 install-stack
mkdir -p var/stacks/${npm_config_vers} && npm install --prefix var/stacks/${npm_config_vers} node-red@${npm_config_vers}

The syntax of the command is incorrect.
```

### Expected Behavior

To work on all supported environments

### Steps To Reproduce

from `flowforge` directory, run `npm run install-stack --vers=2.2.2` 

### Environment

- FlowForge version: 0.4.0 (git)
- Node.js version: 16
- Platform/OS: Windows 11

Steve-Mcl:
fixes #560

### DIFF
--- a/forge/install-stack.js
+++ b/forge/install-stack.js
@@ -1,0 +1,21 @@
+const { exec } = require('child_process')
+const { log, error } = require('console')
+const fs = require('fs')
+const path = require('path')
+const vers = process.env.npm_config_vers
+if (!vers) {
+    throw new Error('command line arg vars is missing')
+}
+const p = path.join('var', 'stacks', vers)
+fs.mkdirSync(p, { recursive: true })
+
+log(`installing stack node-red@${vers}`)
+const npmCmd = `npm install --prefix ${p} node-red@${vers}`
+
+exec(npmCmd, (err, stdout, stderr) => {
+    if (err) {
+        error(err)
+    } else {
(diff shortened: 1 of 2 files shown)


hardillb (approved):
I was meaning to do this at some point, so thanks. Looks good

knolleary:
Some changes are needed to this script.

1. It should be under `scripts` not `forge` as it isn't part of the core forge app.


2. it currently assumes to install under `./var/stacks/<name>/` ie, relative to cwd. That is only true in a dev environment. For a production environment using localfs, it *might* be `/opt/flowforge/var/stacks/...` - that said, in a prod environment, you're not going to be using `npm run ....` and nor do we want to be listing it the `bin` entry to be installed as a proper CLI tool in a production install. So it isn't immediately clear how a user would run it in prod.

 can see now why our docs currently give the manual steps for installing stacks.

I can see scope for a `flowforge-admin` CLI that pulls together different admin tasks... but that's another story.

hardillb:
@knolleary I'm halfway through fixing this, I ran into it while trying to update the docs

Issue #559, opened 2022-05-10
Rebrand: forgot your password page


knolleary:
<img width="465" alt="image" src="https://user-images.githubusercontent.com/51083/167680502-4d21e491-04eb-4b05-849c-e417863fc283.png">

Remember to enable 'allow users to reset their password' via Admin Settings in order to get to this page.

joepavitt:
<img width="918" alt="Screenshot 2022-05-11 at 09 33 01" src="https://user-images.githubusercontent.com/99246719/167805656-b49096d1-4153-4417-9dd0-b2a995fe9d48.png">

PR #564, opened 2022-05-11
Branding updates & fix update expired password routing after success

### LINKED ISSUE #557: Rebrand: password reset page
image

 - log in as admin
 - go to Admin Settings -> Users
 - edit the current user - click expire password (as of right this moment, you'll get an error in the forge app which I'm fixing atm, but the expiry will have happened)
 - click to navigate somewhere else in the app - you'll get prompted to update your password

joepavitt:
Closes #557 and #559

### DIFF
--- a/frontend/src/components/auth/UpdateExpiredPassword.vue
+++ b/frontend/src/components/auth/UpdateExpiredPassword.vue
@@ -56,7 +56,6 @@ export default {
                 return false
             }
             userApi.changePassword(this.input.old_password, this.input.password).then(() => {
-                console.log('DONE!')
                 this.$store.dispatch('account/checkState')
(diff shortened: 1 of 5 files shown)


frontend/src/components/auth/UpdateExpiredPassword.vue
```diff
@@ -59,6 +58,7 @@ export default {
             userApi.changePassword(this.input.old_password, this.input.password).then(() => {
                 console.log('DONE!')
```
knolleary, line 59:
```suggestion
```
Drive-by console.log removal

knolleary:
Tested the various flows locally - all looking good.

PR #588, opened 2022-05-16
Add missing steps to dev:local

Steve-Mcl:
As discussed on slack: https://flowforgeworkspace.slack.com/archives/C032Q63FGG1/p1652345077573039 

In short: adds the following to the end of `dev:local`...

```bash
npm i ../forge-ui-components 
npm run build
npm i ../flowforge-nr-storage
npm i ../flowforge-nr-audit-logger
npm i ../flowforge-nr-auth
```

This ensures symlinked packages have a `node_modules` folder and `forge-ui-components` is built

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
         "cover:unit": "nyc --silent npm run test:unit",
         "cover:system": "nyc --silent --no-clean npm run test:system",
         "cover:report": "nyc report --reporter=text --reporter html",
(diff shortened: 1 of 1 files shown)


knolleary (approved):
Tentatively approving so we can keep things moving forward.

At this stage, I'd prefer to have a script in the `scripts` directory that did the work, rather than end up with hard to maintain, super-long task definitions.

PR #604, opened 2022-05-18
Auto generate credential secret for NR project to remove warnings in log

Steve-Mcl:
closes This https://github.com/flowforge/flowforge-nr-launcher/issues/16

NOTES This PR is linked with flowforge-nr-launcher [16-auto-gen-cred-secret](https://github.com/flowforge/flowforge-nr-launcher/tree/16-auto-gen-cred-secret) branch & should be reviewed/merged as one

### DIFF
--- a/forge/db/controllers/ProjectTemplate.js
+++ b/forge/db/controllers/ProjectTemplate.js
@@ -1,4 +1,5 @@
 const validSettings = [
+    'credentialSecret',
     'disableEditor',
     'httpAdminRoot',
     'codeEditor',
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -206,6 +206,7 @@ module.exports = async function (app) {
                 newSettings.nodes = sourceSettings.nodes
             }
             const options = request.body.sourceProject.options
+            const newCredentialSecret = generateCredentialSecret()
             if (options.flows) {
                 const sourceFlows = await app.db.models.StorageFlow.byProject(sourceProject.id)
                 if (sourceFlows) {
@@ -225,18 +226,19 @@ module.exports = async function (app) {
                     const origCredentials = await app.db.models.StorageCredentials.byProject(sourceProject.id)
                     if (origCredentials) {
                         // There are existing credentials to copy
-                        const sourceCreds = JSON.parse(origCredentials.credentials)
-                        const newKey = crypto.randomBytes(32).toString('hex')
-                        const newCreds = recryptCreds(sourceCreds, sourceSettings._credentialSecret, newKey)
-                        const creds = await app.db.models.StorageCredentials.create({
-                            credentials: JSON.stringify(newCreds),
+                        const srcCredentials = JSON.parse(origCredentials.credentials)
+                        const srcCredentialSecret = await sourceProject.getSetting('credentialSecret') || sourceSettings._credentialSecret
+                        const newCredentials = recryptCreds(srcCredentials, srcCredentialSecret, newCredentialSecret)
+                        const credentials = await app.db.models.StorageCredentials.create({
+                            credentials: JSON.stringify(newCredentials),
                             ProjectId: project.id
                         })
-                        await creds.save()
-                        newSettings._credentialSecret = newKey
+                        await credentials.save()
+                        newSettings._credentialSecret = newCredentialSecret
                     }
                 }
             }
+            await project.updateSetting('credentialSecret', newCredentialSecret)
             const settings = await app.db.models.StorageSettings.create({
                 settings: JSON.stringify(newSettings),
                 ProjectId: project.id
@@ -256,8 +258,9 @@ module.exports = async function (app) {
                     })
                 })
             }
-
             await project.updateSetting('settings', newProjectSettings)
+        } else {
+            await project.updateSetting('credentialSecret', generateCredentialSecret())
         }
 
         await app.containers.start(project)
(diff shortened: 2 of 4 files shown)


forge/db/controllers/ProjectTemplate.js
```diff
@@ -1,4 +1,5 @@
 const validSettings = [
+    'credentialSecret',
```
knolleary, line 2: This isn't needed - we aren't adding `credentialSecret` to the settings a Template can provide. It is being stored separately in ProjectSettings.

test/unit/forge/db/controllers/ProjectTemplate_spec.js
```diff
@@ -17,6 +17,7 @@ describe('Project Template controller', function () {
     describe('Validate Settings', function () {
         it('only allows settings from the known list', async function () {
             const result = app.db.controllers.ProjectTemplate.validateSettings({
+                credentialSecret: 'aaaaaaaabbbbbbbbccccccccdddddddd',
```
knolleary, line 20: This isn't quite the right set of tests - this is testing providing the credentialSecret via Template settings - which is different to being provided via ProjectSettings.

knolleary (commented):
This doesn't quite meeting the design described in the original story.

We do not want to include `credentialSecret` via the Template settings - it should not be accessible via the `/settings` endpoint.

The secret should be passed to the launcher as an env var (or command-line argument) in the same way we pass in the project token and the like.

To do that will require some changes to the drivers to ensure they set the right thing when running the launcher.

Steve-Mcl:
> The secret should be passed to the launcher as an env var (or command-line argument) in the same way we pass in the project token and the like.

> We do not want to include credentialSecret via the Template settings

@knolleary Can i clarify please - we do NOT want `credentialSecret` written to the project folders `settings.js` (asking because the `projectToken` is written to the `settings.js` : https://github.com/flowforge/flowforge-nr-launcher/blob/d6019f895b509e037c18559464f2e8a7d04f6f41/lib/runtimeSettings.js#L65 ) - im probably bundling the statements in to 1

knolleary:
The launcher does need to write the credentailSecret to node-red's settings.js file - that's the only means to pass it in.

The matter here is how we get the secret to the launcher so that it can write it to the file.

forge/routes/api/project.js
```diff
@@ -688,6 +691,9 @@ module.exports = async function (app) {
      * @name /api/v1/project/:id/export
      * @memberof forge.routes.api.project
      */
+    /*
+       @Steve-Mcl @hardillb "will need updating to check if there is a credentialSecret in the ProjectSettings tables first and only look in the StorageSettings entry if it's not found"
+    */
```
knolleary, line 696: With the work in #603, this particular piece of code has moved out of here - and the code in #603 will need updating to take into account this logic.

Rather than add an inline todo comment, this should be added as a comment to the other PR so it is implemented before that PR is merged.
```suggestion
```

forge/routes/api/project.js
```diff
@@ -225,18 +226,19 @@ module.exports = async function (app) {
...
+                        const srcCredentialSecret = await sourceProject.getSetting('credentialSecret') || sourceSettings._credentialSecret
+                        const newCredentials = recryptCreds(srcCredentials, srcCredentialSecret, newCredentialSecret)
+                        const credentials = await app.db.models.StorageCredentials.create({
+                            credentials: JSON.stringify(newCredentials),
                             ProjectId: project.id
                         })
-                        await creds.save()
-                        newSettings._credentialSecret = newKey
+                        await credentials.save()
+                        newSettings._credentialSecret = newCredentialSecret
```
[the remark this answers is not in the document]
knolleary: In fact, we should explicitly delete `newSettings._credentialSecret` just to ensure the value isn't hanging around from a legacy project that got it set.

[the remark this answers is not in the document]
Steve-Mcl: I left it in for loading & copying legacy projects (where the new setting would be empty) but if that's the direction I can remove it throughout all projects & ensure in place upgrade occurs where found?

forge/routes/api/project.js
Steve-Mcl: NP. I will do.  However, before I do,  Ben had asked me to put that comment in as there was something he needed to do in this area (so was a reminder to him)

@hardillb before I remove this comment, could you let me know if I should also remove the existing (commented out code) below it?

Thanks.

knolleary: @Steve-Mcl the commented out code gets removed in #603 - please don't touch it otherwise we will get unnecessary merge conflicts to sort out.

Steve-Mcl: Ah. I should have looked. Np.  Sorry for the noise.

forge/routes/api/project.js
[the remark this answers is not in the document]
knolleary: This code path is specifically for creating a new project (which should get a new credentialSecret generated and stored in the new setting), copying content from an existing project (which might have the new credentialSecret, or might have the legacy _credentialSecret).

The *new* project must end up with the new credentialSecret setting and not the legacy _credentialSecret. If you leave both in place, then the runtime will end up thinking it has to migrate things - which isn't needed.

There is an entirely separate discussion to be had about if we want to migrate *existing* projects. Whether we migrate or not - it is vital that we have tested this with existing projects.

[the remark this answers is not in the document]
Steve-Mcl: Understand. I'll fix this up first thing in morning.

knolleary (approved):

PR #613, opened 2022-05-23
Initial theming support

Steve-Mcl:
Adds elements to support basic theming. This PR works in tandem with https://github.com/flowforge/flowforge-nr-launcher/pull/37

<br>

_UPDATED 21-06-2022_

<br>

## foreword
At some point, we will want to include the installation of `@flowforge/nr-theme` alongside `nr-audit-logger`, `nr-auth` & `nr-storage` into `flowforge-nr-launcher/node_modules/@flowforge`

## Implementation Details...

Updates to include minimal theme related settings

#### forge/db/controllers/ProjectTemplate.js
* theme
* page_title
* page_favicon
* header_title
* header_url
* nodesDir

#### frontend/src/pages/admin/Template/sections/Editor.vue
1 new fields:
* editable.settings.header_title == Editor Title
 
#### frontend/src/pages/admin/Template/utils.js
* theme: Defaults to 'forge-light'
* page_title: Defaults to 'FlowForge'
* page_favicon: Defaults to  ''
* header_title: Defaults to 'FlowForge'
* header_url: - defaults to ''
* nodesDir: not configurable at this time - defaults to `{launcher}/node_modules/@flowforge/nr-theme`

## Screenshots...
image

image


knolleary:
> Install the theme to each stack using npm install @flowforge/flowforge-nr-theme

We don't require the admin to install our other `flowforge-nr-*` modules in the stack - they are installed as dependencies of the launcher. This should be no different to those.

knolleary:
Just as I posted that, I realised the issue what that approach - the theme is implemented as a plugin that NR has to load for itself, rather than a module the settings file can require directly.

We need to reconsider that approach. I think requiring the admin to install our code into the stack, rather than just being a vanilla NR install, complicates things.

At the very least, we would need a PR that updates all the doc/scripts we provide for creating stacks to include this extra step... but I'm not sure its the right approach.

Steve-Mcl:
What about updating the `install-stack` to also include this package?

knolleary:
Yes, that is one of the things I was referring to about the docs/script updates we need in place to do this.

Steve-Mcl:
> I would like to find a way to let the nr-launcher apply the theme without requiring it to be installed in the stack - at the least, for our core/default theme.

Only think popping into my head right now is to check for & if necessary generate sym links to the package - perhaps on FF launch? or as a post install npm script?

Do you have any ideas to explore Nick?

joepavitt:
This is looking great @Steve-Mcl. Couple of minor things I'd like to try if we can:

- No capital letters on the project name (and reduce the font size). Maybe even bolden it up. The full caps works because it's a single word (and company branding), but not for a project name.
- Not sure the dark background works on the actual node canvas. Something lighter would be more comfortable to the eye, and consistent with the FlowForge UI too?

Steve-Mcl:
Thank you @joepavitt 


Forgive me for this - It was always my hope you (master jedi-ui) would be picking this up once it was in place 👼 🤣 

I will of course tweak any points you raise to get it "ready" to go. 

> No capital letters on the project name (and reduce the font size). Maybe even bolden it up. The full caps works because it's a single word (and company branding), but not for a project name.

Can you suggest CSS please? It is currently set here: https://github.com/flowforge/flowforge-nr-theme/blob/79166faef623fc1a6456cbf3a6d2558ffad533a1/common/forge-common.css#L55-L59  - you might notice I have changed the CSS already from `text-transform: uppercase` to `text-transform: capitalize` 


> Not sure the dark background works on the actual node canvas. Something lighter would be more comfortable to the eye, and consistent with the FlowForge UI too

Can you suggest a color value please? I simply sampled the flowforge header & applied a lightening/darkening factors to it for various parts of the editor for consistency / as a starting point - but I see what you mean about the workspace background being darker than the FF general background.

Steve-Mcl:
@joepavitt 

> How do I get this up and running locally?

For the time being (until I figure out how to better integrate this) you need to...
* switch your local env to branch `add-theming` 
  * on projects `flowforge-nr-launcher` and `flowforge`
* git clone `flowforge-nr-theme` locally (into same parent dir as `flowforge-nr-launcher` and `flowforge`)
* run `npm i ../../../flowforge-nr-theme` in your stacks  e.g... 
  ```
  cd var/stacks/2.2.2
  npm i ../../../flowforge-nr-theme
  ```

I am more than happy to assist over a zoom/huddle - would be good to get a 2nd set of eyes

knolleary:
> Not sure the dark background works on the actual node canvas. Something lighter would be more comfortable to the eye, and consistent with the FlowForge UI too?

Just to clarify as it isn't spelt out explicitly, there are two separate themes here - `forge-light` and `forge-dark`. The two screenshots above show the two themes in action. The default is the `forge-light` theme which is much closer to what the core platform looks like. The `dark` theme is the one I think you're commenting on - but as a dark theme, it is intended to have a dark canvas. This is consistent with the other 3rd party dark themes that are available in the community.

joepavitt:
Have just pulled in master, what do I need to do in order to activate/test this? Not seeing any sign of it just restarting...

PR #615, opened 2022-05-23
Add active/target snapshot properties to Devices

knolleary:
This PR is targeting the `snapshot-ui` branch. This PR should be merged before #608 is merged

It adds two properties to the Device model
 - `activeSnapshot` - the id of the ProjectSnapshot the device is currently running
 - `targetSnapshot` - the id of the ProjectSnapshot the device *should* be running

It adds a new `ProjectSetting` called `deviceSettings` that is a JSON blob of configuration. Currently that contains a single item - `targetSnapshot`.

This is managed via the new API end point:

 - `GET/POST /api/v1/projects/:id/devices/settings`

When a device is assign to a Project, its `targetSnapshot` property is set to whatever the Project's `deviceSettings.targetSnapshot` property is set to.

When a device is removed from a Project, its `targetSnapshot` property is set to `null`.

When a Project's `targetSnapshot` is changed, all of the devices assigned to the Project have their `targetSnapshot` updated to match.

Whilst there is some redundancy here (having `targetSnapshot` on both the Project *and* the Device) it would allow a future model where a project is more selectively rolled out to its devices.

### UI

This is a bit rough and ready to get things moving forward.

#### Project Device table

Shows what the active/target snapshots for each device:

image

#### Project Snapshot table

If Device feature is enabled, shows which snapshot is the current target.

**TODO**: the 'device' column is not yet implemented - will be a count of devices running that snapshot.

image

The option menu for each snapshot has a 'set as device target' option, with accompanying confirmation dialog

image

image


hardillb (approved):
Initial review looks good. Will use is as base for pulling stuff with the device agent

PR #618, opened 2022-05-23
Add live device endpoints

knolleary:
This builds on top of the `device-snapshot` branch and PR #615 

This adds new API end points for devices to use to report their status and access the project snapshot they should be running.

 - `POST /api/v1/devices/:id/live/state` - report state from device
 - `GET /api/v1/devices/:id/live/snapshot` - get the active project snapshot for the device

#### `POST /api/v1/devices/:id/live/state` 

The payload should look like:
```
{
   snapshot: "xyz"
}
```

This should be the id of the snapshot the device is running. If the device is not yet configured to run a snapshot, it should either be absent or `null`.

The status code response will be:
 - `200` - all is good
 - `409` - the snapshot is incorrect. The response payload will look like:
    ```
    { "error":"incorrect-snapshot", "snapshot":"ABC" }
    ```

#### `GET /api/v1/devices/:id/live/snapshot`

This will return the active snapshot for the device - if there is one.

The format of the payload matches that defined here: https://github.com/flowforge/flowforge/issues/587#issuecomment-1129775673 - with the addition of an `id` property - which is what the device should store locally and use in post requests to the `live/state` end point.

## Remaining work

 - Tests - getting this PR raised early to unblock the edge agent work
 - Token security - currently the live end points are not restricted to device tokens. They need to be locked down.
    - On that topic, also need to verify the scope of project/device tokens generally - to ensure they don't allow cross-project/device access.

 -


hardillb:
We need to send the credentialSecret to the Device Agent.

Can I suggest we encrypt it with the device token so it can safely be downloaded?

knolleary:
> The GET to /api/v1/devices/:id/live/snapshot doesn't include the snapshot id

It will when I push the commit I must have forgotten to push....

> We need to send the credentialSecret to the Device Agent.

I was waiting to land the other credentialSecret PR to ensure some consistency.

I think this should be generated and passed over as part of the token/credentials object the user is given to manually apply to the device, rather than include it in the API response.

knolleary:
I've updated this PR to target `main`. So this PR will replace all of the others in the chain.

It now adds a `credentialSecret` column to the Device table, and generates that alongside the access token:

image

The Device snapshot's `credentials` property is encrypted using the device key.

There are some more tests to write - but I'm inclined to get this merged now so we can keep moving forward this week.

I also want to refactor the credential encrypt/decrypt code - we currently have it copy and pasted in more than one place. Would be much better at this point to have it centralised - with some more targeted tests.

hardillb:
Do you want to fix the 2 failed tests first?

knolleary:
> Do you want to fix the 2 failed tests first?

Already have :)

In terms of merging - whilst this targets main, I would like us to not use this PR to close the others... we should merge the other PRs in order:
 1. #615 
 2. #608 
 3. Then marge #603 as ready - and merge it.

Once all that is done, this can be merged.

hardillb (changes requested):
We have multiple copies of the credentials de/en/reencrypt functions now, one in `forge/routes/api/project.js` and one in `forge/db/controllers/Project.js`

We should probably just use the version in db/controllers everywhere.

But that and dropping the console.logs looks ok so far.

hardillb (approved):

PR #630, opened 2022-05-25
Add exportCredentials function on Project controller to handle re-encryption

knolleary:
This PR adds `app.db.controllers.Project.exportCredentials` that can be used to re-encrypt a credentials object.

This removes the duplicate code in this area.

There are no tests to add as this functionality is already covered by existing tests - this is an internal refactoring.


hardillb (approved):

Steve-Mcl (approved):
Makes sense.

PR #638, opened 2022-05-27
Add forgeURL to downloaded credentials file

hardillb:
This includes the `app.config.base_url` in the response to `/api/v1/settings` GET

And then includes it in the generated device credentials file.

Should we show it with details in the dialog as well?

### DIFF
--- a/frontend/src/pages/team/Devices/dialogs/DeviceCredentialsDialog.vue
+++ b/frontend/src/pages/team/Devices/dialogs/DeviceCredentialsDialog.vue
@@ -41,6 +41,7 @@
 
 import { ref } from 'vue'
 import deviceApi from '@/api/devices'
+import settings from '@/api/settings'
 
 import FormRow from '@/components/FormRow'
 import { DocumentDownloadIcon } from '@heroicons/vue/outline'
@@ -55,6 +56,9 @@ export default {
         return {
             device: null
         }
+    },
+    async mounted () {
+        this.base_url = (await settings.getSettings()).base_url
     },
     methods: {
         downloadCredentials () {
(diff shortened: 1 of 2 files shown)


frontend/src/pages/team/Devices/dialogs/DeviceCredentialsDialog.vue
```diff
@@ -56,6 +57,9 @@ export default {
             device: null
         }
     },
+    async mounted () {
+        this.base_url = (await settings.getSettings()).base_url
```
knolleary, line 61: The platform settings are preloaded in the Vue store, so you don't need to get them again.

You can use `mapState` to add `settings` as a local computed property, and then you should be able to just use `this.settings.base_url`.

See https://github.com/flowforge/flowforge/blob/main/frontend/src/components/auth/Credentials.vue#L28 for a mapState example that adds `settings`.

hardillb: Updated, (Not 100% sure why it needs to be `mapState('account',['settings')` e.g. why it needs the `account` at the start but it works....

frontend/src/pages/team/Devices/dialogs/DeviceCredentialsDialog.vue
```diff
@@ -76,14 +77,18 @@ export default {
...
     },
     computed: {
+        ...mapState('account', ['settings']),
         hasCredentials: function () {
             return this.device && this.device.credentials
         },
         credentials: function () {
             if (this.device) {
+                console.log(this)
+                console.log(this.settings)
```
hardillb, line 87:
```suggestion
```

knolleary (approved):

Issue #576, opened 2022-05-13
Upgrading individual projects between FF releases


knolleary:
### Description

When we roll out a new version of FF to our cloud, that will include a new version of the `flowforge-nr-launcher` component. This is the executable that wraps Node-RED and provides the interface between the core FF platform and the Node-RED instance.

To do a full and complete upgrade requires the projects to be stopped and restarted. That is not something we should do as part of an admin upgrade - our goal should be for zero downtime for user projects.

Given that, it means we will end up with the core FF platform running version X, and user projects running version X-1, X-2, ....

This has some important consequences we haven't properly spelt out previously - and drives some requires to improve the UX in this area.

- We must not make any changes to the existing API used by the current launcher. `Launcher@X-1` must be able to work with `Forge@X`.
  - That may mean some new features in version X are not available in the Project until the Project is updated
- Users should be notified if there is an update available and encouraged to upgrade when it is convenient
- Admins must be able to know what Launcher version each project is running on

This drives a number of requirements:

 - [x] https://github.com/flowforge/flowforge-nr-launcher/pull/39
   The Launcher `/flowforge/info` HTTP end point must include launcher version information so the platform knows what it is running
 - [x] https://github.com/flowforge/flowforge/pull/642
 - [ ] Add a notification if the launcher is downlevel - *blocked by* needing a way for the platform to know what the current expected level is (see below)
 - [ ] Provide a way for the user to trigger an update of the launcher
   The platform should provide a way to upgrade the launcher (which is different to restarting Node-RED itself). I think it will be sufficient to suspend/resume the project - each driver does it differently, but that will stop the process/container and restart it pulling the new one.

*Blocker*: how should the platform identify what is the current level of the launcher. We can compare the platform version against the launcher version - that will work for 0.x level differences. But we take the approach that the launcher and platform could release fix releases (0.x.**y**) independently.

knolleary:
There is an added complication to this piece - the way to upgrade the launcher varies between the different drivers. For localfs, a suspend/resume is sufficient as that restarts the launcher process - and will pickup the new version on resume.

For docker/k8s however, the launcher is baked into the container that is determined by the stack. To upgrade the launcher requires a change of container, which means the admin must first create a new stack and then the user needs to know to update their stack.

These are two fundamentally different approaches that require different actions by admin and user to achieve the same end result.

Give k8s is the strategic driver we need to prioritise that scenario.

This implies we need to introduce versioning of stacks:

[...]

PR #642, opened 2022-05-30
Add project version info to Project overview

knolleary:
Part of #576 

This adds version information to the Project overview page - if the launcher provides the information.

image

The next task for #576 is to include a notification if the Launcher version is down-level and a restart is needed - we'll add that inline to this version table (in absence of a wider notification system yet)

### DIFF
--- a/frontend/src/pages/project/Overview.vue
+++ b/frontend/src/pages/project/Overview.vue
@@ -6,16 +6,30 @@
 
                 <table class="table-fixed w-full">
                     <tr class="border-b">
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):
All good. Have we captured the " if the Launcher version is down-level and a restart is needed - we'll add that inline to this version table (in absence of a wider notification system yet)" bit elsewhere?

knolleary:
> Have we captured the "...."

Yes - its an item listed under #576, and needs turning into a proper story/task/pr.

Issue #40, opened 2021-08-16
Story: Delete a node-red instance


knolleary:
### Description

**As a:** User

**I want to:** Delete a node-red instance

**So that:** I can remove all trace of the instance

### Acceptance Criteria

 - [ ] UI to delete instance
 - [ ] API to delete instance
 - [ ] Container implementation to delete instance

PR #643, opened 2022-05-31
Map FlowForge logout to nodered auth/revoke

### LINKED ISSUE #442: Logging out of FF should logout of active NR sessions
### Epic

_No response_

### Description

**As a:** user

**I want:** to log out of the FF platform knowing it has also logged me out of any NR editor I was logged into

**So that:** I can be confident that I don't accidentally stay logged in somewhere

---

We currently store login sessions in the FF database (via the NR storage API). When a user logs out of FF, we should be able to identify any sessions in that table for them and their projects.

As NR caches sessions in memory, just updating the database won't do anything until NR is restarted. Instead we would have to post to the NR admin api (`/auth/revoke`) so that it revokes the token, both in memory and in the DB.

For any open editors, the background heartbeat on the websocket connection will cause it to discover the session is invalid and the login dialog shown. That does mean a short delay (up to 30s iirc) before it notices the editor session is invalid (If the user tries to deploy changes etc it will notice immediately).

### Acceptance Criteria

Three scenarios:
 1. Have the editor open in a tab
    - Log out of FF
    - Wait to see the login dialog appear within 30s
 2. Have editor open in a tab
   - Log out of FF
   - Try to deploy a change - it should be rejected
 3. Log into an editor, then close the tab (but copy the url)
   - Log out of FF
   - Reopen the editor - should be prompted to login

Steve-Mcl:
fixes #442 

## NOTE
NOTE: this PR is to be reviewed and merged as 1 of 3 parts. 

1. https://github.com/flowforge/flowforge/pull/643
1. https://github.com/flowforge/flowforge-driver-localfs/pull/48
1. https://github.com/flowforge/flowforge-nr-launcher/pull/40

Also related...
1. https://github.com/flowforge/flowforge-driver-docker/pull/24
1. https://github.com/flowforge/flowforge-driver-k8s/pull/39

## Changes
* Adds code to endpoint `/account/logout` in forge/routes/auth/index.js (STEP 1) which does the following...
  * Gets the `username` from `app.db.models.User.byId(userId)`
  * Gets the storage sessions from `app.db.models.StorageSession.byUsername(user.username)`
  * Iterates each session and calls `app.containers.logoutNodeRED(project, token)` (STEP 2)
* Adds `logoutNodeRED()` to forge/containers/wrapper.js  (STEP 2)
  * This then calls to the driver `this._driver.logoutNodeRED(project, token)` (STEP 3)
* Refer to [localfs driver #48](https://github.com/flowforge/flowforge-driver-localfs/pull/48) for step 3 in this process
* Refer to [launcher #40](https://github.com/flowforge/flowforge-nr-launcher/pull/40) for steps 4 & 5 in this process

### DIFF
--- a/forge/containers/wrapper.js
+++ b/forge/containers/wrapper.js
@@ -182,7 +182,7 @@ module.exports = {
         }
     },
     revokeUserToken: async (project, token) => { // logout:nodered(step-2)
-        if (this._driver.logoutNodeRED) {
+        if (this._driver.revokeUserToken) {
             await this._driver.revokeUserToken(project, token) // logout:nodered(step-3)
         }
     },
--- a/forge/containers/wrapper.js
+++ b/forge/containers/wrapper.js
@@ -183,7 +183,7 @@ module.exports = {
     },
     revokeUserToken: async (project, token) => { // logout:nodered(step-2)
         if (this._driver.logoutNodeRED) {
-            await this._driver.logoutNodeRED(project, token) // logout:nodered(step-3)
+            await this._driver.revokeUserToken(project, token) // logout:nodered(step-3)
         }
     },
     logs: async (project) => {
--- a/forge/containers/wrapper.js
+++ b/forge/containers/wrapper.js
@@ -181,7 +181,7 @@ module.exports = {
             this._driver.restartFlows(project, options)
         }
     },
-    logoutNodeRED: async (project, token) => { // logout:nodered(step-2)
+    revokeUserToken: async (project, token) => { // logout:nodered(step-2)
         if (this._driver.logoutNodeRED) {
             await this._driver.logoutNodeRED(project, token) // logout:nodered(step-3)
         }
--- a/forge/containers/wrapper.js
+++ b/forge/containers/wrapper.js
@@ -181,6 +181,11 @@ module.exports = {
             this._driver.restartFlows(project, options)
         }
     },
+    logoutNodeRED: async (project, token) => { // logout:nodered(step-2)
+        if (this._driver.logoutNodeRED) {
+            await this._driver.logoutNodeRED(project, token) // logout:nodered(step-3)
+        }
+    },
     logs: async (project) => {
         let value = []
         if (this._driver.logs) {
--- a/forge/db/models/StorageSession.js
+++ b/forge/db/models/StorageSession.js
@@ -35,7 +35,7 @@ module.exports = {
                         const sessions = Object.values(session) || []
                         const usersSessions = sessions.filter(e => e.user === username && e.client === 'node-red-editor')
                         return { ProjectId: m.ProjectId, sessions: usersSessions }
-                    }).filter(m => m.sessions.length > 0)
+                    }).filter(m => m.sessions.length > 0 && M.ProjectId)
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


knolleary:
We will also need PRs on the docker/k8s drivers. The implementation will be near identical to the localfs driver change - just make sure you follow the pattern used by the individual drivers to send commands to their launcher.

There is a lot to be said for pulling some of these command actions up into the wrapper level - so we don't end up with near identical code in each driver. But that isn't a refactoring we should consider today.

forge/db/models/StorageSession.js
```diff
@@ -20,6 +20,15 @@ module.exports = {
                         where: { ProjectId: project },
                         attributes: ['id', 'sessions']
                     })
+                },
+                byUsername: async (username) => {
+                    const allStorageSessions = await this.findAll({ attributes: ['sessions', 'ProjectId'] })
```
knolleary, line 25: This won't scale very well when there are a large number of projects.

There are two optimisations of differing complexity:

1. get a list of projects the user has access to, then only get the storage sessions for those projects.
    That is actually quite a lot of DB work

2. optimise this findAll with a `where` query that `sessions` contains the text `username`. That will reduce the amount of data that has to be retrieved from the database in the one query.

Steve-Mcl: I took the path of least resistance as I suspected it would lead to discussion - possibly ending up on a change to DB schema that would scale much better?  However, I bet that is something we want to avoid unless there is no other solution?  I will implement No2 (unless you ping me otherwise? :) )

knolleary:
Please do the PRs. They need to happen and right now you're in the best position to do them. We can test them once the code is written.

forge/containers/wrapper.js
```diff
@@ -181,6 +181,11 @@ module.exports = {
             this._driver.restartFlows(project, options)
         }
     },
+    logoutNodeRED: async (project, token) => { // logout:nodered(step-2)
```
knolleary, line 184:
```suggestion
    revokeUserToken: async (project, token) => { // logout:nodered(step-2)
```

forge/containers/wrapper.js
```diff
@@ -181,6 +181,11 @@ module.exports = {
...
+        if (this._driver.logoutNodeRED) {
+            await this._driver.logoutNodeRED(project, token) // logout:nodered(step-3)
```
knolleary, line 186:
```suggestion
            await this._driver.revokeUserToken(project, token) // logout:nodered(step-3)
```

forge/routes/auth/index.js
```diff
@@ -196,6 +196,29 @@ module.exports = fp(async function (app, opts, done) {
...
+                if (userId != null) {
+                    const user = await app.db.models.User.byId(userId)
+                    const sessions = await app.db.models.StorageSession.byUsername(user.username)
+                    for (let index = 0; index < sessions.length; index++) {
+                        const session = sessions[index]
+                        const ProjectId = session.ProjectId
+                        const project = await app.db.models.Project.byId(ProjectId)
+                        for (let index = 0; index < session.sessions.length; index++) {
+                            const token = session.sessions[index].accessToken
+                            await app.containers.logoutNodeRED(project, token) // logout:nodered(step-2)
```
knolleary, line 215:
```suggestion
                            await app.containers.revokeUserToken(project, token) // logout:nodered(step-2)
```

knolleary (changes requested):
Looks good.

Only minor(ish) thing - I suggest the API is renamed `revokeUserToken`.

I've made suggested changes along those lines that you will be able to batch commit. I'll do that across the whole family of PRs.

knolleary (approved):

PR #644, opened 2022-05-31
Fix device live endpoint to handle null snapshot

knolleary:
If a device has no active snapshot locally, it reports `null`. This was being compared against `undefined` - and failing, making it think it had an update to do.

This PR fixes it so we compare `null` with `null` in that case - so we properly recognise it's in the right state.

Also fixes the token handling to report `401` for invalid tokens, rather than a generic 500.

### DIFF
--- a/forge/routes/auth/index.js
+++ b/forge/routes/auth/index.js
@@ -79,7 +79,6 @@ module.exports = fp(async function (app, opts, done) {
                 await verifyToken(request, reply)
             } else if (!request.context.config.allowAnonymous) {
                 reply.code(401).send({ error: 'unauthorized' })
-                throw new Error()
             }
         } catch (err) {
             reply.code(401).send({ error: 'unauthorized' })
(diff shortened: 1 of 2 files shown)


forge/routes/auth/index.js
```diff
@@ -72,13 +72,17 @@ module.exports = fp(async function (app, opts, done) {
...
-        } else if (request.headers && request.headers.authorization) {
-            await verifyToken(request, reply)
-        } else if (!request.context.config.allowAnonymous) {
+        try {
+            if (request.sid) {
+                await verifySession(request, reply)
+            } else if (request.headers && request.headers.authorization) {
+                await verifyToken(request, reply)
+            } else if (!request.context.config.allowAnonymous) {
+                reply.code(401).send({ error: 'unauthorized' })
```
Steve-Mcl, line 81: Is this line necessary since the throw would result in the same 401 response

knolleary: Good point - but you've also made me spot something else I need to check. Will pick it up in the morning.

knolleary: I've removed the throw. No need to add the overhead of throwing an error that is caught two lines later.

I wanted to double check that *not* throwing an error would properly stop the request being processed any further. I've confirmed that calling `reply.send()` in a `preHandler` means fastify considers the request handled and no further processing is done.

Steve-Mcl (approved):

PR #650, opened 2022-06-01
Show loading graphic when button actions taking place.

### LINKED ISSUE #633: Buttons can be clicked twice to trigger duplicate actions
### Current Behavior

Some buttons in the UI, such as 'create project', trigger an asynchronous call to the API before a result is known.

When running locally, the response time is sufficiently short that it feels responsive.

In FFCloud there can be a noticeable latency before the UI responds. This can lead to the user clicking the button again - triggering the action again. We have seen users create duplicate projects because of this.

We need to review *all* action buttons and ensure they do not allow repeated clicks where inappropriate.

### Expected Behavior

A button that triggers an action cannot be clicked again whilst the first action is running.

### Steps To Reproduce

When running locally, you can use browser tools to simulate a slow network to find where the UI responsiveness is tidied to the network speed.

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

joepavitt:
Closes #633

### DIFF
--- a/frontend/src/pages/admin/Stacks/dialogs/AdminStackEditDialog.vue
+++ b/frontend/src/pages/admin/Stacks/dialogs/AdminStackEditDialog.vue
@@ -1,7 +1,8 @@
 <template>
     <ff-dialog :open="isOpen" :header="(stack ? 'Update' : 'Create') + ' Stack'" @confirm="isOpen = false">
         <template v-slot:default>
-            <form class="space-y-6">
+            <ff-loading v-if="loading" message="Creating Stack..."/>
+            <form v-else class="space-y-6">
                 <FormRow v-if="!stack" :options="options" v-model="input.baseStack">
                     Optionally, use an existing Stack as a starting point:
                 </FormRow>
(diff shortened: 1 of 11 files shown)


frontend/src/pages/admin/Stacks/dialogs/AdminStackEditDialog.vue
```diff
@@ -18,7 +19,7 @@
         </template>
         <template v-slot:actions>
             <ff-button kind="secondary" @click="close()">Cancel</ff-button>
-            <ff-button kind="primary" @click="confirm()" :disabled="!formValid">{{ (stack ? 'Save' : 'Create') }}</ff-button>
+            <ff-button kind="primary" @click="confirm()" :disabled="!formValid || !loading">{{ (stack ? 'Save' : 'Create') }}</ff-button>
```
knolleary, line 22: Missing this whilst reviewing - the logic is flipped here. Should be `loading` not `!loading`.

joepavitt: sorry, just realised th diff you're showing there is my change, not your requested change. Just approved the other PR (https://github.com/flowforge/flowforge/pull/659#pullrequestreview-997641914)

PR #656, opened 2022-06-01
Loading component displayed when loading data from API

### LINKED ISSUE #654: Slow API response times can lead to inconsistent UX
### Current Behavior

When opening any page that hits the forge API to load data, the UI state can be inconsistent until the API responds and then the UI updates to match.

The following are the most obvious ones that need addressing (there may be others, but lets start here...)

1. Team Overview - this is the page most uses land on when they log in. It lists their projects - but until the API call to list the projects returns, the page says 'you have no projects'. That can be distressing to read when you expect lots of projects to be listed.
2. Team Projects page - as above
3. Team Devices page - same again, just devices no projects
4. Project Devices
5. Project Snapshots

### Expected Behavior

Some kind of loading placeholder so we don't show the wrong state whilst API calls are pending.

### Steps To Reproduce

Use Chrome Dev tools to throttle the network link.

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

joepavitt:
Closes #654

### DIFF
--- a/frontend/src/pages/team/Devices/index.vue
+++ b/frontend/src/pages/team/Devices/index.vue
@@ -117,6 +117,7 @@ export default {
     },
     methods: {
         fetchData: async function (newVal) {
+            this.loading = true
             if (this.team.id && !this.project) {
                 const data = await teamApi.getTeamDevices(this.team.id)
                 this.devices.length = 0
@@ -126,6 +127,7 @@ export default {
                 this.devices.length = 0
                 this.devices = data.devices
             }
+            this.loading = false
         },
         showCreateDeviceDialog () {
             this.$refs.teamDeviceCreateDialog.show(null, this.project)


frontend/src/pages/team/Projects.vue
```diff
@@ -18,11 +18,12 @@
                 </ff-button>
             </div>
         </template>
-        <template v-else>
+        <template v-else-if="!loading">
```
knolleary, line 21: This condition needs moving to the top so it takes precedence over the others. As it stands, it is never being reached.

frontend/src/pages/team/Devices/index.vue
```diff
@@ -37,6 +37,7 @@
                 </template>
             </div>
         </template>
+        <ff-loading v-else message="Loading Devices..." />
```
knolleary, line 40: This condition needs moving to the top so it takes precedence over the others. As it stands, it is never being reached.

knolleary (changes requested):
A couple instances where the `v-if` logic isn't being reached to display the loading component.

frontend/src/pages/team/Devices/index.vue
```diff
@@ -92,6 +93,7 @@ export default {
     name: 'TeamDevices',
     data () {
         return {
+            loading: false,
```
knolleary, line 96: We're missing the code to set `loading=true` when retrieving the list of devices

knolleary (changes requested):

joepavitt:
@knolleary fixed the loading variable setting in devices accordingly.

PR #663, opened 2022-06-06
Better handle billing errors whilst deleting projects

### LINKED ISSUE #612: Deleting a project when stripe returns an error leaves project in a hung state
### Current Behavior

If a user deletes a project and stripe doesn't return a successful response to the subscription update (beacuse either stripe is down or the subscription is removed or something else)

The container is deleted but the project object in the database still exists.

You are then unable to access or delete the project in the forge app even though it shows up in the list, the project URL 404s

### Expected Behavior

If stripe fails then the whole project deletion should be stopped and the user should retry later.
Or 
The project should be cleanly deleted and an admin should be alerted to check and manually amend stripe as required..

### Steps To Reproduce

_No response_

### Environment

- FlowForge version: 0.5.2
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Fixes #612 

This PR reorders the sequence we do things when deleting a project.

With this PR, if billing is enabled, our interactions with Stripe are the first thing we do - so if they fail, the whole delete can fail cleanly.

Whilst investigating this issue, I also found that in the unlikely case that a container driver has deleted its internal state for a project, but the overall delete has been cancelled, we didn't protect against errors being thrown when the user tries to interact with the project. This PR adds that error handling.

There will be companion PRs in the localfs, docker and kube drivers to ensure appropriate errors are thrown in those scenarios.

### DIFF
--- a/forge/containers/wrapper.js
+++ b/forge/containers/wrapper.js
@@ -132,9 +132,6 @@ module.exports = {
      * @returns {Promise} Resolves when the project has been stopped
      */
     remove: async (project) => {
-        if (this._driver.remove) {
-            await this._driver.remove(project)
-        }
         if (project.state !== 'suspended') {
             // Only updated billing if the project isn't already suspended
             if (this._app.license.active() && this._app.billing) {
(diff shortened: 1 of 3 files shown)


Steve-Mcl (approved):

Steve-Mcl:
All goodness 👍

PR #665, opened 2022-06-06
Update dependencies and fix test runners

knolleary:
Updates all of the production dependencies of the main app. This is mostly just upping the fix levels we list in package.json.

Some of the `fastify` plugins have migrated to new npm module names - with a major version bump to go with it. The major version change reflects the rename - not a change in behaviour.

I also noticed the core unit tests were not running properly - for both `unit` and `system`, it was only running one of the many test files. The fix was to add quotes to the file glob so that `mocha` can evaluate it properly. I *think* this behaviour changed when #627 was merged - but I can't see why.

Whilst I was in there, I changed the long chains of ` npm run X && npm run Y` to use `npm-run-all` that we already have as a dependency.


Steve-Mcl (approved):
deps update looks pretty straight forward thankfully 👍

PR #667, opened 2022-06-07
Enable device feature by default

### LINKED ISSUE #666: Remove device feature flag
### Description

With the 0.6 release, the Devices feature has reached MVP. It currently requires a feature flag to enable.

Having reached MVP it is appropriate to remove the flag and have the feature enabled by default.

There are two options:

1. Flip the logic so we default the flag to `true` if not otherwise set. This doesn't require lots of code changes
2. Remove all feature flag logic throughout the code base.

I propose we take option 1 - it is the lowest risk, minimal change to achieve this. 

### Epic/Story

#446

knolleary:
Fixes #666

### DIFF
--- a/forge/config/features.js
+++ b/forge/config/features.js
@@ -1,5 +1,5 @@
 const featureFlags = [
-    { name: 'devices', default: false, isPublic: false }
+    { name: 'devices', default: true, isPublic: false }
 ]
 
 module.exports = function (app, config) {


forge/config/features.js
```diff
@@ -1,5 +1,5 @@
 const featureFlags = [
-    { name: 'devices', default: false, isPublic: false }
+    { name: 'devices', default: true, isPublic: false }
```
ZJvandeWeg, line 2: Tracing it back to api/index.js I understand the `isPublic` value doesn't influence anything for enabling this feature.

This line and the flagging should be removed in 0.7 though, I'll create a follow up issue.

ZJvandeWeg (approved):

PR #671, opened 2022-06-07
Allow a stack-less project have a stack applied

### LINKED ISSUE #534: Cannot change project stack when not on a stack
Old projects can change their stack if they're currenlty not a stack. This happens when they were created before stacks existed.

knolleary:
Fixes #534 

Stack-less projects can only exist if they were created in FlowForge 0.2.0 or earlier. Anything more recent will have a stack.

Whilst there are diminishing returns in fixing this issue, the fix is simple enough to just get done.


Steve-Mcl (approved):
👍

PR #702, opened 2022-06-15
Start of K8s install instructions

### LINKED ISSUE #700: Kubernetes install and upgrade guide is missing
### Description

While there's documentation for installing and upgrading on docker and localfs, there's no such guide for Kubernetes.

We should provide install instructions for our users.

hardillb:
Add docs about how to install FlowForge on Kubernetes

Fixes #700

### DIFF
--- a/docs/install/docker/README.md
+++ b/docs/install/docker/README.md
@@ -49,7 +49,7 @@ This container holds the FlowForge App and the Docker Driver
 
 ##### flowforge/node-red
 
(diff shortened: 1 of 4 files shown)


ZJvandeWeg:
@hardillb In the spirit of iteration I would propose we merge this as is, and I'll install it now too and just create a PR with my findings. Ok?

hardillb:
@ZJvandeWeg OK, sounds like a plan.

The AWS EKS instructions can be augmented with the docs in the Cloud Project. I will sanitise them of FF specifics and copy them over to here as we go.

hardillb:
Merging, will iterate later if needed.

Issue #464, opened 2022-04-08
MQTT Broker within FlowForge


sammachin:
### Description

To improve the efficiency of how launchers/devices report status and receive commands, we want to introduce an MQTT Broker into the FlowForge architecture. It will also enable a custom set of nodes described in #662 to provide seamless cross-project communication.

This story does **not** cover external usage of the broker. See https://github.com/flowforge/flowforge/issues/738 for the story covering that

---

 - [x] https://github.com/flowforge/flowforge/pull/706
 - [x] https://github.com/flowforge/flowforge-device-agent/pull/22
 - [x] https://github.com/flowforge/flowforge/issues/662

sammachin:
Some simple options could be, the Aedes node within Node-RED provides a broker inside NR
However this is not workable in FlowForge today because external TCP connections are not permitted.
Solving the TCP connection issue may be an alternative here.

hardillb:
> Solving the TCP connection issue may be an alternative here.

I believe this basically means giving each project their own IP address which I don't think i feasible (Given cost of buying IPv4 address ranges these days).

sammachin:
doesn't have to be a whole IP address, each project could just be given a single random port.

sammachin:
or possibly using SNI if connections were all TLS

hardillb:
The comment was in regard to the the generic TCP access problem, not specifically MQTT.

We will need to look at the options for none HTTP load balancing (this is how K8s talks about this stuff) in AWS EKS

PLCMercenary:
Host a broker cluster and give each project a mount point. we can surely make the subscribe connection to a broker from a project right?

PLCMercenary:
If this issue is still being discussed, I think a simple solution would be to host a broker cluster, it doesnt matter where it is, as long as we can expand to additional compute resources. we ingress everything using a dns address, `mqtt.flowforge.cloud`
then just assign mount points for each admin username, or per the project. it would be simple to generate a user id on creation if the instance, and then provide a key, the mount point is then set in the backend table with a simple update and i believe the process update is a change hook for the table.

ZJvandeWeg:
@knolleary Would appreciate your thoughts here. In general I tend to understand this as: https://github.com/flowforge/flowforge/issues/464#issuecomment-1095317433

We should ship 0.7 with a Bring your own broker, but a working version for FF.cloud.

knolleary:
We need to be very careful to not think BYOB solves this properly.

[...]

PR #706, opened 2022-06-17
Add comms (mqtt) component to forge platform

knolleary:
Part of #464 

This adds the Comms component to the platform, enabling MQTT communication.

The Comms component will enable the following uses:

 - [ ] Devices reporting status and receiving commands from the platform
     - [ ] https://github.com/flowforge/flowforge-device-agent/pull/22
 - [ ] Inter-project communication (EE-only)
     - [ ] https://github.com/flowforge/flowforge/issues/662
 - [ ] **OUT OF SCOPE FOR THIS PR**
     Launchers reporting status and receiving commands from the platform - 
     - This PR enables the broker ACLS for the launchers to send/receive commands, but goes no further.

### Broker Configuration

This requires a properly configured MQTT broker to be running and available to the platform.

We are using mosquitto, with the [mosquitto-go-auth](https://github.com/iegomez/mosquitto-go-auth) plugin.

The following is an example configuration:
```
auth_plugin /Users/nol/code/flowforge/mosquitto/mosquitto-go-auth/go-auth.so
auth_opt_cache true
auth_opt_backends  http
auth_opt_http_host localhost
auth_opt_http_port 3000
auth_opt_http_getuser_uri /api/comms/auth/client
auth_opt_http_aclcheck_uri /api/comms/auth/acl
per_listener_settings false
listener 4880
listener 4881
protocol websockets
```

Note it includes the http host/port of the forge platform, as well as the full path to the go-auth plugin.

### API Routes

The mosquitto auth plugin uses http requests to the platform to authenticate and authorise requests. These are made to two new api routes:

 - `POST /api/comms/auth/client` - validate a given username/password is valid
 - `POST /api/comms/auth/acl` - validate an access request to a given topic is valid for a particular client

### Database Changes

Adds a `BrokerClient` table to store username/passwords for MQTT Clients.

The platform generates its own BrokerClient on initialisation (`forge_platform`) - used to subscribe to status events and publish commands to devices/launchers.

Devices get a BrokerClient generated whenever they generate credentials - and the details are provided along with the other device credentials in the UI.

### Runtime Configuration

A `broker` configuration must be provided via `flowforge.yml`:

```
broker:
  url: mqtt://localhost:1880
```

 - [ ] **Question**: interval vs external broker url (ie, what launchers would use, vs what devices would use)

### Next steps

 - pass client credentials to launcher
    - [ ] localfs: https://github.com/flowforge/flowforge-driver-localfs/pull/54
    - [ ] docker: https://github.com/flowforge/flowforge-driver-docker/pull/30
    - [ ] k8s: https://github.com/flowforge/flowforge-driver-k8s/pull/44
 - [ ] For all of the above PRs - need to also pass through broker connection url. How that is determined (internal vs external access)
 - [ ] **OUT OF SCOPE*** Update launcher to use broker credentials to connect and report status - this needs to be a separate story for a future iteration. It is out of scope for the MVP.


knolleary:
Currently reviewing whether this is the right approach for the security integration with the broker.

The problem with the dynamic security approach is we don't hold the master record of who can access what. We are reliant on the broker storing that information and it always being available.

In the k8s world, that means having a persistent volume - which complicates matters.

There is also the matter of bootstrapping its config file with the admin username/password. We wouldn't be able to mount that file in like we do `flowforge.yml` as mosquitto needs to write to it as well as read.

An alternative approach would be to use https://github.com/iegomez/mosquitto-go-auth which supports talking directly to sqlite/postgres to validate users.

That would allow us to keep the information in our own database - and the mosquitto instance can be more ephemeral.

The added complication is building that plugin (written in Go) for the platforms we want to support and getting it installed properly.

forge/routes/index.js
```diff
@@ -18,5 +18,6 @@ module.exports = fp(async function (app, opts, done) {
     await app.register(require('./setup'), { logLevel: 'warn' })
     await app.register(require('./storage'), { prefix: '/storage', logLevel: 'warn' })
     await app.register(require('./logging'), { prefix: '/logging', logLevel: 'warn' })
+    await app.register(require('./broker'), { prefix: '/api/broker', logLevel: 'warn' })
```
hardillb, line 21: Should this have `/api` prefix? if so should it not be versioned as well?

knolleary:
Rebased this branch to `main` as of the 0.7 release.

hardillb (approved):
Looks good

PR #707, opened 2022-06-17
Bump sqlite node to 5.0.8

hardillb:
This is to fix building with Python 3

At the moment we are pinned at exactly 5.0.2, which expects there to be a `python` on the path to build. Alpine on Arm64 python3 install does not add `python` to the path so it doesn't build


hardillb:
Installing 5.0.4+ works for a clean install, but not an upgrade, still haven't worked out why.

hardillb:
- Install flowforge as is (with 5.0.2 sqlite3)
- In the same directory  run `npm install sqlite3@5.0.8`

```
$ npm install sqlite3@5.0.8
npm ERR! code 127
npm ERR! path /opt/share/projects/flowforge/sqlite-test/node_modules/sqlite3
npm ERR! command failed
npm ERR! command sh -c node-pre-gyp install --fallback-to-build
npm ERR! sh: line 1: node-pre-gyp: command not found

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/hardillb/.npm/_logs/2022-08-01T12_53_27_501Z-debug-0.log
```

Doesn't fail if try and reproduce without FlowForge. Makes me think it's one of the other shared dependencies that is causing the failure.

hardillb:
OK, so it's the bcrypt node that it's clashing with

knolleary:
@hardillb what was the conclusion about the clash?

hardillb:
@knolleary sorry, this got split to the https://github.com/flowforge/security/issues/6 issue.

Copying over:

OK, I think I've tracked down the upgrade problem with sqlite3 5.0.8.
At the moment, we are pinned at version 5.0.2
If you install flowforge with this then try and upgrade to anything newer e.g. 5.0.8 the install fails

```
$ npm install sqlite3@5.0.8
npm ERR! code 127
npm ERR! path /opt/share/projects/flowforge/sqlite-test/node_modules/sqlite3
npm ERR! command failed
npm ERR! command sh -c node-pre-gyp install --fallback-to-build
npm ERR! sh: line 1: node-pre-gyp: command not found

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/hardillb/.npm/_logs/2022-08-01T12_53_27_501Z-debug-0.log
```

It also fails if you edit the package.json and then just run npm install
If you delete the current node_modules dir and edit the package.json to use ^5.0.8 it all installs cleanly
The upgrade path failure is caused by an interaction with the bcrypt modules as if I remove this then the upgrade works. I've not worked out exactly what the problem, both are using the @mapquest/node-pre-gyp (one with ^1.0.0 and the other ^1.0.7)

I propose we just bump the version and once merged we just remember delete node_modules in the development environments and reinstall everything.
All the production builds are from the clean slate and we can document the problem if needed for localfs upgrades

knolleary:
> All the production builds are from the clean slate and we can document the problem if needed for localfs upgrades

In which case, can you add some details to https://github.com/flowforge/flowforge/blob/main/docs/install/upgrading.md as part of this PR.

hardillb:
Trying to come up with some words, but we don't actually have any instructions for how to upgrade a LocalFS install at the moment.

docs/install/upgrading.md
```diff
@@ -25,3 +25,22 @@ users will be able to create new projects:
     2. As a one-time action, set its Project Type to the one just created.
     3. Click 'save'. This will update the stack *and* all existing projects to
        be associated with the new Project Type
+
+
+### Upgrading to 0.8
+
+With the 0.8 release we have updated the version of the SQLite3 node used by the localfs 
+contianer driver. We are moving from v5.0.2 to v5.0.8.
```
knolleary, line 33:
```suggestion
container driver. We are moving from v5.0.2 to v5.0.8.
```

docs/install/upgrading.md
```diff
@@ -25,3 +25,22 @@ users will be able to create new projects:
     2. As a one-time action, set its Project Type to the one just created.
     3. Click 'save'. This will update the stack *and* all existing projects to
        be associated with the new Project Type
+
+
+### Upgrading to 0.8
+
+With the 0.8 release we have updated the version of the SQLite3 node used by the localfs
```
knolleary, line 32:
```suggestion
#### LocalFS Users

With the 0.8 release we have updated the version of the SQLite3 module used by the localfs 
```

docs/install/upgrading.md
```diff
@@ -25,3 +25,22 @@ users will be able to create new projects:
...
+
+```
+npm ERR! path /opt/share/projects/flowforge/sqlite-test/node_modules/sqlite3
+npm ERR! command failed
+npm ERR! command sh -c node-pre-gyp install --fallback-to-build
+npm ERR! sh: line 1: node-pre-gyp: command not found
+```
+
+If you see this then the simplest fix is to remove the `node_modules` directory and reinstall
+all the nodes from clean.
```
knolleary, line 46:
```suggestion
the modules.
```

docs/install/upgrading.md
```diff
@@ -25,3 +25,22 @@ users will be able to create new projects:
...
        be associated with the new Project Type
+
+
+### Upgrading to 0.8
+
+With the 0.8 release we have updated the version of the SQLite3 node used by the localfs 
+contianer driver. We are moving from v5.0.2 to v5.0.8.
+
+There appears to be a clash with the bcrypt module when doing an inplace upgrade of the
+SQLite3 node that gives an error similar to the following:
```
knolleary, line 36:
```suggestion
SQLite3 module that gives an error similar to the following:
```

PR #711, opened 2022-06-21
Admin version

### LINKED ISSUE #655: Access platform version information
### Epic

_No response_

### Description

**As an:** Admin

**I want to:** quickly see the version of the platform I'm using.

**So that:** 
 - Check if an upgrade is available
 - Check an upgrade worked
 - Grab the information when reporting an issue

---

### API Design

We should include this information in the `/api/v1/settings` response for a logged-in user.

### UI Design

For a first iteration, including this on the AdminOverview page would be sufficient. There are further design considerations on how this could be exposed to general users of the platform as well.

### Acceptance Criteria

_No response_

sammachin:
closes #655 

Adds a new box to the admin/overview page displaying the version of the Forge app and the Node version
image

Also added a `-git` label to the version if there is a `.git` folder which is the same pattern as used by Node-RED,
```
[2022-06-21T13:45:17.655Z] INFO: Development mode
[2022-06-21T13:45:17.656Z] INFO: FlowForge v0.6.0-git
[2022-06-21T13:45:17.656Z] INFO: FlowForge running with NodeJS v16.13.0
```

### DIFF
--- a/forge/config/index.js
+++ b/forge/config/index.js
@@ -54,13 +54,12 @@ module.exports = fp(async function (app, opts, next) {
         ffVersion = version
     }
     try {
-        fs.statSync(path.join(__dirname,"..","..",".git"));
-        ffVersion += "-git";
-    } catch(err) {
+        fs.statSync(path.join(__dirname,'..', '..', '.git'))
+        ffVersion += '-git'
+    } catch (err) {
         // No git directory
     }
     app.log.info(`FlowForge v${ffVersion}`)
-
 
     app.log.info(`FlowForge running with NodeJS ${process.version}`)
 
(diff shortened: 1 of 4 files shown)


forge/config/index.js
```diff
@@ -48,13 +48,18 @@ module.exports = fp(async function (app, opts, next) {
...
         // npm start
-        app.log.info(`FlowForge v${process.env.npm_package_version}`)
     } else {
         // everything else
         const { version } = require(path.join(module.parent.path, '..', 'package.json'))
         ffVersion = version
-        app.log.info(`FlowForge v${version}`)
     }
+    try {
+        fs.statSync(path.join(__dirname, '..', '..', '.git'))
```
ZJvandeWeg, line 57: `git describe` is what you're looking for @sammachin

sammachin: this code was lifted out of the Node-RED core to keep the design pattern consistent

PR #713, opened 2022-06-21
Toast Alerts

joepavitt:
Dependent upon: https://github.com/flowforge/forge-ui-components/pull/24

first half of https://github.com/flowforge/flowforge/issues/515 (see this issue for a screen recording demo)

### DIFF
--- a/frontend/src/layouts/Platform.vue
+++ b/frontend/src/layouts/Platform.vue
@@ -20,7 +20,7 @@
 <script>
 import PageHeader from '@/components/PageHeader.vue'
 
-import Alerts from '@/services/alerts.js'
+import alerts from '@/services/alerts.js'
 
 export default {
     name: 'ff-layout-platform',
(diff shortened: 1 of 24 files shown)


joepavitt:
Expecting build to fail at the moment (PR still in draft state) as I'm still referencing my local forge-ui-components in the package json

knolleary:
The basic mechanism looks good. Not 100% on the appearance yet - but we can refine later.

Would be good to capture some docs on how to use it and guidelines on *when* to use it. I know there are a bunch of actions being added in my various PRs that'll merge over the coming days before 0.7 that will need updating to incorporate it.

joepavitt:
Thanks @knolleary  - actually agree on the appearance. I went for the generic toast appearance just to get something in, but it does feel out of place with the rest of the UI which I find a lot cleaner/slicker. Will happily iterate.

All available in Figma if you did want to play yourself too.

joepavitt:
RE: Docs - I have been thinking quite a bit about this, and in particular, _where_ that should live. I feel like a natural home is actually the handbook under a "Design Language" section? Doesn't quite feel right in `forge-ui-components` as it's about more than just the raw Vue components.

Issue #673, opened 2022-06-07
Add postgres to automated test coverage


knolleary:
### Description

We currently run all of our unit tests against sqlite as a convenience to the developer - and rely on manual testing against postgres.

This can lead to postgres-specific issues being discovered late in the release cycle (#672) or worse if they are missed entirely.

We need to add postgres to the automated test setup without sacrificing the speed and convenience of being able to run against sqlite.

PR #719, opened 2022-06-22
Add optional support for running tests against postgres

knolleary:
Part of #673 
This adds the ability to run unit tests against postgres.

I've updated the docs to explain, but for ease of reference, I'm pasting that content below.

This has found a couple issue that are specific to postgres - one a test issue and one a real bug with deleting devices.

I'm not 100% happy about the user having to manually install the `pg` module - as that makes their package.json dirty.

Would probably be better to add it as a dev dependency.

---

By default, the tests use an in-memory sqlite database to test against. This is
the most self-contained way of testing the platform. But it is also necessary to
test against PostgreSQL. To enable the use of PostgreSQL in the tests:

1. Install the `pg` module

        npm install pg

2. Ensure you have an instance of PostgreSQL running locally. For example, via
   docker:

        docker run -it -p 5432:5432 --name ff-postgres -e POSTGRES_PASSWORD=secret postgres

3. Enable PostgrSQL mode by setting the following environment variable:

        export FF_TEST_DB_POSTGRES=true

   The database connection can be set using the following env vars (default values shown)

        export FF_TEST_DB_POSTGRES_HOST=localhost
        export FF_TEST_DB_POSTGRES_PORT=5432
        export FF_TEST_DB_POSTGRES_USER=postgres
        export FF_TEST_DB_POSTGRES_PASSWORD=secret
        export FF_TEST_DB_POSTGRES_DATABASE=flowforge_test


knolleary:
The next step will be figuring out how to add this to our GH actions. The main question is how to get an instance of postgres available to the action.

hardillb:
I need to look closer at the container support in GH actions, we should be able to spin up a Postgres container...

knolleary:
Having thought about it over a sandwich, I've now added `pg` as a dev dependency and removed the need to install it manually from the docs.

hardillb:
I think this looks like what we want

https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers

hardillb (approved):

PR #725, opened 2022-06-23
Add stack versioning apis and admin ui

### LINKED ISSUE #694: Stack Versions
### Epic

#285

### Description

**As an:** admin

**I want to:** create a new version of an existing stack

**So that:** users can be notified there is a new version of what they are using

---

Version History of stacks was in the original concept of #285 but was descoped. It is now time to bring it back.

Currently, users have no way of knowing there is a new version of Node-RED available - or of the NR Launcher used by projects. Admins can create new stacks, and mark existing ones as inactive (so they cannot be picked for new projects). But there is no relationship between old and new.

This story seeks to add such a relationship.

---

### Design

The existing `ProjectStacks` schema looks like this:
```
CREATE TABLE "ProjectStacks" (
	"id"	INTEGER,
	"name"	VARCHAR(255) NOT NULL UNIQUE,
	"active"	TINYINT(1) DEFAULT 1,
	"properties"	TEXT,
	"createdAt"	DATETIME NOT NULL,
	"updatedAt"	DATETIME NOT NULL,
	PRIMARY KEY("id" AUTOINCREMENT)
);
```

To keep things simple, we just need to record whether a ProjectStack has been superseded by another, and the id of that stack. That can be done by adding the column:

 - `"replacedBy" INTEGER NULL

This will be a created as a foreign key to the same table to reference another ProjectStack.

When a stack is created, an option is provided to create it as a replacement for an existing stack. In that case:
 - the new stack (`NEW`) is created
 - the existing stack (`OLD`) has the following changes made:
     - `active` field set to `false`
     - `replacedBy` field set to the `id` of the new stack

Furthermore, any existing stack that has its `replacedBy` field set to`OLD.id` should be updated to point at `NEW.id` - this means we don't have long chains of stacks - old stacks always point to the latest available.

It should only be possible to replace a stack that does not have `replacedBy` set - ie, you can only replace a given stack once, to ensure a single lineage.

### API changes

 - Various APIs will need updating to include the new `replacedBy` field.

 - When creating a stack, it should be possible to provide the id of the stack the new stack replaces. This request should fail if that stack doesn't exist, or already has its `replacedBy` field set.

### UI changes

The admin table of stacks needs to be updated to show where stacks have been replaced - and by what. This will require a bit of experimentation/iteration to get right.

### Acceptance Criteria

_No response_

knolleary:
Closes #694

This PR adds stack versioning. The full API design is described in #694.

The UI changes are primarily in the Admin view for managing stacks.

I've included a video demo of them in action here for feedback:

https://user-images.githubusercontent.com/51083/175297169-1014192d-d1f9-4f84-85fc-06a528767b65.mov

### DIFF
--- a/forge/db/migrations/20220622-01-add-stack-replacedBy.js
+++ b/forge/db/migrations/20220622-01-add-stack-replacedBy.js
@@ -1,0 +1,17 @@
+/**
+ * Add replacedBy column to ProjectStacks table
+ */
+const { DataTypes } = require('sequelize')
+
+module.exports = {
+    up: async (context) => {
+        await context.addColumn('ProjectStacks', 'replacedBy', {
+            type: DataTypes.INTEGER,
+            references: { model: 'ProjectStacks', key: 'id' },
+            onDelete: 'set null',
+            onUpdate: 'cascade'
+        })
+    },
+    down: async (context) => {
+    }
+}
--- a/forge/db/models/ProjectStack.js
+++ b/forge/db/models/ProjectStack.js
@@ -25,12 +25,19 @@ module.exports = {
     },
     associations: function (M) {
         this.hasMany(M.Project)
+        this.hasOne(this, { foreignKey: 'replacedBy' })
     },
-    hooks: {
-        beforeDestroy: async (stack, opts) => {
-            const projectCount = await stack.projectCount()
-            if (projectCount > 0) {
-                throw new Error('Cannot delete stack that is used by projects')
+    hooks: function (M) {
+        return {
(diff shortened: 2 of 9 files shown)


hardillb (approved):

knolleary:
@joepavitt checking the main branch, I don't see the sidebar highlighting anything when in the Admin section.

joepavitt:
Good point - will raise an issue separately. This PR looks good to me in that case.

joepavitt (approved):

PR #730, opened 2022-06-25
initial rollback support

### LINKED ISSUE #587: Snapshot/Rollback
### Epic

#218

### Description

**As a:** owner

**I want to:** be able to rollback my project to a previous deployment.

**So that:** I can undo changes made by others in my team

---

0.6 added the ability to create snapshots.

The focus shifts to allowing a project to be rolled back. See https://github.com/flowforge/flowforge/issues/587#issuecomment-1155156854 for details

---

This is an MVP of version control meant to just be a basic snapshot of the flows, it does not prevent us offering a more complex feature later with git integration and such.
The project should be to keep a certain number or days of snapshots (amount may vary based on Project type in future)

Open questions to consider:
 - Are snapshots automatic and done at certain intervals or on each deploy, or is there a manual intervention to keep a snapshot, (If snapshots are on every deploy this could create a large number)

- How long/how many snapshots do we hold, is this X snapshots for Y days?

These are also potentially differentiators for different tiers so offering more/longer retention on higher tiers.

MVP would be to just rollback the flows file, any nodes installed or removed would be outside the scope of this.

### Acceptance Criteria

_No response_

Steve-Mcl:
fixes #587

This is the first working (minimal) support for rolling back snapshots

### In action...
 
rollback

### Issues 
* UI: settings do not refresh after applying rollback (need to refresh FF web page to see rolled back settings)
* UI: No progress and confirmation of completion
* UI: No way of understanding which snapshot (if any) the currently loaded project matches

### TODO
Once direction of this work is confirmed as the right direction, I will write tests.

### DIFF
--- a/forge/routes/api/projectActions.js
+++ b/forge/routes/api/projectActions.js
@@ -108,4 +108,36 @@ module.exports = async function (app) {
             reply.code(500).send({ error: err.toString() })
         }
     })
+
+    app.post('/rollback', async (request, reply) => {
+        let restartProject = false
+        try {
+            // get (and check) snapshot is valid / owned by project before any actions
+            const snapshot = await app.db.models.ProjectSnapshot.byId(request.body.snapshotId)
+            if (!snapshot) {
+                throw new Error(`snapshot '${request.body.snapshotId}' not found for project '${request.body.snapshotId}'`)
+            }
+            if (snapshot.ProjectId !== request.project.id) {
+                throw new Error(`snapshot '${request.body.snapshotId}' is only valid for project ${snapshot.ProjectId}`)
+            }
+            if (request.project.state === 'running') {
+                restartProject = true
+            }
+            app.db.controllers.Project.setInflightState(request.project, 'rollback')
+            await app.db.controllers.Project.importProjectSnapshot(request.project, snapshot)
+            app.db.controllers.Project.clearInflightState(request.project)
+            await app.db.controllers.AuditLog.projectLog(
+                request.project.id,
+                request.session.User.id,
+                'project.snapshot.rollback',
+                { id: snapshot.hashid }
+            )
+            if (restartProject) {
+                await app.containers.restartFlows(request.project)
+            }
+            reply.send({ status: 'okay' })
+        } catch (err) {
+            reply.code(500).send({ error: err.toString() })
+        }
+    })
 }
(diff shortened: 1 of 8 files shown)


forge/routes/api/projectActions.js
```diff
@@ -108,4 +108,36 @@ module.exports = async function (app) {
             reply.code(500).send({ error: err.toString() })
         }
     })
+
+    app.post('/rollback', async (request, reply) => {
+        let restartProject = false
+        try {
+            // get (and check) snapshot is valid / owned by project before any actions
+            const snapshot = await app.db.models.ProjectSnapshot.byId(request.body.snapshotId)
```
knolleary, line 116:
```suggestion
            const snapshot = await app.db.models.ProjectSnapshot.byId(request.body.snapshot)
```

For the external API, the property should be `snapshot` rather than `snapshotId`. This is more consistent with other places we pass object ids over the api.

frontend/src/api/project.js
```diff
@@ -59,6 +59,12 @@ const updateProject = async (projectId, options) => {
         return res.data
     })
 }
+const rollbackProject = async (projectId, snapshotId) => {
+    const data = {
+        snapshotId: snapshotId
```
knolleary, line 64:
```suggestion
        snapshot: snapshotId
```

For the external API, the property should be snapshot rather than snapshotId. This is more consistent with other places we pass object ids over the api.

frontend/src/pages/project/Snapshots/dialogs/ConfirmSnapshotRollbackDialog.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+    <ff-dialog :open="isOpen" header="Rollback Snapshot">
+        <template v-slot:default>
+            <p>This rollback will overwrite the current project.</p>
+            <p>All changes made since the last snapshot was taken will be lost.</p>
+            <p>Are you sure you want to rollback to this snapshot?</p>
```
knolleary, line 6:
```suggestion
            <p>This rollback will overwrite the current project.</p>
            <p>All changes to the flows, settings and environment variables made since
                   the last snapshot will be lost.</p>
            <p>Are you sure you want to rollback to this snapshot?</p>
```

knolleary (approved):
Looks good.

Minor API tweak and dialog wording change suggested.

PR #734, opened 2022-06-27
Device page backend

hardillb:
Adds DeviceSettings to hold Device Specific properties. Currently this consists of a `env` - to hold device specific env vars.

Adds the following API endpoints

- `/api/v1/device/:deviceId/settings` GET (device:edit) `{ env: [{name: 'FOO', value: 'BAR' }] }` 
- `/api/v1/device/:deviceId/settings` PUT (device:edit) `{ env: [{name: 'FOO', value: 'BAR' }] }` 
- `/api/v1/device/deviceId/live/settings` GET (device token) `{hash: '....', env: { 'FOO':'BAR'}}`


knolleary:
Any chance we can redo the branch to not pull in master? Makes it hard to see the changes.

hardillb:
I'll have a look.

hardillb:
Should be clean now

hardillb:
part of https://github.com/flowforge/flowforge/issues/680

hardillb:
Once merged into @joepavitt's  https://github.com/flowforge/flowforge/tree/device-page branch that should be PR'd to main

hardillb:
It doesn't want to let me change the target branch main for some reason

hardillb:
Re-targeted this to main as it has all the changes in it. Will fix conflicts

forge/db/models/Device.js
```diff
@@ -57,6 +64,37 @@ module.exports = {
...
+                    }
+                    await M.DeviceSettings.bulkCreate(updates, { updateOnDuplicate: ['value'] })
+                    const settings = this.getAllSettings()
+                    this.settingsHash = hashSettings(settings)
+                    await this.save()
+                },
+                async updateSetting (key, value) {
+                    const result = await M.ProjectSettings.upsert({ DeviceId: this.id, key, value })
+                    const settings = this.getAllSettings()
+                    this.settingsHash = hashSettings(settings)
```
knolleary, line 89: missing `await this.save()`

forge/db/migrations/20220624-01-add-devicesettings.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Add DeviceSettings table
+ */
+const { DataTypes } = require('sequelize')
```
knolleary, line 4: The migration needs to create the table.

forge/routes/api/deviceLive.js
```diff
@@ -38,6 +38,13 @@ module.exports = async function (app) {
             })
             return
         }
+        if (request.body.settingsHash !== (request.device.settingsHash)) {
```
knolleary, line 41: We should only do this check if `request.body` includes `settingsHash` - otherwise skip the check. Otherwise we'll break existing agents that will get the 409 but assume its snapshot related.

knolleary (commented):
Is there a reason for burying `env` under an otherwise empty `settings` property in the DeviceSettings table?

We do that for Projects due to the way settings+env are tied to the Template being used for the Project. We don't have the same constraint here, so `env` could be a top-level setting of the device.

forge/db/migrations/20220624-01-add-devicesettings.js
hardillb: Done

knolleary:
I've pushed a commit to fixup the migration - there is some subtle detail on migrations compared to the sequelize model. We need to include the columns that sequelize adds in automatically to the model and the way constraints/relations are setup is slightly different.

forge/routes/api/deviceLive.js
```diff
@@ -38,6 +38,13 @@ module.exports = async function (app) {
             })
             return
         }
+        if (request.body.settingsHash && request.body.settingsHash !== (request.device.settingsHash)) {
+            reply.code(409).send({
+                error: 'incorrect-settings',
+                settingsHash: request.device.settingsHash
+            })
+            return
+        }
```
knolleary, line 47:
```suggestion
        if (request.body.settings && request.body.settings !== (request.device.settingsHash)) {
            reply.code(409).send({
                error: 'incorrect-settings',
                settings: request.device.settingsHash,
                snapshot: request.device.targetSnapshot?.hashid || null
            })
            return
        }
```

On the external API, I think we should call this `settings` not `settingsHash` - we don't use `snapshotHash`.

I also think the response should include both the snapshot and settings values. My suggested change updates one case, but GitHub won't let me extend my suggested change into code you haven't modified - so the response in line 35-37 should be updated to include `settings`.

That will then allow the device to be more selective over whether it grabs both things each time, or just the one that has changed.

knolleary:
Also - in the previous screenshot - not sure why there is a exclamation against the Active snapshot as it matches the Target

joepavitt:
> Have you pulled the latest version of the device agent from the `device-env-vars` branch?
> 
> https://github.com/flowforge/flowforge-device-agent/tree/device-env-vars

That'll be a no

knolleary:
All looking good. I've made some notes on updates to the standalone Device pages (options missing from the Danger page - which we really ought to rename both here and on the Projects page). We can live without them for 0.7 - will raise separate stories to capture properly.

PR #744, opened 2022-06-29
Notify user if a new stack version is available

### LINKED ISSUE #576: Upgrading individual projects between FF releases

knolleary:
Closes #576 

Note this PR is against the `project-type` branch (#739). Once that PR has been merged, this can be retargeted to `main`.

With the ability to create new versions of stacks (#694), this PR adds a notification (in the form of a well placed button) that a new stack version is available for a project.

Clicking the button takes them to the Setting/Danger page which will have a new 'Update Stack' option alongside the 'Change Stack' button.

Clicking that button will update the project to the new stack (using the same backend API as the Change Stack flow)

 - On the project home page:

image

 - On the project settings summary

image

 - The Settings/Danger page

image

### DIFF
--- a/forge/db/models/Project.js
+++ b/forge/db/models/Project.js
@@ -209,7 +209,8 @@ module.exports = {
                                 attributes: ['hashid', 'id', 'name']
                             },
                             {
-                                model: M.ProjectStack
+                                model: M.ProjectStack,
+                                attributes: ['hashid', 'id', 'name', 'links', 'properties', 'replacedBy', 'ProjectTypeId']
                             },
                             {
                                 model: M.ProjectTemplate,
--- a/forge/db/views/Project.js
+++ b/forge/db/views/Project.js
@@ -42,6 +42,7 @@ module.exports = {
                 id: proj.ProjectStack.hashid,
                 name: proj.ProjectStack.name,
                 properties: proj.ProjectStack.properties || {},
+                replacedBy: app.db.models.ProjectStack.encodeHashid(proj.ProjectStack.replacedBy) || undefined,
                 links: proj.ProjectStack.links
             }
         }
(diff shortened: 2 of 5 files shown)


joepavitt (approved):

knolleary:
Yes - looks like github changed the target automatically when I merged #739 - which is a new trick I've not seen happen before. Maybe triggered it when I deleted the `project-types` branch... Either way, this is clear to merge.

PR #790, opened 2022-07-19
First pass at adding @fastify/helmet

hardillb:
Needed to use v8.1.0 as v9.1.0 requires newer fastify

Also needed to disable conentSecurityPolicy to make it run. Need to investigate more.

Fix for https://github.com/flowforge/security/issues/4


hardillb:
Need to work through this list https://helmetjs.github.io/

PR #798, opened 2022-07-22
Component Update: ff-data-table

### LINKED ISSUE #794: Update tables to use the new ff-data-table
### Description

Once the new data table component is available in `forge-ui-components`, we should update the existing tables in FlowForge to utilise it's new functionality (filtering and sorting)

### Epic/Story

_No response_

joepavitt:
Closes #794. Reliant upon https://github.com/flowforge/forge-ui-components/pull/34 and an `npm publish` of `forge-ui-components`


joepavitt:
Examples of the new data-table in action:

https://user-images.githubusercontent.com/99246719/181804431-50b96e8c-1911-44c5-aa00-5937d05de699.mov

joepavitt:
> Also spotted a couple places where the "+ Add Thing" button is in the header and in some places its inline with the search box. Not necessary for this PR, but would be good to standardise.

You're absolutely spot on, there are a couple of page layout inconsistencies in that regard. If I'm honest, I haven't quite worked out why I like it in some places, and not in others. There are some wider layout designs that I want to improve before committing either way, so I'm happy with what we have _for now_

joepavitt:
@knolleary pushed the fix for missing search bars

Issue #705, opened 2022-06-16
Sign in using email address or  username


ZJvandeWeg:
### Description

**As a:** user

**I want to:** be able to sign in with my email address OR my username.

**So that:** I don't have to remember a username

### Acceptance Criteria

- [ ] Username box on login screen accepts email address or username

joepavitt:
Not necessarily got to be an `instead` of here, could it be an `or`?

ZJvandeWeg:
> Not necessarily got to be an instead of here, could it be an or?

I'm fine with this, maybe even best to have an or clause there as long as usernames don't include the @

knolleary:
It should be OR. For local/individual installs we do not require email to be setup, so there is only username available.

sammachin:
Rewritten as a story,

Usernames Must only contain a-z 0-9 - _

So checking for an @ in the string should be sufficient to determine if the input is a username or an email address therefore we can avoid having any kind of user input to set the type

PR #799, opened 2022-07-22
Warn if trying to login with email address

sammachin:
This is a simple iteration towards https://github.com/flowforge/flowforge/issues/705, we should still implement 705 but this will improve the experience for some users now as a stepping stone.

Add a warning when users try to login with an email address instead of a username, currently they just get a generic LOGIN FAILED message and have been contacting us thinking there is an issue with their passwords.

Screen Shot 2022-07-22 at 10 14 41 PM

### DIFF
--- a/frontend/src/components/auth/Credentials.vue
+++ b/frontend/src/components/auth/Credentials.vue
@@ -53,6 +53,10 @@ export default {
                 valid = false
                 this.errors.password = 'Required field'
             }
+            if (this.input.username.includes('@')) {
+                valid = false
+                this.errors.username = 'Username cannot be an email address'
+            }
             if (valid) {
                 this.$store.dispatch('account/login', this.input)
             }


knolleary (approved):
Happy to go with this iteration, but agreed #705 is the proper approach - although I note it isn't currently in plan for 1.0. Given the feedback we've had, should we try to get it in?

sammachin:
> Happy to go with this iteration, but agreed #705 is the proper approach - although I note it isn't currently in plan for 1.0. Given the feedback we've had, should we try to get it in?

@knolleary can you put some design notes against #705 I'm not sure how much backend work is needed to make the API accept email or username, if we've got it designed its certainly something that I would like to drop into a milestone soon if there's capacity

knolleary:
I have added a 1-day sizing. It is a tiny piece of work to do.

PR #803, opened 2022-07-26
 Allow http logging level to be set independently #802 

hardillb:
This allows http route logging to be set independently of all other logging.

It can be controlled from the flowforge.yaml file with:
```yaml
logging:
  level: info
  http: warn
```

Defaults to before the change ('warn') so doesn't log http requests.

Also changes log format to be single line for most output to make parsing it a lot easier


knolleary:
Only nit would be an update to the docs at the same time... but in the interests of getting staging updated lets run with this. Please follow up with a doc pr.

PR #805, opened 2022-07-26
Centralise how tests initialise Forge app

### LINKED ISSUE #770: Postgres test coverage not getting enabled for all tests
### Description

The recent work to add postgres coverage to the tests is not quite reaching all tests.

It is enabled for `test/unit/forge/routes/**` by virtue of the changes added to `test/unit/forge/routes/setup.js`.

However other test areas use their own `setup.js` that doesn't include the postgres enabling code.

We should rationalise the `setup.js` files to avoid duplication and this type of missed update.

### Epic/Story

_No response_

knolleary:
Fixes #770

Reduces code duplication and ensures all tests are enabled for postgres testing

The basic pattern for tests to follow is:

```
// Load the common test utils module
const FF_UTIL = require('flowforge-test-utils')

// Initialise the app with whatever config is required
const app = await FF_UTIL.setupApp({
    license: '....'
})
```

The `FF_UTIL.setupApp` function includes a default app config suitable for most tests. The passed-in configuration augments (rather than replaces) that default config.

It also contains the logic to insert the postgres db config if the appropriate env vars are set to enable postgres testing.

### DIFF
--- a/forge/ee/db/controllers/Subscription.js
+++ b/forge/ee/db/controllers/Subscription.js
@@ -5,7 +5,7 @@ module.exports = {
             customer: customer,
             subscription: subscription
         })
-        newSubscription.setTeam(team)
+        await newSubscription.setTeam(team)
 
         return newSubscription
     },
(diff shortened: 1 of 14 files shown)


knolleary:
As of this PR, the default test configuration is:
```
{
        telemetry: { enabled: false },
        logging: {
            level: 'warn'
        },
        db: {
            type: 'sqlite',
            storage: ':memory:'
        },
        email: {
            enabled: true,
            transport: new LocalTransport()
        },
        driver: {
            type: 'stub'
        },
        broker: {
            url: ':test:'
        }
}
```

hardillb (approved):

PR #807, opened 2022-07-27
Add docs about setting up MQTT broker for K8s/Docker

hardillb:
also add notes about running broker in docker for localfs


docs/install/docker/README.md
```diff
@@ -61,6 +61,39 @@ You also need to update the `VIRTUAL_HOST` entry in the `docker-compose.yml` fil
 
 For more details on the options available, see the [configuration guide](../configuration.md).
 
+### MQTT Broker
+
+FlowForge includes a MQTT broker to allow projects/devices to comunicate with core platform. This will be enabled by default, 
+you can dissable it by commenting out the following lines in the `docker-compose.yml` file
```
knolleary, line 67: Do we really want to spell out how to disable it?

The localfs instructions say it is optional as an eass-of-use thing for quickly trying out the platform. I'm not sure we want to encourage disabling it in the other environments.

docs/install/local/README.md
```diff
@@ -118,7 +118,43 @@ Once installed, you will need to build and install the authentication plugin.
...
+    You will need to customise the values to match your local configuration:
+     - `auth_opt_http_host` value to match the IP address of either the docker0 interface or the external IP address of the host machine
+     - `auth_opt_http_port` if you have changed the port the forge platform is running on
+
+3. Start the container with the following command
+    ```
+    docker run -d -v /full/path/to/mosquitto.conf:/etc/mosquitto/mosquitto.conf -p 1883:1883 -p 1884:1884 -n flowforge-broker iegomez/mosquitto-go-auth
+    ```
+
+    This will map the `1883`/`1884` ports to the host machine so they can be accessed by the
```
knolleary, line 156: Incomplete sentence....

```suggestion
    This will map the `1883`/`1884` ports to the host machine so they can be accessed outside of the container. If you already have an MQTT broker running on port 1883, then you'll need to modify the `-p` options to use a different set of ports. For example: `-p 9883:1883 -p 9884:1884`.
```

docs/install/docker/README.md
hardillb: OK, just wanted to be clear that it's optional

knolleary (approved):

Issue #300, opened 2022-02-15
Allow user to select default team in User Settings


knolleary:
### Description

Add options to the User Settings page to set the default team.

This task depends on #299 being delivered first.

### Epic/Story

#298

knolleary:
Removing from the product board as this is a subtask of #298 which covers the overall sizing/design

Issue #298, opened 2022-02-15
Allow user to set preferred default team


knolleary:
### Description

**As a:** User

**I want to:** set a default team

**So that:** when I log into the FlowForge platform, it takes me to my preferred team, rather than the first team in the list.

Tasks:

 - [x] #299 
 - [ ] #300

Pezmc:
Verified (this is from 0.8 but I'm cleaning up)

PR #810, opened 2022-07-27
Add defaultTeam to User model and enable API to modify

### LINKED ISSUE #299: Add default team property to User model
### Description

As part of #298 we need to be able to store the user's preferred team.

In the scenario a user is invited to a team as well as having a personal team created, the team they then get shown will be whichever is alphabetically first (I think... or it could be in creation order... - either way, it's a bit random)

The MVP is:

 - [ ] Add `defaultTeam` to User model
 - [ ] Create migration to add the column
 - [ ] Update `checkState` in UI to use that value when picking which team to show (if not otherwise determined from the url)

### Epic/Story

#298

knolleary:
Closes #299  #298 #300

 - Adds `defaultTeam` column to User table
 - Update User Settings endpoints modified to accept optional `defaultTeam` value to set it.
    - Validates the user is a member of the team - responds with 400 if not valid
    - Clears the user's defaultTeam if they are removed from that team
 - Adds Default Team to the 'User Settings' page and allows the user to change it
 - When the user logs in, it takes them to this team (if set) rather than the first in the list

image

### DIFF
--- a/frontend/src/pages/account/Settings.vue
+++ b/frontend/src/pages/account/Settings.vue
@@ -1,12 +1,26 @@
 <template>
     <ff-loading v-if="loading" message="" />
     <form v-else class="space-y-6">
-        <FormRow v-model="input.username" :error="errors.username">Username</FormRow>
-        <FormRow v-model="input.name" :placeholder="input.username" :error="errors.name">Name</FormRow>
-        <FormRow v-model="input.email" :error="errors.email">Email</FormRow>
+        <FormRow v-model="input.username" :type="editing?'text':'uneditable'" :error="errors.username">Username</FormRow>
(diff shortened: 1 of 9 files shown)


frontend/src/pages/account/Settings.vue
```diff
@@ -1,12 +1,26 @@
 <template>
     <ff-loading v-if="loading" message="" />
     <form v-else class="space-y-6">
```
joepavitt, line 3: can we add an `@submit` here please? Even if it's just an `@submit.enter.prevent=""` to disable the behaviour.

knolleary: Added

joepavitt (approved):

PR #811, opened 2022-07-27
Handle "@submit" events on forms in dialog boxes

joepavitt:
Closes https://github.com/flowforge/flowforge/issues/334

### DIFF
--- a/frontend/src/pages/admin/Stacks/dialogs/AdminStackEditDialog.vue
+++ b/frontend/src/pages/admin/Stacks/dialogs/AdminStackEditDialog.vue
@@ -2,7 +2,7 @@
     <ff-dialog :open="isOpen" :header="dialogTitle" @confirm="isOpen = false">
         <template v-slot:default>
             <ff-loading v-if="loading" message="Creating Stack..."/>
-            <form v-else class="space-y-6" @submit="confirm()">
+            <form v-else class="space-y-6" @submit.prevent>
                 <div v-if="this.input.replaces">
                     This will create a new stack to replace '{{input.replaces.name}}'.
                     The existing stack will be marked inactive and will not be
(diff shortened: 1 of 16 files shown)


knolleary:
The original problem is that hitting enter in an `<input>` is causing the form to be submitted - which reloads the page rather than save the changes as expected. That sucked and needed fixing. However I'm not 100% on whether this approach is the right one.

My concern is with the approach of this fix is that now hitting enter in *any* input is now the same as clicking the submit button. If I have a form with 5 inputs I may hit enter in the first input expecting to move to the next input and carry on filling in the form, whereas this will trigger the completion of the form. That can lead to premature form submission - and given we have some forms with fields that cannot be edited after they are set, that could lead to unexpected/frustrating behaviour.

Would it be better to just default the submit action all together, and *only* complete the form when the appropriate button is clicked (which can be achieved via tab navigation and hitting 'enter' to ensure keyboard accessibility).

joepavitt:
> Would it be better to just disable the submit action all together, and only complete the form when the appropriate button is clicked

I am okay with this

joepavitt:
@knolleary done

frontend/src/pages/admin/Stacks/dialogs/AdminStackEditDialog.vue
```diff
@@ -144,59 +144,62 @@ export default {
...
-                opts.properties[prop.name] = this.input.properties[prop.name]
-            })
-
-            if (this.stack) {
-                if (this.editDisabled) {
-                    opts = { active: this.input.active }
-                    if (!this.editTypeDisabled && this.input.projectType) {
-                        opts.projectType = this.input.projectType
-                    }
+            console.log('confirm')
```
knolleary, line 147:
```suggestion
```

frontend/src/pages/project/Settings/dialogs/ConfirmProjectDeleteDialog.vue
```diff
@@ -50,15 +50,18 @@ export default {
...
+                this.isOpen = false
+            }
         }
     },
     setup () {
         const isOpen = ref(false)
         return {
             isOpen,
             close () {
+                console.log('close')
```
knolleary, line 64:
```suggestion
```

knolleary (approved):
Couple console.log's to squash, but otherwise LGTM

PR #818, opened 2022-07-29
Added ff- prefix to reflect current filenames

Steveorevo:
This change helped resolve an issue that was mentioned in https://github.com/flowforge/flowforge/issues/806#issuecomment-1198808084

### DIFF
--- a/docs/install/local/stacks.md
+++ b/docs/install/local/stacks.md
@@ -24,7 +24,7 @@ If you are developing FlowForge having checked it out from GitHub then you can r
 the following command in the project root
 
 ```bash
-npm run install-stack --vers=2.2.2
+npm run ff-install-stack --vers=2.2.2
 ```
 
 ### Production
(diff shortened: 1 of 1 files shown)


docs/install/local/stacks.md
```diff
@@ -24,7 +24,7 @@ If you are developing FlowForge having checked it out from GitHub then you can r
 the following command in the project root
 
 ```bash
-npm run install-stack --vers=2.2.2
+npm run ff-install-stack --vers=2.2.2
```
knolleary, line 27:
```suggestion
npm run install-stack --vers=2.2.2
```

When running in the local development setup, the `npm` task is called `install-stack`.

knolleary:
Thanks for this @Steveorevo - I reverted the change to the `npm run` task as that doesn't have the `ff-` prefix. Whether we should align them is another matter. 😉

Issue #662, opened 2022-06-03
Project In / Project Out nodes for interproject communication


ZJvandeWeg:
### Description

**As a:** Developer

**I want to:** Use a custom node to send messages to another gateway

**So that:** it can continue to process the data

### Acceptance Criteria

- [ ] Custom node is introduced: Project In / Project Out like Link In/Link Out
- [ ] MQTT is used as transport, but is transparent (No knowledge of topics or the like)
- [ ] Projects of a team can be selected to transport to, no others.

knolleary:
This is blocked by #464 - and more specifically, the comments on security [here](https://github.com/flowforge/flowforge/issues/464#issuecomment-1146983929)

Steve-Mcl:
All - first up, apologies if the below proposal is a tough read, there are many concepts still floating around and I can not keep iterating in my head - so will post what I have as a basis to grow from. 

I will continue to review and refine in the coming days but would greatly appreciate thoughts and direction.

# Project links design notes

## Concepts
The high level concept of project links is **inbox**/**outbox**  
**inbox** being what is sent directly to a project (a project/device has its own inbox)  
**outbox** being what a project broadcasts to anyone who cares to listen  

Other notes...
* Some nodes will need a project selector that it gets from API
* To support a tree like "picker" (as per link-in/out) nodes, it would be necessary to populate $hidden topics for discoverability
  * This is potentially negated if we permit free form `sub-topic` design
* Node will likely need settings entry in settings.js (to point to API atc)
  * If not found, project-link nodes should not be added to palette & raise error in console
* Instance will likely need Env Vars (device ID etc)


## Topic structures

### Details (From the proposed Subscription ACLs)
Subscription ACLs | Can subscribe to... | Publish ACLs | Can publish to...
-|-|-|-
Receive messages from other projects | `ff/v1/<team>/p/<own-project>/in/+/#` | Send messages to other projects | `ff/v1/<team>/p/<any-project>/in/+/#`
Receive broadcasts from other projects | `ff/v1/<team>/p/<any-project>/out/+/#` | Send broadcast message | `ff/v1/<team>/p/<own-project>/out/+/#`

### project to project
* `ff/v1/<team>/p/<any-project>/in/<level1>/<leveln>`
  * `<any-project>` is the target you want to send to
  * `<level1>` is equivalent to `motor1` in the example `motor/1/speed`
  * `<leveln>` is equivalent to `1` and `speed` in the example `motor/1/speed`
[...]

PR #820, opened 2022-07-29
Allow limited access to project list for anonymous

Steve-Mcl:
part of #662

### DIFF
--- a/forge/routes/api/team.js
+++ b/forge/routes/api/team.js
@@ -22,6 +22,10 @@ module.exports = async function (app) {
         if (request.params.teamId !== undefined) {
             if (request.params.teamId) {
                 try {
+                    // if User is not defined, check to see if request.context.config.allowAnonymous is set
+                    if (request.session?.User === undefined && request.context.config?.allowAnonymous === true) {
+                        return
+                    }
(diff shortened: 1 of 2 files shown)


forge/routes/api/team.js
```diff
@@ -88,10 +92,16 @@ module.exports = async function (app) {
         }
     })
 
-    app.get('/:teamId/projects', async (request, reply) => {
+    app.get('/:teamId/projects', { config: { allowAnonymous: true } }, async (request, reply) => {
```
knolleary, line 95: `allowAnonymous` is too open. That would allow anyone to query the list of projects without any authentication.

We already apply `verifyTokenOrSession` to all api routes, so the project token should be able to access this end point without any special handling. But having realised that, I see that is also too wide open.

I am going to merge this as-is, but will followup with a PR to improve the auth handling more generally and tighten things up for Token access.

knolleary (approved):

forge/routes/api/team.js
Steve-Mcl: Nick, I am sure you realise this but even though i used the name `allowAnonymous` the token must be present otherwise this endpoint returns `Unauthorized` ...

image

... due to existing pathways

PR #824, opened 2022-07-29
Prevent user with unverified email from accessing API

knolleary:
Fixes https://github.com/flowforge/security/issues/3

This restricts the admin api so users with unverified emails cannot call it.
It is possible to poke holes in the api for specific routes by adding `{ config: { allowUnverifiedEmail: true } }`

This also reworks the UX for such a user. It is now consistent with the 'expired password' ux - they don't get in to the platform to try doing anything:

image

### DIFF
--- a/forge/ee/routes/index.js
+++ b/forge/ee/routes/index.js
@@ -5,7 +5,7 @@
  * @memberof forge.ee
  */
 module.exports = async function (app) {
-    app.addHook('preHandler', app.verifyTokenOrSession)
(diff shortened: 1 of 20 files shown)


knolleary:
One minor issue with the UX change here is there is no way for the user to log out. They will forever get this screen until they verify or the session cookie expires.

Bit of an edge case, but if you have multiple accounts and are testing the platform, you find yourself stuck and needing to open incognito browser windows.

joepavitt (approved):

PR #827, opened 2022-07-29
Ensure access tokens can only access specific allowed routes

knolleary:
This removes `verifyTokenOrSession` as we want to be more deliberate and explicit over what routes can be accessed by a token, rather than a login session.

We have stories related to user-generated access tokens. When we come to them, we will need to carefully open it back up.

There are three sets of end points that need to be token-accessible:

 - `/api/v1/user` - used as part of `nr-auth` to get a user's identify when logging into node-red editor
 - `/api/v1/teams/<teamid>/projects` - used by Project Nodes to get a list of projects in the team. I've added logic to verify the token being used belongs to a project owned by the team it is accessing.
 - `/api/v1/devices/<device>/live/*` - used by device agent

I have verified all of those routes behave themselves with this change.

### DIFF
--- a/forge/routes/api/team.js
+++ b/forge/routes/api/team.js
@@ -22,8 +22,27 @@ module.exports = async function (app) {
         if (request.params.teamId !== undefined) {
             if (request.params.teamId) {
                 try {
-                    // if User is not defined, check to see if request.context.config.allowAnonymous is set
-                    if (request.session?.User === undefined && request.context.config?.allowAnonymous === true) {
+                    if (!request.session.User) {
+                        // If request.session.User is not defined, this request is being
+                        // made with an access token. If it is a project access token,
+                        // ensure that project is in this team
+                        if (request.session.ownerType === 'project') {
+                            // Want this to be as small a query as possible. Sequelize
+                            // doesn't make it easy to just get `TeamId` without doing
+                            // a join on Team table.
+                            const project = await app.db.models.Project.findOne({
+                                where: { id: request.session.ownerId },
+                                include: {
+                                    model: app.db.models.Team,
+                                    attributes: ['hashid', 'id']
+                                }
+                            })
+                            // Ensure the token's project is in the team being accessed
+                            if (project && project.Team.hashid === request.params.teamId) {
+                                return
+                            }
+                        }
+                        reply.code(404).type('text/html').send('Not Found')
                         return
                     }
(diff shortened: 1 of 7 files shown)


forge/routes/api/team.js
```diff
@@ -22,8 +22,27 @@ module.exports = async function (app) {
...
+                                    model: app.db.models.Team,
+                                    attributes: ['hashid', 'id']
+                                }
+                            })
+                            // Ensure the token's project is in the team being accessed
+                            if (project && project.Team.hashid === request.params.teamId) {
+                                return
+                            }
+                        }
+                        reply.code(404).type('text/html').send('Not Found')
```
Steve-Mcl, line 45: Nick, the original PR had `if (request.session?.User === undefined && request.context.config?.allowAnonymous === true) {` as there were instances while developing and restarting & jiggling about, the `session` object was not present. 

This ended up in a hard to escape loop.  I didnt manage to re-create this reliably and assumed this was due to the dev work BUT I suspect it was due to log debug periods and session timeout?

Steve-Mcl (commented):
That actually looks far better than just checking the user is `undefined`

Question: when verifying using token, does the `request.session.User` get populated now or will it still be empty? I ask as I use its absence of `User` to do limit the contents of the project listing in the API (Line 118 `const limitResponse = ...` in [forge/routes/api/team.js](https://github.com/flowforge/flowforge/pull/827/files#diff-7864e255717751283432582df0983b856ea2a936271bbef9cebe69c903c199bf))

forge/routes/api/team.js
knolleary: With `allowAnonymous` set to `true` then yes, it would be possible for `request.session` to be undefined because `allowAnonymous` means allow requests that don't have a valid session or token.

That's why I've removed its use here.

Steve-Mcl: ah, makes sense. Thanks for clarifying.

knolleary:
Tokens are not (currently) associated with a user.  At that point in the code, `request.session` **will** be defined because it has passed through authentication, but `request.session.User` will not be defined because it is token based auth.

It would technically be more accurate to use as the check `request.session.ownerType === 'project'` as that is more explicit. I'll update the PR shortly.

knolleary:
Change pushed. This is good for final review/merge.

Steve-Mcl (approved):

hardillb (approved):

PR #829, opened 2022-07-29
Remove all manual show/close logic for ff-dialog (now contained within the component)

joepavitt:
Requires latest forge-ui-components to test locally. Will be releasing new forge-ui-components major version.

This is the first step in updating our dialog workflow. This PR utilises the fact that the internals of `ff-dialog` now better handle the show/close logic and provide control over the default cancel/confirm buttons via properties, rather than having to manually template it each time.

Future work to do here still in making a single JS API that allows devs to call single line or two to open and action a dialog. Discussion on full plans in https://github.com/flowforge/flowforge/issues/830


knolleary (approved):
I think I've touched most of the dialogs here.

Did find one issue with the InviteUser dialog (was a preexisting issue in `main` around the var name used to disable the buttons) - which I've fixed up here.

PR #834, opened 2022-08-01
Show username instead of ID in alert on removal

sammachin:
fixes https://github.com/flowforge/flowforge/issues/833

https://user-images.githubusercontent.com/187645/182179821-228d2cb0-6a12-4030-be4f-6885e6dc4a24.mp4


ZJvandeWeg:
Thanks for the quick fix! LGTM

PR #842, opened 2022-08-03
Ensure expired password users cannot access API

knolleary:
- Adds `allowExpiredPassword` flag for API routes that should be accessible to users with expired passwords.
 - Blocks access to all api end points otherwise
 - Opens access to `/api/v1/user` and `/api/v1/user/change_password`
 - Fixes a focus issue in the Change Password dialog
 - Adds a 'logout' option to both the Change Password and Verify Email dialogs - so users don't get stuck in them

image


Steve-Mcl (approved):
all goodness

PR #845, opened 2022-08-03
Notify user of API errors

joepavitt:
https://user-images.githubusercontent.com/99246719/182592818-19c8d240-00fe-4939-bd8b-c266bce242e1.mov

### DIFF
--- a/frontend/src/api/client.js
+++ b/frontend/src/api/client.js
@@ -1,6 +1,8 @@
 import axios from 'axios'
 import store from '@/store'
 
+import Alerts from '@/services/alerts'
+
 const client = axios.create({
     headers: {
         'Content-Type': 'application/json'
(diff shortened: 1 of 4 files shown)


frontend/src/api/client.js
```diff
@@ -22,4 +25,16 @@ client.interceptors.response.use(function (response) {
     return Promise.reject(error)
 })
 
+// 500 Internal Server Errors
+client.interceptors.response.use(function (response) {
```
knolleary, line 29: Personally would have put this all into the existing interceptor to save every request having to pass through an extra Promise. But this is okay.

knolleary (approved):

frontend/src/api/client.js
joepavitt: Yeah, that's fair, I did it more for cleanliness of code/reading. Will happily merge them in the future.

PR #846, opened 2022-08-03
Select Tiles for "Create Project" & "Admin > Project Types"

### LINKED ISSUE #791: Update selection tile to use ff-selection-tile
### Description

With the new `selection-tile` available in `forge-ui-components` ([PR](https://github.com/flowforge/forge-ui-components/issues/29)) we should update the instances of this UX in FlowForge to utilise this new component and create more consistent visual appearance across the app.

### Epic/Story

_No response_

joepavitt:
Closes #791 

https://user-images.githubusercontent.com/99246719/182600865-95a835fd-bf6c-4d16-8576-8f39b84e5f1e.mov


package.json
```diff
@@ -49,7 +49,7 @@
         "@fastify/csrf-protection": "^4.0.1",
         "@fastify/helmet": "8.1.0",
         "@fastify/static": "^5.0.2",
-        "@flowforge/forge-ui-components": "^0.3.0",
+        "@flowforge/forge-ui-components": "0.3.1",
```
knolleary, line 52:
```suggestion
        "@flowforge/forge-ui-components": "^0.3.1",
```

knolleary (commented):
We had previously said we wouldn't publicise the internal details of the stacks - such as memory and cpu.

Only the localfs driver provides a `node-red` property on its stack - the others do not have that property; they have a container identifier.

This needs a wider discussion as to whether we want to go back on that original decision. @sammachin

package.json
sammachin: agreed, certainly within ff cloud our approach is to use small, medium and not directly quote the underlying resources.
So for this PR I would favour just showing the name of the stack for now, we have the description field within the type where more information can be provided and at present all our stacks within a project type would have the same capacity anyway.

There is some UX improvements to consider around how we present the stack, as just the unique name field isn't the most user friendly, but I would favor adding a description box like the project type and allow free text here.

joepavitt: In that case, I vote a hold on this for the 0.8 Release to make sure we get this right.

knolleary:
Can I suggest a reduced scope that doesn't change the Stack selection ux to use the new tiles until the point around exposing stack properties is decided on, but in the mean time, move forward with replacing the Project Type tiles with the updated UX?

joepavitt:
> Can I suggest a reduced scope that doesn't change the Stack selection ux to use the new tiles until the point around exposing stack properties is decided on, but in the mean time, move forward with replacing the Project Type tiles with the updated UX?

Sounds reasonable. I was also toying with the idea of still using the tiles, but without a description, then it's easier to include it back in later.

There is an example of it in the vide as it's what I did for Templates.

knolleary:
> I was also toying with the idea of still using the tiles, but without a description, then it's easier to include it back in later.

Happy with whichever approach moves things forward.

For the TeamTypes work, I've been digging through a lot of the billing logic and have found some changes to make. Alongside that, I found the user can access the Create Project page even if billing hasn't been setup on the team. This is because the original version of the page let you pick which team to create the project in - so doesn't do any checks on the 'current project'. All of that code is still in the page, but it serves no purpose.. long story short, I need to tidy that up, but would rather have your changes in place first to save the merge conflicts.

joepavitt:
Removed meta/description for now, keeping just the Stack name:

Screenshot 2022-08-10 at 09 03 43

knolleary:
Looks good.

Only issue (which is upstream in ui-components) is how the tile selection overflows. I think `<ff-tile-selection>` needs flex wrap adding?

image


Happy to merge this part as-is.

frontend/src/pages/admin/ProjectTypes/dialogs/ProjectTypeEditDialog.vue
```diff
@@ -1,6 +1,7 @@
 <template>
-    <ff-dialog ref="dialog" :header="dialogTitle" :confirm-label="projectType ? 'Update' : 'Create'" @confirm="confirm()" :disable-primary="!formValid">
+    <ff-dialog ref="dialog" :header="dialogTitle">
         <template v-slot:default>
+            type: {{ projectType ? 'update' : 'create '}}
```
knolleary, line 4:
```suggestion
```

Looks like a debug line?

joepavitt: good catch!

PR #863, opened 2022-08-09
Always respond to stripe webhooks with 200 status unless fatal

### LINKED ISSUE #862: Stripe Webhooks should always return 200 unless they need to be retried
### Description

Currently we respond to stripe webhooks that identify an unknown team with a 404 response. This will cause Stripe to retry the webhook a handful of times over the next day or so.

If the team doesn't exist at the time the webhook is called, then it isn't going to exist later.

I think it would be better to log the invalid team and respond with a 200 as we don't want the webhook to be retried.

This would also reduce issues in the test environment when developers are working in their own test environments at the same time with the same billing keys - as they will be receiving each other's webhooks.

### Epic/Story

_No response_

knolleary:
Fixes #862 

This ensures stripe won't keep retrying webhook calls for teams that don't exist.

### DIFF
--- a/forge/ee/routes/billing/index.js
+++ b/forge/ee/routes/billing/index.js
@@ -65,7 +65,7 @@ module.exports = async function (app) {
                 try {
                     event = stripe.webhooks.constructEvent(request.rawBody, sig, app.config.billing.stripe.wh_secret)
                 } catch (err) {
-                    console.log(err)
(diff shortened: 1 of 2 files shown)


hardillb (approved):

hardillb:
The changes look good, but looks like I had added tests that expected the 404's

knolleary:
Tests fixed

hardillb (approved):

Issue #858, opened 2022-08-08
Non owner user can "Set as Device Target"


Steve-Mcl:
### Current Behavior

It is possible as a non-owner user to operate snapshot action "Set as Device Target"

### Expected Behavior

Should fail with error 403

### Steps To Reproduce

* Log in to FF as non owner
* Select a project with devices
* View snapshots
* Select a snapshot and operate "Set as Device Target"

### Environment

- FlowForge version: 0.8
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

sammachin:
Set as Device Target is akin to deploying a flow just in this case its deploying to a device, Therefore we need to consider if this is the desired behaviour or if this is indeed a bug.

Steve-Mcl:
@sammachin thanks, as discussed I will update https://github.com/flowforge/flowforge/tree/main/docs/user/team so that we have a better footing for this issue and #854

PR #854, opened 2022-08-08
Hide actions menu from non owner

Steve-Mcl:
Ensures the Project -> action menu is not visible to "non owner"

_PR is currently draft as waiting feedback on other items (see comments in issue #843)_

Issue #843, opened 2022-08-03
As a team member, I can see and click "start" and "stop" but nothing happens


Steve-Mcl:
### Current Behavior

Start / Stop menu items are visible & clickable by a team member. 

Clicking these items causes a 403 in browser console but this is not surfaced to user (has no idea it did/did not work)

![image](https://user-images.githubusercontent.com/44235289/182591432-f2ad70ea-2c86-45b5-a5ae-4a7590548cad.png)

### Expected Behavior

* remove or disable entries 
* surface the issue with a toast

### Steps To Reproduce

* log in as team member
* operate project Stop or Start

### Environment

- FlowForge version: 0.7.git
- Node.js version: 16
- npm version:
- Platform/OS: Windows
- Browser: All

sammachin:
Confirmed this issue exists on the current 0.7 production server so it is not a regression, will schedule for 0.9

joepavitt:
Yeah, definitely an issue, I don't know of any reason why that should 403.

knolleary:
The 403 is the correct response.

The issue is that a regular team member (not owner) should not be presented the options to start/stop the project.

Steve-Mcl:
@joepavitt @knolleary 

In working though this I note other places where "owner" actions are visible/enabled.

Please review the 6 concern areas and advise on correct operation.


1. Snapshots -> create snapshot is visible to non owner and results in an unsurfaced 403
![image](https://user-images.githubusercontent.com/44235289/183378107-d9b37d00-4b1d-4385-840e-3e7ecd8183dd.png)

2. Snapshots -> .
    (1). Menu is visible to non owner
    (2). Rollback results in 403
    (3). Set as Device Target actually changes the device target
    (4). Delete Snapshot results in 403
![image](https://user-images.githubusercontent.com/44235289/183379911-b923f1fc-d552-4db0-8d5f-413b1aee5f86.png)


3. Device Settings, General -> "Edit Device" button visible and dialog appears when clicked but results in an unsurfaced 403 upon saving edits
![image](https://user-images.githubusercontent.com/44235289/183378768-a6980cea-2949-45f3-87fa-7d75cb630c0b.png)

4. Device Settings, Env Variables -> non owner can add env vars but upon saving, the variable disappears  and results in an unsurfaced 403
![chrome_sqTNko1p9x](https://user-images.githubusercontent.com/44235289/183378139-05901b69-8f0a-4c0e-92ba-1da467e38135.gif)

[...]

PR #865, opened 2022-08-09
docs: Update RBAC table

Steve-Mcl:
prep for #854, #858

### DIFF
--- a/docs/user/team/README.md
+++ b/docs/user/team/README.md
@@ -52,25 +52,26 @@ describe the different Roles apply for all projects in the team
 | • Activity                           | ✓     | ✓      |
 | • Snapshots...                       |       |        |
 | &emsp; • View Snapshots                 | ✓     | ✓      |
-| &emsp; • Create Snapshot                | ✓     | -      |
-| &emsp; • Rollback                       | ✓     | -      |
+| &emsp; • Create Snapshot                | ✓     | ✓      |
+| &emsp; • Rollback                       | ✓     | ✓      |
 | &emsp; • Set as Device Target           | ✓     | ✓      |
 | &emsp; • Delete Snapshot                | ✓     | -      |
 | • Devices...                         |       |        |
 | &emsp; • View Devices                   | ✓     | ✓      |
 | &emsp; • Register Device                | ✓     | -      |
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
@sammachin please add additional reviewers if you believe them necessary.

sammachin:
Thanks @Steve-Mcl this is really useful, just to confirm this is the current state as of 0.8.0?
In which case I think we should merge this as is and then iterate on where we feel changes need to be made on a case by case basis.

Steve-Mcl:
@sammachin 

> just to confirm this is the current state as of 0.8.0?

Kind of - let me try to clarify

It is mostly`*` the operational state as found when operating buttons / navigating while logged in as a "non owner".  The visual/UI state is often incorrect as of 0.8. 
Example - in `#843` there are [6 places](https://github.com/flowforge/flowforge/issues/843#issuecomment-1207855648) (there are more) where a button/menu/action is visible/clickable but should (probably) not be.  Therefor this PR aims to determine correct RBACs

So, ideally, I need eyes on this to double check my assumptions/findings/sensible guesses as to what it should be both from a front end and back end perspective.

Perhaps we need to add Nick's or Ben's eyes too as they developed many of the features highlighted?


<br>

<small>`*` some things like "viewing environment variables" are the desired condition following discussion with Nick</small>

sammachin:
Few items to change, mostly giving members more access, 
In principle a member should be able to do anything to build and deploy a project to a device but they can't do anything chargeable, destructive, or that changes the membership of the team (other than leave the team themselves)

A Member should be able to 
Snapshots
	Create Snapshot
	Rollback
Devices
	Edit Environment Variables
	Device Row Menu (but not delete device)
	Device Settings
		Edit Environment Variables
Settings
	Edit Environment Variables

They should also be able to assign a device from the team to a project, can't see where that is in the list?

sammachin (changes requested):
as detailed in comments

Steve-Mcl:
> as detailed in comments

Changes applied in [a153cba](https://github.com/flowforge/flowforge/pull/865/commits/a153cba898a210d5e2fbfd395ba0277dd13af688)

PR #869, opened 2022-08-10
E2E Testing Infrastructure

### LINKED ISSUE #816: Finish infrastructure to run E2E testing for the frontend.
### Description

Work was started to get E2E tests running for the frontend, but hit a stumbling block when trying to setup (and reset) different configurations/instances of FlowForge between testing files. This needs to be finished.

### Epic/Story

_No response_

joepavitt:
Closes #816 

Includes the relevant configs & GitHub actions in order to automatically run Frontend E2E tests on git push and PR merges (limited only to changes in the `/frontend` folder or the `test/e2e/frontend` folders).

Included are two `.spec.js` test files with foundational tests in that currently test login and Platform Admin access (need to be fleshed out further). 

We have Cypress `commands` defined:
1. That handle our login via a Cypress `session` (`cy.login()`) which makes switching accounts a lot easier. 
2. We also automate checking of the flurry of API calls triggered by the UI's state management on load of the application in another custom `command` and ensure we reset our view each time - `cy.home()`

Once this is merged (and ideally the [PostHog work](https://github.com/flowforge/flowforge/pull/864) to give us easier access to buttons/actions), our next testing priority should be https://github.com/flowforge/flowforge/issues/867 in order to flesh out the E2E coverage.


.github/workflows/test-ui-e2e.yml
```diff
@@ -0,0 +1,28 @@
+name: UI E2E (Cypress) Tests
+
+on:
+  push:
+    branches: [ main ]
+    paths:
+      - 'frontend/**'
+      - 'test/e2e/frontend/**'
+  pull_request:
+    branches: [ main ]
```
hardillb, line 10: Should we be only running this on PRs that match the same paths as the push?

joepavitt: you're absolutely right - will update

joepavitt:
@hardillb done

hardillb (approved):

hardillb:
Looks to have worked nicely in the when the tests were run for PR so happy for this to be merged

package.json
```diff
@@ -17,6 +18,10 @@
         "test:unit:frontend": "vitest run --config ./config/vitest.config.ts test/unit/frontend/**/*.spec.js",
         "test:unit:frontend-watch": "vitest --config ./config/vitest.config.ts test/unit/frontend/**/*.spec.js",
         "test:system": "mocha 'test/system/**/*_spec.js' --timeout 5000",
+        "test:e2e": "mocha test/system/**/*_spec.js",
+        "test:e2e:frontend": "start-server-and-test serve:test-env http-get://localhost:3000 cy:run",
```
knolleary, line 22: Is there a reason the `test:e2e` script was introduced when we already had `test:system` that does exactly the same thing?

We should use a consistent name. We already had `system` in place and is documented - and there is no docs for the `e2e` name currently. I don't mind which way it goes, but we shouldn't have two names for the same thing.

joepavitt: Legacy calls that were accidentally left in when trialing different testing frameworks. No longer required and will remove with a PR

joepavitt: see: https://github.com/flowforge/flowforge/pull/869

PR #870, opened 2022-08-10
Add billingEnabled flag to team api response to save subscription check

knolleary:
This PR adds a `billingEnabled` flag on the response to `/api/v1/teams/:teamId` (when the billing feature is enabled on the platform). This indicates whether the team has an active billing subscription.

This removes the need for the platform to make the extra call to get the subscription info just to check if it has to redirect to the billing setup page.

If `billingEnabled` is false, it will redirect straight to the billing page for the team.

This PR also tidies up the `createProject` page. In the early days, this page allowed you to pick which team you would create the project in. We removed that as an option some time ago, but the code was still there to load the list of teams and figure out which one it should create the team in. But that meant the page couldn't do a sensible billing check - and a user was enable to enter the url of the page directly to access it and then try to create a project.

The page now uses the `team` from the account store - ie the current team - which means it can also check the `billingEnabled` flag, and redirect to the billing page if needed.

### DIFF
--- a/frontend/src/pages/team/Billing.vue
+++ b/frontend/src/pages/team/Billing.vue
@@ -19,7 +19,7 @@
         <div v-else>
             Billing has not yet been configured for this team. Before proceeding further, you must continue to Stripe and complete this.
             <div class="mt-3">
-                <ff-button size="small" @click="setupBilling()">
+                <ff-button @click="setupBilling()">
                     <template v-slot:icon-right><ExternalLinkIcon /></template>
                     Setup Payment Details
                 </ff-button>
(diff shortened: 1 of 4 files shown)


joepavitt:
@knolleary Being picky on language, is `billingEnabled` the right term? `billingSetup` may be more appropriate? As a platform we enable/disable billing, so technically billing is enabled for all teams. teams may not then have actually configured/setup their own payment details?

knolleary:
> billingSetup may be more appropriate?

Yup - happy with that.

frontend/src/pages/team/createProject.vue
```diff
@@ -229,6 +205,13 @@ export default {
             }
         }, 100)
     },
+    async beforeMount () {
+        if (this.features.billing && !this.team.billingEnabled) {
```
joepavitt, line 209:
```suggestion
        if (this.features.billing && !this.team.billingSetup) {
```

frontend/src/pages/team/Billing.vue
```diff
@@ -91,19 +91,23 @@ export default {
...
         this.loading = true
-        try {
-            const billingSubscription = await billingApi.getSubscriptionInfo(this.team.id)
-            billingSubscription.next_billing_date = billingSubscription.next_billing_date * 1000 // API returns Seconds, JS expects miliseconds
-            this.subscription = billingSubscription
-            this.subscription.items.map((item) => {
-                item.total_price = item.unit_price * item.quantity
-                return item
-            })
+        if (!this.team.billingEnabled) {
```
joepavitt, line 94:
```suggestion
        if (!this.team.billingSetup) {
```

frontend/src/pages/team/index.vue
```diff
@@ -51,18 +49,10 @@ export default {
...
-                } catch (err) {
-                    const path = '/team/' + this.team.slug + '/billing'
-                    // if 404 - no billing setup, but are we running in EE?
-                    if (err.response.status === 404) {
-                        this.$router.push({
-                            path
-                        })
-                    }
-                }
+            if (this.features.billing && !this.team.billingEnabled) {
```
joepavitt, line 52:
```suggestion
            if (this.features.billing && !this.team.billingSetup) {
```

joepavitt (changes requested):

frontend/src/pages/team/createProject.vue
```diff
@@ -229,6 +205,13 @@ export default {
             }
         }, 100)
     },
+    async beforeMount () {
```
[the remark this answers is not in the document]
joepavitt: Ah, ignore me, `CreateProject` is not a route child of `index.vue` - so doesn't inherit this check. Well spotted.

knolleary:
All suggested changes applied

joepavitt (approved):

PR #872, opened 2022-08-10
Rework team create flow to a single screen when billing enabled

knolleary:
This simplifies the Create Team workflow when billing is enabled.

The previous flow involved two screens:

(ignore the TeamType selection in the top image... forgot to remove that when I took the screenshot)

image (2)

Having created lots of teams as part of the TeamTypes work, I find the two screen approach to be cumbersome and doesn't flow very well.

This PR simplifies it to a single screen:

image

By linking to the FF Cloud docs on Billing we can centralise all FAQ about billing. Note that if someone else wanted to run an EE version of FF with billing enabled, they would want to provide their own URL for the FAQ. Easy enough to add in the future what that's an actual requirement.

Under the covers, nothing has changed. Clicking the button will create the team on the platform, and the API call responds with the billingURL. The front end now redirects straight away rather than ask the user to click another button.

When TeamTypes are added, the text above the button will dynamically update to reflect what costs the team type selection will incur. This PR hardcodes it to reflect the status quo of teams not costing anything.

### DIFF
--- a/frontend/src/pages/team/create.vue
+++ b/frontend/src/pages/team/create.vue
@@ -9,8 +9,9 @@
         </SideNavigation>
     </Teleport>
     <main>
-        <ff-loading v-if="loading" message="Creating Team..." />
-        <div v-if="!loading && !needsBilling" class="max-w-2xl m-auto">
+        <ff-loading v-if="redirecting" message="Redirecting to Stripe..." />
+        <ff-loading v-else-if="loading" message="Creating Team..." />
+        <div v-else class="max-w-2xl m-auto">
             <form class="space-y-6" >
                 <FormHeading>Create a new team</FormHeading>
                 <div class="mb-8 text-sm text-gray-500">Teams are how you organize who collaborates on your projects.</div>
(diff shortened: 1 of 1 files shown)


joepavitt (approved):

frontend/src/pages/team/create.vue
```diff
@@ -28,39 +29,22 @@
...
                 </FormRow>
 
-                <ff-button :disabled="!formValid" @click="createTeam()">
+                <template v-if="billingEnabled">
+                    <div class="mb-8 text-sm text-gray-500 space-y-2">
+                        <p>To create the team we need to setup payment details via Stripe, our secure payment provider.</p>
+                        <p>You will not be charged for creating the team. You will be charged for the projects
+                            you create within the team. For more information on billing, please read our <a class="underline" href="https://flowforge.com/docs/cloud/billing/">Billing documentation</a>.</p>
+                    </div>
+                    <ff-button @click="createTeam()">
```
knolleary, line 38:
```suggestion
                    <ff-button :disabled="!formValid" @click="createTeam()">
```

joepavitt (approved):

PR #878, opened 2022-08-11
security: Add security.txt policy

ZJvandeWeg:
For all FlowForge installs there's one central way to report security
issues. As such each FlowForge can expose the same file under the root
route.

Closes https://github.com/flowforge/flowforge/issues/868

### DIFF
--- a/frontend/public/security.txt
+++ b/frontend/public/security.txt
@@ -1,0 +1,5 @@
+Contact: mailto:security@flowforge.com
+Expires: 2023-12-31T22:59:00.000Z
+Preferred-Languages: en
+Policy: https://flowforge.com/handbook/development/security/
+


hardillb:
This has hard coded values for FlowForge inc that will be served up for anybody running FlowForge. I'm not sure this is the right approach.

If we want to deploy this for FlowForge Clould then I can do it from AWS

hardillb:
We already have this for flowforge.com (but it needs the date fixing) here: https://flowforge.com/.well-known/security.txt

hardillb:
And we have a redirect in place for `flowforge.cloud` (but not *.flowforge.cloud) to point to the same file

ZJvandeWeg:
@hardillb if someone runs a FlowForge on `foo.example.com` I want the security.txt file to be discoverable though. The intent is to hard code this.

hardillb:
The problem is that puts us on the hook for somebodies project that if attacking a 3rd party, even if we are not hosting the FlowForge instance it's running on

PR #881, opened 2022-08-11
docs: Update device commands

ZJvandeWeg:
After looking at a demo I found that the copy paste of some commands didn't work
as sudo was missing. In the other case a dollar in front didn't allow for copy+paste work.
This change makes the commands easier to copy and paste.

### DIFF
--- a/docs/user/devices.md
+++ b/docs/user/devices.md
@@ -13,7 +13,7 @@ It can be installed as a global npm module. This will ensure the agent
 command is on the path:
 
 ```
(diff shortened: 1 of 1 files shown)


sammachin:
Thanks, certainly good improvements. 
I'm not totally clear about the use of sudo with `npm -g` it does seem to solve the problem but I wonder if there's a more correct way that something should be setup

hardillb (approved):

Issue #733, opened 2022-06-27
Team Types


sammachin:
### Description

Epic to cover the creation of different Tiers of Team plan.

Features that could be part of the type;
Max Number of members - The basic plan should be limited to 3 members
Collaboration features of FlowForge - Certain features eg team libraries should only be available on certain team types.
Stripe Product/Price ID - Where billing is enabled the stripe id should be setup with the team type

User Stories within this:

- Admin create team Type
- Owner Select Team Type when creating team
- Owner change team type (upgrade/downgrade)
- Accept additional charges when inviting a new user
- increment billing when user accepts invite
- decrement billing when user leaves team

ZJvandeWeg:
Relates to: https://github.com/flowforge/flowforge/pull/739

sammachin:
Needs design work ASAP to assess if it will be required pre 1.0

knolleary:
Updated the description with more details on our TeamTier approach for 0.9.

Summary:
 - TeamTiers are hardcoded - not freely definable like ProjectTypes are.
 - Start with Free and Pro - **TODO:** agree on those names
 - Some required configuration (stripe ids) done via `flowforge.yml`

Does not specify _how_ TeamTiers are used to unlock features - we'll address that when we have a feature to tie to a tier.

knolleary:
Also - key point having discussed at length with @sammachin:

 - Team membership limits apply to current users + invited users.
 - Team membership billing applies only to current users.

This means you cannot _invite_ more users than the tier allows. But you only pay for members who have accepted and have access. This approach simplifies the handling of declined or expired invites.

We could, in the future, look at the approach taken by other platforms where the team owner purchases an allocation of seats and is then left to fill those seats as they wish. But that is not in scope for 1.0.

sammachin:
_Start with Free and Pro - TODO: agree on those names_

I don't think Free is a good term as that implies a hard link to pricing..

Possible names for either tier and a future 3rd tier:

- Basic
- Simple
- Standard
- Pro
- Premium
- Power

Not totally sure about the ordering there, Is Pro higher or lower than Premium.

Basic has the advantage of downselling itself when positioned against Standard or Pro so business users that are not price sensitive are less likely to scrimp on the Basic.

knolleary:
Quick bit of market research of other platform names for their team-equivalent thing:
 - `Free`, `Team`, `Enterprise`
 - `Starter`, `Pro`, `Business`, `Enterprise`
[...]

PR #884, opened 2022-08-11
Initial TeamTypes implementation

knolleary:
This PR adds the core TeamTypes concept to the platform - #733

~There are some deviations from the outline in that issue - I will go back and update the issue dreckly then update this comment to reflect that it has been done.~ - done

### Database Changes

 - A `TeamType` model is added
 - The `Team` model is updated to add a `TeamType` property
 - A migration is included that checks if the `TeamType` table is empty and if so, inserts a default `starter` team type. This migration also then sets the type of all existing teams to that default type.

Default Team Type looks like this currently:

```
 name: 'starter',
 enabled: true,
 description: `Collaborate on projects with a starter team`,
 properties: JSON.stringify({ userLimit : 0 }),
```

~**TODO BEFORE MERGE**: update the description to something more appropriate. Input from @sammachin welcome.~

Note the `userLimit` is set to 0 here. The plan was not to introduce the limit until 1.0 - but the logic is here to support it.

### API Changes

 - Adds `GET /api/v1/team-types` - returns a list of the enabled team types.
 - `GET /api/v1/team/<team-id>` - updated to include a `type` object in the response.
 - Create Team endpoint (`POST /api/v1/team`) now requires `type` parameter to identify the type of team to create

### Billing Interaction & Configuration

When a team is created, the Stripe Product/Price used is set via `flowforge.yml`.

  - `billing.stripe.team_*` provide the default values (this is what we use today)
  - `billing.stripe.teams.<type-name>.*` provide type-specific values
  - `billing.stripe.teams.<type-name>.userCost` is the per user/month cost of the type. This is displayed in the UI side.

Example flowforge.yml config:
```
   billing:
     stripe:
       ...
       team_price: <default team price>
       team_product: <default team product>
       ...
       teams:
         starter:
           price: <starter team price>
           product: <starter team product>
           userCost: 0
         pro:
           price: <pro team price>
           product: <pro team product>
           userCost: 42
```

Whenever the number of team members changes (Invite accepted, or user removed) it updates the quantity of the corresponding Stripe Subscription item to match.

### UI

With billing enabled, the default Create Team looks like:

image

Note that setting `userCost` to `0` in flowforge.yml, gets labelled as a 'free' team, rather than display `$0/user/month`.

In the future, if there is a second team to choose from:

image

Note the text at the bottom is updated to reflect the cost the user is about to agree to.

### What this PR doesn't cover

 - UX around limiting numbers of team members. The API will enforce any limit set, but the UI doesn't warn you and the attempt to invite will fail.
 - UX notifying you of additional charges when you are about to invite a new user.

In summary - the whole 'invitation' UX needs some attention - but we don't need that for 0.9 as we aren't adding non-free team types yet.

 - Documentation updates. For 0.9, there is nothing new for admins or end users to need to know about. The default team type is automatically created and applied to all existing projects. The additional billing config is only required once we add the second team type.


knolleary:
Also currently lacking meaningful unit test coverage of the new stuff. I got all the existing tests working (now that TeamType is an expected property of Teams)... but some more coverage needed around specific TeamType/billing scenarios.

forge/db/migrations/20220808-02-create-default-team-types.js
```diff
@@ -0,0 +1,36 @@
...
+                    description: `Manage your personal projects
+ - Up to 3 team members
+ - More features to highlight
+`,
+                    properties: JSON.stringify({
+                        userLimit: 0
+                    }),
+                    createdAt: new Date(),
+                    updatedAt: new Date()
+                }
```
knolleary, line 26: Need to get consensus on these values for the default 'free' team before this is merged.

forge/ee/lib/billing/index.js
```diff
@@ -1,15 +1,51 @@
...
     app.config.features.register('billing', true, true)
 
     const stripe = require('stripe')(app.config.billing.stripe.key)
 
+    /**
+     * Get the Stripe product/price ids for the given team.
+     *
+     * These are provided via flowforge.yml.
+     *  - billing.stripe.team_* provide the default values.
+     *  - billing.stripe.teams.<team-name>.* provide type-specific values
```
knolleary, line 12:
```suggestion
     *  - billing.stripe.teams.<type-name>.* provide type-specific values
```

forge/ee/routes/billing/index.js
```diff
@@ -172,21 +172,14 @@ module.exports = async function (app) {
         const team = request.team
         const sub = await app.db.models.Subscription.byTeam(team.id)
         if (!sub) {
-            if (!request.session.User.admin) {
```
knolleary, line 175: This check allowed an Admin user to view a team even if billing wasn't setup. But the UX around this was not very good. Far too often it would lead to confusion as to why a team appeared okay - only to realised you were logged in as admin. The real risk was you would go on to create a project which gets in a bad state because the billing isn't setup properly.

forge/db/views/TeamType.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+    teamType: function (app, teamType) {
+        const properties = { ...teamType.properties }
+        if (app.config.billing.stripe.teams?.[teamType.name]) {
```
hardillb, line 4:
```suggestion
        if (app.config.billing?.stripe.teams?.[teamType.name]) {
```

Need to check if billing is enabled at all

hardillb:
OK, looks good so far. Will continue to kick the tires while you are out

knolleary:
Updated with some initial billing unit tests. Have figured out how to mock the `stripe` api client. It isn't ideal, as the tests end up having to know all of the individual stripe api calls the code makes and ensure they are mocked. But its the best option available.

forge/db/migrations/20220808-02-create-default-team-types.js
```diff
@@ -0,0 +1,36 @@
...
+        const count = await context.sequelize.query('select count(id) as count from "TeamTypes"', { type: context.sequelize.QueryTypes.SELECT })
+        if (parseInt(count[0].count) === 0) {
+            await context.bulkInsert('TeamTypes', [
+                {
+                    name: 'starter',
+                    enabled: true,
+                    description: `Manage your personal projects
+ - Up to 3 team members
+ - More features to highlight
+`,
```
knolleary, line 20:
```suggestion
                    description: `Collaborate on projects with a starter team`,
```

knolleary: Removing the bullet points for now and keeping the description fairly generic.

forge/db/migrations/20220808-02-create-default-team-types.js
```diff
@@ -0,0 +1,33 @@
...
+
+module.exports = {
+    up: async (context) => {
+        const count = await context.sequelize.query('select count(id) as count from "TeamTypes"', { type: context.sequelize.QueryTypes.SELECT })
+        if (parseInt(count[0].count) === 0) {
+            await context.bulkInsert('TeamTypes', [
+                {
+                    name: 'starter',
+                    enabled: true,
+                    description: `Collaborate on projects with a starter team`,
```
knolleary, line 17:
```suggestion
                    description: 'Collaborate on projects with a starter team',
```

knolleary:
I've updated the default description:

image

With that done, this PR is good for review and merge.

hardillb (approved):

PR #885, opened 2022-08-12
Add docs on how to test with billing enabled

knolleary:
Adds documentation on how to test with billing enabled.

We don't have lots of details on *how* to setup a stripe account properly and what entities are required. That's out of scope for this update. This is primarily for FlowForge Inc team members who can use the existing Stripe Test environment (details now available in 1Password).

### DIFF
--- a/docs/contribute/README.md
+++ b/docs/contribute/README.md
@@ -198,7 +198,7 @@ billing:
 
 You will also need to install the [Stripe CLI](https://stripe.com/docs/cli/) in order
 to handle webhook callbacks properly. Install the CLI following their documentation, then
-run the following command, with the API key using the valud of `billing.stripe.key` from
+run the following command, with the API key using the value of `billing.stripe.key` from
 above.
 
 ```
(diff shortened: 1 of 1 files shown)


docs/contribute/README.md
```diff
@@ -168,6 +168,48 @@ If you are developing locally and need to enable external email sending, you can
...
+    teams:
+      starter:
+        price: ***
+        product: ***
+        userCost: 0
+```
+
+You will also need to install the [Stripe CLI](https://stripe.com/docs/cli/) in order
+to handle webhook callbacks properly. Install the CLI following their documentation, then
+run the following command, with the API key using the valud of `billing.strip.key` from
```
knolleary, line 201:
```suggestion
run the following command, with the API key using the valud of `billing.stripe.key` from
```

docs/contribute/README.md
```diff
@@ -168,6 +168,48 @@ If you are developing locally and need to enable external email sending, you can
...
+    teams:
+      starter:
+        price: ***
+        product: ***
+        userCost: 0
+```
+
+You will also need to install the [Stripe CLI](https://stripe.com/docs/cli/) in order
+to handle webhook callbacks properly. Install the CLI following their documentation, then
+run the following command, with the API key using the valud of `billing.stripe.key` from
```
knolleary, line 201:
```suggestion
run the following command, with the API key using the value of `billing.stripe.key` from
```

hardillb (approved):

PR #900, opened 2022-08-17
Dialog JS API

### LINKED ISSUE #830: Reduce number of (almost identical) duplicate dialog components
### Description

While working in and around the flowforge frontend, it was noticeable the number of almost identical components for confirmation type tasks.

It would be advantageous for the majority of simple "OK/Cancel" "Please enter a name", "Are you sure?" dialogs to be replaced with a simple utility - something like `const result = await app.showMessageBox(options); if(result === MESSAGEBOX_OK) {...}` 

This then negates the need for separating the "dialog instantiation code" in the page from the "action code" in the component

Additionally, also very useful for something as simple as notifying success or failure of new features (until a more async method is developed)

joepavitt:
Closes #830 

**JS API**
Includes additional JS api to open Dialogs. I've overridden the initial dialogs where appropriate, mostly in "Delete" use cases, e.g.:

```
Dialog.show({
        header: '<header title>',
        kind: (optional) 'danger',
        text: 'show this message in the dialog',
        html: 'instead of "text", you can provide html for more custom appearance and content',
        confirmLabel: '<confirm-label>'
}, async () => {
        // callback goes here
})
```

**Architecture**
The core dialog service lives in `./frontend/src/services/dialog.js`. The key `.show()` function is used across the app in order to show and act upon confirmation of any dialogs.

The `bind` function is called from `./frontend/src/layouts/Platform.vue`. There is an instance of `ff-dialog` here that is driven by the `bind` function. When `show()` is called, the dialog being modified/shown is this instance.

**Testing**
Given that we now have E2E testing up and running, I've also used this as an opportunity to include a bunch of E2E tests that then cover a lot of primary use cases that use these Dialogs. I haven't gone for 100% coverage on this, but it's a significant dent into a backlog of technical testing debt too.


frontend/src/layouts/Platform.vue
```diff
@@ -13,6 +13,10 @@
                                        :type="a.type" :message="a.message"
                                        :countdown="a.countdown || 3000" @close="clear($index)"></ff-notification-toast>
             </TransitionGroup>
+            <ff-dialog ref="dialog" :header="dialog.header" :kind="dialog.kind" :disable-primary="dialog.disablePrimary" :confirm-label="dialog.confirmLabel" @cancel="clearDialog" @confirm="dialog.onConfirm">
+                <p v-if="dialog.text">{{ dialog.text }}</p>
+                <div class="space-y-2" v-html="dialog.html" />
```
Steve-Mcl, line 18:
```suggestion
                <div class="space-y-2" v-html="dialog.html"></div>
```
Self closing DIV is not w3c valid

frontend/src/pages/team/Devices/index.vue
```diff
@@ -160,35 +160,44 @@ export default {
...
+                    kind: 'danger',
+                    text: 'Are you sure you want to delete this device? Once deleted, there is no going back.',
+                    confirmLabel: 'Delete'
+                }, async () => {
+                    await deviceApi.deleteDevice(device.id)
+                    const index = this.devices.indexOf(device)
+                    this.devices.splice(index, 1)
+                })
             } else if (action === 'updateCredentials') {
                 this.$refs.deviceCredentialsDialog.show(device)
```
Steve-Mcl, line 183: Can this not be replaced with `Dialog.Show`? It opens this...
image

joepavitt: Good spot, I missed that one. If you'd like to try and build it yourself, could be a good exercise to try out the API? Also happy to do it myself of course.

Steve-Mcl (approved):
All viewed, ran locally, all good :)

frontend/src/pages/team/Devices/index.vue
Steve-Mcl: @joepavitt can you pick up this last one please?

joepavitt: Having taken a look, whilst that condition looks simple on the surface, actually looking into the [dialog itself](https://github.com/flowforge/flowforge/blob/dialog-api/frontend/src/pages/team/Devices/dialogs/DeviceCredentialsDialog.vue) it does have a lot of complex logic in. We could extract the "Are you sure" into the JS API, but it is tightly coupled with the credentials generation dialog, which in itself would still need to stay as a standalone component

Steve-Mcl: ok, NP.

Steve-Mcl:
All issues resolved, ready to merge @joepavitt 👍

PR #904, opened 2022-08-19
Roundup Billing

sammachin:
Currently we remove the project from stripe billing whenever a container is stopped, this can occur:
- When a user changes the stack of a project (container is stopped then new one is started)
- When a user chooses to suspend the project

(There are possibly other places that the stop is called I have missed.)

When a user chooses to suspend a container we want to ensure that the minimum time charged is 1 day, this is to discourage frequent short runs of a project.

This change introduces a new parameter in the stripe update to set the proration date which is where the charge will be calculated to, we set this to the start time of the container (in HH:MM:SS) on tomorrows date. Which has the effect of rounding up the bill to a full day.

Because there are scenarios where the container is stopped but we don't want to roundup I have added this as a new param flag to the addProject billing call and set the default to false (the current behaviour) so any other places that call this do not need to be explicit.

Opening as draft for discussion with @knolleary and @hardillb on if this is a good approach, another option would be to not touch billing at all for the stack changes but I can see merit in removing and readding billing should the new container fail to start.

Stripe bills by the second so the net result on the bill is zero because they are credited the remaining time when it is stopped then charged that same time when it is started.

### DIFF
--- a/forge/containers/wrapper.js
+++ b/forge/containers/wrapper.js
@@ -91,7 +91,7 @@ module.exports = {
      * @returns {Promise} Resolves when the project has been stopped
      */
 
-    stop: async (project) => {
+    stop: async (project, roundupBilling=false) => {
         if (project.state === 'suspended') {
             // Already in the right state, nothing to do
             return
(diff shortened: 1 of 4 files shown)


knolleary:
Having discussed with @sammachin and @ZJvandeWeg we agreed to hold this out for now.

> When a user chooses to suspend a container we want to ensure that the minimum time charged is 1 day, this is to discourage frequent short runs of a project.

The downside of this approach is that a user who is exploring/experimenting with the platform could get unexpected additional charges (the suspend/resume 3 times, they get billed for 3 full days).

At this stage, we prefer to avoid surprising/upsetting those users. We can look at other methods of discouraging the 'frequent short runs' pattern of behaviour if and when we see that type of usage happening.

PR #909, opened 2022-08-23
add default crt & key instructions

### LINKED ISSUE #894: Docker nginx defaults to first project alphabetically on unknown project name
### Current Behavior

IN the Docker install, if you try to load the url for a project that isn't know in the database eg https://[random-garbage].sammachin.com 
It will redirect you to the first project that exists alphabetically, for example I just created https://aaa.sammachin.com
Which is now served for all unknown project names, (previously it was returning alexa.sammachin.com)

### Expected Behavior

The forge app should be the default application for an unknown project.

### Steps To Reproduce

See current behaviour

### Environment

- FlowForge version: 0.8.0 Docker
- Node.js version: 16.15

sammachin:
fixes #894


hardillb:
Does this fix the picking the first project from the list if no matching names found?

sammachin:
Yes nginx-proxy has config for a server_name of `_` (default) but will only listen on 443 for that if there are default crt and key files.
https://github.com/nginx-proxy/nginx-proxy/blob/main/nginx.tmpl#L237

You can try that by going to https://jkfdhlfkjh.sammachin.com which should now return a 503 as I've added them on my instance

hardillb (approved):

PR #922, opened 2022-08-26
Limit password length to 1024 chars

### LINKED ISSUE #921: UX Bug: Long passwords can lead to failed sign up
### Current Behavior

When signing up for FlowForge a user can create a very long password (over 1mb in length). This leads to an error for the user without specifying why the error occurred.

### Expected Behavior

When a password is too long the submit button is disabled.

### Steps To Reproduce

_No response_

### Environment

- FlowForge version: FlowForge Cloud August 2022

hardillb:
The fastify body parser already limits the body of any post to 1mb
this just makes sure we don't hit that with a stupidly long password.

We can bcrypt hash a 1mb payload in aproxy 102ms with nodejs so
shouldn't be a problem, but 1024 seams like an excsively long password.

fixes #921

### DIFF
--- a/frontend/src/components/auth/Credentials.vue
+++ b/frontend/src/components/auth/Credentials.vue
@@ -53,6 +53,10 @@ export default {
                 valid = false
                 this.errors.password = 'Required field'
             }
+            if (this.input.password.length > 1024) {
+                valid = false
+                this.errors.password = 'Too long'
+            }
             if (valid) {
                 this.$store.dispatch('account/login', this.input)
             }
--- a/frontend/src/pages/account/PasswordReset.vue
+++ b/frontend/src/pages/account/PasswordReset.vue
@@ -52,6 +52,10 @@ export default {
                 this.errors.password = 'Password too short'
                 return false
             }
+            if (this.input.password.length > 1024) {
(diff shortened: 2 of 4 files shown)


frontend/src/components/auth/Credentials.vue
```diff
@@ -53,6 +53,10 @@ export default {
                 valid = false
                 this.errors.password = 'Required field'
             }
+            if (this.input.password.length > 1024) {
+                valid = false
+                this.errors.password = 'Too long'
+            }
             if (valid) {
```
ZJvandeWeg, line 60: We can remove all the `valid` stuff. Don't think we need it

```suggestion
            if (this.errors.username.length === 0 && this.errors.password.length === 0) {
```

ZJvandeWeg:
@hardillb This seems fairly easy to add a unit test for, can you add it?

frontend/src/components/auth/Credentials.vue
ZJvandeWeg: I'm fine with a response like: "ZJ, stop over engineering". But felt like a better way than to do the `valid = false` stuff

ZJvandeWeg:
Thanks for the test! ⭐

PR #924, opened 2022-08-26
Hide loading graphic when project is susspended and viewing logs

hardillb:
When trying to view the logs from a suspended project the loading graphic is shown. It should show the "No logs available" message.


Steve-Mcl (approved):
looks fine.

PR #925, opened 2022-08-27
Permit `a-zA-Z-` in project name

### LINKED ISSUE #546: Project names cannot include upper-case letters
### Current Behavior

You cannot create a project with an upper-case letter in the name.

Whilst this makes some sense as they are used for DNS names, we should support upper-case letters in the name.

It isn't entirely trivial to just relax the regex. The database model constraint is currently used to ensure unique names - but that is a case-sensitive check. We would now need to have a case-**insensitive** check.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

Steve-Mcl:
fixes #546

related...
* https://github.com/flowforge/flowforge-driver-k8s/pull/51
* https://github.com/flowforge/flowforge-driver-docker/pull/37

~~As discussed in #546, this draft PR approaches the issue using a virtual column...~~

> - ~~Do NOT add 2nd field~~
> - ~~Permit mixed case to be written to name field of DB~~
> - ~~Add a virtual field safeName that returns the lcase of the name field~~
> - ~~When determining if a name is already used (.count({ where: { name: name }})) ([ref](https://github.com/flowforge/flowforge/blob/main/forge/routes/api/project.js#L179)) is called, use the safeName field and the lowercase of request.body.name in the where clause. Alternatively, this could be abstracted to a function in the model (e.g. isNameAvailable(name))~~

### UPDATED 30-08-2022, 31-08-2022

As per further discussion in #546 this PR does the following..
* adds a new field `safeName` to the `project` table
  * This field is set in the migration to be both `not null` and `unique`
  * The DB migration routine does 3 steps...
    a. add column without constraints
    b. runs an `UPDATE` query to copy `name` to `safeName`
    ~~c. applies constraints~~
    c. Adds a unique index to the new column
    d. adjusts the model and adds a unique index to the new `safeName` column (in `.indexes[]`) so that any new DB will get the same index as a migrated DB
* adds tests to  
  * ensure a project with duplicate name (in any case) cannot be created
  * ensure a project with any case name can be created (and that the `title` property is defaulted to the `name` the user enteres)

~~TODO: use `project.safeName` where the lowercase version is needed (slug? DNS?)~~ Done.


forge/routes/api/project.js
```diff
@@ -170,15 +170,16 @@ module.exports = async function (app) {
...
 
-        const name = request.body.name
-
-        if (bannedNameList.includes(name)) {
+        const name = request.body.name?.trim()
+        const safeName = name?.toLowerCase()
+        if (bannedNameList.includes(safeName)) {
             reply.status(409).type('application/json').send({ error: 'name not allowed' })
-            return
+            return reply
```
hardillb, line 177: Why are we now returning `reply`?

hardillb (commented):
First pass looks good, just the questions about returning reply when we don't anywhere else

forge/routes/api/project.js
Steve-Mcl: Something I read - 

> Since v4, we are making it mandatory to `return reply`. if you want to call `reply.send` after the route handler has completed its execution.

So thought it no harm to think ahead.  However, I will remove them & we can handle that at a future point (if ever)

forge/routes/api/project.js
```diff
@@ -576,8 +577,19 @@ module.exports = async function (app) {
...
+            const reqSafeName = reqName?.toLowerCase()
+            const projectName = request.project.name?.trim()
+            if (reqName && projectName !== reqName) {
+                if (bannedNameList.includes(reqSafeName)) {
+                    reply.status(409).type('application/json').send({ error: 'name not allowed' })
+                    return reply
+                }
+                if (await app.db.models.Project.isNameUsed(reqSafeName)) {
+                    reply.status(409).type('application/json').send({ error: 'name in use' })
+                    return reply
```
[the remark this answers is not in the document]
Steve-Mcl: resolved in [5f0249b](https://github.com/flowforge/flowforge/pull/925/commits/5f0249b186708bfe16a19f0cf7f6c2b53b289149)

forge/routes/api/project.js
```diff
@@ -576,8 +577,19 @@ module.exports = async function (app) {
...
         } else {
-            if (request.body.name && request.project.name !== request.body.name) {
-                request.project.name = request.body.name
+            const reqName = request.body.name?.trim()
+            const reqSafeName = reqName?.toLowerCase()
+            const projectName = request.project.name?.trim()
+            if (reqName && projectName !== reqName) {
+                if (bannedNameList.includes(reqSafeName)) {
+                    reply.status(409).type('application/json').send({ error: 'name not allowed' })
+                    return reply
```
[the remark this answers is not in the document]
Steve-Mcl: resolved in 5f0249b

forge/routes/api/project.js
Steve-Mcl: resolved in 5f0249b

forge/routes/api/project.js
```diff
@@ -170,15 +170,16 @@ module.exports = async function (app) {
...
             reply.status(409).type('application/json').send({ error: 'name not allowed' })
-            return
+            return reply
         }
-        if (await app.db.models.Project.count({ where: { name: name } }) !== 0) {
+
+        if (await app.db.models.Project.isNameUsed(safeName)) {
             reply.status(409).type('application/json').send({ error: 'name in use' })
-            return
+            return reply
```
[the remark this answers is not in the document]
Steve-Mcl: resolved in 5f0249b

forge/routes/api/project.js
hardillb: OK, I've found the comment in the fastify migration guide, but it only applies to async responses which we don't have.

hardillb (approved):

forge/db/models/Project.js
```diff
@@ -185,6 +212,13 @@ module.exports = {
                 }
             },
             static: {
+                isNameUsed: async (name) => {
+                    const safeName = name?.toLowerCase()
+                    const found = await this.findOne({
+                        where: { safeName: safeName }
+                    })
+                    return !!found
```
knolleary, line 220:
```suggestion
                    const count = await this.count({
                        where: { safeName: safeName }
                    })
                    return count !== 0
```

knolleary: Using `count` is a more optimal choice (as per the original code we had that checked for duplicates) - it removes the need to retrieve the whole database object

Issue #774, opened 2022-07-10
Set Cutom Dashboard Path


sammachin:
### Description

**As a:** project owner

**I want to:** set the path that my dashboard is available on

**So that:** I don't have to serve the dashboard on /ui 

### Acceptance Criteria

- [ ] criteria 1
- [ ] criteria 2
- [ ] criteria 3
- ...

Steve-Mcl:
Tested locally ok.
On staging, the launcher is not updated so cannot test but the setting is present, is locked when should be locked and settable/retained when it should be.

Steve-Mcl:
Now verified on production


![Image](https://user-images.githubusercontent.com/44235289/187938678-924b1ea4-bdf8-4057-8d5f-7eb216b51bb3.png)

PR #926, opened 2022-08-30
permit custom dashboard path to be set from FF

Steve-Mcl:
part 1 of 2 #774

### DIFF
--- a/frontend/src/pages/admin/Template/utils.js
+++ b/frontend/src/pages/admin/Template/utils.js
@@ -18,7 +18,7 @@ const templateFields = [
 const defaultTemplateValues = {
     disableEditor: false,
     httpAdminRoot: '',
-    dashboardUI: '',
+    dashboardUI: '/ui',
     codeEditor: 'monaco',
     theme: 'forge-light',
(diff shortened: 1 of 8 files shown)


frontend/src/pages/admin/Template/utils.js
```diff
@@ -17,6 +18,7 @@ const templateFields = [
 const defaultTemplateValues = {
     disableEditor: false,
     httpAdminRoot: '',
+    dashboardUI: '',
```
knolleary, line 21: This should reflect the actual default value:
```suggestion
    dashboardUI: '/ui',
```

knolleary (commented):
When viewing an existing project's settings with this code applied, it shows 'No Value' against this setting:
image

This is misleading as the default value will be applied.

This is a broader issue than just this setting - but it doesn't look right.

knolleary:
Doing some more testing, by setting the default value to `/ui` then the ui does the right thing and shows that value. I had got into a odd state by testing the original PR with the default set to `''` - which does cause 'No Value' to be shown (which is in itself a misleading statement... but that's a different issue.

PR #927, opened 2022-08-30
Update Stack Deleting Error Text

### LINKED ISSUE #901: Broken Stack State Logic
### Current Behavior

In trying to delete a stack, I am presented with the following error:

```
Error: Cannot delete stack is active replacing another stack
```

The stack was created as a replacement of an existing stack which has a project associated to it. When creating, the existing stack successfully moved to "Inactive". 

I then went to test deletion and the changing of active/inactive state of a stack. I was unable to delete or de-activate the newly created stack.

If I created a new version, of the new version, then I could delete the first copy fine. 

### Expected Behavior

Either:

~1. Allow deletion/inactivate state for the newly created Stack~
2. Clearer user feedback as to why the deletion/editing of active/inactive state was not possible

I suspect this behaviour is in place to ensure we always have the top of the hierarchy of stacks available to users, in which case, (2) should be enacted.

### Steps To Reproduce

1. Create a stack
2. Click the kebab-menu options and select "Create New Version"
3. Then try to delete the newly created stack.

### Environment

- FlowForge version: 0.8.0 (`main`)
- Platform/OS: Mac
- Browser: Chrome

sammachin:
Closes #901

### DIFF
--- a/forge/db/models/ProjectStack.js
+++ b/forge/db/models/ProjectStack.js
@@ -37,7 +37,7 @@ module.exports = {
                 }
                 const replacedByCount = await M.ProjectStack.count({ where: { replacedBy: stack.id } })
                 if (replacedByCount > 0) {
-                    throw new Error('Cannot delete stack is active replacing another stack')
+                    throw new Error('Cannot delete stack that is the latest version of an active stack')
                 }
             }
         }


knolleary:
It's a better error message, but still not quite 100% accurate as the 'active' state is irrelevant. I think it's a good enough improvement to go with now. There are some bits around stack lifecycle that could do with tightening up which would impact this - but we don't need to go there now.

Issue #853, opened 2022-08-08
Editing Stacks silently fails


hardillb:
### Current Behavior

If you try to edit a stack property e.g. CPU % for an already deployed stack the dialogue allows you to make and submit the change and updates the table with the new value even though the change has been silently dropped.

If you refresh the page then the value reverts to it's original value.

### Expected Behavior

An error should be shown and the change should not be made to the table (or it should reload from the backend to always show the true values)

### Steps To Reproduce

- Deploy a stack with Docker/K8s backend
- Try and edit the stack CPU/Memory values

### Environment

- FlowForge version: 0.8
- Node.js version: 16.x
- npm version: 8.x
- Platform/OS: Linux/Alpine/Docker
- Browser: Chrome

sammachin:
@hardillb is is possible to change the values of the stack once it has been created? should these fields just not be editable once created?

hardillb:
@sammachin It should definitely not be possible to change these properties on a stack that has any deployed projects.

But yes, having it be uneditable would be a possible solution.

sammachin:
This error also occurs when trying to change a stacks active state

sammachin:
Verified on staging

PR #928, opened 2022-08-30
Fix Stack update logic

sammachin:
Fixes https://github.com/flowforge/flowforge/issues/853

This now checks if the Project Type isn't set against the stack and then changes it and only throws an error if the admin is trying to change an existing project type, so other values eg CPU/Mem/Active can be changed on a stack that does not have deployed projects

### DIFF
--- a/forge/routes/api/stack.js
+++ b/forge/routes/api/stack.js
@@ -164,21 +164,22 @@ module.exports = async function (app) {
             }
         }
         if (request.body.projectType) {
-            if (stack.ProjectTypeId) {
+            if (!stack.ProjectTypeId) {
+                // This is assigning the stack to a project type for the first time
+                // We'll allow that as part of the migration of legacy stacks
+                const projectTypeId = app.db.models.ProjectType.decodeHashid(request.body.projectType)
+                if (projectTypeId) {
+                    stack.ProjectTypeId = projectTypeId[0]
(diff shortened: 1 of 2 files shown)


hardillb:
This doesn't fix #853. If you try to change a value it still silently fails in UI and shows the changes as applied until the page is reloaded.

sammachin:
@hardillb what do you see in the developer tools for the PUT to `/api/v1/stacks/[STACKID]`

hardillb:
```
PUT http://localhost:3000/api/v1/stacks/47E0MWzNxZ 400 (Bad Request)
{error: 'Cannot change stack project type'}
```

But the point is not what shows up in the console (where a normal user will not look). The error as raised is that as far as UI is concerned the action succeeded, the memory value shows as changed in the table of stacks and no error is displayed. You only see that it hasn't worked if you reload the stacks page and notice the value you changed reverts back to it's initial value.

The fix for this is going to be in the front end code.

sammachin:
It should be possible to change cpu/memory values if the stack does not have any deployed projects, that error was only introduced if trying to modify the project Type but its catching other changes 

We have the same issue with trying to deactivate a stack.

sammachin:
@hardillb can you retest, this now works with a localfs install trying to change the active state

hardillb (approved):

PR #945, opened 2022-09-05
Basic user suspend

### LINKED ISSUE #222: Suspend Accounts
### Description

As an admin I want to be able to suspend;

- - a users account, 
- - a team  
- - a single project 

- such that they are unable to regain access

- a configurable message is displayed to the user on who to contact.
 
- The user(s) should be emailed this message as well. 
 
- As an admin I should be able to choose if running projects are stopped and the configuration/data saved so that the account can be restored.
 
- Containers should be terminated so that minimal system resources are consumed by suspended accounts.

hardillb:
- [x] TODO:  message to user when suspended
- [x] TODO: show message if user tries to login
- [x] TODO: check for suspension when verifying session

Only issue I've found so far is that if the user is logged into a Project we can't stop them deploying changes as the storage API uses a Project scoped token.

I'm also looking at the oAuth code to see if there is anything in there we can do.

fixes #222

Added new column to admin user view

Screenshot from 2022-09-08 11-31-32

Added locked setting to suspend a user

Screenshot from 2022-09-08 11-31-41

This uses a new config option in the yaml file to provide either a string or URL pointing to the `support_contact`. If the `support_contact` is not set it will use the email address of the first admin user, if that is not present then it will fall back to the string `the administrator`


hardillb:
I see how to invlidate oauth sessions has been considered but not implemented before

https://github.com/flowforge/flowforge/blob/90ef0a19b3eb7e2ed3b12a3ff386b428eb8fb58a/forge/db/controllers/Team.js#L102

sammachin (approved):

hardillb:
Re-running Cypress tests as they all pass when run locally

forge/routes/api/shared/users.js
```diff
@@ -30,6 +30,40 @@ module.exports = {
...
+                        if (app.postoffice.enabled()) {
+                            // Send email
+                            const context = {}
+                            if (app.config.support_contact) {
+                                console.log('from yaml')
+                                context.support = app.config.support_contact
+                            } else {
+                                const admin = await app.db.models.User.scope('admins').findOne()
+                                if (admin?.email) {
+                                    console.log('from admin email')
```
knolleary, line 46:
```suggestion
```

forge/routes/api/shared/users.js
```diff
@@ -30,6 +30,40 @@ module.exports = {
...
                 }
+
+                if (request.body.suspended !== undefined) {
+                    if (request.body.suspended === true) {
+                        await app.db.controllers.User.suspend(user)
+                        if (app.postoffice.enabled()) {
+                            // Send email
+                            const context = {}
+                            if (app.config.support_contact) {
+                                console.log('from yaml')
```
knolleary, line 41:
```suggestion
```

forge/routes/api/shared/users.js
```diff
@@ -30,6 +30,40 @@ module.exports = {
...
+                                }
+                            }
+                            try {
+                                context.url = new URL(context.support)
+                                if (context.url.protocol === 'mailto:' || context.url.protocol === 'tel:') {
+                                    context.support = context.url.pathname
+                                }
+                            } catch (err) {
+                            }
+                            console.log(context)
```
knolleary, line 60:
```suggestion
```

forge/routes/auth/index.js
```diff
@@ -178,6 +179,9 @@ module.exports = fp(async function (app, opts, done) {
                 reply.setCookie('sid', session.sid, cookieOptions)
                 reply.send({ status: 'okay' })
                 return
+            } else {
+                reply.code(403).send({ error: 'User Suspended' })
```
knolleary, line 183: Are we sure 'user suspended' is the only reason `session` may be null?

test/unit/forge/routes/api/users_spec.js
```diff
@@ -349,4 +349,24 @@ describe('Users API', async function () {
             membersAfterA.members.filter(e => e.username === 'elvis').should.have.property('length', 0)
         })
     })
+
+    describe('Suspend User', async function () {
+        it('Suspend/Resume elivis', async function () {
+            app.db.controllers.User.suspend(TestObjects.elvis)
```
knolleary, line 355:
```suggestion
            await app.db.controllers.User.suspend(TestObjects.elvis)
```

knolleary (changes requested):

forge/routes/auth/index.js
hardillb: yep, we've already checked that there is a user in the authentication on line 171 above so that when we call `createUserSession()` we know there is a user

```
    createUserSession: async function (app, username) {
        const user = await app.db.models.User.byUsernameOrEmail(username)
        if (user && !user.suspended) {
            return app.db.models.Session.create({
                sid: generateToken(32, 'ffu'),
                expiresAt: Date.now() + DEFAULT_WEB_SESSION_EXPIRY,
                UserId: user.id
            })
        }
        return null
    },
```

knolleary (approved):

PR #946, opened 2022-09-05
Set starter team userLimit to 3 and enforce member limits in api

### LINKED ISSUE #781: Limit max team size
The license must set a limit on the maximum number of members in a team.
This is used in the CE to limit to the same size as the free team tier on Cloud Edition.

In EE licenses this would normally be set to the same value as the number of users on the platform making it moot but allows for other commercial models to be adopted with EE licensing too

knolleary:
Closes #781 

This PR modifies the default `starter` team type to have a limit of 3 members.

It also adds additional checks to both the Team and Invitation controllers to ensure the userLimit is honoured.


hardillb (commented):
Code changes look OK, but need to work out why the tests are failing

knolleary:
Fixed the test. This one test resulted in a team having 4 members - which is no longer possible without a proper license. Changed the test to use a different team so as not to hit the limit.

hardillb (approved):

Issue #780, opened 2022-07-11
Community Edition license defaults


sammachin:
Where a platform does not have a license file (Community Edition) then default limits must be applied 
50 Projects
50 Teams
50 Devices
150 Users
Max Team Size 3

PR #947, opened 2022-09-06
Add limits to license and enforce Project limit

### LINKED ISSUE #776: License limits number of Projects on the platform
The license file must impose a limit on the maximum number of projects that can be created on the platform.

If a user attempts to create a project that exceeds this amount the request must fail with an appropriate error.

knolleary:
Closes #776 #780 

This PR builds on #946 - that should get merged first.

This adds claims to the license to cover projects, users, teams and devices.

If no license is applied (CE edition), they default to the values described in #780 

If a license is applied, it uses the values from the license, defaulting to `0` if not present.

When the platform starts up, it reports on the counts:

```
[2022-09-06T18:45:57.482Z] INFO: License verified:
[2022-09-06T18:45:57.482Z] INFO:   ****************************
[2022-09-06T18:45:57.482Z] INFO:   * Development-mode License *
[2022-09-06T18:45:57.482Z] INFO:   ****************************
[2022-09-06T18:45:57.482Z] INFO:  Org:     FlowForge Inc. Development
[2022-09-06T18:45:57.482Z] INFO:  Valid From : 2022-09-06T00:00:00.000Z
[2022-09-06T18:45:57.482Z] INFO:  Expires    : 2223-02-04T23:59:59.000Z
[2022-09-06T18:45:57.491Z] INFO: Usage:
[2022-09-06T18:45:57.491Z] INFO:  Users    : 3/150
[2022-09-06T18:45:57.491Z] INFO:  Teams    : 3/50
[2022-09-06T18:45:57.491Z] INFO:  Projects : 5/50
[2022-09-06T18:45:57.492Z] INFO:  Devices  : 4/50
```

In the second commit, changes are added to enforce the Project limit. This is done as a hook on the model itself - so any code path that tries to create a project in the database will be handled.

The UI side does a fairly default Alert notification that the license limit has been reached. That can be refined as I work through policing the other limits.

# NOTE TO DEVELOPERS

Once this PR is merged, you will need to update the development license you are running with locally otherwise you will no longer be able to create projects (unless you run with no license at all). As before, this is provided in `forge/licensing/index.js`.

### DIFF
--- a/forge/db/controllers/Invitation.js
+++ b/forge/db/controllers/Invitation.js
@@ -2,13 +2,7 @@ const { Roles } = require('../../lib/roles.js')
 module.exports = {
     createInvitations: async (app, invitor, team, userList) => {
         const externalInvitesPermitted = app.postoffice.enabled() && !!app.settings.get('team:user:invite:external')
-        const currentTeamMemberCount = await team.memberCount()
-        const currentTeamInviteCount = await team.pendingInviteCount()
-        const userLimit = team.TeamType.properties?.userLimit
-        if (userLimit > 0 && currentTeamMemberCount + currentTeamInviteCount >= userLimit) {
-            throw new Error('Team user limit reached')
-        }
-
+        const pendingInvites = []
         const results = {}
         for (let i = 0; i < userList.length; i++) {
             const userDetail = userList[i]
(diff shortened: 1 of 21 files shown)


knolleary:
Ah - the test attempts to invite a 4th user to a team. Which is not possible anymore due to the new limits.

Steve-Mcl:
Kinda proves it works. :)

knolleary:
Reordered the test a bit so that it never tries to exceed the team membership limit.

Steve-Mcl (approved):

PR #950, opened 2022-09-07
Apply license limits to number of users

### LINKED ISSUE #777: License limits number of Users on the platform
The license file must impose a limit on the maximum number of users that can be created on the platform.

If an admin attempts to create a user that exceeds this amount the request must fail with an appropriate error.
If an owner attempts to invite an external user to the platform that exceeds this amount the request must fail with an appropriate error.
If a user attempts to register a new account that would exceed this amount the request must fail with an appropriate error

knolleary:
Closes #777 

Builds on #947 - that should be reviewed and merged first.

This PR adds enforcement of the User limit on the platform as a whole (rather than within individual teams).

It adds unit tests around the paths that can create new users.

### DIFF
--- a/test/unit/forge/db/models/User_spec.js
+++ b/test/unit/forge/db/models/User_spec.js
@@ -6,112 +6,124 @@ const { Roles } = FF_UTIL.require('forge/lib/roles')
 describe('User model', function () {
     // Use standard test data.
     let app
-    beforeEach(async function () {
-        app = await setup()
-    })
 
     afterEach(async function () {
-        await app.close()
-    })
-
-    it('User email can be null', async function () {
-        await app.db.models.User.create({ username: 'nullEmail', password: '12345678' })
-        await app.db.models.User.create({ username: 'nullEmail2', password: '12345678' })
-    })
-
-    it('User email, if set, must be unique', async function () {
-        await app.db.models.User.create({ username: 'duplicateEmail', email: 'duplicate@email.com', password: '12345678' })
-        try {
-            await app.db.models.User.create({ username: 'duplicateEmail2', email: 'duplicate@email.com', password: '12345678' })
-            throw new Error('Created user with duplicate email')
-        } catch (err) {
-            /SequelizeUniqueConstraintError/.test(err.toString()).should.be.true()
+        if (app) {
+            await app.close()
+            app = null
         }
     })
 
-    it('User email, if set, must be valid email', async function () {
-        try {
-            await app.db.models.User.create({ username: 'badEmail', email: 'not-n-email', password: '12345678' })
-            throw new Error('Created user with invalid email')
-        } catch (err) {
-            /SequelizeValidationError/.test(err.toString()).should.be.true()
-        }
+    describe('Model properties', function () {
+        beforeEach(async function () {
+            app = await setup()
+        })
+
+        it('User email can be null', async function () {
+            await app.db.models.User.create({ username: 'nullEmail', password: '12345678' })
+            await app.db.models.User.create({ username: 'nullEmail2', password: '12345678' })
+        })
+
+        it('User email, if set, must be unique', async function () {
+            await app.db.models.User.create({ username: 'duplicateEmail', email: 'duplicate@email.com', password: '12345678' })
+            try {
+                await app.db.models.User.create({ username: 'duplicateEmail2', email: 'duplicate@email.com', password: '12345678' })
+                throw new Error('Created user with duplicate email')
+            } catch (err) {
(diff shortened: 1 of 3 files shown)


test/unit/forge/routes/auth/index_spec.js
```diff
@@ -0,0 +1,132 @@
...
+                    email: `u${i}@example.com`
+                })
+                resp.statusCode.should.equal(200)
+            }
+            await expectRejection({
+                username: 'u6',
+                password: '12345678',
+                name: 'u6',
+                email: 'u6@example.com'
+            }, /license limit reached/)
```
Steve-Mcl, line 129: A little puzzled here.

> This license has limit of 5 users (1 created by default test setup (test/unit/forge/routes/setup.js))

| user | count |
| - | - |
| default user | 1 |

> // Register 4 more users to reach the limit
> `for (let i = 1; i <= 4; i++) {` 

Then `for (let i = 1; i <= 4; i++) {` adds four more named ```username: `u${i}\` ```
| user | count |
| - | - |
| default user | 1 |
| u1 | 2 |
| u2 | 3 |
| u3 | 4 |
| u4 | 5 |

At this point there are 5 users.

> ```
>             await expectRejection({
>                 username: 'u6',
> ```


Where does user 'u6' get created?

forge/licensing/index.js
```diff
@@ -40,8 +50,18 @@ module.exports = fp(async function (app, opts, next) {
...
+            if (!key) {
+                return activeLicense
+            }
+            if (activeLicense) {
+                if (Object.hasOwn(activeLicense, key)) {
+                    return activeLicense[key]
+                }
+                return undefined
+            }
+            return defaultLimits[key]
```
Steve-Mcl, line 62: `licenseApi.get(key)` seems to return default values i no `activeLicense` however if a key is not provided, the empty `activeLicense` is returned 

Should this be return `activeLicense || defaultLimits` ?

<br>

Also, the defaults see a little high (150/50/50/50) - if a user doesnt apply a licence they get large defaults? 
I assume this is acceptable since without a license the product is CE (not EE)?

Steve-Mcl (commented):
Small points raised - some of them may well be my understanding - otherwise, looks good

test/unit/forge/routes/auth/index_spec.js
Steve-Mcl: My point is I can not even see how or where there is even an attempt to create a `u6`

> Where does user 'u6' get created?

Steve-Mcl: or user `u5` for that matter (the for loop only generates u1~u4)

knolleary: The test never attempts to create `u5`... lets just call that an off-by-one error. The important part is the actual number of users created.

The function `expectRejection` wraps the call to create the user and also verifies it fails properly.

Steve-Mcl: ok, now it makes sense. 

Apart from the non descriptive function naming  of `expectRejection` and the off by one error on "u1" ~ "u4" , I now see this will work as expected.

Feel free to mark this comment as resolved however for the sake of future comprehension, I would recommend altering the loop to generate `user${i+1}`  (i.e. user2, user3...user5) and would also recommend renaming  `expectRejection` to something like `addUserExpectRejection({ username: 'user6',...`

test/unit/forge/db/models/User_spec.js
```diff
@@ -6,112 +6,124 @@ const { Roles } = FF_UTIL.require('forge/lib/roles')
...
+
+        it('User Real Name with URL', async function () {
+            let user
+            try {
+                user = await app.db.models.User.create({
+                    username: 'foo',
+                    password: '123456abc',
+                    email: 'foo@example.com',
+                    name: 'http://example.com'
+                })
```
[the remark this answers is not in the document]
Steve-Mcl: I would recommend (while we are in here reordering the code) that we accept the test re-name as it improves comprehension. 

Otherwise, please make as resolved.

forge/licensing/index.js
knolleary: > Also, the defaults see a little high (150/50/50/50) - if a user doesnt apply a licence they get large defaults?
I assume this is acceptable since without a license, the product is CE (not EE)?

Those are the agreed defaults for the unlicensed, ie CE, platform.

knolleary: IF the user doesn't have a license applied and it asks for a license value, it returns the default.

If the user does have a license applied, it returns whatever the license specifies (defaulting to `0` if the license doesn't say anything).

PR #951, opened 2022-09-07
Add ability to update T+Cs to cause a re-review / reaccept for users 

### LINKED ISSUE #451: Be able to update T&Cs and prompt users to accept update
### Epic

_No response_

### Description

**As a:** Admin

**I want to:** be able to update the T&Cs of the platform and require users to accept the new terms

**So that:** I know who has/hasn't accepted the latest terms

---

Follow on to #363 and #443 which allows the admin to require T&C acceptance on sign up.

High level implementation thoughts

 - [ ] sign-up api should require a flag to confirm T&C acceptance rather than just a UI block on submitting the form
 - [ ] update User model to record a date of when they accepted T&Cs
 - [ ] provide option to admin to trigger update of t&c
 - [ ] when user logs in, compare the t&c date with the user-accepted date. If t&cs newer - trigger a reconfirmation prompt.

### Acceptance Criteria

_No response_

Steve-Mcl:
fixes #451

TODO:
* update design on story
* tests

### DIFF
--- a/forge/db/views/User.js
+++ b/forge/db/views/User.js
@@ -37,6 +37,9 @@ module.exports = {
         if (user.password_expired) {
             result.password_expired = true
         }
+        if (user.tcs_accepted) {
+            result.tcs_accepted = user.tcs_accepted
+        }
         result.email_verified = user.email_verified
         if (user.defaultTeamId) {
             result.defaultTeam = app.db.models.Team.encodeHashid(user.defaultTeamId)
--- a/frontend/src/pages/admin/Settings/General.vue
+++ b/frontend/src/pages/admin/Settings/General.vue
@@ -29,9 +29,19 @@
             </template>
         </FormRow>
         <FormRow v-if="input['user:tcs-required']" v-model="input['user:tcs-url']" type="text" :error="errors.termsAndConditions">
-            Terms &amp; Conditions URL:
-            <template #description>
-                The URL for the Terms &amp; Conditions to load when a user wishes to view them.
+            Terms &amp; Conditions URL
+        </FormRow>
+        <FormRow v-if="input['user:tcs-required']">
+            Date of Terms &amp; Conditions applied
+            <template #input>
+                <div class="flex justify-between items-center" style="width: 100%">
+                    <div class="font-mono text-md grow">
+                        <span class="text-blue-700">{{tcsDate}}</span>
+                    </div>
+                    <div class="grow-0">
+                        <ff-button size="small" :disabled="loading" :kind="'secondary'" @click="updateTermsAndConditions">Update...</ff-button>
+                    </div>
+                </div>
             </template>
         </FormRow>
         <FormHeading>Teams</FormHeading>
@@ -47,7 +57,7 @@
             Allow users to invite external users to teams
             <template #description>
                 <p>Users can invite existing users to join a team. If they provide
-                    an email address of an unregistered user, the invitiation will be
+                    an email address of an unregistered user, the invitation will be
                     sent to that email address.</p>
             </template>
         </FormRow>
@@ -74,7 +84,8 @@
 
 <script>
 import settingsApi from '@/api/settings'
-
+import Dialog from '@/services/dialog'
+import Alerts from '@/services/alerts'
 import FormRow from '@/components/FormRow'
 import FormHeading from '@/components/FormHeading'
 import { mapState } from 'vuex'
@@ -85,6 +96,7 @@ const validSettings = [
     'user:team:auto-create',
     'user:tcs-required',
     'user:tcs-url',
+    'user:tcs-date',
     'team:create',
     'team:user:invite:external',
     'telemetry:enabled'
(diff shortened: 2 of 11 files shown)


Steve-Mcl:
@knolleary this is now ready for review.

NOTE: e2e test failing on team membership as your licence limits PR

```
  FlowForge - Team Membership
    ✓ loads team members into the data table (3777ms)
    ✓ member cannot can invite or remove a user (3108ms)
    ✓ owner/admin can invite user (3578ms)
    ✓ loads invitations into the data table (2318ms)
    1) user can accept a team invite
    2) admin/owner can remove a member from the team


  4 passing (27s)
  2 failing

  1) FlowForge - Team Membership
       user can accept a team invite:
     AssertionError: Timed out retrying after 4000ms: Too many elements found. Found '1', expected '0'.
      at Context.eval (http://localhost:3000/__cypress/tests?p=test/e2e/frontend/cypress/tests/team-membership.spec.js:142:60)

  2) FlowForge - Team Membership
       admin/owner can remove a member from the team:

      AssertionError: Timed out retrying after 4000ms: Not enough elements found. Found '3', expected '4'.
      + expected - actual

      -3
      +4
```

... this should be fixed when [9921b43](https://github.com/flowforge/flowforge/pull/950/commits/9921b4358dad90c0a79a7b79fa194dc759d06fac) is merged

knolleary:
We already have a migration in `main` that has the prefix `forge/db/migrations/20220905-01-`. It's complete filename comes *after* the one added by this PR, which will mean this migration will be skipped in any local install that has already run the `main` code.

Can you rename the migration file to today's date?

forge/db/views/User.js
```diff
@@ -38,6 +38,9 @@ module.exports = {
         if (user.password_expired) {
             result.password_expired = true
         }
+        if (user.tcs_accepted) {
```
knolleary, line 41: We should only include this in the response if `app.settings.get('user:tcs-required')` is set.

frontend/src/pages/admin/Settings/General.vue
```diff
@@ -147,20 +171,47 @@ export default {
...
+        },
+        async updateTermsAndConditions () {
+            // don't save the T&C's if not required
+            if (!this.input['user:tcs-required']) {
+                return
+            }
+            Dialog.show({
+                header: 'Update Terms and Conditions',
+                kind: 'danger',
+                html: '<p>This action will update the timestamp associated with the Terms and Conditions effectively causing existing users to be prompted to re-review and reaccept the updated Terms and Conditions.</p><p>Are you sure?</p>',
```
knolleary, line 199:
```suggestion
                html: '<p>This action will require all existing users to reaccept the Terms and Conditions the next time they access the platform.</p><p>Are you sure?</p>',
```

I think this message can be simplified. The fact it's a timestamp is an internal implementation detail.

knolleary:
Sorry - I mucked up resolving the merge conflict. Have pushed the fix.

knolleary:
Generally looks good - but I think some UX changes are needed on the admin settings page.

I have worked up the following. LMK what you think:

image

Note I have proposed here that changing the T&Cs url should also trigger an update. How does that sound? Or do we want the admin to be able to change the url without triggering re-acceptance?

Steve-Mcl:
> Generally looks good - but I think some UX changes are needed on the admin settings page.
> 
> I have worked up the following. LMK what you think:
> 
> image
> 
> Note I have proposed here that changing the T&Cs url should also trigger an update. How does that sound? Or do we want the admin to be able to change the url without triggering re-acceptance?

Very much agree - and i did try (and failed) so left it as was 

If you have the necessary html to how you achieved the cleaner look - please send it my way (or update the PR)

Either way let me know.

frontend/src/pages/admin/Settings/General.vue
Steve-Mcl: much better

Steve-Mcl:
> We already have a migration in `main` that has the prefix `forge/db/migrations/20220905-01-`. It's complete filename comes _after_ the one added by this PR, which will mean this migration will be skipped in any local install that has already run the `main` code.
> 
> Can you rename the migration file to today's date?

What are the odds we both wrote DB migrations on the same date lol

Will push update shortly

knolleary:
I have pushed the UI changes.

One more piece to add (updating the date when the url is changed)... will update here when done

forge/db/views/User.js
Steve-Mcl: resolved in [9e58e97](https://github.com/flowforge/flowforge/pull/951/commits/9e58e97fa516f3faf262245ef094b0d4017a680f)

knolleary:
I've updated the logic around saving admin settings to ensure `tcs-updated` gets set at the right times.

There was also an undiscovered bug around detecting if the tcs url has changed in the `saveEnabled` computed property.

Steve-Mcl:
Tests updated to ensure presence of/absence of tcs settings in the user API

Added a couple more tests that seem appropriate...

* Terms and Conditions
   * admin can enable Terms and Conditions
   * admin can update Terms and Conditions date
   * user can accept Terms and Conditions
   * user API does not return T&Cs properties if T&Cs are disabled

Steve-Mcl:
@knolleary - finally passing e2e tests once more.  

The modification you pushed that sets the T+Cs date upon URL change caused some headaches.  

Sorted now.

PR #954, opened 2022-09-08
Add full path to invitations on email

### LINKED ISSUE #953: Finding invites from email isn't obvious
### Description

Feedback from a customer that the invite email isn't as clear as it could be and finding the invitations isn't obvious.

Improve the email template to add a link directly to the invitations.

### Epic/Story

_No response_

sammachin:
fixes #953

### DIFF
--- a/forge/postoffice/templates/TeamInvitation.js
+++ b/forge/postoffice/templates/TeamInvitation.js
@@ -5,11 +5,11 @@ module.exports = {
 
 You've been invited to join team {{{invite.team.name}}} on the FlowForge platform.
 
-{{{ signupLink }}}
+{{{ signupLink }}}/account/teams/invitations
 `,
     html:
 `<p>Hello!</p>
 <p>You've been invited to join team {{invite.team.name}} on the FlowForge platform.</p>
-<p><a href="{{{ signupLink }}}">{{{ signupLink }}}</a></p>
+<p><a href="{{{ signupLink }}}/account/teams/invitations">{{{ signupLink }}}/account/teams/invitations</a></p>
 `
 }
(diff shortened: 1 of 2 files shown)


knolleary (changes requested):
This is not the right way to change the url.

In the case of an invitation sent to a non-existing user, this change will result in a url of `http://example.com/account/create?email=foo@example.com/account/teams/invitations`

To change the url, you should change the code that sets `signupLink` in `forge/routes/api/teamInvitations.js`

In the case of a non-existing user, we cannot direct them straight to the invitations page - they have to signup first.

In the case of an existing user, if they are not logged in, this will take them to the login page and then to the default user view. We don't currently have a mechanism to redirect back to a page *after* login.

sammachin:
I saw the 2 template files, `TeamInvitation` and `UnknownUserInvitation` and assumed the latter was for non-existing users, but I realise now thats for inviting to the platform in general, will move the change to the code that sets the link.

forge/postoffice/templates/TeamInvitation.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
 
 You've been invited to join team {{{invite.team.name}}} on the FlowForge platform.
 
-{{{ signupLink }}}
+{{{ signupLink }}}/account/teams/invitations
```
knolleary, line 8: We shouldn't have this partial url in the template. It should be set entirely via `signupLink`.

sammachin: sorry missed removing it on the plain text template, fix now

knolleary:
Looks good.

There is one part of the UX that needs considering as a separate issue:
>In the case of an existing user, if they are not logged in, this will take them to the login page and then to the default user view. We don't currently have a mechanism to redirect back to a page after login.

sammachin:
> Looks good.
> 
> There is one part of the UX that needs considering as a separate issue:
> 
> > In the case of an existing user, if they are not logged in, this will take them to the login page and then to the default user view. We don't currently have a mechanism to redirect back to a page after login.

Yes this is perhaps a more general UX question about sharing links inside the FF platform where a user is not logged in, something that could be fixed across the whole app perhaps?

Issue #955, opened 2022-09-08
'Create personal team' option on /admin/users/create does not work


knolleary:
### Current Behavior

The admin page for creating a user directly has an option 'Create personal team' that is intended to create a team for that user - skipping the invitation step.

It is broken since 0.8 as the call to create the team doesn't include a TeamType.

The user does get created, but the api call returns:

```
{"error":"TypeError: Cannot read properties of null (reading 'properties')"}
```

The team is created, but as it is missing the TeamType, it breaks all subsequent calls to view teams.

---

Fix:

1. The 'create personal team' option should be extended with a dropdown of available team types. The value should be included in the api call.
2. More protection is needed to ensure a Team cannot be created without a TeamType - as doing so will break things. Need to look at how an association can be marked as required in sequelize - and what that means for the db table and what migrations might be necessary. Failing that, a `beforeCreate` hook can be used to validate things.

### Expected Behavior

The user is created with a team successfully.

### Steps To Reproduce

Go to `/admin/users/create`, fill in the form and tick the 'create personal team' box.

### Environment

- FlowForge version: 0.8+

knolleary:
A similar issue exists if the `user:team:auto-create` option is enabled to create a team automatically when a user registers.

That isn't an option we enable in FF-Cloud, but it is a valid configuration we have provided since 0.1.

In the interests of convention over configuration, we could chose to remove that option entirely - although we don't know if any CE users make use of it today. The easier approach will be to default to the 'starter' team for now.

PR #959, opened 2022-09-08
Enforce limits on number of teams that can be created

### LINKED ISSUE #778: License limits the number of Teams on the Platform
The license file must impose a limit on the maximum number of teams that can be created on the platform.
If a user attempts to create a new team that would exceed this the request must fail with an appropriate error message

knolleary:
Closes #778 #955 

This limits how many teams can be created on the platform.

The limit is applied in the `beforeCreate` hook of the Team model.

There are three ways a team can be created:

1. A user creates a team - this is policed via the `beforeCreate` hook and an error returned to the ui
2. An admin creates a user and ticks 'create personal team' - by the time it comes to create the team, the user has already been created - so too late to reject the request. So this does its own check of the limit before attempting to create the user.
3. A user registers on the platform and the 'auto-create team' option is enabled - same as above (does explicit limit check)

Aside: for 2 & 3, I identified a bug (#955) that was stopping them from working. This PR addresses that bug by ensuring the TeamType is set on the projects they create. In fact, I have added a `beforeCreate` check to ensure `TeamTypeId` is always provided when creating a team. This required a couple tests to be updated as they weren't doing so.


Steve-Mcl (approved):
all good

Steve-Mcl:
@knolleary code looks fine but there is an error in Project Lifecycle test...

```
Project Lifecycle
       "before all" hook for "login user":
     Error: Cannot create team without TeamTypeId
      at Function.beforeCreate (forge/db/models/Team.js:21:27)
      at Function.runHooks (node_modules/sequelize/lib/hooks.js:94:18)
      at model.save (node_modules/sequelize/lib/model.js:2378:30)
      at async Function.create (node_modules/sequelize/lib/model.js:1344:12)
      at async Context.<anonymous> (test/system/100-project-lifecycle_spec.js:51:29)
```

Can I ask if this ran ok under SQLITE locally?

knolleary:
@Steve-Mcl fix pushed. I verified the unit tests but overlooked the system test folder when looking for calls to `Team.create` that needed updating.

Steve-Mcl (approved):

Steve-Mcl (approved):

Issue #782, opened 2022-07-11
Device Billing


sammachin:
### Epic

#446

### Description

implement billing for devices where the platform has billing enabled.

Devices are a new subscription product within Stripe.
There is only one type of device at a single price point
The number of devices on the team will be reflected by the quantity of the device product in the subscription

Team Tiers #733 may define that a team type includes a number of free devices, in which case this should be taken into account when setting the quantity on the subscription.

### Acceptance Criteria

- [ ] When a new device is created the team must be billed for the device by incrementing the quantity in the device product attached to the team subscription. If there is no device product as part of the subscription one should be created
- [ ] When a device is deleted from the team the quantity in the device product subscription should be decremented.
- [ ] The owner should be shown a check box and asked to accept additional charges when creating a new device.

knolleary:
This is what our pricing guide current says (bulleted for ease of reference):

> 1. The team is billed per device at the same rate regardless of tier or number of devices.
> 2. A team plan may include a number of free devices.
> 3. The EE License will be for a defined number of devices on the platform regardless of teams.
> 4. The CE license will allow 50 devices.

From (1) - a single stripe product/price ID will be provided via `flowforge.yml` for the device object.

The CE(4)/EE(3) License determines the total number of devices allowed on the platform (implemented via #961)

From (2) - the `TeamType` needs to be able to specify how many devices are free. We have a similar concept for `userLimit` - which is set to `3` for the default `starter` team (although we aren't talking about a limit here - just a free allocation of devices). Let's call the new property `deviceFreeAllocation` - until I think of something better.

I think the following describes how that property will be handled. @sammachin please let me know if this makes sense to you:
1. On CE, `deviceFreeAllocation` is the limit of how many devices you can have in the team
2. On EE (without billing enabled), `deviceFreeAllocation` has no effect - you can have as many devices as you want in a team (you are already paying for them via the license fee)
3. On EE (with billing enabled), `deviceFreeAllocation` is the free allocation, anything after that is then billed in addition (whilst staying within the overall platform-wide limit on devices as set by the license).

sammachin:
> * On CE, `deviceFreeAllocation` is the limit of how many devices you can have in the team

I don't believe we specified a limit on devices per team in CE, just 50 across the platform. So if you run CE with 1 team that team could have 50 devices.

I'm not opposed to putting a limit on CE teams I'm just not sure its really required, it feels like additional complexity to explain to a customer.

knolleary:
[...]

PR #964, opened 2022-09-09
Add TeamType.properties.deviceLimit to restrict devices in team

knolleary:
Part of #782 

This PR adds `TeamType.properties.deviceLimit` as a way to limit how many devices may be added to a team of a given type. If the property isn't set, no limit is applied.

- Adds a TeamType.getProperty() accessor function to make it easier to access individual properties
- When creating a device in a team, if checks if `TeamType.properties.deviceLimit` is set, and if so, checks the limit won't be exceeded.

### DIFF
--- a/forge/db/controllers/Invitation.js
+++ b/forge/db/controllers/Invitation.js
@@ -47,10 +47,10 @@ module.exports = {
             opts.invitorId = invitor.id
             pendingInvites.push({ userDetail, opts })
         }
-        if (team.TeamType.properties?.userLimit > 0) {
+        if (team.TeamType.getProperty('userLimit') > 0) {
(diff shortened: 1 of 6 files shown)


Steve-Mcl (approved):

Steve-Mcl:
@knolleary 

I missed the `.only` in my review : 

https://github.com/flowforge/flowforge/pull/964/files#diff-6ac2d7b38d1e1030740415f71b67508a18ee8f929f347e4abfa919a5871b13d6R160

I assume this is a debugging left over?


<br>

PS, I dont often use `it.only` but instead use the mocha test explorer
Code_8JwHJhc7NQ

PR #965, opened 2022-09-09
Add httpNodeAuth settings

hardillb:
Allows the setting of a single username/password in the template

Fixes #https://github.com/flowforge/flowforge/issues/578

Requires https://github.com/flowforge/flowforge-nr-launcher/pull/64

Screenshot from 2022-09-09 14-03-17

### DIFF
--- a/forge/db/controllers/ProjectTemplate.js
+++ b/forge/db/controllers/ProjectTemplate.js
@@ -1,3 +1,4 @@
+const { hash } = require('../utils')
 const validSettings = [
     'disableEditor',
     'httpAdminRoot',
@@ -166,6 +167,9 @@ module.exports = {
                     result.palette.denyList = paletteDenyList
                 }
             }
+        }
+        if (typeof result.httpNodeAuth?.pass === 'string' && result.httpNodeAuth.pass.length > 0) {
+            result.httpNodeAuth.pass = hash(result.httpNodeAuth.pass)
         }
         return result
     },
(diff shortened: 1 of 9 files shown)


hardillb:
Does the password bcrypt in page so only the hash is ever sent to the backend and stored in the database

hardillb:
@knolleary I have pushed what might be a slightly naive fix for the password getting overwitten on any other change to the template.

knolleary:
I've reworked how passwords are handled in template settings.

We now hash them in the backend before they are stored. We never return the value in the api - we return a boolean to indicate if a non-blank value exists.

The UI then does the Right Thing to show a placeholder or not.

I have verified the settings are applied properly by the launcher - and that the values can be overridden in the project.

There is one PR to come against the launcher to avoid righting blank settings into httpNodeAuth.

knolleary:
Huh - @hardillb as the person who raised this PR originally, it will now let me add you as a reviewer (which is why I tried unassigning it from you first...). Please take a look.

hardillb:
Yep, was already looking, got confused for a second as to why there was no change to package.json to remove the `bcryptjs`, but makes sense now

hardillb:
Code looks OK, but 2 of the UI tests are with something to do with httpNodeAuth

hardillb:
All the tests pass locally, so I'm happy to merge this (and we need to set aside some time next sprint to fix the postgres tests and work out what's happening here)

PR #967, opened 2022-09-09
Prevent last admin being removed from platform

### LINKED ISSUE #952: Prevent the removal of the last admin
### Description

Update DELETE `/api/v1/users/:id` to fail if trying to remove the last user with the admin flag set.

### Epic/Story

_No response_

### LINKED ISSUE #966: No feedback when an API error occurs editing user
### Current Behavior

Server-side validation errors are **not** shown to admin user

### Demo show where an attempt to change the name of user **test5** to **user4** - note **user4** already exists but no error is show.
chrome_FyITRIblTL

The same issue is present for API errors
* "Invalid username"
* "email must be unique"
* unknown errors

### Expected Behavior

Dialog to remain open and to inform user of issue via the validation text

NOTE: the validation text is already present but is impossible to show since the `ff-dialog` closes upon confirm regardless of the API call success

Related: https://github.com/flowforge/forge-ui-components/issues/48

### Steps To Reproduce

with 3 users named **admin1** (admin) & **user1** **user2** (non admin)...
1. login as **admin1**
2. edit user1
3. set user1 name to user2

dialog is dismissed and no change has occured.

### Environment

- FlowForge version: 0.9.0-git
- Node.js version: NA
- npm version: NA
- Platform/OS: All
- Browser: All

Steve-Mcl:
fixes #952, closes #966,

depends on https://github.com/flowforge/forge-ui-components/pull/49 

chrome_w51OGSvxOk

### DIFF
--- a/forge/db/models/User.js
+++ b/forge/db/models/User.js
@@ -53,7 +53,14 @@ module.exports = {
                     user.name = user.username
                 }
             },
-            beforeUpdate: (user) => {
+            beforeUpdate: async (user) => {
+                if (user._previousDataValues.admin === true && user.admin === false) {
+                    const currentAdmins = await app.db.models.User.scope('admins').findAll()
+                    console.log(currentAdmins)
+                    if (currentAdmins.length <= 1) {
+                        throw new Error('Cannot remove last Admin user')
+                    }
+                }
                 if (user.avatar.startsWith(`${process.env.FLOWFORGE_BASE_URL}/avatar/`)) {
                     user.avatar = generateUserAvatar(user.name || user.username)
                 }
--- a/forge/db/models/User.js
+++ b/forge/db/models/User.js
@@ -56,7 +56,6 @@ module.exports = {
             beforeUpdate: async (user) => {
                 if (user._previousDataValues.admin === true && user.admin === false) {
                     const currentAdmins = await app.db.models.User.scope('admins').findAll()
-                    console.log(currentAdmins)
                     if (currentAdmins.length <= 1) {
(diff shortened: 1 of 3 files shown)


Steve-Mcl:
@hardillb please feel free to review

If ok, let me know and I will write suitable tests.

forge/db/models/User.js
```diff
@@ -53,7 +53,14 @@ module.exports = {
                     user.name = user.username
                 }
             },
-            beforeUpdate: (user) => {
+            beforeUpdate: async (user) => {
+                if (user._previousDataValues.admin === true && user.admin === false) {
+                    const currentAdmins = await app.db.models.User.scope('admins').findAll()
+                    console.log(currentAdmins)
```
hardillb, line 59:
```suggestion
```
Left over `console.log()`

hardillb (commented):
Rest looks ok

hardillb (approved):

forge/db/models/User.js
```diff
@@ -53,7 +53,13 @@ module.exports = {
                     user.name = user.username
                 }
             },
-            beforeUpdate: (user) => {
+            beforeUpdate: async (user) => {
+                if (user._previousDataValues.admin === true && user.admin === false) {
+                    const currentAdmins = await app.db.models.User.scope('admins').findAll()
```
knolleary, line 58: It would be better to use `.count()` rather than `findAll()` - let the database do the work to count, rather than transfer all of the data back to the app for us to found.

PR #970, opened 2022-09-12
Add billing to device lifecycle

knolleary:
Part of #782 

This adds a billing step to the device lifecycle.

If `billing.stripe.device_product/device_price` are set, then whenever a device is added/removed from team, the billing subscription item is updated for the team to reflect how many devices are in the team.

If the team type includes a `deviceFreeAllocation` property, this is taken into account when identifying how many 'billable' devices a team has.

This PR does not update the UI - that will come in the final PR for this story.

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -17,6 +17,8 @@ module.exports.init = async function (app) {
      *       ...
      *       team_price: <default team price>
      *       team_product: <default team product>
+     *       device_price: <default device price>
+     *       device_product: <default device product>
      *       ...
      *       teams:
      *         starter:
@@ -29,7 +31,11 @@ module.exports.init = async function (app) {
     function getBillingIdsForTeam (team) {
         const result = {
             price: app.config.billing.stripe.team_price,
-            product: app.config.billing.stripe.team_product
+            product: app.config.billing.stripe.team_product,
+            device: {
+                price: app.config.billing.stripe.device_price,
+                product: app.config.billing.stripe.device_product
+            }
         }
         if (app.config.billing.stripe.teams?.[team.TeamType.name]) {
             result.price = app.config.billing.stripe.teams[team.TeamType.name].price || result.price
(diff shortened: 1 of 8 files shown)


test/unit/forge/ee/lib/billing/index_spec.js
```diff
@@ -167,4 +167,141 @@ describe('Billing', function () {
             updateData.should.have.property('proration_behavior', 'always_invoice')
         })
     })
+
+    describe.only('updateTeamDeviceCount', async function () {
```
Steve-Mcl, line 171:
```suggestion
    describe('updateTeamDeviceCount', async function () {
```

remove `.only`

knolleary: 🤦🏻

forge/ee/lib/billing/index.js
```diff
@@ -211,6 +217,51 @@ module.exports.init = async function (app) {
...
+        updateTeamDeviceCount: async (team) => {
+            const billingIds = getBillingIdsForTeam(team)
+            if (!billingIds.device.product) {
+                return
+            }
+            const subscription = await app.db.models.Subscription.byTeam(team.id)
+            if (subscription) {
+                const deviceCount = await team.deviceCount()
+                const deviceFreeAllocation = team.TeamType.getProperty('deviceFreeAllocation') || 0
+                const billableCount = Math.max(0, deviceCount - deviceFreeAllocation)
```
Steve-Mcl, line 229: Question: I cannot see where `deviceFreeAllocation` is ever set (different PR? Future? From the License?)

Steve-Mcl: Also, should `deviceFreeAllocation` be covered in tests?

Steve-Mcl (commented):
1st review pass complete

forge/ee/lib/billing/index.js
knolleary: `deviceFreeAllocation` will be defined as a property of the `TeamType` - (ref: https://github.com/flowforge/flowforge/issues/782#issuecomment-1241742266)

It defaults to 0 - as it stands we haven't decided what values we will apply (if any). The purpose of that part of this PR is to ensure we have this capability.


I will add some additional tests for it - just requires a bit of juggling around of things.

knolleary:
I have now added the final test for deviceFreeAllocation.

PR #980, opened 2022-09-13
Add notice about user limit in team invite dialog

### LINKED ISSUE #781: Limit max team size

knolleary:
Closes #781 

This PR adds better UI around the team user limit.

If a team type has a user limit, the invite user dialog includes information on what that limit is:

image

If a team as at (or exceeds) the limit, the dialog doesn't allow any more invites to be sent:

image


knolleary:
Just adding here that I really want to give the invite workflow/UX an overhaul - these dialogs are functional but clunky. But that is for another day.

Steve-Mcl (approved):

Issue #940, opened 2022-09-02
Admin User RBAC


sammachin:
### Description

We need to review what an Admin user can access in terms of projects and teams, The Admin user is not really a user in the scope of RBAC, admin is a separate flag that is system wide, owner/user is a flag that is at the team level.

There are some use cases where an admin of the platform needs to carry out specific actions on a team or project, for example, suspending or deleting a project, removing a team, etc.

There are 3 possible approaches to consider:

_Updated by @knolleary 14/Sept - see my comment below for details_

1) ~The admin has the ability to invite users to any team, and the ability to set the users access level. Therefore they use this to invite themselves into a team and then promote themselves to owner and carry out other tasks.~

2) The admin had direct access to the actions they need in any team/project page regardless of membership.

3) ~We create new views and controls within the admin pages for the actions required and the admin has no access to any of the team pages. (one of these controls may well be to add themselves to a team as an owner)~

Option 1 is the previous situation but this has been lost with the recent RBAC cleanup, also it would introduce complexity with chargeable teams where the account is billed/limited based on number of members in a team so the admin joining the team would be a chargeable event.

Option 2 is tricky to maintain as there are a number of places that these checks need to be made both UI and API, in effect we would need to consider the admin to be a 3rd level on the RBAC.

Option 3 would provide a cleaner experience and avoid confusion for admins that are able to see things they are not members of, but would require additional development work. It also raises the question of how the actions are invoked, do we just check for admin on the same API or create new Admin scoped APIs.

~Following discussion on the product call we are going with option 3 and will raise stories for specific admin actions that are needed~

_Update 14/Sept_ Option 2 is the current plan

knolleary:
Following some development work on the Admin Project view it became clear taking option 3 was simply too much work to reproduce admin-only views.

We are going to take Option 2 - allow an admin to access the team view and perform any action. There is one exception - an admin will not be allowed to login to the Node-RED instance itself.

To achieve that, we need to do the following:

 - [ ] #987 
 - [ ] Do not allow admins to log in to Node-RED instances in teams they are not members
 - [ ] Check all logic around teamMembership levels in the UI to verify Admin is treated the same as Owner. We typically use the check `this.teamMembership.role === Role.Owner` to test if they are an owner. If we want to treat Admins as Owners, we could probably just change that to `this.teamMembership.role >= Role.Owner` given how the Role constants are defined. But will need to see what makes sense in the code
 - [ ] Update RBAC table in docs

[...]

PR #991, opened 2022-09-15
Ensure admins have team-owner level access to all views in frontend

knolleary:
Part of #940 

(includes #987)

This takes advantage of how roles are defined in `forge/lib/roles.js`

```
const Roles = {
    None: 0,
    Member: 30,
    Owner: 50,
    Admin: 99
}
```

Rather than check `role === Roles.Owner || role === Roles.Admin` we can more concisely do `role >= Roles.Owner`.


hardillb (approved):
I had to look up what Teleport did, but the rest looks good.

PR #993, opened 2022-09-15
Add case-insensitive unique constraint on User.username and email

### LINKED ISSUE #983: User names can be same (but different case)
**NOTE** PR https://github.com/flowforge/flowforge/pull/984 should be merged before commencing this works to avoid potential merge issue and duplicate works.

<br> 
### Current Behavior

You can currently have user `test`, `Test` and user `TEST`  all with the same email address.

image

image

The user edit page says only characters `a-z0-9-` are permitted but under the hood, performs a case insensitive check (`!/^[a-z0-9-_]+$/i.test(v)` 

NOTE: Sam has already pointed out the production DB has mixed case names so consideration will be needed when squaring up any field validations. 

NOTE2: The email field has similar issue - a user who signed up with `MrDead@HarryAndChums.com` cannot log in with  `mrdead@harryandchums.com`  - AND - another user could have the Email address of `MRdead@Harryandchums.com`

### Expected Behavior

Name and email should be unique (case insensitive test)

i.e. if user `ted` exists then a new or 2nd user CANNOT be (re)named to `TED`
Same should go for the Email address

### Steps To Reproduce

edit a users name or email address to be the same as another (but change letter case)

### Environment

- FlowForge version: 0.9.0-git

knolleary:
Closes #983 

It took some digging, but I found a sequelize unit test for adding a unique index that uses the `lower` function. This means we can add a unique index to the table, without having to add another column:

```
indexes: [
   { name: 'user_username_lower_unique', fields: [sequelize.fn('lower', sequelize.col('username'))], unique: true }
]
```

I have tested and verified this constrain works as required on both sqlite and postgres.

Alongside this, I've updated the `User.byUsername/email` getter functions to do case insensitive queries and improved the feedback in the Admin Create User page and User Registration page.

### DIFF
--- a/forge/routes/auth/index.js
+++ b/forge/routes/auth/index.js
@@ -292,7 +292,11 @@ module.exports = fp(async function (app, opts, done) {
             reply.send({ status: 'okay' })
         } catch (err) {
             let responseMessage
-            if (err.errors) {
+            if (/user_username_lower_unique/.test(err.parent?.toString())) {
+                responseMessage = 'username not available'
+            } else if (/user_email_lower_unique/.test(err.parent?.toString())) {
+                responseMessage = 'email not available'
+            } else if (err.errors) {
                 responseMessage = err.errors.map(err => err.message).join(',')
             } else {
                 responseMessage = err.toString()
--- a/forge/db/controllers/User.js
+++ b/forge/db/controllers/User.js
@@ -1,17 +1,18 @@
 const { compareHash, sha256 } = require('../utils')
 const jwt = require('jsonwebtoken')
+const { fn, col, where } = require('sequelize')
 
 module.exports = {
     /**
      * Validate the username/password
      */
     authenticateCredentials: async function (app, username, password) {
-        let clause = { username: username }
-        if (/.+@.+/.test(username)) {
-            clause = { email: username }
-        }
+        const column = /.+@.+/.test(username) ? 'email' : 'username'
         const user = await app.db.models.User.findOne({
-            where: clause,
+            where: where(
+                fn('lower', col(column)),
+                username.toLowerCase()
+            ),
             attributes: ['password']
         })
         // Always call compareSync, even if no user found, to ensure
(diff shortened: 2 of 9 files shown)


forge/routes/auth/index.js
```diff
@@ -292,7 +292,11 @@ module.exports = fp(async function (app, opts, done) {
             reply.send({ status: 'okay' })
         } catch (err) {
             let responseMessage
-            if (err.errors) {
+            if (/user_username_lower_unique/.test(err.parent?.toString())) {
+                responseMessage = 'username not available'
+            } else if (/user_email_lower_unique/.test(err.parent?.toString())) {
+                responseMessage = 'email not available'
```
hardillb, line 298: This doesn't match the error thrown from `forge/route/api/users` for the same error condition and is causing a test failure

See above

forge/routes/auth/index.js
```diff
@@ -292,7 +292,11 @@ module.exports = fp(async function (app, opts, done) {
             reply.send({ status: 'okay' })
         } catch (err) {
             let responseMessage
-            if (err.errors) {
+            if (/user_username_lower_unique/.test(err.parent?.toString())) {
+                responseMessage = 'username not available'
```
hardillb, line 296: This doesn't match the error thrown from forge/route/api/users for the same error condition and is causing a test failure

```
            if (/user_username_lower_unique/.test(err.parent?.toString())) {
                responseMessage = 'username not unique'
            } else if (/user_email_lower_unique/.test(err.parent?.toString())) {
                responseMessage = 'email not unique'
            } else if (err.errors) {
```

hardillb (commented):
Looks like the errors from register/user api/v1/users post should be the same?

forge/routes/auth/index.js
knolleary: Have standardised on `xxx not available` in both places and updated the test to match.

Issue #463, opened 2022-04-08
User list not refreshing after changing user details


ZJvandeWeg:
### Current Behavior

As an admin you can expire user passwords. When you do, the user list page is not updated and doesn't display the users password as expired. After a refresh of the page, the field does read "true".

### Expected Behavior

List is refreshed automatically.

### Steps To Reproduce

Expire a password in the admin interface.

### Environment

- FlowForge version: 0.3

sammachin:
Refreshing of all views to be revistied after re-brand of app.

sammachin:
issue still exists in 0.8 after the table re-design
@joepavitt is there a quick fix to refresh the table after a model closes?

knolleary:
Yes - its s simple call to refresh the data after the dialog closes - we do that in other views.

PR #997, opened 2022-09-20
Prevent admin from suspending themselves

hardillb:
Similar to not being able to remove last admin

As part of testing/reviewing #222 found it was possible for an admin to suspend themselves, this prevents that.

Formatting for the diff is a bit strange, but this is basically an `if` block round the section that updates the user.

### DIFF
--- a/forge/routes/api/shared/users.js
+++ b/forge/routes/api/shared/users.js
@@ -35,32 +35,37 @@ module.exports = {
                 }
 
                 if (request.body.suspended !== undefined) {
-                    if (request.body.suspended === true) {
-                        await app.db.controllers.User.suspend(user)
-                        if (app.postoffice.enabled()) {
-                            // Send email
-                            const context = {}
-                            if (app.config.support_contact) {
-                                context.support = app.config.support_contact
-                            } else {
-                                const admin = await app.db.models.User.scope('admins').findOne()
-                                if (admin?.email) {
-                                    context.support = `mailto:${admin.email}`
+                    if (request.user.id !== user.id) {
+                        if (request.body.suspended === true) {
+                            await app.db.controllers.User.suspend(user)
+                            if (app.postoffice.enabled()) {
+                                // Send email
+                                const context = {}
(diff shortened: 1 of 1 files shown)


forge/routes/api/shared/users.js
```diff
@@ -35,32 +35,37 @@ module.exports = {
...
-                            } catch (err) {
+                                app.postoffice.send(user, 'UserSuspended', context)
                             }
-                            app.postoffice.send(user, 'UserSuspended', context)
+                        } else {
+                            user.suspended = false
                         }
                     } else {
-                        user.suspended = false
+                        reply.code(400).send({ error: 'can not susspend self' })
```
knolleary, line 67:
```suggestion
                        reply.code(400).send({ error: 'cannot suspend self' })
```

knolleary:
For good measure, can we have a unit test for this?

forge/routes/api/shared/users.js
hardillb: Should be fixd

hardillb:
Test added

knolleary (approved):

forge/routes/api/shared/users.js
```diff
@@ -35,32 +35,37 @@ module.exports = {
...
-                        if (app.postoffice.enabled()) {
-                            // Send email
-                            const context = {}
-                            if (app.config.support_contact) {
-                                context.support = app.config.support_contact
-                            } else {
-                                const admin = await app.db.models.User.scope('admins').findOne()
-                                if (admin?.email) {
-                                    context.support = `mailto:${admin.email}`
+                    if (request.user.id !== user.id) {
```
knolleary, line 38: Missed this in the review - `request.user` is the user being updated - which at the same point is the same as `user`.

This should be `request.session.User.id !== user.id`. In hindsight I now see the unit test was verifying admin cannot suspend themselves, but we didn't have a test verifying admin can suspend *anyone else*. Ooops.

Fix incoming as part of #463

PR #1001, opened 2022-09-20
Extend login sessions to one week with idle timeout

### LINKED ISSUE #972: Keep sessions valid for 2 weeks
### Description

Everytime I use FlowForge I feel like I have to sign in again. I would like not to sign in often, once every 2 weeks should be enough.

knolleary:
Fixes #972 

This modifies how we handle login sessions as described in https://github.com/flowforge/flowforge/issues/972#issuecomment-1252266637

More specifically:

 - adds `idleAt` timestamp to Session object
 - sets `Session.expiresAt` to `now + 7 days`
 - sets `Session.idleAt` to `now + 32 hours`
 - When accessing API:
    - if `Session.expiresAt` is in the past - destroy session and logout
    - if `Session.idleAt` is in the past - destroy session and logout
    - if `Session.idleAt` is < 31 hours in the future, update it to `now + 32 hours` - this keeps extending the session as long as the user is 'active', whilst avoiding hammering the db by updating the idle time with every single request.

 - Removes the `remember me` checkbox from the UI and API - we didn't have the logic right for it.

### DIFF
--- a/forge/db/controllers/Session.js
+++ b/forge/db/controllers/Session.js
@@ -1,6 +1,14 @@
 const { generateToken } = require('../utils')
 
-const DEFAULT_WEB_SESSION_EXPIRY = 1000 * 60 * 60 * 24 // One day session
+const HOUR = 1000 * 60 * 60
+// This is the maximum time a user can be logged in for before being asked to reauth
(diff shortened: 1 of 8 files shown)


sammachin (approved):

hardillb:
Can I merge this? Just planning ahead for another staging build tomorrow

PR #1006, opened 2022-09-21
First pass at Coupon support

### LINKED ISSUE #978: Coupon Codes in Signup Links
### Epic

_No response_

### Description

**As a:** user

**I want to:** be able to embed the coupon code in the signup URL and have that auto-populate/apply the coupon code at the stripe checkout

**So that:** I can make better use of coupon code

_No response_

### Acceptance Criteria

- [ ] Signup URL can contain a query param eg `/signup?coupon=SOMECODE`
- [ ] This is persisted (cookie?) through the users signup journey and when they are redirected to the Stripe checkout sesison the coupon is passed in the call to stripe.
- [ ] Coupon is automatically applied to the subscription
- ...

hardillb:
http://forge.example.com/account/create?code=[CouponCode]

Sets the `ff_coupon` cookie with a week long life

closes #978

### DIFF
--- a/frontend/src/pages/team/Billing.vue
+++ b/frontend/src/pages/team/Billing.vue
@@ -112,8 +112,6 @@ export default {
     },
     watch: { },
     async mounted () {
-        // TODO remove
-        console.log(await window.cookieStore.get('ff_coupon'))
         this.loading = true
         if (!this.team.billingSetup) {
             this.coupon = (await window.cookieStore.get('ff_coupon'))?.value.split('.')[0]
--- a/frontend/src/pages/team/Billing.vue
+++ b/frontend/src/pages/team/Billing.vue
@@ -18,6 +18,12 @@
         </div>
         <div v-else>
             Billing has not yet been configured for this team. Before proceeding further, you must continue to Stripe and complete this.
+            <div v-if="coupon">
+                <div class="mb-8 text-sm text-gray-500 space-y-2">Will apply coupon code <span v-text="coupon"></span> at checkout</div>
+            </div>
+            <div v-if="errors.coupon">
+                <div class="ml-9 text-red-400 inline text-xs">{{errors.coupon}}</div>
+            </div>
             <div class="mt-3">
                 <ff-button @click="setupBilling()" data-action="setup-payment-details">
                     <template v-slot:icon-right><ExternalLinkIcon /></template>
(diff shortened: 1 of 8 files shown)


hardillb:
OK, works, but might want a little UI polishing, I've picked what I think are the right classes, but the spacing might be a little off

sammachin (approved):

frontend/src/pages/team/Billing.vue
```diff
@@ -97,13 +105,18 @@ export default {
...
-            }]
+            }],
+            coupon: false,
+            errors: {}
         }
     },
     watch: { },
     async mounted () {
+        // TODO remove
+        console.log(await window.cookieStore.get('ff_coupon'))
```
knolleary, line 116:
```suggestion
```

frontend/src/pages/team/Billing.vue
```diff
@@ -116,8 +129,9 @@ export default {
                 })
                 this.loading = false
             } catch (err) {
-                // check for 404 and redirect if 404 returned
-                if (err.response.status === 404) {
+                // check for 402 and redirect if 402 returned
+                if (err.response.status === 402) {
+                    this.coupon = (await window.cookieStore.get('ff_coupon'))?.value.split('.')[0]
```
knolleary, line 134: `window.cookieStore` is not supported widely enough for us to use - https://developer.mozilla.org/en-US/docs/Web/API/CookieStore#browser_compatibility

knolleary (changes requested):
A couple initial comments without looking at the visuals of it.

frontend/src/pages/team/Billing.vue
hardillb: Pushed update without `window.cookieStore`

frontend/src/pages/team/Billing.vue
```diff
@@ -133,12 +145,31 @@ export default {
...
                 }
             }
             window.open(billingUrl, '_self')
+        },
+        getCookie (name) {
+            if (document.cookie) {
+                const cookies = document.cookie.split(';')
+                for (let i = 0; i < cookies.length; i++) {
+                    const cookie = cookies[i]
+                    if (cookie.split('=')[0] === name) {
```
knolleary, line 167: Given multiple cookies, the `;` is space-padded between entries so need to trim.

```suggestion
                    if (cookie.split('=')[0].trim() === name) {
```

knolleary (approved):

PR #1005, opened 2022-09-21
Add Viewer role

knolleary:
Adds a Viewer role to the platform.

They are full members of a team, but with read-only access to everything.

---

Remaining work:

 - [ ] tests around api access. Not sure how extensive to make it yet as `needsPermission` does all the heavy lifting
 - [ ] doc updates
 - [x] give user read-only access to the editor

PR #1008, opened 2022-09-22
Include role in Team Invitation

knolleary:
Part of #1005 

This adds the ability to select the role of a user when inviting them - rather than default to Member.

This requires adding a column to the invitation table to store the role info.

Also updates the various views to include the role name in the UI.

### DIFF
--- a/config/vitest.config.ts
+++ b/config/vitest.config.ts
@@ -6,18 +6,18 @@ import Vue from '@vitejs/plugin-vue'
 
 export default defineConfig({
     plugins: [
-      Vue(),
+        Vue()
     ],
     test: {
         globals: true,
         environment: 'jsdom',
         coverage: {
-          src: ['./frontend/src'],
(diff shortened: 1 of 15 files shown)


hardillb (approved):

hardillb:
The code reads ok, but the failure in the sqlite test implies that `Roles` is undefined which looks to be set right.

hardillb:
Test also fails locally for me, so not some GitHub strangeness

hardillb:
`RoleNames` is there, just `Roles` is empty which is strange

knolleary:
Fix pushed.

Looks like vitest cannot do the alias importing like webpack can. I found the test was already working around that limitation by having a hardcoded copy of the roles data. But that meant it was old data - so I've changed it to load the real data properly.

hardillb:
Will merge once tests have finished

PR #1013, opened 2022-09-23
Ensures stack/template are preselected on Create Project page

### LINKED ISSUE #989: Default Stack and Templates
### Description

Make the 1st Stack and Template on the new project screen selected be default so a user doesn't have to make a choice.

This is in keeping with our priniciple of Sensible Defaults

### Epic/Story

_No response_

knolleary:
Closes #989 

 - Ensures `ProjectType.defaultStack` is preselected when the ProjectType is selected.
 - Falls back to selecting first in list if defaultStack isn't found
 - Hides the Template selection entirely if only 1 template found
 - Ensures the `ProjectType.defaultStack` is updated when a stack is replaced
 - Ensures the options are selected properly in the copy-project flow

### DIFF
--- a/frontend/src/pages/team/createProject.vue
+++ b/frontend/src/pages/team/createProject.vue
@@ -52,7 +52,7 @@
                     </ff-tile-selection>
                 </div>
                 <!-- Template -->
-                <div class="flex flex-wrap gap-1 items-stretch">
+                <div v-if="templates.length !== 1" class="flex flex-wrap gap-1 items-stretch">
                     <label class="w-full block text-sm font-medium text-gray-700 mb-1">Template</label>
                     <label v-if="!input.projectType && !input.stack" class="text-sm text-gray-400">Please select a Project Type &amp; Stack first.</label>
                     <label v-if="errors.template" class="text-sm text-gray-400">{{ errors.template }}</label>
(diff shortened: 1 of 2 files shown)


frontend/src/pages/team/createProject.vue
```diff
@@ -186,26 +200,22 @@ export default {
...
-
+        this.$nextTick(() => {
             if (this.projectTypes.length === 0) {
                 this.errors.projectTypes = 'No project types available. Ask an Administrator to create a new project type'
             }
 
             if (!this.sourceProjectId) {
-                this.input.stack = this.stacks.length > 0 ? this.stacks[0].value : ''
-                this.input.template = this.templates.length > 0 ? this.templates[0].value : ''
+                this.input.stack = this.stacks.length > 0 ? this.stacks[0].id : ''
```
hardillb, line 211: Is this going to override the default stack if it's not the first one in the array?

knolleary: No - this code is marginally redundant. It will get called before the user has selected a ProjectType, so in fact this.stacks will always be empty. When they select a ProjectType it triggers the watcher (line 163) which loads the stacks for that type and selects the right one.

hardillb (approved):

PR #1021, opened 2022-09-26
Standardise format of error responses

### LINKED ISSUE #956: Inconsistent api response format for errors
### Current Behavior

We have an inconsistent style of response errors in the api.

We have a lot of:
```
{ error: 'invalid username' }
```

But we also have some:
```
{ status: 'error', message: err.toString() }
```

We need to standardise the format and make it consistent across the whole api surface.

The frontend should not rely on the particular spelling of an error message in order to detect it as that makes it impossible to consider i18n in the future.

### Expected Behavior

A consistent format for api errors

### Steps To Reproduce

_No response_

### Environment

_No response_

knolleary:
Closes #956 

All error responses must now be of the format:

```
{
   code: 'an_error_code',
   error: 'A human readable message'
}
```

I've added words to the API design pages on the format of this response and what `code` should be.

This PR generally doesn't touch the existing error messages, just adds appropriate `code` properties.

One bigger change is the response of the Create Invitation end point - the one that was most different to the rest. This brings it more into line with this standard. It also fixes the reporting of the error in the UI

### DIFF
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -196,7 +196,7 @@ module.exports = async function (app) {
                 url: ''
             })
         } catch (err) {
-            reply.status(400).type('application/json').send({code: 'unexpected_error', error: err.message })
+            reply.status(400).type('application/json').send({ code: 'unexpected_error', error: err.message })
             return
         }
 
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -133,7 +133,7 @@ module.exports = async function (app) {
         // If we have roles that limit creation, that will need to be checked here.
 
         if (!teamMembership) {
-            reply.code(401).send({ error: 'Current user not in team ' + request.body.team })
+            reply.code(401).send({ code: 'unauthorized', error: 'unauthorized' })
             return
         }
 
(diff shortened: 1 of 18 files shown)


forge/routes/api/projectActions.js
```diff
@@ -62,14 +62,14 @@ module.exports = async function (app) {
             app.db.controllers.Project.clearInflightState(request.project)
             reply.send(result)
         } catch (err) {
-            reply.code(500).send({ error: err.toString() })
+            reply.code(400).send({ code: 'unexpected_error', error: err.toString() })
```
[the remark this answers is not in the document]
knolleary: I have reverted this codes to 500.

forge/routes/api/project.js
```diff
@@ -196,7 +196,7 @@ module.exports = async function (app) {
                 url: ''
             })
         } catch (err) {
-            reply.status(400).type('application/json').send({ error: err.message })
+            reply.status(400).type('application/json').send({code: 'unexpected_error', error: err.message })
```
hardillb, line 199:
```suggestion
            reply.status(400).type('application/json').send({ code: 'unexpected_error', error: err.message })
```

forge/routes/api/project.js
```diff
@@ -360,7 +360,7 @@ module.exports = async function (app) {
             )
             reply.send({ status: 'okay' })
         } catch (err) {
-            reply.code(500).send({ error: err.toString() })
+            reply.code(400).send({ code: 'unexpected_error', error: err.toString() })
```
hardillb, line 363:
```suggestion
            reply.code(500).send({ code: 'unexpected_error', error: err.toString() })
```
Sorry, should have tagged all of the ones that really are backend errors

forge/routes/api/projectActions.js
```diff
@@ -136,7 +138,7 @@ module.exports = async function (app) {
             }
             reply.send({ status: 'okay' })
         } catch (err) {
-            reply.code(500).send({ error: err.toString() })
+            reply.code(400).send({ code: 'unexpected_error', error: err.toString() })
```
hardillb, line 141:
```suggestion
            reply.code(500).send({ code: 'unexpected_error', error: err.toString() })
```

hardillb (changes requested):
I think that's all the 400 vs 500 errors

hardillb (approved):

PR #1037, opened 2022-10-03
add manual step (verify page) between email and verification API call

### LINKED ISSUE #1027: Email Verification should include a manual step
### Epic

_No response_

### Description

**As a:** FlowForge Admin

**I want to:** ensure that email verification involves human interaction

**So that:** I am sure the new user explicitly verifies their email address

Certain email platforms (e.g. hotmail) follow all links in received email to check for malicious content. At the moment this triggers email verification.

A potential solution would be to have the verification link load a page with a button that actually completes the process, ensuring a human interaction.

All Users, (CE)

### Acceptance Criteria

- [ ] No more automatic verification for Hotmail users.

Steve-Mcl:
fixes #1027

Adjusts the link sent in verify account email to open `/account/verifyemail/` which shows this to the user...
image

If the user is a new user (not member of team), it looks like this...
msedge_Doyc8rOwR7
_The continue button can be clicked to avoid wating for toast to complete_

If the user is a member of team, it looks like this...
msedge_AGmOMxDUoY


knolleary:
I would suggest skipping the 'continue' step - as that's now adding two steps rather than one.

Clicking Verify should be all they need to do and if successful, take them straight on to the team page (or the 'you need to be in a team' page).

Steve-Mcl:
> I would suggest skipping the 'continue' step - as that's now adding two steps rather than one.

It is not _really_ 2 steps as after the toast times out, the "continue" action is operated regardless & user is redirected.  

The "continue" is there as a visual cue and a catch all since it replaces the "verify" button.  During development, it was possible (due to forced errors) to remain on the page in an out of step state & this felt like the best way to handle it.

Happy to remove it if you insist?

hardillb:
Trying this locally, I'm just getting redirected straight to the login page.

Will dig a little deeper between sessions.

knolleary:
> It is not really 2 steps as after the toast times out, the "continue" action is operated regardless & user is redirected.

The user is presented a verify button, they are then presented a continue button - that is two steps. Users wont know the page will click continue for them if they just wait a bit - that would be unexpected behaviour and may cause a user to feel they have missed something if the page changes suddenly.

Also, having a toast/alert notification saying 'email verified' show on a page that also says 'email verified' is redundant. The alert is sufficient.

We do not need the continue view.

Steve-Mcl:
I have removed the delay. Now after completion it immediately "continue"s.

I did however leave the "continue" section in the view in the incase the user gets back to this page, in which case the "continue" button will be shown instead of the "verify" button

hardillb:
Still not getting shown the new page.

- enable external sign up
- sign up new users
- follow link in email

End up at login screen, without seeing the verify email button. I do see the console log at line 104 in frontend/src/pages/account/routes.js, but that's it.

Steve-Mcl:
@hardillb ah - ok, so I have been flipping the "email_verified" bit on existing accounts. My Bad. I should have tried creating new sign ups. DOH 

I will fix this up and ping you when done later.

Steve-Mcl:
@hardillb should be fixed up now.

hardillb:
Testing looks good. Will walk the code

forge/routes/auth/index.js
```diff
@@ -372,12 +380,12 @@ module.exports = fp(async function (app, opts, done) {
...
                 'VerifyEmail',
                 {
-                    confirmEmailLink: `${app.config.base_url}/account/verify/${verifyToken}`
+                    confirmEmailLink: `${app.config.base_url}/account/verifyemail/${verifyToken}`
                 }
             )
             reply.send({ status: 'okay' })
         } else {
-            reply.code(400).send({ status: 'error', message: 'Email already verified' })
+            reply.code(400).send({ code: 'invalid_request', error: 'Email already verified' })
```
hardillb, line 388: If it's already verified I think it would be better to just redirect to the login page

Steve-Mcl: Are you proposing I set the API to send a non 4xx code & then use that in the `vue` to re-direct?
Something return like "202 Accepted" (The request has been received but not acted upon) and simply redirect user to home (which will either show them their home or the login dialog if they are not logged in)

docs/contribute/workflows/signup.md
```diff
@@ -2,28 +2,36 @@
...
-    Runtime->>Runtime: Generate Email Verification Token
-    Runtime->>UserEmail: Send email containing verification link /account/verify/{token}
-    Runtime-->>-UI: { status: 'okay' }
+    US->>UI: Enters details on sign-up page
+    US->>UI: Clicks Sign-up
+    UI->>+RT: POST /account/register
+    RT->>DB: Create User
+    RT->>RT: Generate Email Verification Token
+    par Runtime to UserEmail
+        RT->>UE: Send email containing verification link /account/verifyemail/{token}
```
knolleary, line 18: Is there a reason for having changed this initial url? I would personally have chosen to keep it as `/account/verify/{token}` rather than a separate end point.

knolleary (commented):
I'm not clear why we need to add `verifyEmailInFlight` and `verifyEmailToken` to the store and I'm wary of adding lots of single-use properties to the store.

The only place I can see you reading `verifyEmailToken` is when the user clicks the button to do the verification. But as the token could be obtained directly from the router params at that point, I'm not clear why it has to come from the store instead.

Steve-Mcl:
> I'm not clear why we need to add `verifyEmailInFlight` and `verifyEmailToken` to the store and I'm wary of adding lots of single-use properties to the store.

Because without the inflight, the internal redirections vue does I was losing the token and the fact we were validating. Also, there are 2 scenarios - 1 where a new user is creating an account and 2 where the user already has an account but needs to re-verify email (flag was cleared)

It may be my initial attempts and layout etc were the main reason I needed  `verifyEmailInFlight` and `verifyEmailToken` but I am wearing of removing them due to the fights I had with the router on this final iteration. 

Happy to be advised.

docs/contribute/workflows/signup.md
knolleary: I don't see the issue with existing sent emails. Nothing will have changed about the fact the user has to visit the url `/account/verify/{token}` - it's what happens when they get there that is changing.

Steve-Mcl: Can I clarify please - is the suggestion we should change the API route 9move it out of the way) and leave the frontend route the same?

hardillb: Can it not just be GET for the webpage and POST to do the update?

knolleary:
Currently `GET /account/verify/{token}` is handled entirely on the server - despite the fact it appears in `routes.js` in the frontend (I think that is a legacy bit of code that is unused). The request then redirects to `/` regardless of success/failure.

The new sequence should be:
 - `GET /account/verify/{token}` takes you to the new 'VerifyEmail' page - which AFAICT doesn't involve any url changing/redirecting.
    - If the user is currently logged in and `user.email_verified` is true, then bounce straight to `/`
    - If no user currently logged in, or current user not verified, then stay on `VerifyEmail` page
 - The button on the page triggers a `POST` to `/account/verify/{token}` that responds with okay or failure.
 - If okay, I suggest doing a full reload to to `/`.
 - If not okay, show an appropriate error message

Given that sequence, I just don't see why you need to store anything. The `VerifyEmail` page has access to the router so can access the `{token}` part of the current url - that's all it needs to generate the url its `POST` should be sent to.

Steve-Mcl:
Yes, it makes sense but I ended up where I was because of (i think most likely) the way I initially had the new page embedded as a template in home.vue & the URL gets re-written (by vue) and the token was lost (not available in `this.$route`) when arriving at the script code in the component. 

But I see your point and it looks sane - so I will give it another try.


~1 point though needs a little more info -~
> GET /account/verify/{token} takes you to the new 'VerifyEmail' page 

~- are you saying the servers [API get Endpoint](https://github.com/flowforge/flowforge/blob/main/forge/routes/auth/index.js#L315) should redirect? Surely the backend API should observe separation of concern - it should be only used to "verify" a token?~
~OR~
~Should are you saying ...~

Never mind - my brain has just kicked back in.

On the PR I changed the backend API `GET /account/verify/:token` to a `POST` - so it wouldnt clash with the VUE route `/account/verify/{token}` anyhow 🤦

docs/contribute/workflows/signup.md
Steve-Mcl: @hardillb yes, I think so - as I wrote below in response to Nick, my brain has just kicked in 🙄

Steve-Mcl:
@joepavitt 

I have removed the inflight store vars & setup the code as per our "working expectation"

I have left a few comments in the code as clues/notes but these are found by searching `// TODO: Remove` 

Handing over now. Thanks.

Steve-Mcl:
@joepavitt 

The workflow for testing is...

### New User
1. Invite a user by email
3. Grab link from email
4. Open fresh browser page & should you see "VerifyEmail.vue"
5. Clicking the button should set the `email_verified` flag
6. user should be directed to "Create Account"
 
### Existing User
1. Clear the email_verified flag from user (in the SQL table or via user admin)
2. Sign in as user
3. Click button to generate new email
4. Grab link from email
5. Open fresh browser page & should you see "VerifyEmail.vue"
7. Clicking the button should set the `email_verified` flag

knolleary:
@Steve-Mcl I spoke to Joe earlier - I'm going to pick this up as I'm already up to speed on the issue.

knolleary:
I think I've got things largely in order now. There are some follow-up issues (#1057 and #1055 ) that need fixing as well.

I've documented the UX flows here: https://www.figma.com/file/mkOzyvy2b5Iu86xBpFaBZA/User-SignUp-Flows?node-id=0%3A1

I found we had some duplicate views depending on whether the user was logged in or not. I've removed some of that duplication so its a more consistent experience (and few paths to follow).

From an overall UX, I think there's a lot to be improved, but this is the incremental improvement to address the original issue - not a complete overhaul.

I'll push the changes shortly for review

knolleary:
@Steve-Mcl changes pushed. I would appreciate if you could run through some tests with it? Check the figma linked above for the expected sequences for the new user and existing user paths.

Steve-Mcl (commented):
### Working as expected for new user and existing user.

* ran on local dev
* tested the 2 scenarios outlined above


### Other
* Mermaid is out of date (review comment incoming)

Steve-Mcl:
@knolleary 

Tested and ready for review signoff (I can't "confirm" it as I was PR author)

frontend/src/App.vue
```diff
@@ -15,7 +15,7 @@
             </main>
         </template>
         <!-- Platform Entry Point -->
-        <template v-else-if="user && !user.password_expired && !termsAndConditionsRequired">
+        <template v-else-if="user && !user.password_expired && !termsAndConditionsRequired && user.email_verified !== false">
```
joepavitt, line 18: small gripe, but why not have `user.email_verified === true` - it's easier to read?

joepavitt (approved):

frontend/src/App.vue
knolleary: At the point of writing that code I couldn't remember if that was one of the properties that we only set on the user object if it is the 'unexpected' value. Having checked the User view code, I can see that isn't the case, so just setting for `user.email_verified` would be suitable.

For reference, we only set `password_expired` on the User object if the password is expired - otherwise it is undefined. But there we can lean into `undefined === false` so not a problem.

joepavitt: Happy to leave it as is - will merge.

PR #1052, opened 2022-10-10
/s/notifictions/notifications/

Pezmc:
Fixes a typo in the platform layout class names

### DIFF
--- a/frontend/src/layouts/Platform.vue
+++ b/frontend/src/layouts/Platform.vue
@@ -9,7 +9,7 @@
                 <div id="platform-banner"></div>
                 <slot></slot>
             </div>
-            <TransitionGroup class="ff-notifications" name="notifictions-list" tag="div">
+            <TransitionGroup class="ff-notifications" name="notifications-list" tag="div">
                 <ff-notification-toast v-for="(a, $index) in alertsReversed" :key="a.timestamp"
                                        :type="a.type" :message="a.message"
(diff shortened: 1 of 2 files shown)


Steve-Mcl:
Hi @Pezmc 

If this gets merged, then PR #1037 may also need correcting. (as the same code was taken from `Platform.vue` to permit toasts in a `ff-layout-box` container).

I would also suggest @joepavitt reviews this one as I think potentially all of the `.notifictions-xxx` entries should be `.ff-notifications-xxx` (or they are redundant?)

joepavitt:
@Steve-Mcl they're a bit of an oddball case because of how Vue Transition Groups work. The `name` attribute automatically gets converted to a group of classes that enable you to control animation styling during particular events/transitions for each of the elements, e.g. when an element in the notification list does an `enter`, `move`, `leave`.

As a result, we do in fact use the `.notification-list...` classes as they're generated from the `name="notification-list"`.

I left them without the `ff-` prefix in order to make sure we could see where they were originating from (as I thought confusion would arise is the CSS showed use of a `ff-notification-list-move` class that was never explicitly defined. At least by leaving `notification-list` in, then it was unique to that transition group.

joepavitt:
@Pezmc they classes aren't bound to the `ff-notification-toast` component as such. In FlowForge we've defined there enter/exit/move animations as these, but I didn't want to force that upon _anyone_ using `forge-ui-components`, so the underlying component library purely provides the individual toast notification and nothing else.

joepavitt (approved):

PR #1054, opened 2022-10-11
Hide editor links if viewing the project as a platform admin

### LINKED ISSUE #1028: Remove open editor links if viewing a project as an administrator
### Description

If the user is viewing the project page because they are an admin and not a member of the team they have access to all functionality with the exception of opening the editor.

Currently if they do load the editor Node-RED sits in a retry loop without any obvious feedback (to fix upstream)

We should remove the links to Open Editor and not make the Editor URL clickable if the user is viewing as an admin to further re-enforce this limitation
image

### Epic/Story

_No response_

Pezmc:
Follow up to https://github.com/flowforge/flowforge/issues/940, fixes #1028 

### After

Screenshot 2022-10-11 at 16 11 22

### Before

Screenshot 2022-10-11 at 16 11 44

I went on a deep dive of the e2e tests with cypress, before ultimately concluding this doesn't need explicit coverage.

### DIFF
--- a/frontend/src/pages/project/Overview.vue
+++ b/frontend/src/pages/project/Overview.vue
@@ -62,7 +62,7 @@ import AuditLog from '@/components/AuditLog'
 
 export default {
     name: 'ProjectOverview',
-    props: ['project'],
+    props: ['project', 'isVisitingAdmin'],
     computed: {
         options: function () {
             return [
--- a/frontend/src/pages/project/Overview.vue
+++ b/frontend/src/pages/project/Overview.vue
@@ -8,7 +8,15 @@
                     <tr class="border-b">
                         <td class="w-1/4 font-medium">Editor</td>
                         <td>
-                            <a v-if="editorAvailable" :href="project.url" target="_blank" class="forge-button-secondary py-1 mb-1"><span class="ml-r">{{project.url}}</span><ExternalLinkIcon class="w-4 ml-3" /></a>
+                            <div v-if="editorAvailable">
+                                <div v-if="isVisitingAdmin" class="my-2">
+                                    {{project.url}}
+                                </div>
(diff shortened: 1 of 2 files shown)


frontend/src/pages/project/index.vue
```diff
@@ -22,7 +22,7 @@
             </template>
             <template v-slot:tools>
                 <div class="space-x-2 flex">
-                    <a v-if="editorAvailable" :href="project.url" target="_blank" class="ff-btn ff-btn--secondary" data-action="open-editor">
+                    <a v-if="editorAvailable && !isVisitingAdmin" :href="project.url" target="_blank" class="ff-btn ff-btn--secondary" data-action="open-editor">
```
Pezmc, line 25: I was originally included to disable the button (and add a title/tooltip explaining why), but decided against this after discussing with @joepavitt.

frontend/src/pages/project/Overview.vue
```diff
@@ -8,7 +8,15 @@
                     <tr class="border-b">
                         <td class="w-1/4 font-medium">Editor</td>
                         <td>
-                            <a v-if="editorAvailable" :href="project.url" target="_blank" class="forge-button-secondary py-1 mb-1"><span class="ml-r">{{project.url}}</span><ExternalLinkIcon class="w-4 ml-3" /></a>
+                            <div v-if="editorAvailable">
+                                <div v-if="isVisitingAdmin" class="my-2">
```
joepavitt, line 12: This suggests in your "after" screenshot, we should be seeing the `project.url`, but I do not?

Pezmc: Unless I'm missing something, that's the `http://localhost:7881`

joepavitt: yep - sorry, you're absolutely right, and I just deleted my comment in realising that :D

joepavitt (approved):

PR #1056, opened 2022-10-11
Allow Flows/Credential to be imported to a project

### LINKED ISSUE #835: Import Credentials
### Epic

_No response_

### Description

**As a:** user

**I want to:** import a credentials file from another Node-RED into my project

**So that:** I can migrate my projects to FlowForge more easily.

Need to decide what to do with existing credentials in the project, ideally they would be merged, although for MVP it would be viable to make this only available if a project does not already have any credentials.
Ideally this would be available from the project settings page at any time after a project has been created, but it could also be part of project creation.
The user should have a dialog to upload or paste a credentials file and then enter the credentials secret.
There should be validation of the secret as part of the import.

- [X] All Users, (CE)
- [ ] Specific Project Types (EE)
- [ ] Specific Team Tiers (EE)
- [ ] Other (See comments)

### Acceptance Criteria

_No response_

hardillb:
Fixes #835


hardillb:
@joepavitt is there a way to disable the comfirm button on a dialog?

I want to disable it if there are no flows or credential files provided and if the files provided failed initial tests (e.g. not valid JSON)

joepavitt:
> @joepavitt is there a way to disable the comfirm button on a dialog?

checkout the first example in the docs: https://flowforge.github.io/forge-ui-components/#ff-dialog

hardillb:
@joepavitt Thanks, sorted.

Last major issues
-  is how to clear the file inputs if cancel is pushed. 
- How to handle a bad credentials secret

joepavitt:
> is how to clear the file inputs if cancel is pushed.
`@cancel="myFunction()"` on the `ff-dialog` will let you run functionality when a user presses the Cancel button

joepavitt:
> How to handle a bad credentials secret

Assume this is after the confirm button is hit? If so, close the dialog 9which would be done anyway) and then display a notification via the `alerts` service (https://flowforge.com/handbook/development/frontend/services/#alerts) should do the job

hardillb:
> @cancel="myFunction()" on the ff-dialog will let you run functionality when a user presses the Cancel button

Yep, but I'm not sure how to get a handle to the input objects (which are in the component) to set the value to null.

The error is appearing as a toast at the moment, but it probably needs trapping and cleaning up as it shows the JSON parse error.

joepavitt:
> Yep, but I'm not sure how to get a handle to the input objects (which are in the component) to set the value to null.

I'm assuming you've got something like:

```
<ff-dialog>
   ... stuff in here to define the form...
    <FormRow v-model="input.credential" />
</ff-dialog>
```

In which case, the scope of `input.credential` is still accessible at the top parent level that is displaying `ff-dialog`

hardillb:
It's not working that way at the moment (fully accept there might be a better way to do it), the component has it's own local data that is uses for the filenames and then pushes it back into a variable passed in. Have a look at `ImportProjectComponent.vue` and `ImportProjectDialog.vue`

frontend/src/pages/project/Settings/dialogs/ImportProjectDialog.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+    <ff-dialog ref="dialog" header="Import Flows" confirm-label="Import Project" @confirm="confirm()" kind="danger" :disablePrimary="disabled">
+        <template v-slot:default>
+            <form class="space-y-6" @submit.prevent>
+                <div class="mt-2 space-y-2">
+                    <p class="text-sm text-gray-500">
+                        Replace current flows by uploading a new flow file. This will restart the project to pick up the new flow or credentials
+                    </p>
+                </div>
+                <ImportProjectComponents id="importSettings" v-model="parts"/>
```
joepavitt, line 10: You can define a `clear()` function in `frontend/src/pages/project/components/ImportProjectComponents.vue`, then here,
add a `ref="project-components"`.

In the `@cancel` handler call `this.$refs['project-components'].clear() and it'll run the relevant functionality defined in that component

hardillb: VSCode complains if I try and use `v-model="localModelValue"` with a input of type file (in the FormRow.vue I had to add a new case for the file type)

joepavitt: You don't need to change anything re: `v-model`. I'm happy to put the relevant functionality in myself, or can talk through in a huddle.

joepavitt:
The test is failing because the `.ff-dialog-box button.ff-btn.ff-btn--danger` selector is picking up both the Delete button in the "Delete" dialog _*and*_ the newly introduced Import Project Dialog.

I'll put in a more precise selector

joepavitt (approved):

joepavitt:
@knolleary I've approved from my said, but @hardillb mentioned you may want to review/take a look before we merge

knolleary:
@joepavitt thanks - on my list for tomorrow

joepavitt:
@knolleary bump

test/unit/forge/routes/api/project_spec.js
```diff
@@ -1457,4 +1457,272 @@ describe('Project API', function () {
...
+                y: 140,
+                wires: []
+            }
+        ]
+        const credentials = {
+            $: '00fd012ce6da285e6f0fd83e7360afebP/e6OeYPMqTUPXO7L+9e3SUgvJ9NaoCtZbUPme8Y4GMbep8oQzZd64r3J4i1BVvxR5mBZD3kst++ke3114qhm/MmmxyT6V0mogRB0d7z/AapbjnVliXdvE4vGO1tsSHnYa4uf5k//STkK90='
+        }
+        const credsSecret = 'd8bc017ef274d0418725b23d86cdf7a65f8e3699340dd6b65e26600719ad2ac6'
+
+        it('Import Flow', async function () {
```
knolleary, line 1666: Whilst these tests verify the API returns the expected response, then should also be verifying that the Project's flows have actually been updated.

knolleary (changes requested):
Looks good, but I think the tests are missing an important part to verify the updates are actually applied to the `StorageFlow/Credentials` table

test/unit/forge/routes/api/project_spec.js
hardillb: Fair. I'll have a look shortly

hardillb: Test update (and found a bug)

knolleary (approved):
Just resolved a merge conflict. Happy to get this merged once the tests pass.

forge/db/controllers/Project.js
```diff
@@ -182,6 +182,59 @@ module.exports = {
...
+        }
+        if (project.state === 'running') {
+            app.db.controllers.Project.setInflightState(project, 'restarting')
+            project.state = 'running'
+            await project.save()
+            const result = await app.containers.restartFlows(project)
+            app.db.controllers.Project.clearInflightState(project)
+            return result
+        }
+    }
```
hardillb, line 236:
```suggestion
    },
```

hardillb:
All tests now passed :partying_face: merging

PR #1069, opened 2022-10-14
Update Docker README.md

robmarcer:
Edits to the Docs to cover Let's Encrypt support in the Docker build.


hardillb (approved):

hardillb:
@robmarcer just a FYI, the public docs don't get updated until we do a release at the moment.

Issue #1076, opened 2022-10-14
Inconsistent errors returned from the API


Pezmc:
### Description

While working on #929 I noted that there are several different standards that are used for returning errors from the API, including but not limited to:

Code and error as JSON:

```js
reply.code(400).send({ code: 'invalid_source_project', error: 'Source Project Not Found' })
```

Error explicitly as html

```js
reply.code(404).type('text/html').send('Not Found')
```

Error returned as html/json based on request type:

```js
reply.code(404).send('Source Project not found')
```

Error sent as a status and message:

```js
reply.code(400).send({ status: 'error', message: 'Email not enabled - cannot reset password' })
```

Error as error and description:

```js
reply.code(400).send({
    error,
    description
})
```

Error and details:

```js
reply.code(409).send({
    error: 'incorrect-project',
    project: request.device.Project?.id || null,
    snapshot: request.device.targetSnapshot?.hashid || null,
    settings: request.device.settingsHash || null
})
```

### Proposed Solution

Error `code` and `error` seems the most common format in the codebase and is informative for the API consumer.

 - [x] Add this standard to the handbook and ensure we're using it going forward
 - [ ] Update all other responses to match this format - _lower priority_

The consistency will make development against the API easier for the FlowForge team and other consumers.

Given we might have to update the front-end codebase to handle any changes to error codes, updating the existing responses might not be of value, but we should definitely agree on one standard for any newly introduced end points.

hardillb:
I thought @knolleary had been thought and standardised on `reply.code(xxx).send({error: '...', code:'....'}) recently.

It may be that he didn't get them all (or only did a specific subset)

knolleary:
See https://github.com/flowforge/flowforge/pull/1021 standardised this in the vast majority of cases and updated the API docs to outline the approach.

It didn't touch 404s as they are a bit more context specific - but I agree they could be tidied up further. In general, there aren't paths where the frontend will hit a 404 in the API.

The 409 example you've highlighted is a specific case of an API contract between the platform and device agent that cannot be easily changed for compatibility reasons.

Pezmc:
I somehow overlooked this in the docs [docs/contribute/api-design.md](https://github.com/flowforge/flowforge/blob/main/docs/contribute/api-design.md#error-formats) is the relevant link for anyone following along.

I'll do another pass to check if anything was missed but sounds like this is likely already solved.

knolleary:
[...]

PR #1079, opened 2022-10-14
Catch any unhandled  XHR errors that bubble up to the view

### LINKED ISSUE #929: 4xx Errors not shown in App
### Current Behavior

Certain actions return a 400 error, for example trying to change a project type on a stack (see #853)

While this particular error shouldn't have been returned (and is now fixed in #928) plus the UI prevents editing the project type once set, it does highlight that there is no indication to the user that the error occurred.

In theory it should not be possible for a user to make a bad request using the forge app as the front end should prevent this being sent, however as a belt & braces approach it would be better to show an error messsage if it does occur to aid in tracking down issues.

### Expected Behavior

400 Errors should trigger a notification toast.

### Steps To Reproduce

Make a request that results in a 400 error, 

Once way to do this is to edit a stack and then remove the disabled flag from the project type dropdown so that it can be changed.

### Environment

- FlowForge version: 0.8.0
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

Pezmc:
Fixes #929. If errors are not handled (with a try/catch or .catch) they bubble up to vue, which displays them in the UI using the Alerts service.

If the API has returned the error in the [standard format](https://github.com/flowforge/flowforge/blob/0dc8b449036ed3310b8d1c3ab9337a057ca174c6/docs/contribute/api-design.md?plain=1#L116-L125) that is displayed: 

Screenshot 2022-10-14 at 13 56 45

Otherwise, it falls back to the error from Axios:

Screenshot 2022-10-14 at 13 56 53

### DIFF
--- a/frontend/src/main.js
+++ b/frontend/src/main.js
@@ -1,10 +1,12 @@
 import { createApp } from 'vue'
+import { AxiosError } from 'axios'
 
 import ForgeUIComponents from '@flowforge/forge-ui-components'
 import '@flowforge/forge-ui-components/dist/forge-ui-components.css'
 import LottieAnimation from 'lottie-web-vue'
 
 import router from '@/routes'
+import Alerts from '@/services/alerts'
 import store from '@/store'
 import App from '@/App.vue'
 import '@/index.css'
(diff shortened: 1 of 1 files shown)


Pezmc:
Given this handling should solely be a fallback if error handling hasn't been implemented for the specific request, I've not added test coverage. It was validated using the following code which covers all error formats that exist in the codebase today (see #1076).

Shared for future reference: 

```js
const output = Math.floor(Math.random() * 13)
        switch (output) {
        case 0:
            return reply.code(400).send({ code: 'invalid_source_project', error: 'Source Project Not Found' })
        case 1:
            return reply.code(401).send({ code: 'unauthorized', error: 'unauthorized' })
        case 2:
            return reply.code(403).send({ error: 'User Suspended' })
        case 3:
            return reply.code(404).type('text/html').send('Not Found')
        case 4:
            return reply.code(404).send('Source Project not found')
        case 5:
            throw new Error('An error')
        case 6:
            return reply.code(404).send()
        case 7:
            return reply.code(405).send()
        case 8:
            return reply.code(500).send({ code: 'unexpected_error', error: (new Error('bla')).toString() })
        case 9:
            return reply.code(200).send()
        case 10:
            return reply.code(409).send({
                error: 'incorrect-settings',
                project: request.device.Project?.id || null,
                settings: request.device.settingsHash || null,
                snapshot: request.device.targetSnapshot?.hashid || null
            })
        case 11:
            return reply.code(400).send({ status: 'error', message: 'Email not enabled - cannot reset password' })
        default:
            return reply.code(501)
        }
```

joepavitt (approved):

PR #949, opened 2022-09-07
Add device count and project counts by type to admin

Pezmc:
Fixes #880 

Screenshot 2022-09-07 at 11 51 25

PR #1080, opened 2022-10-14
Hide the projects stats if empty or undefined

Pezmc:
Fixes a tiny UI bug introduced by me in #949

### Before

Screenshot 2022-10-14 at 15 17 46

### After

Screenshot 2022-10-14 at 15 09 58

This isn't needed for the users section as there will always be at least one admin to be able to visit this page.


joepavitt:
Am I going crazy, or have you got the same UI twice in those screenshots? Can't see the difference...

Issue #1077, opened 2022-10-14
Community limits should be displayed in the admin panel


ZJvandeWeg:
### Description

When starting an unlicensed FlowForge it has limitations for the community edition. When booting the server a lot message will tell the administrator how far along the limits are:

```
[2022-10-14T10:12:29.895Z] INFO: Usage:
[2022-10-14T10:12:29.895Z] INFO:  Users    : 0/150
[2022-10-14T10:12:29.895Z] INFO:  Teams    : 0/50
[2022-10-14T10:12:29.895Z] INFO:  Projects : 0/50
[2022-10-14T10:12:29.895Z] INFO:  Devices  : 0/50
```

It would be great to also show this on the administrator overview page in the GUI.

Jozefik:
Hey, is this issue free ? I can work on it.

knolleary:
From an API point of view, the existing `/stats` admin endpoint is used by the admin overview page to get this type of information and where the limits info should get added.

ZJvandeWeg:
@Jozefik Jep, totally!

Jozefik:
You want it inside License field or should I create new field ?

hardillb:
I think there are 2 choices: 

1. in the license section
2. in each of the boxes at the top, e.g. users 3/100 meaning 3 of 100 avaiable.


@joepavitt Any comment?

Jozefik:
I have done both. Just tell me which one is better. 

![image](https://user-images.githubusercontent.com/1617822/195853994-acfca077-5b63-4d60-82a5-93766982f989.png)

knolleary:
In the top boxes gets my vote.

Thanks @Jozefik - looks great.

Jozefik:
Okay fine, I will leave it in top boxes. And also before i make PR those limits only apply to Community edition yeah ?

knolleary:
A licensed version of the platform will have its own limits to report - so the `/stats` admin end point should be returning what the limits are (regardless of whether they are community or enterprise licensed).

Jozefik:
Yeah i added limits to `/stats` but comunity license doesn't provide that info with `app.license.get()` it only returns `null`.

knolleary:
I think stats should return for both types. Need to remind myself how the community defaults are picked up. Let me know if you want any pointers.

Jozefik:
@knolleary I will make PR and you can check how i fixed it. Also if I can ask, is it better to fork and make PR or create new branch and make PR ?

[...]

PR #1082, opened 2022-10-14
Added limits to admin panel

Jozefik:
Fixed issue #1077

### DIFF
--- a/forge/routes/api/admin.js
+++ b/forge/routes/api/admin.js
@@ -4,7 +4,7 @@ module.exports = async function (app) {
     app.get('/stats', async (request, reply) => {
         const userCount = await app.db.models.User.count({ attributes: ['admin'], group: 'admin' })
         const projectStateCounts = await app.db.models.Project.count({ attributes: ['state'], group: 'state' })
-        const license = await app.license.get() || {}
+        const license = await app.license.get() || app.license.defaults
         const result = {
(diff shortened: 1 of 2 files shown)


forge/routes/api/admin.js
```diff
@@ -4,13 +4,18 @@ module.exports = async function (app) {
     app.get('/stats', async (request, reply) => {
         const userCount = await app.db.models.User.count({ attributes: ['admin'], group: 'admin' })
         const projectStateCounts = await app.db.models.Project.count({ attributes: ['state'], group: 'state' })
+        const license = await app.license.get() || {}
```
hardillb, line 7: I feel we should find a way to populate this with the default values if no license present, then we don't have hard coded values in the UI portion that we may need to change later.

hardillb:
Sorry, meant to say thanks before suggesting changes....

forge/routes/api/admin.js
hardillb:
```suggestion
        const license = await app.license.get() || app.license.defaults
```

Jozefik: Sure i will change it. Thanks for the suggestion.

hardillb:
@knolleary happy for me to merge this? All tests have passed.

knolleary (approved):

PR #1089, opened 2022-10-17
Upgrade actions to drop NodeJS v12

hardillb:
There has been a warning on recent action runs saying that v2 versions of `actions/checkout` and `actions/setup-node` 

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12

```
runner-job (16.x)
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/setup-node, actions/checkout
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,9 +26,9 @@ jobs:
       matrix:
         node-version: [16.x]
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


hardillb:
As well as updating the actions to version that no longer use NodeJS v12 I've also locked the Postgres container to v14 (v15 released at the weekend and broke the docker compose setup)

PR #1090, opened 2022-10-17
Add palette.modules to Project Settings

### LINKED ISSUE #405: Preinstall NPM Modules
### Epic

#141

### Description

**As a:** user

**I want to:** specify NPM modules that are pre-installed as part of a template 

**So that:** I can start a project with my commonly used nodes

### Acceptance Criteria

- [ ] Template has list of 'pre-install modules'
- [ ] strings (npm package names) can be added or removed from the list
- [ ] version can be defined with `@1.2.3` syntax
- ...

knolleary:
Closes #405

This adds the ability to list npm modules that should get preinstalled in the project before Node-RED starts.

Note: there will be a companion PR against the launcher to match.

The Project-Settings-Palette view gets a new 'Installed Modules' table. This lists the additional modules that have been installed:

image

### Adding a new module

If the user is allowed to modify the installed modules, then the `+ add module` button is shown. Clicking on it reveals another row in the table:

image

Both fields use regex validation:

image

Note that we don't care if its a Node-RED node module or not - it just needs to be a valid npm module. This means it can be used to install `nrlint` and other plugins that don't otherwise show up in the palette manager.

### Editing an existing module

The pencil icon can be used to edit the version of a given module. Clicking it turns the version string into an editable input:

image

### Removing a module

The trashcan icon can be used to remove a module

### Applying the changes

After any changes are made, the user must click 'save settings' to... save the settings. They then must restart Node-RED to apply these changes (as is the case with *all* settings changes... a bit of UX to improve IMHO)

### Editing the palette in Node-RED

The user can still install new modules in the editor. We catch the update when Node-RED updates its runtime settings and make any necessary changes to our own record of what should be installed.

### DIFF
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -76,6 +76,7 @@ module.exports = async function (app) {
      * @memberof forge.routes.api.project
      */
     app.get('/:projectId', async (request, reply) => {
+        // request.project.runtimeSettings = await app.db.models.StorageSettings.byProject(request.project.id)
         const result = await app.db.views.Project.project(request.project)
         const inflightState = app.db.controllers.Project.getInflightState(request.project)
         if (inflightState) {
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -76,7 +76,6 @@ module.exports = async function (app) {
      * @memberof forge.routes.api.project
      */
     app.get('/:projectId', async (request, reply) => {
-        // request.project.runtimeSettings = await app.db.models.StorageSettings.byProject(request.project.id)
         const result = await app.db.views.Project.project(request.project)
         const inflightState = app.db.controllers.Project.getInflightState(request.project)
         if (inflightState) {
--- a/forge/db/controllers/Project.js
+++ b/forge/db/controllers/Project.js
@@ -76,6 +76,7 @@ module.exports = {
         moduleList.forEach(module => {
             modules[module.name] = module.version
         })
+        result.palette = result.palette || {}
         result.palette.modules = modules
         result.env = env
         return result
@@ -188,7 +189,7 @@ module.exports = {
     mergeProjectModules: async function (app, project, moduleList) {
         let changed = false
         let newProjectModuleList
-        const currentProjectSettings = await project.getSetting('settings')
+        const currentProjectSettings = await project.getSetting('settings') || {}
         if (currentProjectSettings?.palette?.modules) {
             // The project has an existing list of modules - need to resolve any
             // changes between the two lists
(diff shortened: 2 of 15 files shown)


knolleary:
Need to add some tests to this (and fix some existing tests...). Getting it raised now to get feedback on UI (Joe) and underlying approach (Ben)

joepavitt:
I am conscious here that we have a table of data that isn't using our `ff-data-table` component. That being said, visually, I think this is very clean and intuitive.

Any particular reason you avoided the data-table? (interested from a dev UX perspective more than anything else)

knolleary:
> Any particular reason you avoided the data-table? 

I did do an initial pass with ff-data-table, but it was getting quite involved to have the custom buttons in the cells that could invoke methods on the parent component. At the point I was doing `this.$parent.$parent.$parent.emit('...')` it felt like there must be a better way of doing things that I was missing - but I wasn't sure how best to distill that down to a specific requirement on data-table.

Given we already had this table layout done for the Environment Variables, I lifted that over to get things moving.

knolleary:
@joepavitt  One specific requirement here was to be able to have the last row in the table to be the 'add new entry' row - that needed very different content to the rest of the table - so didn't quite fit the model provided by ff-data-table at the top level (ie passing in `rows` and `columns` and letting it do all the work.

But now I think about it, I think it may have been doable if I used the `<template v-slot:rows>...</template>` to provide my own set of `<ff-data-table-row>`... hmmm... I may revisit that to see.

forge/routes/api/project.js
```diff
@@ -76,6 +76,7 @@ module.exports = async function (app) {
      * @memberof forge.routes.api.project
      */
     app.get('/:projectId', async (request, reply) => {
+        // request.project.runtimeSettings = await app.db.models.StorageSettings.byProject(request.project.id)
```
hardillb, line 79: Is this just saved for later?

hardillb (commented):
First pass looks good, need to pull it and give it a try when I have finished what I'm working on.

hardillb:
This isn't working for me when I run it locally, it's never updating the package.json in the project. (I have also pulled and checked out the manage-modules branch of the flowforge-nr-launcher in the dev env)

hardillb:
I've spun up a totally new flowforge-dev-env and things are a lot better (not installing multiple nr-launcher instances), but the "save settings" button is not enabling when I add a node on the settings page

hardillb:
> Have you 'confirmed' the add by clicking the 'tick'?

Yep. even tried adding a second node

forge/routes/api/project.js
knolleary: Removed

hardillb:
Looks to be all working now

knolleary:
@joepavitt I have raised #1100 to capture the need to migrate over to ff-data-table at some point.

I'm going to merge this as-is so we can get it into staging sooner rather than later for further validation

joepavitt (approved):

PR #1094, opened 2022-10-18
Update package dependencies

knolleary:
We've fallen behind on keeping our dependencies up to date.

A separate issue has already been raised on the handbook on the question of how we want to better maintain dependencies across all of the repos through process and automation.

This PR bumps the packages for a number of them. It doesn't cover *everything* that needs updating, but does a first pass of the more straight-forward version bumps that do not require any changes on our part - at least to the extent my checking of their change-logs and our unit tests are able to determine.

I expect to have follow-up prs for the remaining module updates to clear more of the backlog.


knolleary:
I've backed off updating eslint/eslint-standard as it introduces a new rule (`Expected property shorthand - object-shorthand`) that we don't pass cleanly on.

We *should* update those, but its a wider change that needs its own PR

Pezmc:
@knolleary I'm hesitant to jump quite so many major releases at once, but if you can get the tests to pass and it's merged before we start QA of 1.0, then we should catch anything that falls through the test coverage.

package.json
```diff
@@ -48,50 +48,49 @@
     "dependencies": {
         "@aws-sdk/client-ses": "^3.100.0",
         "@aws-sdk/credential-provider-node": "^3.100.0",
-        "@fastify/autoload": "^5.2.0",
```
Pezmc, line 51: Is autoload no longer a requirement?

knolleary:
> I'm hesitant to jump quite so many major releases at once

I have reviewed all of the major version bumps included here and they are pretty much all major due to dropping old node versions that we don't support anyway.

The only functional change was the csrf module that changed its default cypher choice, but that should have no impact on us.

package.json
knolleary: Correct - we used it early on but removed it. It was overlooked here.

Pezmc (approved):
Assuming you get the tests passing either by rolling back the rules or fixing the complaint 👍

knolleary:
@Pezmc yeah - working on the tests. This is a can of worms I partially wish I had delegated 😉

knolleary:
I'm abandoning this PR.

#1099 addresses eslint. Once that's done, I'll come back to a PR that updates sem-ver minor level updates.

PR #1099, opened 2022-10-19
Update eslint and fixup errors

knolleary:
- Updates to latest eslint and associated plugins
 - Sets object-shorthand to error, to ensure it is caught
 - Fixes up all resulting linting errors
    - All were related to object-shorthand and auto-fixes
    - One was the use of `Menu` as a component name in `DropdownMenu` - so renamed the component when exported to the template.


hardillb (approved):

knolleary:
@Pezmc oops - saw Ben had approved and went on auto-pilot. Apologies if you were looking already.

Issue #1065, opened 2022-10-13
Enable search over large data sets


joepavitt:
### Description

**As a:** Admin User

**I want to:** be able to query users of the FlowForge Platform

**So that:** I can find particular users/teams in order to assist with support tickets

Currently, the [`ff-data-table`](https://flowforge.github.io/forge-ui-components/#ff-data-table) component filtering only supports filtering of locally loaded content. Given that our API uses pagination, when we have large data sets, the filtering is limited to searching only the already loaded data.

As a result, we need to implement the following path:

1. As a user types into the filter input, call the relevant API, filtering by the search term, where that search term could cover any of the parameters detailed in the table. The `search-fields` property on `ff-data-table` will also provide data as to which fields can be assessed when filtering.
2. Given the response from the API, populate the table with the provided data.

### Acceptance Criteria

- [x] API - Add `?query=` support to paginated APIs (#1104)
- [x] FE - Call the API when a user types into the filter box (#1110)
- [x] FE - Filter the Admin > Users table as per the result of the new search API functionality (#1110)

hardillb:
We don't have any entities on Staging that are more than one page in a table so can't easily test this at the moment

knolleary:
Closing as all prs are merged

Steve-Mcl:
Added 100 random users (bash script below for future testing/reference). 

Some slight niggles - but is working...
* table empties and reloads with every character typed
* when clicking `more` the last entry is shown but the `more` button is still visible & then disappears after 1 more click.
* sorting is client side only (e.g. sorting by name does not reveal the last user `Zumba Zulu`)

bash script for generating names from "Arron Aardvark" to "Zumba Zulu" (all names are auto generated / fake)

```bash
[...]

PR #1104, opened 2022-10-19
Adds query capability to all paginated apis

knolleary:
Part of #1065 

This extends our standard for paginated APIs to include a `query=` parameter that can be used to filter the result.

For this initial MVP of search, we do case insensitive search against specific columns of the database. This isn't going to scale massively, but will be sufficient for the time being.

This PR implements this on all for our APIs that we already had pagination on. There are still some APIs that do not support pagination - but they collections that are typically much smaller in size and will fit within the default limit of 30 results.

A new utility function (`buildPaginationSearchClause`) is provided that can generate the appropriate sequelize `where` object that applies both the pagination and search logic. It is passed the pagination/search options from the url, any additional where constraints, and a list of the column names that should be searched:

```
where = buildPaginationSearchClause(pagination, where, ['User.username', 'User.name', 'User.email'])
```

## Note - pagination format has changed

This also updates our pagination design to remove the top-level `count` property. This required a number of updates in the frontend to use the length of the array returned, rather than the count.

Why? Because in writing the tests I found it was not returning the right value as soon as you were on the second page of results.

We were using `findAndCountAll()` which helpfully returned both the query results, *and* a full count from the table. When using `limit` as part of pagination, that would mean we could get the full count of objects, whilst only returning `limit` worth of them.

However, that returned count is actually the count of results as determined by the where clause. Our pagination relies on using a where clause to select objects *after* a particular object - which now means the count returned reflects the count of objects that match that criteria, *not* the full table count.

If we wanted to keep the count property in the API, we'd have to do a second query for every request to get that count.

Currently in the UI there are some places that really want the full count - and I've updated them to use the length of the array returned. This is okay for small collections that don't exceed a page size - but most of those places didn't take pagination into account anyway, so we're no worse off.

We need a task to follow-up here to fix those places to handle paginated responses better. Some of them just want the count and don't care about the actual results (for example, where we display the number of invitations a user has). Right now they are getting the full data, counting it and discarding the data - very inefficient. So some sensible optimisations can be made to provide a way to get the counts without returning the full data as well. But that's all for a separate issue/PR.

### DIFF
--- a/test/unit/forge/routes/api/team_spec.js
+++ b/test/unit/forge/routes/api/team_spec.js
@@ -1,7 +1,5 @@
 const should = require('should') // eslint-disable-line
 const setup = require('../setup')
-// const FF_UTIL = require('flowforge-test-utils')
-// const { Roles } = FF_UTIL.require('forge/lib/roles')
 
 describe('Team API', function () {
     let app


knolleary:
I appreciate this is a big PR. I recommend reviewing the commits in turn.

 - https://github.com/flowforge/flowforge/pull/1104/commits/1092cf50568c93d8faac11f699c5fbb6f2fb8ee1 - adds the search functionality to the paginated apis
 - https://github.com/flowforge/flowforge/pull/1104/commits/dd48635e987da9d341539cd71cf210dded0d911a - removes `count` from pagination api
 - https://github.com/flowforge/flowforge/pull/1104/commits/472f7c790cc9cf1521ef3bcc7c0bf61dd37ca55e - adds tests for search on user/teams pagination APIs.

Note I haven't written tests for *all* the searchable pagination apis - just the two we have the immediate need to support in the Admin console.

Pezmc:
> If we wanted to keep the count property in the API, we'd have to do a second query for every request to get that count.

Given the cache would be warm from the find query, I'm not sure there'd be significant extra load from running a second request, particularly when you consider that count is relatively cheap. I'd propose focusing on ensuring the columns being used for filtering had appropriate indexes set, rather than worrying about this extra query.

All that's to say I don't know if this is a problem we should be trying to engineer around, the only way to get the full size of the collection is to count it. I'd support returning the count property, and in the follow up you've proposed add a parameter to return only count (and/or without count?) for sections where it might not be relevant.

knolleary:
All good points @pez. Going to revisit the removal of `count`.

test/unit/forge/routes/api/team_spec.js
```diff
@@ -1,7 +1,7 @@
 const should = require('should') // eslint-disable-line
 const setup = require('../setup')
-const FF_UTIL = require('flowforge-test-utils')
-const { Roles } = FF_UTIL.require('forge/lib/roles')
+// const FF_UTIL = require('flowforge-test-utils')
+// const { Roles } = FF_UTIL.require('forge/lib/roles')
```
hardillb, line 4:
```suggestion
```

hardillb (approved):

Issue #876, opened 2022-08-11
Don't automatically start a copied project


sammachin:
### Description

When I create a new project from an existing one I want to be able to choose how it starts up

This might be the project is in a suspended state so that I can edit environment variables first or possibly the flows are stopped so that I can make changes to my flows before it is started.

sammachin:
Just creating the copy but in the suspended state doesn’t really help as you can’t edit the flow without starting, this would need the ability to start the project with the flows stopped from the forge app,

Possible options:
Expose a start/stop flows in the forge app and copy the project with the container running allowing the user to start the flows from the forge app
Copy the project with the flows stopped and expect the user to login to the editor and start the flows from there.

The 2nd one is potentially confusing for a user who may copy a project and then not understand why it is not running, so the messaging around copying a project would need to make this clear.

hardillb:
OK, looking at the stop flows feature in NR to implement this. It would require more than just the next deploy to start the flows.

It looks like this has not been shipped yet (target 3.0.3?) so not sure we can continue at the moment.

knolleary:
Copying over comment added to the PR (following discussion):

We do not currently expose flow state in the platform UI. If the flows are stopped, the project state says 'running' (as Node-RED is running).

There's a wider piece of work needed here to expose flow state to the platform properly. Rather than rush that in at this point, I propose we hold this out of 1.0.

Furthermore, and UX changes around how we report state should take #1046 into account as they are closely related and imminent pieces of work.

ZJvandeWeg:
@hardillb Any update on this issue? It's close to the release date, and it's still open.

knolleary:
I have label this issue as blocked due to my comments made [here](https://github.com/flowforge/flowforge/issues/876#issuecomment-1289118108)

Specifically, we don't currently expose the **flow state** to the user in any way. Until we do that, we don't have a meaningful way to create the project with the flows stopped, but the editor available.

As such, moving this back to the Product board and will raise an issue to cover the missing state handling

PR #1107, opened 2022-10-20
First pass at Starting copies in stopped state

hardillb:
Start of #876

For Stacks based on NR 3.0.2 or later this start the project with the flows in a stopped state requiring somebody to log into the editor and start the flows.

Not sure if it needs more.

### DIFF
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -264,6 +264,9 @@ module.exports = async function (app) {
                     }
                 }
             }
+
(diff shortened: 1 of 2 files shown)


hardillb:
@knolleary it's purely consumed by Node-RED, it reads it when it loads the runtime settings at startup (and resets it when you start the flows in the editor)

hardillb:
I propose closing the PR as it's been around since October 2022 and we've not needed it. We can raise a new work item if it comes up again.

PR #1108, opened 2022-10-20
Update README.md

robmarcer:


hardillb:
This will also need the mention of mosquitto (and using docker on windows) moving from section 4 of the "Manual Install" section.

robmarcer:
@hardillb I think it's OK to leave section 4 of the manual install  just needs a note saying it's optional? I think we should do the same for section 2 of the docker install too. OK?

robmarcer:
Not sure if the links to anchors I've added will work? It's valid MD as far as I can see.

PR #1114, opened 2022-10-21
Consistent Menu Item pointers

### LINKED ISSUE #1113: Mouse pointer inconsistent
### Current Behavior

menus on same page show different pointers

chrome_EgsM2iza98

### Expected Behavior

same pointer on all drop downs

### Steps To Reproduce

* Visit `project/xxx-xxx-xxx-xxx/snapshots` or `project/xxx-xxx-xxx-xxx/device`
* inspect the action menu and table kebab menu

### Environment

- FlowForge version: latest-git
- Browser: chrome

Steve-Mcl:
fixes #1113 

1. Ensure headlessui MenuItem pointer for active `menuitem` anchors are set to `pointer`
2. Ensure disabled menu items do NOT get a `pointer` 

NOTE: This also lead to finding a bug where the menu could be navigated and operated by keyboard due to lack of `disabled` attribute on the `MenuItem` - also fixed.

BEFORE: items have faint hover, are keyboard navigable, and menu is dismissed when operated

chrome_QO7eEGio1E

AFTER: disabled items and separators are no longer clickable nor dismiss the menu & no longer keyboard navigable. 

chrome_szSbHJUfVU

### DIFF
--- a/frontend/src/components/DropdownMenu.vue
+++ b/frontend/src/components/DropdownMenu.vue
@@ -17,7 +17,7 @@
         >
             <MenuItems :class="[edge === 'left' ? 'left-0 origin-top-left' : 'right-0 origin-top-right','z-50 absolute w-56 mt-1 bg-white divide-y divide-gray-100 rounded overflow-hidden shadow-lg ring-1 ring-black ring-opacity-10 focus:outline-none']">
                 <div class="apx-1 apy-1">
-                    <MenuItem v-for="(item, $index) in options" v-slot="{ active }" :key="$index">
+                    <MenuItem v-for="(item, $index) in options" v-slot="{ active }" :key="$index" :disabled="!item || item.disabled == true ? true : false">
                         <template v-if="item == null">
                             <hr/>
                         </template>
--- a/frontend/src/stylesheets/common.scss
+++ b/frontend/src/stylesheets/common.scss
@@ -54,3 +54,8 @@ label {
     color: $ff-grey-400 !important;
     opacity: 1; /* Firefox */
 }
+
+/* Ensure headlessui MenuItem pointer for active `menuitem` anchors */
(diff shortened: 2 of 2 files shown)


knolleary:
Not quite the approach I was imagining, although I realise I wasn't explicit in my comment on the issue. I had envisaged getting rid of the old Dropdown component entirely, rather than fix it.

Could you see how feasible that is to do?

Steve-Mcl:
~~I guess you mean get rid of the pre-built component menu and use `ff-kebab-menu`?~~

HOH! Just spotted your comment on the "issue"...

> This is because the main project action menu is still using the old DropdownMenu component - not the component provided by forge-ui-components.
> As such, this is not an issue for forge-ui-components - I will move over to the flowforge repo.

The issue I have with the `ff-kebab-menu` is that it is not keyboard navigable, does set `role` or aria attributes etc. I feel the headlessui menu is far more capable and developed than the `ff-kebab-menu` 

However I guess so long as the issues are raised on the components lib - we will get them resolved - so yeah, makes sense, I will take a 2nd shot at it.

knolleary:
@Steve-Mcl at this point, I don't want this to turn into a huge piece of work. If there are limitations of the ui-components menu, lets get those raised and resolved first, rather than ending up with a worse UX than what's there.

knolleary:
Which is to say, your current approach is pragmatic to fix the cursor and we can go with it now.

In parallel, raise the issues against ui-components and once those are addressed we can remove Dropdown menu - but that doesn't have to be a 1.0 task.

Steve-Mcl:
ok, np.

Then ready for review.

I've updated the comment to demonstrate the fixes in this PR

joepavitt:
@Steve-Mcl please link here, in the `forge--ui-components` issue once you've raised it too please.

PR #1115, opened 2022-10-21
Standardize 404 response format

### LINKED ISSUE #1076: Inconsistent errors returned from the API

knolleary:
Fixes #1076 

This updates our response format of 404 errors to adhere to our standard response format.

It also tidies up some other responses to ensure they have both `code` and `error` properties.

### DIFF
--- a/forge/routes/api/device.js
+++ b/forge/routes/api/device.js
@@ -16,17 +16,17 @@ module.exports = async function (app) {
                 try {
                     request.device = await app.db.models.Device.byId(request.params.deviceId)
                     if (!request.device) {
-                        reply.code(404).type('text/html').send('Not Found')
+                        reply.code(404).send({ code: 'not_found', error: 'Not Found' })
                         return
                     }
                     if (request.session.User) {
                         request.teamMembership = await request.session.User.getTeamMembership(request.device.Team.id)
                     }
                 } catch (err) {
-                    reply.code(404).type('text/html').send('Not Found')
+                    reply.code(404).send({ code: 'not_found', error: 'Not Found' })
                 }
             } else {
-                reply.code(404).type('text/html').send('Not Found')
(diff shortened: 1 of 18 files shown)


knolleary:
Force pushed having rebased on `main` to figure out the (unrelated) unit test failure - which is now fixed.

Pezmc (approved):
I've compared against the proposed principles in [your comment](https://github.com/flowforge/flowforge/issues/1076#issuecomment-1286776155) and it looks good! I think this closes #1076.

knolleary:
Have resolved the merge conflict. Once the tests pass this can be merged.

Issue #990, opened 2022-09-15
When a project is started/restarted launcher details not populated


hardillb:
### Current Behavior

When a project is starting for the first time or is being restarted after being suspended the version details are empty.

![Screenshot from 2022-09-15 10-52-05](https://user-images.githubusercontent.com/1800845/190374100-63316112-99f4-433b-9c74-9c3fef2e9e20.png)

Once it enters started state this does not get updated.

### Expected Behavior

The launcher, nodejs, and node-red version are populated when the project enters started state.

![Screenshot from 2022-09-15 10-54-56](https://user-images.githubusercontent.com/1800845/190374589-9d0f7378-dafe-4860-87c3-1c3690fefb7d.png)

### Steps To Reproduce

Start a new project

Or suspend a project, navigate to the team overview, back to the project and start it.

### Environment

- FlowForge version: 0.9
- Node.js version: 16.x
- npm version: 8.x
- Platform/OS: K8s/Linux
- Browser: Chrome

Steve-Mcl:
@hardillb looking at this for last 45 mins - tried about 20 different combinations with 2 FF installs and 3 different projects - I cannot get this to happen.
I only spend as long as I did as I remember seeing this myself about 2 weeks ago.

My gut feeling is it was fixed when recent PR to wait for project running status change was merged with all other changes in 0.10.0

Could you please re-test this at your side on latest/git version?

sammachin:
@Steve-Mcl I'm still seeing this on staging when creating a new project, If you're testing with localfs it may be hard to reproduce as the project starts much quicker than a k8s install

Steve-Mcl:
Thanks for that @sammachin - definitely something to look at

Steve-Mcl:
@hardillb I've noticed when running localfs the update comes through relatively quickly however on cloud there is no update at all (have to browse away then come back). 
I suppose I would start looking at the differences between localfs and k8s?

hardillb:
yep, and given the install demo this moring I think the delay going from starting to runnning is in a similar spot.

hardillb:
I'm going to say this is done, staging is behaving

PR #1131, opened 2022-10-25
Ensure page polls status when "installing" state

Steve-Mcl:
status polling improvements for #990


hardillb (approved):
Will merge when tests finish

PR #1134, opened 2022-10-25
Update template policy defaults

### LINKED ISSUE #1043: Settings should all be  enabled by default when creating a new template
When a new template is created (or new options are added to the platform) the default value for Editable are set to Locked, They should default to Editable.

joepavitt:
Closes #1043


knolleary:
We have a bit of code duplication, with an equivalent list in `forge/db/controllers/ProjectTemplate.js` that needs to be kept in sync.

This is ripe for pulling out to a shared lib as we do for UserRoles etc - I'll pick that up now as I'm between tasks and push to this branch.

joepavitt:
thanks for flagging - it worked locally for me when testing, so hadn't clocked a need to update elsewhere too.

knolleary:
@joepavitt  I've pushed the refactor - but as you raised the PR I cannot add you as a reviewer. Please can you take a look? Local testing works.

joepavitt:
> Local testing works.

verified - looks good

PR #1135, opened 2022-10-25
Update FormRow styling for disabled state in Project Settings

### LINKED ISSUE #1041: Improve Template creation & Edit Project Settings UX
### Epic

_No response_

### Persona 1:
**As a:** Platform Admin 

**I want to:** be able to easily setup a template project

**So that:** I can setup projects, without being overwhelmed by options.

### Persona 2:
**As a:** Flowforge User

**I want to:** be able to configure my project and not be overwhelmed by options

**So that:** I can configure my project to my liking, within the constraints of the respective template.

### Conclusion:
Template settings should all be enabled by default (not disabled). Currently, we overwhelm end users with options that they cannot edit (by default) if disabled in a template. 

Whilst the admin can go in and enable them in the underlying template one-by-one, in creating a default behaviour of these settings being editable, we reduce the onboarding friction, and the need for an admin to go through all of the available options (which will only grow) and understand each of them.

Additionally, we should not display locked/disabled settings in the Project > Settings view to a user. There is no value in doing so, and just overwhelms the user with options that they cannot action.

All Users, (CE)

### Acceptance Criteria

- [ ] #1043
- [x] #1042

joepavitt:
Blocked by https://github.com/flowforge/flowforge/pull/1134

Closes #1041


joepavitt:
@knolleary - noticing this test now failing, pretty sure this is the test needing an update, rather than the code being wrong?

```
1) Project Template controller
       Validate Settings
         only passes through settings a template policy allows:
     AssertionError: expected Object {
  disableEditor: true,
  codeEditor: 'monaco',
  palette: Object { allowInstall: true, nodesExcludes: 'ex.js' },
  modules: Object { allowInstall: true },
  env: Array [ Object { name: 'ALLOWED', value: '123' } ],
  httpAdminRoot: '/foo',
  dashboardUI: '/dashfoo'
} not to have property httpAdminRoot of '/foo' (false negative fail)
      at Assertion.value [as property] (node_modules/should/cjs/should.js:368:17)
      at Context.<anonymous> (test/unit/forge/db/controllers/ProjectTemplate_spec.js:94:36)
```

knolleary:
Test failure introduced by https://github.com/flowforge/flowforge/pull/1134 - oddly the actions didn't run against my commit on that so didn't spot the issue. Will PR the fixes shortly.

knolleary:
Ahh - PR #1134 merged the changes into this branch, *not* to `main` - which is why the tests are passing for me locally... I'm not running the right code.

joepavitt:
> Ahh - PR https://github.com/flowforge/flowforge/pull/1134 merged the changes into this branch, not to main

Apologies, should have explicitly mentioned that

knolleary:
Fix for tests pushed

knolleary (approved):

PR #1140, opened 2022-10-26
Have Admin view use Stack labels

### LINKED ISSUE #1137: Admin Stack table doesn't use label
### Current Behavior

You can now set human readable (none unique) labels for stacks.

These are not used in the Admin Stack table

### Expected Behavior

Admin Stack table should use Stack Label

### Steps To Reproduce

Add Label to stack

### Environment

- FlowForge version: 1.0
- Node.js version: 16.x
- npm version: 8.x
- Platform/OS: K8s/Linux
- Browser: All

hardillb:
Fixes #1137

### DIFF
--- a/frontend/src/pages/admin/Stacks/dialogs/AdminStackEditDialog.vue
+++ b/frontend/src/pages/admin/Stacks/dialogs/AdminStackEditDialog.vue
@@ -8,25 +8,33 @@
                     The existing stack will be marked inactive and will not be
                     available for use by new projects.
                 </div>
-                <div v-if="this.editDisabled">
-                    This stack is being used by projects. Its properties cannot
-                    be modified, other than to change its active state.
-                </div>
-                <FormRow v-model="input.name" :error="errors.name" :disabled="editDisabled">Name</FormRow>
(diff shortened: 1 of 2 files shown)


knolleary:
I have pushed an update that allows the label to be edited for in-use stacks.

It also now hides the uneditable properties of the stack in the edit dialog (if the stack is in-use). This makes the dialog clearer in terms of what is editable or not at that stage.

Also fixes an issue where editing a stack's properties would get reflected in the stack table even if the dialog was cancelled.

knolleary:
Also added descriptions to the name/label fields to help explain what they are - as there is some overlap here.

I think we should look at what role the 'name' field plays now. The reason we had to add the label field was the fact the db model has a unique constraint applied - and removing that constraint via a migration is tricky to get right (sqlite deleting things...).

image

hardillb:
Yep, we could look at totally hiding them and generating them going forward

knolleary:
I'm happy for this to go in now. Don't want to usurp @joepavitt as the marked reviewer, but if you could give the nod we can get this in.

PR #1141, opened 2022-10-26
Allow device to subscribe to all-project broadcast

knolleary:
The existing checks handled devices subscribing to a specific project's broadcast, but not all-projects (`+` in the topic).

This adds a check to ensure the device is in the team and assigned to a project.


Steve-Mcl (approved):
Yep, confirmed locally👍

PR #1144, opened 2022-10-26
Replace redundant <form> tags with <div>s

### LINKED ISSUE #1143: Pressing return in search box reloads page
### Current Behavior

While filtering a table (e.g. users, devices, teams or projects) if you type in the search box and then press enter it reloads the page (and looses the search term)

### Expected Behavior

Not reload the page

### Steps To Reproduce

Open any page with a table + search option
Start typing in the search box
Press enter

### Environment

- FlowForge version: 1.0.0
- Node.js version: 16.x
- npm version: 8.x
- Platform/OS: NA
- Browser: Chrome

joepavitt:
Closes #1143

### DIFF
--- a/frontend/src/pages/team/Settings/General.vue
+++ b/frontend/src/pages/team/Settings/General.vue
@@ -1,5 +1,5 @@
 <template>
-    <form class="space-y-6">
+    <div class="space-y-6">
         <FormRow v-model="input.teamName" :type="editing.teamName?'text':'uneditable'" :error="errors.teamName" id="teamName" ref="name-row">
             <template #default>Name</template>
             <template #description>
@@ -37,8 +37,7 @@
                 </div>
             </template>
         </div>
-
-    </form>
+    </div>
 </template>
 
 <script>
--- a/frontend/src/pages/team/Settings/General.vue
+++ b/frontend/src/pages/team/Settings/General.vue
@@ -1,5 +1,5 @@
 <template>
-    <div class="space-y-6">
+    <form class="space-y-6">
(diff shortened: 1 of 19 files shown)


frontend/src/pages/team/Settings/General.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <form class="space-y-6">
+    <div class="space-y-6">
```
Pezmc, line 2: For this (and the some of other settings changes) semantically `form` does feel like a better fix than div. This part of the page does indeed contain a form, even if we're not using HTML form handling.

joepavitt: very fair point - in most cases where we are using form - it isn't a form. Happy to revisit each of them and sanity check their usage.

knolleary: Whilst I do agree (which is why I used them originally), this is also a pragmatic fix for the issue as we don't rely on any of the browser-defined `<form>` behaviours.

We should aspire to restore `<form>` where appropriate - but will need to solve the UX issue around unexpected/inappropriate form submission.

knolleary (approved):

joepavitt:
@knolleary just a note as I've just pushed post-approval, that I've just reverted some based on @Pezmc's feedback

frontend/src/pages/team/Settings/General.vue
Pezmc: I'm :+1: to merge this as is and follow up (or create an issue) to restore the forms where appropriate.

Pezmc (approved):

PR #1151, opened 2022-10-27
Fix template create and edit

### LINKED ISSUE #1148: Description of template not saved when created
### Current Behavior

1. description is not saved when creating a new template
2. template edit form (sometimes) shows as "dirty" when simply opening the template. (see steps to reproduce)

chrome_htKUmac5pS

### Expected Behavior

1. description should be stored when template is created
2. template edit should be clean when opened/unedited

### Steps To Reproduce

1. create template with description
   a. description is not saved
1. edit template description & save
   a. description is set
1. From this point (until a full page refresh) opening the settings of any template marks the form dirty (e.g. the "Discard Changes" button is shown)

### Environment

- FlowForge version: git
- Node.js version: 16
- Platform/OS: windows
- Browser: chrome latest

Steve-Mcl:
fixes #1148 

* adds missing fields `description` and `active`
  * fixes newly created template has no description
  * fixes newly created template (without every setting the `active` checkbox) being considered active and selectable for use
* ignores `module_templates` in the `admin/settings/x` routes
  * fixes template edit "Discard Changes" when there are no changes
* spelling fix
* component name fix

### DIFF
--- a/frontend/src/pages/admin/Template/index.vue
+++ b/frontend/src/pages/admin/Template/index.vue
@@ -108,6 +108,11 @@ export default {
                     changed = changed || this.editable.changed.description
 
                     templateFields.forEach(field => {
+                        if (field === 'palette_modules') {
+                            // Don't check `palette_modules` for changes.
+                            // They are part of the template but are not edited here in admin/settings
+                            return
+                        }
                         this.editable.changed.settings[field] = this.editable.settings[field] !== this.original.settings[field]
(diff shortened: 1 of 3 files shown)


Steve-Mcl:
We may want to consider setting the `Active` check `true` by default if that is what `undefined` is considered to be?

By that I mean - before this PR, the `template` variable was missing the `active` flag.  So when a template was created it got `1` in the DB...
image

Issue #1164, opened 2022-10-28
Pagination broken on stacks page


Pezmc:
### Current Behavior

Pagination keeps returning the first page from the API

![2022-10-28 13 39 57](https://user-images.githubusercontent.com/507155/198579059-9269d81d-7113-460b-83d4-5e0205569316.gif)

### Expected Behavior

Should return new results and add them to the bottom of the list

### Steps To Reproduce

Have enough stacks to trigger pagination for stacks, click load more.

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

robmarcer:
I'm assuming this is a bug for the purpose of the release blog. I'll also add the bug tag.

PR #1173, opened 2022-11-02
Refactor stacks logic to only hold one list client side

Pezmc:
**Warning:** Merge target is https://github.com/flowforge/flowforge/pull/1167 do not merge until after that PR.

This has negligible impact on the time complexity for calculating inactiveStacks/activeStacks (from `O(n)` to `O(n2)` but still linear), with the following advantages:

- Easier to maintain by removing all the checks for `active`
- Maintaining one rather than three lists that can end up out of sync (see #1164).
- `O(1)` lookup of any stack using [`Map.get()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/get) - previously required searching over array with `indexOf`

I've tested with 1000 stacks locally and the page is equally performant

### DIFF
--- a/frontend/src/pages/admin/Stacks/index.vue
+++ b/frontend/src/pages/admin/Stacks/index.vue
@@ -70,13 +70,26 @@ const StackName = {
     props: ['id', 'name', 'label', 'description', 'projectTypeName'],
     components: { DesktopComputerIcon }
 }
+
+function comparator (A, B) {
+    if (A.projectType === B.projectType) {
+        return A.createdAt.localeCompare(B.createdAt)
+    } else {
+        if (!A.projectType) {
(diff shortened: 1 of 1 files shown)


joepavitt (approved):
Great work

joepavitt:
@Pezmc just realised the E2E tests didn't run on this PR because you weren't merging with `main` - upon merging, the E2Es have failed: https://github.com/flowforge/flowforge/actions/runs/3376566923/jobs/5604443717

can you investigate please?

Pezmc:
@joepavitt I was hoping they'd run when the base was changed! Looking now, but will also open a PR to change our GitHub config.

PR #1177, opened 2022-11-02
Run CI checks against all PR's

Pezmc:
I often have PR's that depend on each other so set the bases accordingly. However, that means the tests might not run (see #1173).

GH Workflows will now run for all PR's in this repo, regardless of target. 

I imagine this will have a relatively negligible impact on our GH actions runtime, but I don't have access to the stats, so will ask @knolleary or @ZJvandeWeg to check in a week.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,6 @@ on:
       - 'test/system/**'
       - 'package.json'
   pull_request:
-    branches: [ main ]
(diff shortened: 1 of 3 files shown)


hardillb:
I don't think we are close to the limits at the moment, so I'd rather run tests more than miss things. Go for it

joepavitt (approved):

PR #1182, opened 2022-11-03
Change editor auth to use AccessToken rather than Session

knolleary:
This PR changes how the editor authentication is managed in the platform.

We have two tables in the database related to access control - `AccessToken` and `Session`.

 - `Session` - represents a login session in the UI by a real user. For each request, it uses the `sid` cookie to look up the appropriate Session object to get the User. If valid, it gives the request the full range of access that User has.
 - `AccessToken` - this represents a token that can be passed in via the `authorisation` header to authenticate a request.

When logging into the Node-RED editor, the oauth flow between the `nr-auth` plugin and the platform would store its state as a `Session`. However the storage/logging apis expect the token to be provided in the auth header. This mean we had a special case in the code where, if the token 'looked' like an nr-auth token, it would divert to the Session table.

This PR changes the oauth flow to store things in the AccessToken table instead of Session. This removes the need to have a special case for the editor token and is more logically consistent.

It also steps closer to allowing the `AccessToken` to have an associated `scope` (already in the table... just not used widely) - meaning a token could be generated for a user that has only a specific set of permissions, rather than full access.

### DIFF
--- a/forge/db/models/AccessToken.js
+++ b/forge/db/models/AccessToken.js
@@ -17,7 +17,7 @@ module.exports = {
             type: DataTypes.STRING,
             allowNull: false,
             get () {
-                const rawValue = this.getDataValue('scope')
+                const rawValue = this.getDataValue('scope') || ''
                 return rawValue.split(',')
             },
             set (value) {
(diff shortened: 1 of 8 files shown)


forge/routes/auth/oauth.js
```diff
@@ -169,12 +169,22 @@ module.exports = async function (app) {
...
                     state: requestObject.state
                 })
+                console.log('/account/complete redirect', responseUrl.toString())
                 reply.redirect(responseUrl.toString())
                 return
             }
         }
         return badRequest(reply, 'access_denied', 'Access Denied')
     })
+    app.get('/account/reject/:code', async function (request, reply) {
```
knolleary, line 179: This slipped in and isn't strictly part of the PR - but will be part of the next PR in this area :)

forge/db/models/AccessToken.js
```diff
@@ -29,18 +29,35 @@ module.exports = {
...
+        ownerType: { type: DataTypes.STRING },
+        refreshToken: {
+            type: DataTypes.STRING,
+            set (value) {
+                if (typeof value === 'string') {
+                    this.setDataValue('refreshToken', sha256(value))
+                }
+            }
+        }
     },
```
hardillb, line 41: I'm not sure what the if block actually does here?

forge/routes/auth/index.js
```diff
@@ -44,18 +44,16 @@ module.exports = fp(async function (app, opts, done) {
                 if (scheme !== 'Bearer') {
                     reply.code(401).send({ code: 'unauthorized', error: 'unauthorized' })
                 }
-                if (/^ff[td]/.test(token)) {
```
hardillb, line 47: What is this going to do to live sessions at the point of upgrade?

Are we going to log all existing sessions out?

knolleary: A side-effect will be to invalidate existing edit sessions. We *might* be able to add a migration that moves the appropriate tokens from one table to the other.... will take a look.

forge/db/models/AccessToken.js
knolleary: If `refreshToken` is `null` then we shouldn't try to hash it.

knolleary:
I'm going to get this merged now to move forward with the plugin auth, otherwise things will get backed up.

PR #1199, opened 2022-11-09
Rename `values.yml` to `customizaton.yml`

hardillb:
Just to make it more obvious and not clashing with the `values.yaml` with default in the chart


ZJvandeWeg:
So simple, it's nearly briljant. Thanks for the PR

PR #1201, opened 2022-11-10
Form Row

### LINKED ISSUE #1133: Update FormRow
### Description

Replace/update the `FormRow` component to use the relevant `forge-ui-components` components, e.g. `<select>` > `ff-dropdown`.

The FormRow component itself needs it's own `ff-form-row` class which will give us more consistent styling control (currently pure tailwind which is resulting in a lot of duplicated CSS and state management (e.g. `disabled`)

### Epic/Story

_No response_

joepavitt:
closes #1133 

- ~Requires https://github.com/flowforge/forge-ui-components/pull/94~ and consequently a ~forge-ui-components version bump~ once those changes are merged
- forge-ui-components release: https://github.com/flowforge/forge-ui-components/releases/tag/v0.4.4

**Updates:**
- Remove Tailwind Forms dependency
- Bump ui-components version
- Add comments in `FormRow.vue` to help decipher the if/else logic easier
- Remove radio-button option in FormRow. Instead, we use `ff-radio-group` from `forge-ui-components`. This is because there is dependency across each "row" and it's much easier to work with a dedicated component for handling this logic.
- Update standard HTML `<input type="checkbox" />` to `<ff-checkbox>`
- Update standard `<select>` to use `<ff-dropdown>`
- Unit test coverage for Form Rows


test/unit/frontend/components/form-row.spec.js
```diff
@@ -0,0 +1,197 @@
...
+                type: 'file'
+            },
+            shallow: true
+        })
+
+        expect(wrapper.find('ff-checkbox').exists()).toBe(false)
+        expect(wrapper.find('input[type="file"]').exists()).toBe(true)
+        expect(wrapper.find('ff-dropdown').exists()).toBe(false)
+        expect(wrapper.find('ff-text-input').exists()).toBe(false)
+        expect(wrapper.find('.uneditable').exists()).toBe(false)
```
Pezmc, line 39: I'm slightly concerned that asserting a thing doesn't exist with a non class specific classname is kind of brittle, but not blocking.

test/unit/frontend/components/form-row.spec.js
```diff
@@ -0,0 +1,197 @@
...
+        expect(FormRow).toBeTruthy()
+
+        const wrapper = mount(FormRow, {
+            props: {
+                type: 'checkbox'
+            },
+            shallow: true
+        })
+
+        // console.log(wrapper.html())
```
Pezmc, line 16: ✂️

Pezmc (approved):
This was a lot, I've gone though commit by commit and things look good!

Pezmc:
Possible follow up, update https://github.com/flowforge/flowforge/pull/1257 to use the new dropdown.

test/unit/frontend/components/form-row.spec.js
joepavitt: Assuming you're just referring to the `console.log` here, not the full 4 lines included in the comment?

joepavitt:
> Possible follow up, update https://github.com/flowforge/flowforge/pull/1257 to use the new dropdown.

If you're using `FormRow` - that'll happen automatically, once all merged. Otherwise, the `ff-dropdown` component is already available in `forge-ui-components`, so can be used directly now as part of that PR

test/unit/frontend/components/form-row.spec.js
joepavitt: https://github.com/flowforge/flowforge/pull/1201/commits/835855731d4f8661b98627310b4f9884cc4e7289

test/unit/frontend/components/form-row.spec.js
joepavitt: Resolved in https://github.com/flowforge/flowforge/pull/1201/commits/835855731d4f8661b98627310b4f9884cc4e7289

Pezmc:
It's using the `ff-dropdown` component directly via slots right now: https://github.com/flowforge/flowforge/pull/1257/files#diff-db516fb72d730bd703bffb9cae9a98344ece1caabd83b48bcb5f7fc10342981dR28-R42

PR #1203, opened 2022-11-10
Add route to test Bearer tokens are active

hardillb:
Used by the File Server project

### DIFF
--- a/forge/routes/auth/index.js
+++ b/forge/routes/auth/index.js
@@ -539,10 +539,14 @@ module.exports = fp(async function (app, opts, done) {
         }
     })
 
-    app.get('/token/test', {
+    app.get('/token/test/:projectId', {
         onRequest: verifyToken
     }, async (request, reply) => {
-        reply.code(200).send()
+        if (request.session.ownerType === 'project' && request.params.projectId === request.session.ownerId) {
+            reply.code(200).send()
+        } else {
+            reply.code(401).send({ code: 'unauthorized', error: 'unauthorized' })
+        }
     })
 
     done()
--- a/forge/routes/auth/index.js
+++ b/forge/routes/auth/index.js
@@ -539,16 +539,6 @@ module.exports = fp(async function (app, opts, done) {
         }
     })
 
-    app.get('/token/test/:projectId', {
-        onRequest: verifyToken
-    }, async (request, reply) => {
-        if (request.session.ownerType === 'project' && request.params.projectId === request.session.ownerId) {
-            reply.code(200).send()
-        } else {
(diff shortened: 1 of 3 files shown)


hardillb:
Realised last night this also needs to say which project the token belongs, will re-open shortly

forge/routes/auth/index.js
```diff
@@ -539,6 +539,16 @@ module.exports = fp(async function (app, opts, done) {
         }
     })
 
+    app.get('/token/test/:projectId', {
```
knolleary, line 542: All of the token generation routes live in `oauth.js` rather than `index.js`. This should move over there.

We have different types of token - not just project tokens - so the path ought to be scoped so we can space to add similar routes for devices etc in the future. 

I suggest: `/account/check/:ownerType/:ownerId`

knolleary (changes requested):

forge/routes/auth/oauth.js
```diff
@@ -301,4 +301,18 @@ module.exports = async function (app) {
             badRequest(reply, 'invalid_request', "Invalid grant_type. Only 'authorization_code' and 'refresh_token' are supported")
         }
     })
+
+    app.get('/account/check/:ownerType/:ownerId', {
+        onRequest: async (request, reply) => {
+            await app.verifyToken(request, reply)
```
hardillb, line 307: For some reason I had to wrap this to get it to work

forge/routes/auth/oauth.js
```diff
@@ -301,4 +301,18 @@ module.exports = async function (app) {
...
+        }
+    }, async (request, reply) => {
+        console.log(request.params)
+        console.log(request.session)
```
knolleary, line 311:
```suggestion
```

knolleary (approved):

knolleary (commented):
Looks good - but would benefit from a couple tests being added, especially as this is around security.

hardillb:
Will see what tests I can write this afternoon

knolleary (approved):

test/unit/forge/routes/auth/index_spec.js
```diff
@@ -141,4 +141,45 @@ describe('Accounts API', async function () {
...
+                method: 'GET',
+                url: '/account/check/project/random',
+                headers: {
+                    authorization: `Bearer ${authTokens.token}`
+                }
+            })
+            response.statusCode.should.equal(401)
+        })
+
+        it('Fail to verify with random project id', async function () {
```
Steve-Mcl, line 172: Duplicate test names - causes issues with test runners and makes problem diagnosis that tiny bit harder ;)

Issue #1174, opened 2022-11-02
Deployments: Rename devices section and tweak table layout


Pezmc:
### Description

![image](https://user-images.githubusercontent.com/507155/200310255-c27d2804-15e5-4f53-9af8-9bdebed0b118.png)

### Epic/Story

#1046

Pezmc:
Fixed in #1204

PR #1204, opened 2022-11-10
Project > Deployments Page

Pezmc:
Continued work on #1174, fixes https://github.com/flowforge/flowforge/issues/1176, this PR:

- Reorders the devices table columns
- Adds a cloud deployments table (title not settled on yet) to the top of the page
- Adds information regarding the last flow deployment to the projects API
- Refactors all the inline vue components into separate files

Screenshot 2022-11-10 at 17 33 30

### DIFF
--- a/forge/containers/stub/index.js
+++ b/forge/containers/stub/index.js
@@ -29,7 +29,7 @@ module.exports = {
         this._app = app
 
         const projects = await this._app.db.models.Project.findAll()
-        projects.forEach(async (project) => {
+        projects.forEach(project => {
             if (project.state !== 'suspended') {
                 const p = {
                     id: project.id,
--- a/forge/containers/stub/index.js
+++ b/forge/containers/stub/index.js
@@ -102,13 +102,6 @@ module.exports = {
                 const stubProjectToken = forgeUtils.generateToken(8)
                 await project.updateSetting('stubProjectToken', stubProjectToken)
             }
-            const existingFlow = await this._app.db.models.StorageFlow.byProject(project.id)
-            if (!existingFlow) {
-                await this._app.db.models.StorageFlow.create({
-                    flow: JSON.stringify([]),
-                    ProjectId: project.id
-                })
-            }
             if (project.name === 'stub-fail-start') {
                 return new Promise((resolve, reject) => {
                     setTimeout(() => {
--- a/forge/containers/stub/index.js
+++ b/forge/containers/stub/index.js
@@ -29,7 +29,7 @@ module.exports = {
         this._app = app
 
         const projects = await this._app.db.models.Project.findAll()
-        projects.forEach(project => {
+        projects.forEach(async (project) => {
             if (project.state !== 'suspended') {
                 const p = {
                     id: project.id,
@@ -101,6 +101,13 @@ module.exports = {
             if (await project.getSetting('stubProjectToken') === undefined) {
                 const stubProjectToken = forgeUtils.generateToken(8)
                 await project.updateSetting('stubProjectToken', stubProjectToken)
+            }
+            const existingFlow = await this._app.db.models.StorageFlow.byProject(project.id)
+            if (!existingFlow) {
+                await this._app.db.models.StorageFlow.create({
+                    flow: JSON.stringify([]),
+                    ProjectId: project.id
+                })
             }
             if (project.name === 'stub-fail-start') {
                 return new Promise((resolve, reject) => {
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -74,8 +74,14 @@ module.exports = async function (app) {
      * @memberof forge.routes.api.project
      */
     app.get('/:projectId', async (request, reply) => {
-        const result = await app.db.views.Project.project(request.project)
+        const [result, projectFlow] = await Promise.all([
+            await app.db.views.Project.project(request.project),
+            await app.db.models.StorageFlow.byProject(request.project.id)
+        ])
         const inflightState = app.db.controllers.Project.getInflightState(request.project)
+
+        result.flowLastUpdatedAt = projectFlow?.updatedAt
+
         if (inflightState) {
             result.meta = {
                 state: inflightState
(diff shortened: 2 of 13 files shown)


joepavitt:
@knolleary do we have an API call available to us for when the Node-RED instance was last deployed? See the "last deployed" column in the "Cloud" table

Pezmc:
@hardillb And I went looking for an equivalent, and the best we found was to look for the last time a Projects Flows had been updated, which happens when a snapshot is restored or the deploy button is clicked; as far I could find, this is the closest equivalent we have for a project being deployed.

The screenshot shows never as I was using a stubbed driver (i.e. there are no flows to deploy), if using localfs if works as expected.

forge/routes/api/project.js
```diff
@@ -74,8 +74,14 @@ module.exports = async function (app) {
...
      */
     app.get('/:projectId', async (request, reply) => {
-        const result = await app.db.views.Project.project(request.project)
+        const [result, projectFlow] = await Promise.all([
+            await app.db.views.Project.project(request.project),
+            await app.db.models.StorageFlow.byProject(request.project.id)
+        ])
         const inflightState = app.db.controllers.Project.getInflightState(request.project)
+
+        result.flowLastUpdatedAt = projectFlow?.updatedAt
```
Pezmc, line 83: Using the Projects Flows Storage last updated at to track deploys (i.e. any time a projects flows have been written to disk, they've been "deployed").

frontend/src/pages/project/Deployments.vue
```diff
@@ -1,94 +1,151 @@
 <template>
```
Pezmc, line 1: View this file with [?w=1](https://github.com/flowforge/flowforge/pull/1204/files?w=1) for easier review

forge/containers/stub/index.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
         this._app = app
 
         const projects = await this._app.db.models.Project.findAll()
-        projects.forEach(project => {
+        projects.forEach(async (project) => {
```
hardillb, line 32: Why does this need to be async? it's not doing any work, just pushing to a local variable

frontend/src/api/project.js
```diff
@@ -12,6 +12,7 @@ const getProject = (projectId) => {
     return client.get(`/api/v1/projects/${projectId}`).then(res => {
         res.data.createdSince = daysSince(res.data.createdAt)
         res.data.updatedSince = daysSince(res.data.updatedAt)
+        res.data.flowLastUpdatedSince = daysSince(res.data.flowLastUpdatedAt)
```
hardillb, line 15: Does this handle a null/undefined value?

hardillb (commented):
The backend stuff looks ok, just the question about existing projects with no flows in the table

forge/containers/stub/index.js
```diff
@@ -102,6 +102,13 @@ module.exports = {
                 const stubProjectToken = forgeUtils.generateToken(8)
                 await project.updateSetting('stubProjectToken', stubProjectToken)
             }
+            const existingFlow = await this._app.db.models.StorageFlow.byProject(project.id)
+            if (!existingFlow) {
+                await this._app.db.models.StorageFlow.create({
+                    flow: JSON.stringify([]),
+                    ProjectId: project.id
+                })
+            }
```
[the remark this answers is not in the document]
knolleary: Given other drivers don't do this, there is a risk here we mask in our testing the edge case where a project has been created but nothing yet deployed.

If we want to ensure there's something in the StorageFlow table for every project, we should do this bit of work as part of the Project creation flow in the core, so it happens regardless of the driver.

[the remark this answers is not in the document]
hardillb: Doh, I'd missed this was the stub and assumed it was the wrapper, so yeah, more reason why we may have no flow in the table for all the other drivers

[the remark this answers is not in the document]
Pezmc: Yes, I struggled to find a better place to put this. I want the stub driver to be able to emulate a "deploy" having happened so we can fully test the UI even in stub mode. Do either of you have a better suggestion on how to handle that? @hardillb @knolleary

frontend/src/api/project.js
Pezmc: It just returns empty string https://github.com/flowforge/flowforge/blob/619565bcb9dcb9213cd217d12502fd0164c47a88/frontend/src/utils/daysSince.js#L3-L6

forge/containers/stub/index.js
Pezmc: Looks like a leftover from some changes I refactored. Thanks!
```suggestion
       projects.forEach(project => {
```

forge/containers/stub/index.js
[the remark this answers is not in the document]
hardillb: Can we just include it in the tests by having it post `[]` to the `/storage/flows` endpoint for the project with it's token as part of setup? I can help with the format if needed

joepavitt (commented):
**Open Editor Button:**
Would be nice to have it styled without stretching. `inline-flex` on the button seems to do the trick (as per screenshot attached), although would also be good to have it right-aligned in the cell.

Screenshot 2022-11-15 at 16 19 31

**Project Header**
I believe we spoke about this anyway, but would like the see the main project editor removed across the pages (except for "Overview". Then the "Project" text in the side bar, should instead say the project name (trimmed with ellipsis).

**Target Snapshot**
Rather than navigate to the "Snapshots" view, where it is currently then unintuitive to click the kebab menu for a snapshot to then set as device target. It would be nice to see a dialog popup, where the user can select from the available snapshots.

Pezmc:
Editor Button: On it
Project Header: Coming in a follow up
Target Snapshot: Being handled separately in https://github.com/flowforge/flowforge/issues/1175

Pezmc:
@joepavitt This PR has the editor button, the others I'll do in a follow-up.

joepavitt:
holding out from erge in case there are any outstanding server-side changes required. If not, then feel free to proceed @Pezmc

PR #1209, opened 2022-11-11
Add pagination to devices page

### LINKED ISSUE #1207: Devices list not showing all devices and pagination
### Current Behavior

I have added 50 devices but it is showing only 30 devices on devices page.
ff-devices-list

### Expected Behavior

Devices page should show all devices with proper pagination.

### Steps To Reproduce

Add 30+ devices

### Environment

- FlowForge version: v1.0.2
- Node.js version: v16.17.0
- npm version:
- Platform/OS: Windows 11 Docker
- Browser: Chrome

Pezmc:
Fixes #1207, ~but should not be merged before I update forge-ui-components~

### DIFF
--- a/frontend/src/pages/team/Devices/index.vue
+++ b/frontend/src/pages/team/Devices/index.vue
@@ -113,8 +113,10 @@ export default {
     watch: {
         team: 'fetchData'
     },
-    mounted () {
-        this.pollForData()
+    async mounted () {
+        await this.fetchData()
+        this.loading = false
+        this.checkInterval = setTimeout(this.pollForData, 10000)
     },
     unmounted () {
         clearInterval(this.checkInterval)
@@ -122,10 +124,7 @@ export default {
     methods: {
         async pollForData () {
             try {
-                if (this.team.id) {
-                    await this.fetchData(null, true) // to-do: For now, this only polls the first page...
-                    this.loading = false
-                }
+                await this.fetchData(null, true) // to-do: For now, this only polls the first page...
             } finally {
                 this.checkInterval = setTimeout(this.pollForData, 10000)
             }
@@ -227,7 +226,14 @@ export default {
             ]
         }
     },
-    props: ['team', 'teamMembership'],
+    props: {
+        team: {
+            required: true
+        },
+        teamMembership: {
+            required: true
+        }
+    },
     components: {
         TeamDeviceCreateDialog,
         DeviceCredentialsDialog,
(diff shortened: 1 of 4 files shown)


frontend/src/pages/team/Devices/index.vue
```diff
@@ -95,36 +102,52 @@ export default {
...
     methods: {
-        fetchData: async function (newVal) {
-            if (this.team.id) {
-                const data = await teamApi.getTeamDevices(this.team.id)
-                this.devices.length = 0
-                this.devices = data.devices
+        async pollForData () {
+            try {
+                if (this.team.id) {
+                    await this.fetchData(null, true) // to-do: For now, this only polls the first page...
```
Pezmc, line 126: I want to call this out explicitly: As implemented now, polling is only updating the first page of results. Given server side updates is in the roadmap, it didn't seem worth entering this rabbit hole for what should be a bug-fix PR. 

I'll spin up an issue to either re-think polling on this page, or remove this logic once server side events is in place. For now, it should be considered a known limitation of the devices page.

joepavitt:
If I sorted the devices, it only sorts the first page of context, rather than _all_ devices. I know this has come up before, but we would need to resolve this.

frontend/src/pages/team/Devices/index.vue
ZJvandeWeg: Why would we want updates from devices that are not on the rendered page?

While I understand we're accruing technical debt, it seems I'm missing something implicit in your comment why it's an issue?

frontend/src/pages/team/Devices/index.vue
```diff
@@ -38,7 +38,7 @@
                     </ff-button>
                 </div>
             </template>
-            <template v-if="devices.length === 0">
+            <template v-if="this.devices.size === 0">
```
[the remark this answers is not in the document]
Pezmc: There's no native wrapper, but there's no reason we couldn't add one. The change was needed as `this.devices` is no longer an array, instead it's a `Map`, and `Map` uses size rather than length to count the number of objects.

frontend/src/pages/team/Devices/index.vue
Pezmc: The devices table (and data-table) doesn't currently have pages or page numbers visually. The load more button simply loads the next page from the API and appends it to the bottom of the table. Perhaps the comment could be clearer.

joepavitt:
> We definitely need to take a step back and re-think how to implement a combination of search, pagination and filtering across the app, this is a pattern that's going to come up again and again and it'd great to get off on the right foot and keep it consistent.

Agreed entirely, will open new issue for that, and resolve this

PR #1219, opened 2022-11-15
docs: Add --port usage for the device agent

ZJvandeWeg:
A feature on the device agent was implemented to change the port the Node-RED instance binds too. It wasn't documented however, which this change adds.

### DIFF
--- a/docs/user/devices.md
+++ b/docs/user/devices.md
@@ -16,7 +16,9 @@ command is on the path:
 sudo npm install -g @flowforge/flowforge-device-agent
 ```
 
-### Configuration directory
+## Configuration
+
+### Execution directory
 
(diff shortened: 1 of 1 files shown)


docs/user/devices.md
```diff
@@ -29,6 +31,17 @@ sudo mkdir /opt/flowforge-device
 sudo chown -R $USER /opt/flowforge-device
 ```
 
+### Listen Port
+
+By default Node-RED will listen to port `1880`. The device agent has a flag to
+change this behaviour and listen on another port of choosing: `--port`. This can
```
hardillb, line 37:
```suggestion
change this behaviour and listen on another port of choosing: `-p/--port`. This can
```

To match the other examples

hardillb (approved):

PR #1237, opened 2022-11-17
Projects: Move the header to only the Overview Page

Pezmc:
Fixes https://github.com/flowforge/flowforge/issues/1235

### Simplified version on overview page
Screenshot 2022-11-17 at 14 46 56

### All other pages have no header
Screenshot 2022-11-17 at 14 47 03
Screenshot 2022-11-17 at 14 47 11

### DIFF
--- a/frontend/src/pages/project/Overview.vue
+++ b/frontend/src/pages/project/Overview.vue
@@ -1,5 +1,26 @@
 <template>
-    <div v-if="project.meta" class="space-y-4">
+    <div class="space-y-4">
+        <SectionTopMenu>
+            <template #hero>
+                <div class="h-full flex items-center">
+                    <div class="text-gray-800 text-xl font-bold">
+                        {{ project.name }}
+                    </div>
+                </div>
+            </template>
+            <template #tools>
+                <div class="space-x-2 flex">
+                    <!--  && !isVisitingAdmin -->
+                    <a v-if="editorAvailable" :href="project.url" target="_blank" class="ff-btn ff-btn--secondary" data-action="open-editor">
+                        Open Editor
+                        <span class="ff-btn--icon ff-btn--icon-right">
+                            <ExternalLinkIcon />
+                        </span>
+                    </a>
+                    <DropdownMenu v-if="hasPermission('project:change-status')" buttonClass="ff-btn ff-btn--primary" alt="Open actions menu" :options="options" data-action="open-actions">Actions</DropdownMenu>
+                </div>
+            </template>
+        </SectionTopMenu>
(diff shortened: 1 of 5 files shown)


knolleary:
What happens if the project name is longer than the space available in the sidebar?

This was one of the original rationales for having the common header on all project pages.

joepavitt:
> This was one of the original rationales for having the common header on all project pages.

I know it was, but many times I've found myself lost for what project I'm in and my go-to is that side bar navigation, even though it's right there in that header across the top - I just don't see it, so i wanted to try it out there instead. It'd have to be a very long project name to cut over, and would cut off with ellipsis (as things stand)

knolleary:
I've no issue moving it to the sidebar - just wanted to check what had been done to handle long names.

I don't know if its handled elsewhere (don't see a specific issue for it), but the Device pages use a similar pattern with the Device name in the page rather than sidebar.

Pezmc:
Screenshot 2022-11-18 at 09 52 23

frontend/src/pages/project/Overview.vue
```diff
@@ -62,43 +82,80 @@
...
-            return [
-                { name: 'Start', action: async () => { await projectApi.startProject(this.project.id) } },
-                { name: 'Restart', action: async () => { await projectApi.restartProject(this.project.id) } },
-                { name: 'Suspend', action: () => { this.$router.push({ path: `/project/${this.project.id}/settings/danger` }) } }
+            const flowActionsDisabled = !(this.project.meta && this.project.meta.state !== 'suspended')
+
+            const result = [
+                {
+                    name: 'Start',
+                    action: () => this.$emit('project-start'),
```
knolleary, line 133: I see this removes the explicit call to `projectApi.startProject(...)` in favour of emitting this event - but I don't see anything added to handle that event? Is that something that has already been added elsewhere?

Pezmc: The project-start event bubbles up to the project index view:
 https://github.com/flowforge/flowforge/blob/514cb4dc87ee238f0b1c479e0847cff58d098815/frontend/src/pages/project/index.vue#L51

https://github.com/flowforge/flowforge/blob/514cb4dc87ee238f0b1c479e0847cff58d098815/frontend/src/pages/project/index.vue#L201-L204

knolleary: @Pezmc see comment above

joepavitt:
All looks good, although I've realised I don't have `disabled` state on `list-item` - for some reason i've set it as the disabled class. As such, I'm making a new cut for forge-ui-components to include this (and include example in the docs)

joepavitt (approved):

PR #1241, opened 2022-11-18
Move auditLog apis to top-level component of the main app

knolleary:
This PR (targting the branch `origin/auditlog-encapsulation-1183`) modifies how the Audit Log helper functions are exposed in the app.

Given how widely they are used in the code base, we want them to be readily accessible. As such, it makes sense to decorate the top level `app` object with an `auditLog` component.

This means that rather than have to explicitly `require` the audit loggers (and have to call `getLoggers` to pass in an `app` instance), they can be called directly.

So where before you would do:
```
const { getTeamLogger } = require('../../lib/audit-logging/index.js')
...
const teamAuditLog = getTeamLogger(app)
...
teamAuditLog.team.user.invite.accepted(user, null, invitation.team, user, role)
```

You can now just do:

```
app.auditLog.Team.team.user.invite.accepted(user, null, invitation.team, user, role)
```

The loggers are exposed as:

```
app.auditLog.Team
app.auditLog.Platform
app.auditLog.Project
app.auditLog.User
app.auditLog.Device
```

Similarly, `app.auditLog.formatters` is exposed for where that is needed.

**Update** - I appreciate this loses some of the automatic intellisense magic that VSCode can do. From an architectural POV, this is the approach we use everywhere else, so I'd like to find a way forward with it.

---

This was motivated by trying to get the unit tests to work reliably for the loggers. The current approach was to unit test at a very fine grain the individual logging functions. The difficulty was this meant the tests had to stub out internal functions of the logger - and that is hard to do with the current approach.

I've started to rewrite the tests to consider the audit logger functions at a slightly coarser grain of unit. The basic unit of work is 'if I call this log function, do I get an entry in the AuditLog table with the right properties'.

This PR current reworks the Platform unit tests - the others haven't been touched yet so will likely be broken until we finished working through them.

Along the way I found the audit log was exposing raw database IDs - which must not happen. So have added son sanitisation logic to the view code.


knolleary:
@Steve-Mcl this is now updated - just have `test/unit/forge/auditLog/user_spec.js` for you to update to the new style.

Steve-Mcl:
@knolleary 

I have finished this & found a couple of issues causing (random test fails, incorrect signatures, missing await etc). 
Now working 100%

Additionally, for my own sanity I added a variable with type inference so I could quickly navigate between calls and implementations (and to see function signatures at a glance) - I hope you dont mind.

image

image

PR #1245, opened 2022-11-18
Allow snapshots to be created with api provided flows/credentials

knolleary:
This updates the `POST /api/v1/projects/:id/snapshots` endpoint to allow the request to include flows, credentials and list of modules - rather than pulling the current values from active project.

This allows a snapshot to be created outside of the platform and pushed in.

Notes:

 - if `credentials` are encrypted, the payload must include `credentialSecret` so they can be decrypted and reencrypted with the projects own key. Otherwise they must be posted unencrypted.
 - The snapshot will still include the current project settings, include env values. In other words, it isn't possible, with the work in this PR, to customise the env vars included in the snapshot. That may come in a future iteration.

### DIFF
--- a/test/unit/forge/db/controllers/ProjectSnapshot_spec.js
+++ b/test/unit/forge/db/controllers/ProjectSnapshot_spec.js
@@ -1,0 +1,113 @@
+const should = require('should') // eslint-disable-line
+const setup = require('../setup')
+const crypto = require('crypto')
+// const FF_UTIL = require('flowforge-test-utils')
+// const { Roles } = FF_UTIL.require('forge/lib/roles')
+
+function decryptCreds (key, cipher) {
+    let flows = cipher.$
+    const initVector = Buffer.from(flows.substring(0, 32), 'hex')
+    flows = flows.substring(32)
+    const decipher = crypto.createDecipheriv('aes-256-ctr', key, initVector)
+    const decrypted = decipher.update(flows, 'base64', 'utf8') + decipher.final('utf8')
+    return JSON.parse(decrypted)
+}
+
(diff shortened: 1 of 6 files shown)


hardillb (approved):

test/unit/forge/db/controllers/ProjectSnapshot_spec.js
```diff
@@ -0,0 +1,113 @@
...
+            snapshot.flows.should.have.only.keys('flows', 'credentials')
+            snapshot.flows.flows.should.have.length(1)
+            snapshot.flows.flows[0].should.have.property('id', '456')
+            snapshot.flows.credentials.should.have.only.keys('$')
+            const keyHash = crypto.createHash('sha256').update(credSecret).digest()
+            const creds = decryptCreds(keyHash, snapshot.flows.credentials)
+            creds.should.have.only.keys('456')
+        })
+    })
+    // describe('exportProject', function () {
```
ZJvandeWeg, line 108: @knolleary @hardillb Can we remove this?

test/unit/forge/db/controllers/Project_spec.js
```diff
@@ -203,6 +204,54 @@ describe('Project controller', function () {
...
+            const result = app.db.controllers.Project.exportCredentials(credentials, null, newKey)
+
+            result.should.only.have.keys('$')
+            ;(typeof result.$).should.equal('string')
+
+            const decrypted = decryptCreds(newHash, result)
+            decrypted.should.only.have.keys('foo')
+        })
+    })
     // describe('exportProject', function () {
```
ZJvandeWeg, line 255: Should this too be removed?

Issue #1194, opened 2022-11-04
Add project type/template/stack stage to Setup process


knolleary:
### Description

**As a:** user installing FlowForge for the first time

**I want to:** be helped to create my first Project Type, Template and Stack

**So that:** I have to read less documentation and get a running Node-RED project as quickly as possible.

---

Our current install process requires an Admin to create a Project Type, Template and Stack before any projects can be created. We don't provide much help in doing that and expect a new user to quickly learn what those things are and what needs doing.

The Setup wizard should provide help so much of this can be automated with a minimum of input from the user.

1. Project Type - we can create a default Project Type with a suitable name/description.
2. Template - we can create a default template using all default values
3. Stack - this gets a *little* tricker as it will rely on the user doing *something* to get the stack into place. The wizard should be able to provide accurate instructions for the chosen driver. For localfs, that should be the appropriate set of commands to run (with all the correct paths set) that can be copy/pasted into a terminal. For docker/k8s, something else is needed - tbd.

### Which customers would this be availble to

All Users, (CE)

### Acceptance Criteria

- [ ] Project Type/Stack/Template are setup as part of the first-run setup wizard
- [ ] Install documentation updated to reflect the change
- [ ] First-run doc screenshots updated

ZJvandeWeg:
> I want to: be helped to create my first Project Type, Template and Stack

> The Setup wizard should provide help so much of this can be automated with a minimum of input from the user.

@knolleary Was brainstorming with @joepavitt on this during our 1:1 and we figured: "Why put this in the wizard and not just install a default?" That seems to cut down the scope of this issue, and be more effective for the end user too. What do you think?

knolleary:
For ProjectType/Template, we can. ProjectStack is more complicated, because the user has to do *something* outside of the platform, or provide a piece of information (stack container id, or version number) that we don't know.

But the goal is to minimise the work as much as possible.

knolleary:
Notes from design call between @hardillb and me

 - Localfs
   - [ ] Update `install-stacks` script to be able to install `latest` and have it work out the version number to rename the stack dir properly (avoid conflict if already exists)
   - [ ] Remove question about installing stack from the install script and just do the install

 - [ ] Add `getDefaultStackProperties` api to each container driver (localfs/docker/k8s) that returns the appropriate default values for that driver.
   - [ ] Localfs can scan the stacks directory and pick up the latest stack if finds (should only be one)
[...]

PR #1249, opened 2022-11-19
Developer demo follow-ups (docs)

Steve-Mcl:
Commit 1 (docs/install/first-run.md)
* numbering issues corrected
* odd non printable "space like" character preventing header rendering
* clean up duplicate information about logging in
* adds link back to step 2

Commit 2 (docs/install/README.md)
* heading nesting levels corrected
* add missing windows commands
* mark optional parts with **(Optional)** (as an in-your-face visual cue)
* added **Windows** code block in **Running FlowForge** section

---

**UPDATE 21/11/2022**

### ~~Not done...~~

~~Ponting to a specific download (instead of asking user to determine x.y.z).  Steps 2 and 3 in `## Installing FlowForge` currently reads...~~
 - [x] removal of x.y.z done

### DIFF
--- a/docs/install/first-run.md
+++ b/docs/install/first-run.md
@@ -5,9 +5,13 @@ through the initial setup.
 
 By default, you can access the platform at [http://localhost:3000](http://localhost:3000).
 
+## 1. Begin
+Open FlowForge in your browser [http://localhost:3000](http://localhost:3000).  
+Click the **START SETUP** button
+
 <img src="images/setup-01.png" width=500 />
 
-## 1. Create an Administrator
+## 2. Create an Administrator
 
 The first user you create will be an Administrator. They will have full access
 to the platform, be able to set platform-wide configuration and manage users and teams.
@@ -19,7 +23,7 @@ outside of the platform in a future release.
 
 <img src="images/setup-02-user.png" width=500 />
 
-## 2. Upload a license
+## 3. Upload a license
 
 FlowForge Community Edition is Open Source and can be used freely without a license.
 
@@ -27,7 +31,7 @@ If you have a FlowForge Enterprise Edition license you can upload it here.
 
 <img src="images/setup-03-license.png" width=500 />
 
-## 3. Platform Options
+## 4. Platform Options
 
 There are some runtime configuration options that can be configured at this stage.
 
@@ -40,14 +44,14 @@ anonymous usage information from running instances.
 
 The Administrator can opt-out of this feature.
 
-## Login
+## 5. Done, lets login
 
-Once you complete the setup, you will be able to login as the Administrator user
-you created in the first step.
+Once you complete the setup, you will be able to login as the Administrator
+user that you created in [Step 2](#2-create-an-administrator).
 
-## Next Steps
+## 6. Next Steps
 
-Once logged in, there are three remaining tasks the Administrator must complete
+There are three remaining tasks the Administrator must complete
 before projects can be created
 
 ### 1. Create a Project Type
@@ -84,7 +88,7 @@ chosen deployment model.
  - [Docker Stacks](./docker/stacks.md)
  - Kubernetes Stacks - *coming soon*
 
-### 2. Create a Project Template
+### 3. Create a Project Template
 
 A Project Template provides a default set of Node-RED settings, such as the path
 the editor is served from and whether users can install new nodes.
--- a/docs/install/local/README.md
+++ b/docs/install/local/README.md
@@ -122,17 +122,6 @@ If you want to allow access from other devices on the network, you must edit the
 change the `host` setting to 0.0.0.0. NOTE: We do not support changing the `host` value once you have created a project.
 For more information on all of the options available, see the [configuration guide](../configuration.md).
 
-## Before running FlowForge (Windows Only)
-Some Windows applications like Hyper-V and WSL can reserve TCP ports. This can prevent FlowForge from running correctly. If necessary, adjust the starting port variable `driver.options.start_port` in the FlowForge configuration yaml file.
-
-You can see which ports are reserved on a Windows machine using the command...
-```console
-netsh interface ipv4 show excludedportrange protocol=tcp
-```
-
-IMPORTANT: When modifying `start_port` (default `12080`), be aware that FlowForge uses port numbers incrementally and in pairs separated by 1000. e.g. If `start_port` is set to `6000`, it will allocate ports `6000` and `7000` to the 1st project then ports `6001` and `7001` to the next project (and so on). 
-
-_INFO: Ports between 10000 ~ 49000 on windows are typically not included in the default reservation_
 
 ## Running FlowForge
 
--- a/docs/install/local/README.md
+++ b/docs/install/local/README.md
@@ -3,9 +3,9 @@
 This version of the FlowForge platform is intended for running on a single machine
 for a smaller deployment (e.g. evaluation or home user).
 
-### Prerequisites
+## Prerequisites
 
-#### Operating System
+### Operating System
 
 The install script has been tested against the following operating systems:
 
@@ -17,13 +17,11 @@ The install script has been tested against the following operating systems:
  - MacOS Big Sur & Monterey on Intel & Apple M processors
  - Windows 10 & 11
 
-[^1]: Arm6 devices, such as the original Raspberry Pi Zero and Zero W are not supported.
-
-#### Node.js
+### Node.js
 
 FlowForge requires ***Node.js v16***.
 
-##### Linux
+#### Linux
 
 The install script will check to see if it can find a suitable version of Node.js.
 If not, it will offer to install it for you.
@@ -31,27 +29,23 @@ If not, it will offer to install it for you.
 It will also ensure you have the appropriate build tools installed that are often
 needed by Node.js modules to build native components.
 
-##### Windows/MacOS
+#### Windows/MacOS
 
 If the install script cannot find a suitable version of Node.js, it will exit.
 
 You will need to manually install it before proceeding. Information about
-how to do this can be found on the Node.js website here:
-
-[https://nodejs.org/en/download](https://nodejs.org/en/download)
+how to do this can be found on the Node.js website here: [https://nodejs.org/en/download](https://nodejs.org/en/download)
 
 You will also need to install the appropriate build tools.
 
-On Windows, the standard Node.js installer will offer to do that for you.
+* **Windows**: the standard Node.js installer will offer to do that for you.
+* **MacOS**: you will need the `XCode Command Line Tools` to be installed. 
+  This can be done by running the following command:
+    ```
+    xcode-select --install
+    ```
 
-On MacOS, you will need the `XCode Command Line Tools` to be installed. This can
-be done by running the following command:
-
-```
-xcode-select --install
-```
-
-### Installing FlowForge
+## Installing FlowForge
 
 1. Create a directory to be the base of your FlowForge install. For example: `/opt/flowforge` or `c:\flowforge`
 
@@ -66,7 +60,7 @@ xcode-select --install
     mkdir c:\flowforge
     ```
 
-2. Download the lastest [Installer zip file](https://github.com/flowforge/installer/releases/latest).
+2. Download the latest [Installer zip file](https://github.com/flowforge/installer/releases/latest).
 
 
 3. Unzip the downloaded file into a temporary location and copy its contents to
@@ -81,6 +75,7 @@ xcode-select --install
 
    For Windows:
     ```
+    mkdir cd c:\temp
     cd c:\temp
     tar -xf flowforge-installer-x.y.z.zip
     xcopy /E /I flowforge-installer-x.y.z c:\flowforge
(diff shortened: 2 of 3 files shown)


docs/install/local/README.md
```diff
@@ -81,6 +75,7 @@ xcode-select --install
 
    For Windows:
     ```
+    mkdir cd c:\temp
```
knolleary, line 78: As discussed on the review, if `c:\temp` isn't the standard temporary file location on Windows, then there's no reason for these instructions to require its use.

I believe you said `cd %Temp%` would take you to a more appropriate temp location?

docs/install/local/README.md
```diff
@@ -109,13 +105,27 @@ If you select this option, it will ask if you want to run the service as the
 current user, or create a new `flowforge` user. If you choose to create the
 user, it will also change the ownership of the FlowForge directory to that user.
 
-### Configuring FlowForge
+## Configuring FlowForge (optional)
```
knolleary, line 108:
```suggestion
## Configuring FlowForge
```

I think this section contains important information (the location of the config file) that a user should be encouraged to read. I do not believe we should mark it optional.

I suggest changing the text after the file locations to say:

> The default configuration file already contains everything you need to get started with FlowForge.
> 
> To customise your installation, see the [configuration guide](../configuration.md).

However, we *do* have the issue with the `host` property - if the user leaves it as the default, and create a project, they will be stuck when they realise they want it to be more widely accessible as we don't support changing the `host` value after projects are made.

docs/install/local/README.md
```diff
@@ -109,13 +105,27 @@ If you select this option, it will ask if you want to run the service as the
...
 
-The default FlowForge configuration is provided in the file `/opt/flowforge/etc/flowforge.yml`.
+The default FlowForge configuration is provided in the file `flowforge.yml`
+* Linux/MacOS: `/opt/flowforge/etc/flowforge.yml`
+* Windows: `c:\flowforge\etc\flowforge.yml`
 
 For more details on the options available, see the [configuration guide](../configuration.md).
 
-### Running FlowForge
+## Before running FlowForge (Windows Only)
```
knolleary, line 116: If we're singling this out, then we should be consistent and provide this warning for all OS as there might be things using the ports we default to using.

Or we update the install script to make this check - otherwise this is adding a lot of complication to the install for a 'just in case' scenario.

Or we change the default `start_port` to be high enough to not be a concern on any OS and remove the need for any of this.

docs/install/local/README.md
```diff
@@ -248,20 +259,26 @@ you can use a pre-built docker image that provides everything needed.
...
+    * `npm install @flowforge/flowforge@x.y.z` (Windows)
  4. Check the release notes for any additional steps needed to upgrade the particular version
  5. Restart FlowForge `sudo service flowforge start` [^2]
  
 If you are running as your normal user you can drop the `sudo -u flowforge` and just run `npm install @flowforge/flowforge@x.y.z`
 
 ---
 
+
+[^1]: Arm6 devices, such as the original Raspberry Pi Zero and Zero W are not supported.
```
knolleary, line 282: Given these footnotes aren't rendered properly by our docs, we either need to add whatever markdown plugin is needed on the website, or stop using them and add these things as inline comments.

Steve-Mcl: Agree. I found the necessary Marked info for adding footnote support to Marked so what i did here was to group the footnotes at the footer in preparation.

I would like to keep this change and (eventually) the website will catch up.

docs/install/local/README.md
Steve-Mcl: In the review, we (I) said...

> Don’t have a temp directory on windows
> * Can use %temp% (decided to include mkdir c:\temp instead)
> * Can use a target location with `tar`, the next command
> * S: We should mkdir c:\temp
> * Steve to fix this ([PR 54](https://github.com/flowforge/installer/pull/54/))
> * ZJ: Extracting installer files to the right location takes more time than expected

This is a bit more nuanced than it first appears. Typically, a less experienced user would not know what `%temp%` is and if they managed to browse there, they would be fairly horrified at the crud they find in there (`%temp%` typically resolves to `c:\Users\<my-account>\AppData\Local\Temp` & on any given day contains a ton of weird crud) 

So it felt more aligned with Linux/MacOS to specify a fixed location like `c:\temp`

I can give this more thought if need be?

Also open to suggestions.

docs/install/local/README.md
Steve-Mcl: NP, will make the necessary adjustments shortly.

docs/install/local/README.md
Steve-Mcl: > we should be consistent and provide this warning for all OS

If I follow this route, do you know necessary Linux/MacOS command to view reserved ports?
Does Linux/MacOS have such a concept of ranges of reserved ports?

<br>

> Or we update the install script to make this check

That is possible but would require us asking (and explaining) to users how to run an administrative account.  
Also, would we need to do the same to the Linux/MacOS script?


<br>

> Or we change the default start_port to be high enough to not be a concern on any OS and remove the need for any of this.

Also possible, but there are a fair number of places where this is written verbatim (i think) - but certainly do-able


---

Please advise.

hardillb: The only "reserved" ports on Linux are those less than 1024 which normally only root can open, but there is no other concept of reserved ports.

docs/install/local/README.md
Steve-Mcl: Issue raised here: https://github.com/flowforge/website/issues/250

docs/install/local/README.md
knolleary: Notes from chat on this topic:

 - [ ] Update localfs to set start point to `12080` - @hardillb 
 - [ ] Update flowforge.yml to document 12080 - and config page in docs - @hardillb 
 - [ ] Update PR to remove the section about docs - @Steve-Mcl 
 - [ ] Update install readme remove the section on windows ports - @Steve-Mcl

Steve-Mcl: Info RE WSL + ports for future reference: `https://github.com/microsoft/WSL/issues/5514`

docs/install/local/README.md
Steve-Mcl: @knolleary @hardillb 
Regarding the `host` comment - I see 3 options...

1. Dont mention it (not ideal)

1. Mention it in line (in the README)...

> The default configuration file already contains everything you need to get started with FlowForge.
> 
> To customise your installation, see the [configuration guide](https://github.com/flowforge/flowforge/configuration.md) in particular, not setting the `host` value will mean that projects will not be accessible outside of the FlowForge server and we currently do not support changing the `host` value for existing projects.

1. Mention it in the [configuration guide](https://github.com/flowforge/flowforge/configuration.md).

> 
> Option | Description
> -------|------------
> `host` | The address to serve the web ui on. This defaults to `localhost` which means the ui will only be available when browsing from the same server that is running the platform. To make it accessible to other devices on the network, set it to `0.0.0.0`.  NOTE: We currently do not support changing the `host` value for existing projects.

--- 

please comment :)

docs/install/local/README.md
knolleary: > I found the necessary Marked info for adding footnote support to Marked

Is that info shared in the issue or are you planning to PR the necessary changes?

Steve-Mcl: No, not yet Nick. Will update the issue next.

I can also self assign the issue if you wish?

docs/install/local/README.md
hardillb: But if `host` is changed `base_url` also needs changing as this is used to find the oauth endpoints when signing into a project and defaults to `http://locahost:3000`

Steve-Mcl: > But if host is changed base_url also needs changing as this is used to find the oauth endpoints when signing into a project and defaults to http://locahost:3000

:doh:

Do you have a suggestion pls?

docs/install/first-run.md
```diff
@@ -5,9 +5,13 @@ through the initial setup.
...
 By default, you can access the platform at [http://localhost:3000](http://localhost:3000).
 
+## 1. Begin
+Open FlowForge in your browser [http://localhost:3000](http://localhost:3000).  
+Click the **START SETUP** button
+
 <img src="images/setup-01.png" width=500 />
 
-## 1. Create an Administrator
+## 2. Create an Administrator
```
ZJvandeWeg, line 14: We should just remove these numbers. That's much more stable

docs/install/first-run.md
```diff
@@ -40,14 +44,14 @@ anonymous usage information from running instances.
...
+user that you created in [Step 2](#2-create-an-administrator).
 
-## Next Steps
+## 6. Next Steps
 
-Once logged in, there are three remaining tasks the Administrator must complete
+There are three remaining tasks the Administrator must complete
 before projects can be created
 
 ### 1. Create a Project Type
```
ZJvandeWeg, line 57: @hardillb Does your work in adding these steps to the wizard remove these lines? Or move them out of `first-run.md` at least?

docs/install/first-run.md
```diff
@@ -84,7 +88,7 @@ chosen deployment model.
  - [Docker Stacks](./docker/stacks.md)
  - Kubernetes Stacks - *coming soon*
 
-### 2. Create a Project Template
+### 3. Create a Project Template
```
ZJvandeWeg, line 91: Not needed for a first run?

docs/install/local/README.md
hardillb: Add a note to the `host` entry to say that if changed to update `base_url` to `http://[ip-address-of-host]:3000`

docs/install/first-run.md
hardillb: Not just yet, there is an outstanding bit of work in #1194 to add these in the setup wizard.

docs/install/local/README.md
Steve-Mcl: ok, looks like this...
image

Gonna push that in 3... 2... 1...

Steve-Mcl: I believe this is now resolved/outdated. Will mark as resolved for sake of clarity & re-request reviews shortly after some housekeeping

docs/install/first-run.md
Steve-Mcl: @knolleary @hardillb 

I guess we (I) should remove this section now? in this PR?

docs/install/first-run.md
Steve-Mcl: There are references back to Step x

Will not be doing this now (but I am not opposed if you want to pick it up personally ;) )

docs/install/first-run.md
knolleary: I will be updating this section via #1250 to keep the changes together.

docs/install/first-run.md
knolleary: Will be resolved via #1250

Steve-Mcl:
As of writing: 

* commit [b0980b7](https://github.com/flowforge/flowforge/pull/1249/commits/b0980b7529553911795c2cf3089a24e5e0185e79) implements options 2 & 3 around `host` as suggest in Nicks comment.

* commit [4992fc0](https://github.com/flowforge/flowforge/pull/1249/commits/4992fc0623d4cec22539675ffb47111ae42a07c3) adds a comment about `base_url` as picked up by Ben

* commit [41c0a70](https://github.com/flowforge/flowforge/pull/1249/commits/41c0a70f48114c2f96a56040ffad113324a46816) removes the x.y.z from zip location and clarifies the whole tmp/temp path think better (I think - re-review tell)

Apart from @ZJvandeWeg comment about removing the "Create a Project Template" and "There are three remaining tasks" this should be pretty much ready now.

docs/install/local/README.md
knolleary: @Steve-Mcl if you know what needs doing, then please get it done.

knolleary:
@Steve-Mcl as mentioned, the 'create a project template' etc sections will get updated via #1250 once this one is merged and I can make the changes without unnecessary merge conflicts.

docs/install/local/README.md
knolleary: I'm still seeing the section on Windows ports/Hyper-V etc. The reason for changing our default port to 12080 was to remove the need for this section in the install guide. This type of detail can go in the configuration docs - it is not needed here.

docs/install/local/README.md
Steve-Mcl: NP

Steve-Mcl:
Perfect. 

I will re-request reviews.

### Remaining Tasks (async)
 - [ ] Sync installer README with final outcome of this PR - @Steve-Mcl
    - BLOCKED: Requires this PRs "Approval"
 - [ ] Support Footnotes in Marked docs
    - https://github.com/flowforge/website/issues/250 - @Steve-Mcl
    - In progress

knolleary:
@Steve-Mcl I added a comment for your attention; the whole section on windows ports needs removing now we've changed our default port to be out of danger of WSL etc.

docs/install/local/README.md
Steve-Mcl: weird - it is gone in local!?!?

Gonna re-check

Steve-Mcl: That was freaky, no amount of `git pull` synced up my local env!

Ended up using `.` web vscode! 

Resolved in [9465992](https://github.com/flowforge/flowforge/pull/1249/commits/9465992b4947c8b919d66eec970d281aaf2888ff)

PR #1252, opened 2022-11-21
Change localfs default port to 12080

hardillb:
Also update the config docs to show new port


Steve-Mcl (approved):
Double checked head - only 2 references to 7880 found 👍

PR #1254, opened 2022-11-21
Refactor test seeds config for E2E tests

Pezmc:
Extra coverage needs to be introduced for the deployments screen, which involves lots of additions to this test setup and it was becoming very unwieldy and hard to follow. This PR refactors the test setup to use the factory pattern. It also fixes an issue where team2 was using project template/types/stacks owned by Alice who is on team 1.

Follow-up PR which introduces the config for testing deployments coming shortly.

### DIFF
--- a/test/e2e/frontend/environments/standard.js
+++ b/test/e2e/frontend/environments/standard.js
@@ -21,12 +21,10 @@ class TestModelFactory {
             name: 'unnamed-team',
             TeamTypeId: defaultTeamType.id
         }
-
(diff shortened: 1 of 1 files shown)


knolleary:
Looks good. We should consider making the factory available at a higher level for other test areas that use this pattern. Not a priority for 1.1 though.

Pezmc:
Will look into test failures once other PR is open

Pezmc:
@knolleary I like that idea, a single factory shared across unit/integration tests.

PR #1255, opened 2022-11-21
Update token verification to use verifySession

knolleary:


knolleary:
The original PR was created before #1187 was merged that got rid of `verifyToken` in favour of `verifySession`.

hardillb (approved):

PR #1257, opened 2022-11-21
Project > Deployments: Assign snapshot

Pezmc:
Fixes https://github.com/flowforge/flowforge/issues/1175

Screenshot 2022-11-21 at 16 34 01

### DIFF
--- a/frontend/src/components/FormRow.vue
+++ b/frontend/src/components/FormRow.vue
@@ -21,7 +21,7 @@
         <!-- Single Line File Selection -->
         <template v-else-if="type==='file'">
             <label v-if="hasTitle" :for="inputId" class="text-sm font-medium text-gray-800"><slot></slot></label>
-            <label data-el="form-row-title" v-if="hasTitle" :for="inputId" class="text-sm font-medium text-gray-700"><slot></slot></label>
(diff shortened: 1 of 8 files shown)


Pezmc:
@joepavitt A couple of UX decisions I'd appreciate your input on:

Right now the dropdown always defaults to the placeholder so any selection is explicit: 
Screenshot 2022-11-21 at 16 41 49

Related, the current target snapshot is always shown in the list:

Screenshot 2022-11-21 at 16 41 57

This means a user could select the target snapshot that's always assigned, but having it not unexpectedly hidden from the list and requiring the user to always choose from the dropdown, rather than setting the default, felt best to me.

Pezmc:
@joepavitt I've added that, a loading state, and a basic no snapshots exist state to the dialog:

Screenshot 2022-11-22 at 12 18 24
Screenshot 2022-11-22 at 12 04 56

joepavitt:
@Pezmc what's the "Add" option here? Will that just go and take a snapshot of whatever we have live now?

Pezmc:
That was just the name of one of my snapshots in testing! That could be a nice to have to add in the future though. For now you need to navigate to the project snapshots page to add.

PR #1267, opened 2022-11-21
Add housekeeper component to run regularly scheduled tasks

### LINKED ISSUE #1181: Add housekeeping process to remove expired sessions/tokens periodically
### Description

We have some objects in the database that can expire - primarily Sessions and AccessTokens.

Currently we only delete an expired session/token if a request is made that tries to use it.

In many cases, a user will just close the window and let things expire naturally. This leaves the expired objects lying around in the database.

We should have a housekeeper process that runs periodically (once a day?) that seeks and destroys expired entities.

Architecturally it may be nice to have a common component in the app that any part can register a task that should be run to a schedule. But that is probably overkill for the immediate need to keep the access tokens and sessions tidy.

### Epic/Story

_No response_

knolleary:
Closes #1181

Need to figure out a test strategy for the individual tasks as they will get run automatically whenever running the app, which makes test cases a bit tricky.. ie they run before the test will get a chance to populate the database with anything the task may want to operate on.

But getting this raised so it can be reviewed as-is and merged once I get some tests added later.

### DIFF
--- a/forge/housekeeper/index.js
+++ b/forge/housekeeper/index.js
@@ -7,12 +7,14 @@ const { scheduleTask } = require('cronosjs')
  * Runs regular maintenance tasks to keep things clean and tidy
  *
  * A task is defined as an object with the properties:
+ * ```js
  * {
  *    name: 'name of the task',
  *    startup: true/false, // should the task be run whenever the platform starts
  *    schedule: '@daily', // cron string to define schedule. '@daily' means midnight
  *    run: async function(app) { } // the task itself
  * }
+ * ```
  */
 module.exports = fp(async function (app, _opts, next) {
     const tasks = {}
--- a/forge/housekeeper/index.js
+++ b/forge/housekeeper/index.js
@@ -1,0 +1,55 @@
+const fp = require('fastify-plugin')
+const { scheduleTask } = require('cronosjs')
+
+/**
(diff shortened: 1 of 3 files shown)


forge/housekeeper/index.js
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * Runs regular maintenance tasks to keep things clean and tidy
+ *
+ * A task is defined as an object with the properties:
+ * {
+ *    name: 'name of the task',
+ *    startup: true/false, // should the task be run whenever the platform starts
+ *    schedule: '@daily', // cron string to define schedule. '@daily' means midnight
+ *    run: async function(app) { } // the task itself
+ * }
```
Steve-Mcl, line 15:
```suggestion
 * ```js
 * {
 *    name: 'name of the task',
 *    startup: true/false, // should the task be run whenever the platform starts
 *    schedule: '@daily', // cron string to define schedule. '@daily' means midnight
 *    run: async function(app) { } // the task itself
 * }
 * ```
```

Nip picking.

Turns this...
image



into this...
image

Steve-Mcl (changes requested):
Very neat solotuion. 👍

Steve-Mcl (approved):

PR #1268, opened 2022-11-21
Doc: Copy project

ZJvandeWeg:


frontend/src/pages/project/Settings/Danger.vue
```diff
@@ -46,11 +46,11 @@
             </div>
         </div>
 
-        <FormHeading>Export Project</FormHeading>
+        <FormHeading>Copy Project</FormHeading>
         <!-- Hiding for now (0.5) -->
```
ZJvandeWeg, line 50: Not sure what this is, or what needs to happen with it?

hardillb: This is to move a project out of FlowForge (exports flows and creds). Given we now have projects not sure we need this at the moment.

frontend/src/pages/project/Settings/Danger.vue
```diff
@@ -46,11 +46,11 @@
...
         <!-- <div>
             <div>
-                <ff-button kind="secondary" @click="showExportProjectDialog()">Export Project</ff-button>
+                <ff-button kind="secondary" @click="showExportProjectDialog()">Copy Project</ff-button>
```
hardillb, line 53:
```suggestion
                <ff-button kind="secondary" @click="showExportProjectDialog()">Export Project</ff-button>
```

Reverting this as Export is right for this action if it was re-enabled.

frontend/src/pages/project/Settings/dialogs/ExportProjectDialog.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <ff-dialog ref="dialog" header="Export Project" confirm-label="Export Project" @confirm="confirm()">
+    <ff-dialog ref="dialog" header="Copy Project" confirm-label="Copy Project" @confirm="confirm()">
```
hardillb, line 2:
```suggestion
    <ff-dialog ref="dialog" header="Export Project" confirm-label="Export Project" @confirm="confirm()">
```

Again reverting as Export is the correct way to describe this (if it gets re-enabled). We should schedule a review of this feature for after Thursday.

hardillb (approved):

Issue #1200, opened 2022-11-10
Enable documentation updates without doing full releases


knolleary:
### Description

Currently the documentation on the website is built from the latest release of the `flowforge` repository. This makes it hard to make documentation updates without doing a new release.

We should be able to make multiple doc updates every day with as little friction as possible.

It isn't appropriate to pull the content from the `main` branch as that will include docs for unreleased work in the current milestone.

We don't want to pull the docs into its own repository as that will break the link between features and the docs getting merged together.

## Updated Proposal

 - All doc updates are made in the `main` branch as normal via PR
 - If a doc update is also intended to be deployed live, the PR should be labelled `docs-publish`
 - Change our maintenance process so we have a single branch to publish maintenance releases from - rather than create a new maintenance branch with each release. Call this branch `maintenance` (`maint` is a more succint option, but that's one character off `main` and could cause accidents)
 - Update the website automation to pull the docs from the `maintenance` branch
 - Update release automation of `flowforge/flowforge` to force-push `main` to `maintenance` when a new release is created from the `main` branch
 - A new automation is created that is triggered whenever a PR is merged with the `docs-publish` label.
      - It creates a new PR that cherry-picks the commits to the `maintenance` branch and assigns the PR for review by the author of the original PR. (Decided it would be safer to have a PR step, rather than direct cherry-picking to avoid merge-conflict headaches)

<details><summary>Old Proposal</summary>
 - Create a `live-docs` branch. This will be the branch the website always pulls from.
 - All doc updates are made in the `main` branch as normal via PR
 - If a doc update is also intended to be deployed live, the PR should be labelled `live-docs`
 - A new automation is created that is triggered whenever a PR is merged with that label.
      - It cherry-picks the commits to the `live-docs` branch and triggers a website rebuild.
      - (maybe) It cherry-picks the commits to the current maintenance release branch
 - The release automation script, when operating on the `main` branch (not maintenance release branch), force pushes `main` to `live-docs` and triggers website rebuild

If someone forgets to add the `live-docs` label to the docs PR, they will have to manually PR the update to the `live-docs` branch themselves and trigger a website rebuild.
</details>

knolleary:
I have updated the proposal above following discussions with Ben and ZJ.

TL;DR: change our maintenance process to use a single branch for maintenance releases, and have the website always pull from that branch. That will mean just merging a docs change to the branch will get it published to the website, without requiring a full maintenance release.

knolleary:
[...]

PR #1272, opened 2022-11-21
Add gh action to backport any pr labeled backport

knolleary:
Part of #1200 

This adds an action that is triggered whenever a PR is closed or labeled.

If the PR is confirmed to have been merged and has the `backport` label applied, the script will:

 - create a branch called `backport-<#pr>` based on the `maintenance` branch (which I've just created based on `1.x-maint`, our current maintenance branch that will get retired via #1200)
 - cherry-pick the commits from the PR to that branch
 - pushed the branch to origin and creates a PR

The PR title is a copy of the original PR title with (`backport <#pr>`) appended. This ensures the generated changelog is meaningful.

The body of the PR includes a link back to the original PR, plus the body of the original PR.

It *should* get triggered if an already merged PR is retrospectively labelled with `backport`.

Note I have tested the `backport.sh` script locally (see https://github.com/flowforge/flowforge/pull/1271 as an example). But it's hard to do the last mile integration of a GitHub Action without actually doing it. So this PR is that attempt to see if I've got the action right.

Given the action won't do anything until a PR is merged, we'll need to get this PR merged and then I can test be adding `backport` to a suitable candidate PR.

### DIFF
--- a/.github/scripts/backport.sh
+++ b/.github/scripts/backport.sh
@@ -1,0 +1,66 @@
+#!/bin/sh
+
+
+function usage() {
+   echo "Backport a PR to the `maintenance` branch, via a new PR"
+   echo
+   echo "USAGE"
+   echo "  backport <#PR>"
+   echo
+   echo "ARGS"
+   echo "  #PR: The PR number to backport"
+   echo
+   exit 1
+}
+
+if [ -z $1 ]
+then
+    usage
+fi
+
+CURRENT_BRANCH=`git branch --show-current`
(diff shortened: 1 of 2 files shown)


hardillb (approved):

knolleary:
Thanks @hardillb 

Now, having adding the `backport` label to https://github.com/flowforge/flowforge/pull/1266, I was expecting to see the action get triggered, but it has been a couple minutes and nothing is showing.

Will pick this up in the morning.

PR #1283, opened 2022-11-22
Move linux service start instructions

hardillb:
setting up the service is a question asked by the linux install.sh which is why it's not mentioned anywhere else in the docs

### DIFF
--- a/docs/install/local/README.md
+++ b/docs/install/local/README.md
@@ -132,16 +132,18 @@ To run it manually, you can use:
     /opt/flowforge/bin/flowforge.sh
     ```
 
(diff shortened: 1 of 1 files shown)


hardillb:
@Steve-Mcl can I merge this?

Steve-Mcl (changes requested):
Ben, I thought we were to change the wording as per your screenshot...

image

hardillb:
Sorry, thought I'd already pushed it, check now please

Steve-Mcl (approved):
Much betterer 👍

PR #1297, opened 2022-11-23
Update AWS K8s install instructions

hardillb:
This includes the missing files and removes a bunch of the notes about things that should have been finxed before it was published


docs/install/kubernetes/setup-rds.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash -x
```
ZJvandeWeg, line 1: <3 Thanks for the script

ZJvandeWeg (approved):

PR #1299, opened 2022-11-23
Allow email transport to retry connection without having to restart the platform

### LINKED ISSUE #1159: Email features do not pick up if email server is offline when FF launches
### Current Behavior

When starting up FF, if the mailbox is offline, but comes online soon/anytime after FF is started, FF mailbox features remain disabled until forge is restarted.

### Expected Behavior

For forge to recover automatically / gracefully

### Steps To Reproduce

* stop SMTP server (or make it appear offline)
* start FF
* start SMTP
* no amount of time or refreshes restore mailbox features in FF

### Environment

- FlowForge version: 1.0
- Node.js version: 16
- Platform/OS: Windows
- Browser: Chrome

Steve-Mcl:
closes #1159

In a nutshell, 

This PR only concerns itself with retrying the email transport initialisation IF it is enabled and configured but failed to connect.

It does not attempt to reconnect if a transport failure occurs later.

Currently, the retry is hard coded at 5min. We may wish to expose this in the config yml.

### DIFF
--- a/forge/postoffice/index.js
+++ b/forge/postoffice/index.js
@@ -8,30 +8,32 @@ module.exports = fp(async function (app, _opts, next) {
     let mailTransport
     let exportableSettings = {}
     let EMAIL_ENABLED = app.config.email.enabled
+    let poStartupCheck
     const isConfigured = EMAIL_ENABLED && (app.config.email.smtp || app.config.email.transport || app.config.email.ses)
     const mailDefaults = { from: app.config.email.from ? app.config.email.from : '"FlowForge Platform" <donotreply@flowforge.com>' }
-    if (EMAIL_ENABLED) {
-        let poStartupCheck
-        init(false, (err, enabled) => {
-            if (!err && enabled) {
+
+    // Startup initialisation
+    init(false, (err, connected) => {
+        if (!err && connected) {
+            clearInterval(poStartupCheck)
+        }
+    })
+
+    if (isConfigured) {
+        poStartupCheck = setInterval(() => {
+            const notReady = app.config.email.enabled === true && EMAIL_ENABLED === false
+            if (notReady) {
+                init(true, (err, connected) => {
+                    if (!err && connected) {
+                        clearInterval(poStartupCheck)
+                    }
+                })
+            } else {
                 clearInterval(poStartupCheck)
             }
-        })
-        if (isConfigured) {
-            poStartupCheck = setInterval(() => {
-                const notReady = app.config.email.enabled === true && EMAIL_ENABLED === false
-                if (notReady) {
-                    init(true, (err, success) => {
-                        if (!err && success) {
-                            clearInterval(poStartupCheck)
-                        }
-                    })
-                } else {
-                    clearInterval(poStartupCheck)
-                }
-            }, 1000 * 60 * 5) // check every 5 minutes until successful
-        }
+        }, 1000 * 60 * 5) // check every 5 minutes until successful
     }
+
     function init (retry, callback) {
         if (EMAIL_ENABLED || retry) {
             if (retry && mailTransport) {
--- a/forge/postoffice/index.js
+++ b/forge/postoffice/index.js
@@ -27,7 +27,7 @@ module.exports = fp(async function (app, _opts, next) {
                         }
                     })
                 } else {
-                    clearTimeout(poStartupCheck)
+                    clearInterval(poStartupCheck)
                 }
             }, 1000 * 60 * 5) // check every 5 minutes until successful
         }
--- a/forge/postoffice/index.js
+++ b/forge/postoffice/index.js
@@ -134,7 +134,7 @@ module.exports = fp(async function (app, _opts, next) {
                 })
             }
         }
-        if (isConfigured && app.config.email.debug) {
+        if (app.config.email.debug) {
             app.log.info(`
 -----------------------------------
 to: ${mail.to}
--- a/forge/postoffice/index.js
+++ b/forge/postoffice/index.js
@@ -7,68 +7,93 @@ const templates = {}
 module.exports = fp(async function (app, _opts, next) {
     let mailTransport
     let exportableSettings = {}
-
-    // mailTransport = nodemailer.createTransport(require("./localDelivery"));
-
-    // Use a local var as we may decide to disable email if the transport
-    // fails to verify
     let EMAIL_ENABLED = app.config.email.enabled
-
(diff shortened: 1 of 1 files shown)


hardillb:
Are you mixing `clearInterval()` with `clearTimeout()` here?

forge/postoffice/index.js
```diff
@@ -7,68 +7,93 @@ const templates = {}
...
-                    app.log.error('Failed to verify email connection: %s', err.toString())
-                    EMAIL_ENABLED = false
+        })
+        if (isConfigured) {
+            poStartupCheck = setInterval(() => {
+                const notReady = app.config.email.enabled === true && EMAIL_ENABLED === false
+                if (notReady) {
+                    init(true, (err, success) => {
+                        if (!err && success) {
+                            clearTimeout(poStartupCheck)
```
hardillb, line 26:
```suggestion
                            clearInerval(poStartupCheck)
```

forge/postoffice/index.js
```diff
@@ -7,68 +7,93 @@ const templates = {}
...
+                        }
+                    })
+                } else {
+                    clearTimeout(poStartupCheck)
```
hardillb, line 30:
```suggestion
                    clearInterval(poStartupCheck)
```

forge/postoffice/index.js
Steve-Mcl: oops

Steve-Mcl: funny thing - it worked in testing 🤔

forge/postoffice/index.js
Steve-Mcl: Thanks Ben

forge/postoffice/index.js
```diff
@@ -7,68 +7,93 @@ const templates = {}
...
-                    app.log.error('Failed to verify email connection: %s', err.toString())
-                    EMAIL_ENABLED = false
+        })
+        if (isConfigured) {
+            poStartupCheck = setInterval(() => {
+                const notReady = app.config.email.enabled === true && EMAIL_ENABLED === false
+                if (notReady) {
+                    init(true, (err, success) => {
+                        if (!err && success) {
+                            clearInerval(poStartupCheck)
```
hardillb, line 26:
```suggestion
                            clearInterval(poStartupCheck)
```

Steve-Mcl:
Ready for review. no more pushes (or typos)

Steve-Mcl:
Well that is an unexpected failure!

```
1) Project/Device API
       Update project device settings
         can update the project target snapshot:
     TypeError: Cannot read properties of null (reading 'should')
      at Context.<anonymous> (test/unit/forge/routes/api/projectDevices_spec.js:167:42)
```

Doesnt happen locally & cant see how it is connected to this?

hardillb:
@Steve-Mcl The test ran clean on repeat

knolleary:
Not intended to merge this for 1.1

hardillb:
Do we want to merge this now?

forge/postoffice/index.js
```diff
@@ -106,15 +133,15 @@ module.exports = fp(async function (app, _opts, next) {
...
-                app.log.info(`
-    -----------------------------------
-    to: ${mail.to}
-    subject: ${mail.subject}
-    ------
-    ${mail.text}
-    -----------------------------------`)
-            }
+        }
+        if (isConfigured && app.config.email.debug) {
```
knolleary, line 137: By adding `isConfigured` here, it is no longer possible to enable debug mode without also providing a valid smtp/transport/ses configuration. That is not the intended behaviour - if you're developing locally it should be possible to enable email in debug mode without requiring an external email config to match.

forge/postoffice/index.js
```diff
@@ -7,68 +7,95 @@ const templates = {}
...
+    const isConfigured = EMAIL_ENABLED && (app.config.email.smtp || app.config.email.transport || app.config.email.ses)
     const mailDefaults = { from: app.config.email.from ? app.config.email.from : '"FlowForge Platform" <donotreply@flowforge.com>' }
 
-    if (EMAIL_ENABLED) {
-        if (app.config.email.smtp) {
-            const smtpConfig = app.config.email.smtp
-
-            mailTransport = nodemailer.createTransport(smtpConfig, mailDefaults)
+    // Startup initialisation
+    init(false, (err, connected) => {
```
knolleary, line 16: In the case of `app.config.email.transport`, it looks like the callback is called synchronously to this call to `init`. That means `clearTimeout()` will get called *before* the `setInterval` is reached on line 23. So the interval will get setup even though init has been completed. Is that a problem? I'm not sure to be honest - it looks like 5 minutes later the interval will fire, find `EMAIL_ENABLED === true` and call `clearInterval` on itself to stop further checks.
Please have a second look to make sure I've understood that logic correctly.

knolleary (changes requested):

forge/postoffice/index.js
Steve-Mcl: I can address that, however, for info, I did this since prior to this modification if email transport failed, EMAIL_ENABLED would get toggled off and no debug was possible at all (settings were essentially mutually exclusively e.g. have "EMAIL_ENABLED and zero config" or "working config" )

I can certainly re-address this & make debug independent of configuration or SMTP status if thats preferrable?.

forge/postoffice/index.js
Steve-Mcl: > In the case of app.config.email.transport, it looks like the callback is called synchronously to this call to init

Yes, thats correct.

>  That means clearTimeout() will get called before the setInterval 

The clearInterval call does occur but has zero effect (no exception etc) & in this case, when the interval does occurs, if the transport is unhealthy, the init is called with retry flag - otherwise the clearInterval is called.

It effectively takes care of both the sync and async methods.


> Please have a second look to make sure I've understood that logic correctly.

You got it :)

forge/postoffice/index.js
Steve-Mcl, line 137:
```suggestion
        if (app.config.email.debug) {
```

As suggested

Steve-Mcl: Addressed in [e4334c1](https://github.com/flowforge/flowforge/pull/1299/commits/e4334c10484f5936659886f0bc393e4bd6ae89b5)

knolleary:
I think the only thing missing is a bit more logging to confirm when the connection has been verified. It is quite hard to test without positive feedback the connection has been retried.

PR #1300, opened 2022-11-23
Fix backport action to checkout full branch history

knolleary:
The backport action failed because it couldn't find the `maintenance` branch. This is because, by default, the checkout action only grabs the pr branch history.

By setting `fetch-depth` to `0`, it grabs the full git history so we can access the maintenance branch.

In addition, this updates the release-publish action to force push `main` to `maintenance` if the release is published on `main`.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
       - uses: actions/setup-node@v3
         with:
           node-version: 16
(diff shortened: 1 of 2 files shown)


hardillb (approved):

.github/workflows/release-publish.yml
```diff
@@ -17,4 +19,11 @@ jobs:
...
         with:
-          token: ${{ secrets.NPM_PUBLISH_TOKEN }}
\ No newline at end of file
+          token: ${{ secrets.NPM_PUBLISH_TOKEN }}
+      - name: Resync Maintenance:
+        if: ${{ github.ref == 'refs/heads/main' }}
+        run:
+          git checkout maintenance
+          git reset --hard main
+          git push --force origin maintenance
```
ZJvandeWeg, line 28: This works, but note `--force-with-lease` is a thing, and

ZJvandeWeg (approved):

PR #1310, opened 2022-11-24
Add specific version upgrade instructions

hardillb:
I've added instructions for each version.

They are a bit rough but they are a starting point


ZJvandeWeg:
Starting points are great. Thanks!

PR #1311, opened 2022-11-24
Remove ff:tooltip from Snapshot component

### LINKED ISSUE #1302: Deployments UI crashes after deploying a snapshot to devices
### Current Behavior

FlowForge's interface seems to crash once I use the button to deploy a snapshot in the new Deployments section.

https://d.pr/v/bwbuKS

### Expected Behavior

The interface should remain responsive to user interaction and data coming back from the devices.

### Steps To Reproduce

On staging use the Target Snapshot interface to deploy a snapshot to the devices. Once you've done this the UI should stop responding until you manually refresh your browser.

### Environment

- FlowForge version: Staging for 1.1
- Node.js version: Staging for 1.1
- npm version: Staging for 1.1
- Platform/OS: Mac M1 latest
- Browser: Chrome latest

knolleary:
Fixes #1302

Tracked down the cause of the hanging UI to be related to the new tooltip directive on the containing span.

Removing the tooltip to get it working.

### DIFF
--- a/frontend/src/pages/project/components/cells/Snapshot.vue
+++ b/frontend/src/pages/project/components/cells/Snapshot.vue
@@ -3,7 +3,6 @@
         <span
             v-if="activeSnapshot?.id || updateNeeded"
             class="flex items-center space-x-2 text-gray-500 italic"
-            v-ff-tooltip:left="'Target snapshot has not yet been deployed to this device.'"
         >
             <ExclamationIcon
                 v-if="updateNeeded"


Pezmc (approved):
@hardillb @knolleary And I dug into this, the root cause seems to be the tooltip manipulating innerHTML which seems to confuse vue when trying to swap the two icons when the snapshot state changes. That leads to Vue.js not being able to find the correct nodes and locking up the UI.

https://github.com/flowforge/forge-ui-components/blob/021fbad67445d61d58bfcbef795c2b8d52385181/src/directives/Tooltip.js#L3-L12

Issue on forge-ui-components: https://github.com/flowforge/forge-ui-components/issues/107

PR #1326, opened 2022-11-28
Improve User Settings email validation and error flow

sdirosa:
This PR solves https://github.com/flowforge/flowforge/issues/1281

**Changes**
- In the users api:
  - added `invalid_email` error code when email validation fails
  - added unit test to verify whether sending an invalid email returns the new error code 
- Improved a few things in the frontend for the settings view 
  - In the template, the email field's `type` attribute is now set to `email`, instead of `text`
  - The form will remain in `editing` mode when receiving errors from remote
  - The `invalid_email` error code will result in a message being shown under the email field

### DIFF
--- a/forge/routes/api/shared/users.js
+++ b/forge/routes/api/shared/users.js
@@ -155,9 +155,11 @@ module.exports = {
             } else {
                 responseMessage = err.toString()
             }
-            console.log(err.toString())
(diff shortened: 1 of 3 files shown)


knolleary:
Thanks @sdirosa

PR #1327, opened 2022-11-28
docs: Render docs of running instance under /docs

ZJvandeWeg:
While the repository always comes with docs, these aren't rendered. If FlowForge is either outdated or in an environment without internet access it's hard to read what's contained. With this change it's rendered under /docs as if it were `flowforge.com/docs`.

### DIFF
--- a/forge/routes/docs/index.js
+++ b/forge/routes/docs/index.js
@@ -1,0 +1,45 @@
+var MarkdownIt = require('markdown-it'), md = new MarkdownIt();
+const path = require('path')
+const { stat } = require('fs')
+const { readFile } = require('fs/promises')
+
+const docsBasePath = path.normalize(path.join(__dirname, "..", "..", "..", "docs"))
+const indexFile = "README.md"
+
+// User requests are unsafe and need to be validated so there's no path
+// traversal or related issues.
+function safeDocRequestPath(unsafeSuffix) {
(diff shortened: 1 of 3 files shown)


forge/routes/index.js
```diff
@@ -18,5 +18,7 @@ module.exports = fp(async function (app, opts, done) {
     await app.register(require('./setup'), { logLevel: app.config.logging.http })
     await app.register(require('./storage'), { prefix: '/storage', logLevel: app.config.logging.http })
     await app.register(require('./logging'), { prefix: '/logging', logLevel: app.config.logging.http })
+    await app.register(require('./docs'), { prefix: '/docs', logLevel: app.config.logging.http })
+    await app.register
```
Pezmc, line 22: ✂️

forge/routes/docs/index.js
```diff
@@ -0,0 +1,45 @@
...
+ * @namespace docs
+ * @memberof ???
+ */
+module.exports = async function (app) {
+	app.get('/*', async (request, response) => {
+		const cleanPath = safeDocRequestPath(request.params["*"])
+
+		// Assume rendering of the index is requested
+		let mdFile = mdFile = "${cleanPath}.md"
+		let mdFile = path.join(cleanPath, indexFile)
```
ZJvandeWeg, line 33: Only works for directories right now

PR #1348, opened 2022-12-01
Cypress 11

joepavitt:
Upgrade to Cypress 11 & fix config & a few, now failing, tests as the tests run in a different order (top level tests first, then nested tests in folders, used to be inverse in v9)


Pezmc (approved):
I've rebased onto main to get the fixes in asap, we can rebase audit-log-pr back onto head of pain once this is merged.

PR #1351, opened 2022-12-02
Show Team ID in Team > Settings > General

joepavitt:
- Also cleaned up an excessive object structure in there for `editing` state


Pezmc (approved):

Pezmc (changes requested):
Does data.editing not need to be updated?

joepavitt:
> Does data.editing not need to be updated?

right you are, although it does work without the change.

joepavitt:
had some other local changes that hadn't pushed too - woops. Thanks for flagging.

Pezmc (approved):
That looks better! 👯

PR #1356, opened 2022-12-03
Dns docs

hardillb:
First pass at some DNS docs for local development

### DIFF
--- a/docs/install/dns-setup.md
+++ b/docs/install/dns-setup.md
@@ -6,13 +6,7 @@ This is because FlowForge uses hostname based routing to know which project you 
 
 By default the project name is used in combination with a supplied domain. In this document I will use `example.com` as the domain. (It doesn't need to be a "whole" domain, it could also be a sub domain of one you already own. e.g. `ff.example.com`).
 
-Because DNS is all about mapping hostnames to IP addresses (and vice versa) I will be using the `192.168.0.22` IPv4 address to represent the host machine running Docker, or one of the nodes in the Kubernates cluster running the Ingress Controller.
-
 If you are running Docker/Kubernetes on the same machine as the DNS server and Web Browser do not use `127.0.0.1` as the IP address to point the wild card domain at. This is because the host names will be looked up by the Web Browser, the Forge Application, and the Project Node-RED instances. The last 2 are running in containers and `127.0.0.1` will resolve to the container, not the entry point. 
-
-For Docker on Linux you can use `172.17.0.1` which is the IP address assigned to the `docker0` interface.
-
-On MacOS you can alias a private IP address to the loop back interface e.g. `10.128.0.1` with `sudo ifconfig lo0 alias 10.128.0.1`
 
 ## Production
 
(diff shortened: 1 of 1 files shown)


docs/install/dns-setup.md
```diff
@@ -0,0 +1,157 @@
...
+
+This is because FlowForge uses hostname based routing to know which project you want to access.
+
+By default the project name is used in combination with a supplied domain. In this document I will use `example.com` as the domain. (It doesn't need to be a "whole" domain, it could also be a sub domain of one you already own. e.g. `ff.example.com`).
+
+If you are running Docker/Kubernetes on the same machine as the DNS server and Web Browser do not use `127.0.0.1` as the IP address to point the wild card domain at. This is because the host names will be looked up by the Web Browser, the Forge Application, and the Project Node-RED instances. The last 2 are running in containers and `127.0.0.1` will resolve to the container, not the entry point. 
+
+## Production
+
+For a production deployment you will need to talk to who ever owns your DNS infrastructure.
```
robmarcer, line 13: I'd assume most customers (with a domain) would CRUD their DNS via a web based control panel so will set up the DNS record themselves. My point is, I think we should just tell them what they need to set up rather than suggest they seek help.

docs/install/dns-setup.md
```diff
@@ -0,0 +1,157 @@
...
+sudo pihole restartdns
+```
+
+If running Pi Hole in Docker then you will need to create the file on the host and mount it to the `/etc/dnsmasq.d/02-flowforge.conf` location.
+
+### No Local DNS server
+
+If you really can't run dnsmasq then there is a possible alternative.
+
+You can set the domain FlowForge uses to the following pattern `172.17.0.1.sslip.io`
```
robmarcer, line 155: How does a user actually do this? Do they need to set up a DNS record?

docs/install/dns-setup.md
hardillb: We do in the next 2 sections, but we don't know if it's going to be a A record or a CNAME record until we know where/how they have deployed.

docs/install/dns-setup.md
hardillb: No this is just internet "magic". The next line explains

Try it out

```
dig 172.17.0.1.sslip.io


dig www.127-0-0-1.sslip.io
```

robmarcer: Thanks, what I mean to say is that it implies the user needs to do something but I think you are saying it just works. I think we should make it clear that they don't need to do anything if that's the case.

docs/install/dns-setup.md
robmarcer: Sorry if it's not clear, I'm saying that this sentence implies the reader can't set up the DNS records themselves but I'd expect in most cases they can. If I could remember how to suggest an edit I would say it should be changed to...

For a production deployment you will need to have access to your DNS control panel, if you are not sure how to set up DNS records talk to whoever manages your DNS.

hardillb:
closed flowforge/docker-compose#60

docs/install/dns-setup.md
hardillb: I'm saying you need to configure FlowFlorge to match, all the earlier examples use `example.com` which is what the shipped configuration files have as a default.

docs/install/dns-setup.md
```diff
@@ -0,0 +1,156 @@
...
+
+If running Pi Hole in Docker then you will need to create the file on the host and mount it to the `/etc/dnsmasq.d/02-flowforge.conf` location.
+
+### No Local DNS server
+
+If you really can't run dnsmasq then there is a possible alternative.
+
+You can set the `domain` entry in the `flowforge.yml` configuration file to the following pattern `172.17.0.1.sslip.io`
+
+This works because the `sslip.io` domain is set up to always return the IP address embedded in the hostname queried.
```
robmarcer, line 156:
```suggestion
### No Local DNS server

If you really can't run dnsmasq then there is a possible alternative.

A really useful service called sslip.io allows you to test FlowForge even when you cannot use dnsmasq. You can read more about [sslip.io on their web site](https://sslip.io).

You will need to set the `domain` entry in the `flowforge.yml` configuration file to the following pattern `172.17.0.1.sslip.io`. Don't forget to replace the IP address with the correct one for your configuration.

This will work because the `sslip.io` domain is set up to always return the IP address embedded in the hostname queried. 
```

PR #1361, opened 2022-12-05
API support for setting a projects hostname

Pezmc:
Part of https://github.com/flowforge/flowforge/issues/1344

This PR adds a project hostname setting, which is stored on the ProjectSettings table via the existing hasMany.

Due to some existing assumptions that a project has only one ProjectSettings, some re-jigging was required first. This is likely easier to review commit by commit as a result.

### DIFF
--- a/forge/db/controllers/Project.js
+++ b/forge/db/controllers/Project.js
@@ -1,7 +1,7 @@
 const crypto = require('crypto')
 const semver = require('semver')
 
-const { SETTINGS_KEY } = require('../models/ProjectSettings')
+const { KEY_SETTINGS } = require('../models/ProjectSettings')
 
 /**
  * inflightProjectState - when projects are transitioning between states, there
@@ -61,7 +61,7 @@ module.exports = {
                 })
             }
         }
-        const projectSettingsRow = project.ProjectSettings.find((projectSetting) => projectSetting.key === SETTINGS_KEY)
+        const projectSettingsRow = project.ProjectSettings.find((projectSetting) => projectSetting.key === KEY_SETTINGS)
         if (projectSettingsRow) {
             const projectSettings = projectSettingsRow.value
             result = app.db.controllers.ProjectTemplate.mergeSettings(result, projectSettings)
--- a/forge/db/controllers/Project.js
+++ b/forge/db/controllers/Project.js
@@ -1,6 +1,8 @@
 const crypto = require('crypto')
 const semver = require('semver')
 
+const { SETTINGS_KEY } = require('../models/ProjectSettings')
+
 /**
  * inflightProjectState - when projects are transitioning between states, there
  * is no need to store that in the database. But we do need to know it so the
@@ -59,8 +61,9 @@ module.exports = {
                 })
             }
         }
-        if (project.ProjectSettings[0]?.key === 'settings') {
-            const projectSettings = project.ProjectSettings[0].value
+        const projectSettingsRow = project.ProjectSettings.find((projectSetting) => projectSetting.key === SETTINGS_KEY)
+        if (projectSettingsRow) {
+            const projectSettings = projectSettingsRow.value
             result = app.db.controllers.ProjectTemplate.mergeSettings(result, projectSettings)
             const envVars = app.db.controllers.Project.insertPlatformSpecificEnvVars(project, result.env)
             // convert  [{name: 'a', value: '1'}, {name: 'b', value: '2'}]  >> to >>  { a: 1, b: 2 }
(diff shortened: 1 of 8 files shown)


forge/db/controllers/Project.js
```diff
@@ -59,8 +61,9 @@ module.exports = {
                 })
             }
         }
-        if (project.ProjectSettings[0]?.key === 'settings') {
-            const projectSettings = project.ProjectSettings[0].value
+        const projectSettingsRow = project.ProjectSettings.find((projectSetting) => projectSetting.key === KEY_SETTINGS)
```
Pezmc, line 64: I'd really prefer if these finder methods were somehow on the ProjectSettings model, but at runtime project.ProjectSettings` is a naive array object. The closest I could find would be adding a static method to the ProjectSettings prototype, but this breaks Sequelize conventions and feels like a hack.

forge/db/models/ProjectSettings.js
```diff
@@ -39,6 +44,18 @@ module.exports = {
...
+        return {
+            static: {
+                isHostnameUsed: async (hostname) => {
+                    const count = await this.count({
+                        where: { key: KEY_HOSTNAME, value: hostname.toLowerCase() }
+                    })
+                    return count !== 0
+                }
+            }
+        }
```
hardillb, line 57: This feels like it's missing something.

What happens if the user enters a hostname with our exisitng domain?

e.g. foo.flowforge.cloud

I think we need a test to see if the supplied domain ends with `app.config.domain` and also fail. This prevents them trying to be clever on the existing system to get round the list of banned names

https://github.com/flowforge/flowforge/blob/01c52ced147ee1c594f0ecc2828a2df0f490df87/forge/routes/api/project.js#L22-L35

forge/lib/validate.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+    isFQDN (string) {
+        const regex = /(?=^.{4,253}$)(^((?!-)[a-zA-Z0-9-]{0,62}[a-zA-Z0-9]\.)+[a-zA-Z]{2,63}\.?$)/
```
hardillb, line 3: I'm going to need this regex explaining

Pezmc: It boils down to:

- `(?=^.{4,253}\.?$)` total FQDN between 4 and 254 characters
-  `(` - start new group
- `^((?!-)` does not begin with a `-` (otherwise valid)
- `[a-zA-Z0-9-]{1,63}` Up to 1-63 alphanumeric characters plus `-`
- `(?<!-)` not ending in a `-`
- `\.` literal .
- `)+` repeat this group unlimited times 
- `[a-zA-Z]{2,63}` Up to 2-63 alphabetic letters for the TLD

Adapted from https://stackoverflow.com/questions/11809631/fully-qualified-domain-name-validation

forge/db/models/ProjectSettings.js
Pezmc: @hardillb Added in [dc880ee](https://github.com/flowforge/flowforge/pull/1361/commits/dc880ee90979ef990025e6b2c69a2181d1da6c73)

hardillb (approved):

test/unit/forge/routes/api/project_spec.js
```diff
@@ -1103,6 +1105,83 @@ describe('Project API', function () {
...
+
+                // call "Update a project" with a new hostname
+                const response = await app.inject({
+                    method: 'PUT',
+                    url: `/api/v1/projects/${TestObjects.project1.id}`,
+                    payload: {
+                        hostname: 'Already-In-Use.FlowForge.com'
+                    },
+                    cookies: { sid: TestObjects.tokens.alice }
+                })
```
hardillb, line 1163: `already-in-use.flowforge.dev` != `Already-In-Use.FlowForge.com`

hardillb:
```suggestion
                existingProject.updateSetting(KEY_HOSTNAME, 'already-in-use.flowforge.dev')

                // call "Update a project" with a new hostname
                const response = await app.inject({
                    method: 'PUT',
                    url: `/api/v1/projects/${TestObjects.project1.id}`,
                    payload: {
                        hostname: 'Already-In-Use.FlowForge.dev'
                    },
                    cookies: { sid: TestObjects.tokens.alice }
                })
```

`.com` vs `.dev`

Pezmc:
Rebase to squash the test fixes back into [bcf099e](https://github.com/flowforge/flowforge/pull/1361/commits/bcf099eb749e65f96c9fce99d6be17c5fc941ba5)

PR #1371, opened 2022-12-06
doc: License upgrade

ZJvandeWeg:


hardillb (approved):

hardillb:
do we want to move https://github.com/flowforge/flowforge/blob/main/docs/install/upgrading.md in to the new `upgrade` directory as well?

ZJvandeWeg:
@hardillb Update the path of the upgrade docs. Thanks!

hardillb (approved):

Issue #226, opened 2022-01-24
Support Single Sign On


knolleary:
### Description

We need to support external authentication systems for the core platform.

This will allow easier integration with existing systems, rather than having to create users manually.

Node-RED supports arbitrary SSO options via the use of pluggable PassportJS. We need to consider if that is the right approach, or do we want to be a bit more curated in what we support.

We have already had a query about Active Directory support.

SSO solves authentication, but we also need to consider how this relates to authorisation. For example, should Team membership be controllable via an external source of truth (LDAP etc).

---

#### Platform Configuration

 - An Administrator will be able to associate a particular email domain with an SSO SAML configuration
    - Need to scope whether this is done via Admin Settings in UI, or via yml config files.
       - For self-hosted, there is likely to only be a single SSO config that doesn't change - so can be done via yml
       - For FF Cloud, if done via yml, will require restart of the platform to pickup any changes... not ideal
 - **Future** 
    - An Administrator can disable user/password login entirely - requiring all users to login via SAML. Although need to consider the potential impact of locking Admins out if the SSO service is unavailable.

#### User login

 - Any user trying to login with a registered email domain will be redirected via SSO - even if they had a previously created an account with password
    - Should we allow Admins to still login via password as a safety net? If the SSO service goes down, we don't want them unable to login to fix things. Alternative here is a warning to ensure Admins use a different email domain that isn't covered by SSO... but that may not satisfy security requirements.
 - A SSO User will *not* be able to change their email address - as that is their primary identity

#### Adding Users

 - For self-hosted instances, an Admin will be able to manually add users to the platform, rather than auto-creating accounts when a user tries to login the first time.
 - For FF Cloud, given customers will need to contact us (as the admins) to setup the SSO configuration for them, we could also take on the task to create users from a list they provide. It's manual work, but keeps the scope of 1.2 manageable. This will be something to keep under review through 1.2 to see if we can do something better.

 - **Future**
    - If enabled for the SSO configuration, auto-create a user when they login for the first time via SSO.

----

### UX

Two options for how to expose SSO in the log-in dialog:

- **Option 1** - Add prompt to log-in dialog to use SSO instead (this is what Figma does)
  - Pro: for a mixed platform, doesn't add friction for non-sso users
  - Pro: its the pattern Figma uses, so not completely unreasonable
  - Con: Does a user know if they need to use SSO?

[...]

PR #1383, opened 2022-12-08
Add sso_enabled flag to user model

knolleary:
Part of #226 

 - This adds a new flag to the User model, `sso_enabled` that indicates this user has logged in via SSO.

 - If the flag is true for a given user, they cannot edit their email address.

image

 - Adds an sso_enabled column to the admin table of users

image

With this PR, nothing sets this property to true. That'll come in the next PR that introduces SSO login.

### DIFF
--- a/forge/db/views/User.js
+++ b/forge/db/views/User.js
@@ -46,6 +46,9 @@ module.exports = {
         if (user.defaultTeamId) {
             result.defaultTeam = app.db.models.Team.encodeHashid(user.defaultTeamId)
         }
+        if (app.config.features.enabled('sso')) {
+            result.sso_enabled = !!user.sso_enabled
+        }
         return result
     },
 
--- a/test/unit/forge/routes/api/user_spec.js
+++ b/test/unit/forge/routes/api/user_spec.js
@@ -22,10 +22,14 @@ describe('User API', async function () {
         }
     }
     beforeEach(async function () {
-        app = await setup({ features: { devices: true } })
+        const setupConfig = { features: { devices: true } }
+        if (this.currentTest.license) {
+            setupConfig.license = this.currentTest.license
+        }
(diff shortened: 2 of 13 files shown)


hardillb (approved):

forge/db/views/User.js
```diff
@@ -46,6 +46,9 @@ module.exports = {
         if (user.defaultTeamId) {
             result.defaultTeam = app.db.models.Team.encodeHashid(user.defaultTeamId)
         }
+        if (app.config.features.enabled('sso')) {
+            result.sso_enabled = !!user.sso_enabled
```
hardillb, line 50: Only sets the property if true. Test checks it's false

test/unit/forge/routes/api/user_spec.js
```diff
@@ -101,7 +105,36 @@ describe('User API', async function () {
...
+                method: 'GET',
+                url: '/api/v1/user',
+                cookies: { sid: TestObjects.tokens.alice }
+            })
+            response.statusCode.should.equal(200)
+            const result = response.json()
+            result.should.have.property('id', TestObjects.alice.hashid)
+            result.should.have.property('username', TestObjects.alice.username)
+            result.should.have.property('email', TestObjects.alice.email)
+            result.should.have.property('sso_enabled', false)
```
hardillb, line 123: This fails because the view only adds the SSO field if it's set to true

hardillb (changes requested):

test/unit/forge/routes/api/user_spec.js
knolleary: Odd - tests passes locally. Will investigate.

knolleary:
Have pushed a bit more to the branch to enable the missing bits.

hardillb (approved):

PR #1396, opened 2022-12-09
Downgrade Axios to 1.1.3

Pezmc:
Addresses* https://github.com/flowforge/flowforge/issues/1390 by downgrading Axios to a version without issues handling Brotli.

See: https://github.com/axios/axios/issues/5357 and https://github.com/axios/axios/issues/5346 for the issues with 1.2.0 and 1.2.1 respectively.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
         "@flowforge/localfs": "^1.1.0",
         "@headlessui/vue": "1.7.3",
         "@heroicons/vue": "1.0.6",
-        "axios": "^1.2.0",
+        "axios": "^1.1.3",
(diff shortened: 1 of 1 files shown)


Pezmc:
@hardillb / @knolleary Should we be cutting a new release of FF for this fix or will it go into 1.2?

knolleary:
We only use axios to access our own APIs... so I'm not sure if we're exposed here to warrant pushing a release out.

hardillb:
This is only an issue when running in nodejs, anything in the browser is safe by the look of things

hardillb (approved):

PR #1410, opened 2022-12-09
Add SSO login support

knolleary:
This adds support for SSO login based on a hardcoded list of SAML configurations.

That hardcoded list is empty in this PR, but the next PR will deal with how the SAML is provided by an Admin - either via YML, or a new SSO config screen in Admin Settings.

This code has been successfully tested against OneLogin and Google Workspaces as Identity Providers - by manually adding the appropriate provider configurations in the source code.

Whilst the list of providers is hardcoded, it is impractical to write test cases for this as it requires the ability to inject provider config and to have a functioning identity Provider to test against. So test coverage will come in the later PR that adds the ability to configure providers.


forge/ee/lib/sso/index.js
```diff
@@ -2,18 +2,92 @@ module.exports.init = async function (app) {
...
     app.config.features.register('sso', true, true)
 
+    /**
+     * A temporary hardcoded list of SAML configurations.
+     * This will eventually either come from .yml or via a db table
+     */
+    const PROVIDERS = {
+        // provider1: {
+        //     id: 'provider1',
+        //     domainFilter: '^.*@flowforge.com$',
```
hardillb, line 12: I know this is commented out, but the `.` in `flowforge.com` should probably be escaped because as it is it represents any char. (Just something to think about when we validate customer supplied regex)

forge/ee/lib/sso/index.js
```diff
@@ -2,18 +2,92 @@ module.exports.init = async function (app) {
...
+        }
+        if (user) {
+            if (await isSSOEnabledForEmail(user.email)) {
+                if (request.body.username.toLowerCase() !== user.email.toLowerCase() || request.body.password) {
+                // A SSO enabled user has tried to login with their username, or have provided a password.
+                // We need them to provide just their email address to avoid
+                // us exposing their email domain
+                    reply.code(403).send({ code: 'sso_required', error: 'Please login with your email address' })
+                } else {
+                    reply.code(403).send({ code: 'sso_required', redirect: `/ee/sso/login?u=${user.email}` })
```
hardillb, line 78: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403

>This status is similar to [401](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401), but for the 403 Forbidden status code, re-authenticating makes no difference. The access is tied to the application logic, such as insufficient rights to a resource.

I don't understand why 403 , "re-authenticating makes no difference" implies you shouldn't try and auth again?

forge/ee/lib/sso/index.js
knolleary: Good point to keep in mind. This whole block will go away in the next PR when we figure out the right way to provide this configurations.

forge/ee/lib/sso/index.js
knolleary: Good point. Do you think 401 would be more appropriate here?

hardillb: yes, 401 does feel better in this caes

PR #1417, opened 2022-12-12
Handle Stripe subscription cancellations

Pezmc:
Part of https://github.com/flowforge/flowforge/issues/1162

This PR:

 - Adds a new column, status, to subscriptions
 - Updates that status when FlowForge receives a Stripe callback for the subscription cancellations 
 - Blocks starting of projects and flows if the subscription is marked as cancelled
 - Updates project actions to always reset inflight state on request failure

### DIFF
--- a/forge/db/migrations/20221208-01-add-state-to-subscription.js
+++ b/forge/db/migrations/20221208-01-add-state-to-subscription.js
@@ -6,7 +6,12 @@ const { DataTypes } = require('sequelize')
 
 /** @type {import('sequelize-cli').Migration} */
 module.exports = {
-    async up (queryInterface, Sequelize) {
+    async up (queryInterface) {
+        const tableExists = await queryInterface.tableExists('Subscriptions')
+        if (!tableExists) {
+            return queryInterface.sequelize.log('Skipping Subscription migration as table does not exist')
+        }
+
         await queryInterface.addColumn('Subscriptions', 'status', {
             type: DataTypes.ENUM,
             values: ['active', 'canceled'], // Full list from stripe: trialing,active,incomplete,incomplete_expired,past_due,canceled,unpaid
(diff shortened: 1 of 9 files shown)


forge/db/migrations/20221208-01-add-state-to-subscription.js
```diff
@@ -0,0 +1,20 @@
...
+            type: DataTypes.ENUM,
+            values: ['active', 'canceled'], // Full list from stripe: trialing,active,incomplete,incomplete_expired,past_due,canceled,unpaid
+            defaultValue: 'active'
+        })
+    },
+
+    async down (queryInterface, Sequelize) {
+        await queryInterface.removeColumn('Subscriptions', 'status')
+    }
+}
```
hardillb, line 20: Should this be in `forge/ee/db/migrations` since it should only be applied if there is a license and billing is enabled

hardillb (changes requested):
First pass looks OK, apart from where the migration should be

forge/db/migrations/20221208-01-add-state-to-subscription.js
knolleary: We do not currently have any concept of ee-only migrations - because a license could get applied at any time to enable the EE components.

The issue here is the migration will get applied the first time FF runs and it will expect to find the Subscriptions table to exist - which would only be true if it had previously run with a license.

When billing was introduced, it looks like no migration was added to create the Subscriptions table - so it is left to Sequelize to do it on demand. Which means the table won't have been created by the time this migration runs on a clean system.

This migration needs to check to see if the Subscriptions table exists and only then attempt to add the column.

Pezmc: In [b65b0a6](https://github.com/flowforge/flowforge/pull/1417/commits/b65b0a6c02de65a59a2e2b0436686162e3b77686) I've added a guard locally and verified it works by deleting the subscription table and migration from the meta versions.

hardillb (approved):

PR #1419, opened 2022-12-13
@enter function correction on SSO username

joepavitt:
When pressing `enter` after typing out my username, I expect it to attempt the login, then show me the password field. Fairly sure I'd already fixed this as part of a review, but it appears the `@enter` call was reverted for some reason.

### DIFF
--- a/frontend/src/pages/Login.vue
+++ b/frontend/src/pages/Login.vue
@@ -5,7 +5,7 @@
             <template v-else>
                 <div>
                     <label>username / email</label>
-                    <ff-text-input ref="login-username" label="username" :error="errors.username" v-model="input.username" @enter="focusPassword"/>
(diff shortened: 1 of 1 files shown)


knolleary:
I think it got reverted by mistake during a rebase/merge conflict resolution.

However I think the logic needs to be slightly different as the form can be in one of two states:

1. just showing the username - hitting enter should submit
2. showing bother username and password fields - hitting enter in username should focus password

The current code only handles case 2 - your proposed change only handles case 1.

joepavitt:
very fair point - I can change it to a purpose built function instead, although I would expect "enter" to submit a form generally, and tab to iterate to the next field as it does currently.

knolleary (approved):
Whilst it may not be 100% to my taste, I can see that submitting the form when hitting enter is probably the more expected behaviour. Happy to go with this PR as-is given the current behaviour is just wrong.

PR #1426, opened 2022-12-14
Audit log accordions

joepavitt:
Closes #1341 

Accordions disabled in Project Overview, enabled elsewhere.

https://user-images.githubusercontent.com/99246719/207592656-a8589ce3-5d81-423c-a168-886b0cda108c.mov

PR #1428, opened 2022-12-14
Format Audit Entry "Error" messages 

### LINKED ISSUE #1352: Improved rendering of audit entry error
### Description

https://github.com/flowforge/flowforge/issues/1340#issuecomment-1335113987

As per @Steve-Mcl's suggestion, we should look at a cleaner design when rendering an audit entry with an attached error code. Steve's suggestion was to use a `detail` tag.

### Epic/Story

#428

joepavitt:
~Depends on #1426~ merged

Closes #1352 

https://user-images.githubusercontent.com/99246719/207638098-0f464914-bf4f-4915-9712-b9fff6274482.mov

### DIFF
--- a/frontend/src/components/Accordion.vue
+++ b/frontend/src/components/Accordion.vue
@@ -41,7 +41,7 @@ export default {
         contentHeight: function () {
             if (this.isOpen) {
                 const content = this.$refs.content
-                return content.scrollHeight + 'px'
+                return (2 * content.scrollHeight) + 'px'
             } else {
                 return null
             }
--- a/frontend/src/components/audit-log/AuditEntryVerbose.vue
+++ b/frontend/src/components/audit-log/AuditEntryVerbose.vue
@@ -338,11 +338,22 @@
     </template>
 
     <template v-if="error">
-        <span>{{ entry.body }}</span>
+        <details class="ff-audit-entry--error">
+            <summary>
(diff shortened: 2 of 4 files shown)


joepavitt:
not sure why this isn't updating to the latest changes in `main` - I haven't changed 9 files here

frontend/src/components/Accordion.vue
```diff
@@ -41,7 +41,7 @@ export default {
         contentHeight: function () {
             if (this.isOpen) {
                 const content = this.$refs.content
-                return content.scrollHeight + 'px'
+                return (2 * content.scrollHeight) + 'px'
```
Pezmc, line 44: Why is doubling the scrollHeight needed here? `contentHeight` is setting the maxHeight, right, which is used for the accordion animation, or is there something else it's handling I'm missing?

Pezmc (approved):
One open question, but good to go otherwise!

frontend/src/components/Accordion.vue
joepavitt: It's doubled due to the new error `<detail>` sections introduced - it was my dirty quick fix. When you expand the `<summary>`, the panel size doesn't update with it, if we set the max height to double, it gives us plenty of wiggle room to expand multiple error messages, without fear of exceeding. Alongside this, i also double the animation time, so that then behaves/animates at the speed we would have wanted prior to introducing the doubling.

PR #1431, opened 2022-12-14
Add SAMLProvider model for sso handling

knolleary:
Part of #226 

This PR introduces the `SAMLProvider` database model that is used to store SSO configurations, along with the admin api needed to manage them. The UI changes will follow in the next PR for ease of review.

 - `SAMLProvider` model added, including migration file
 - `forge/ee/lib/sso/index.js` updated to pull from the database rather than the hardcoded list of providers
 - Moved the SSO flow logic out of `forge/ee/routes/sso/index.js`  and into `forge/ee/routes/sso/auth.js` - this was needed to keep the additional middlewares/decorators used by SSO encapsulated to just the SSO routes and not the provider CRUD routes.
 - Added `/ee/sso/providers/*` routes for CRUD operations on the providers (in `forge/ee/routes/sso/index.js`)
 - Added tests for the CRUD routes and internal api handling of the providers.

Some additional notes:

 - `domainFilter` was previously treated as a regex. I have changed that to now require it to be a fully qualified domain and to start with an `@`. We compare the full domain of the user email against the list to find the provider.
 - I've add the ability to register new route permissions at runtime - so the EE permissions can be added dynamically rather than have them hardcoded into the CE code.


forge/ee/db/models/SAMLProvider.js
```diff
@@ -0,0 +1,96 @@
...
+                    value = '@' + value
+                }
+                this.setDataValue('domainFilter', value)
+            }
+        },
+        active: { type: DataTypes.BOOLEAN, defaultValue: true },
+        acsURL: { type: DataTypes.VIRTUAL, get () { return process.env.FLOWFORGE_BASE_URL + '/ee/sso/login/callback' } },
+        entityID: { type: DataTypes.VIRTUAL, get () { return process.env.FLOWFORGE_BASE_URL + '/ee/sso/entity/' + this.hashid } },
+        options: {
+            type: DataTypes.TEXT,
```
hardillb, line 23: There is a DataTypes.JSON that will do the get/set stuff for you, it's good on all the DBs we use (and are likely to use)

https://sequelize.org/docs/v6/other-topics/other-data-types/#json-sqlite-mysql-mariadb-oracle-and-postgresql-only

knolleary:
Pushed a minor, but important, update.

If a user with an sso-enabled email attempts to login with their username, we request they login with their email to trigger sso. However, that could leave an Admin user locked out if the SSO configuration is wrong, or the provider is unavailable.

So I've pushed a changed that allows an admin user to login with their username/password and bypass SSO. If they try to login with their email, it will go via SSO.

hardillb (approved):

PR #1432, opened 2022-12-14
Add Admin Settings for SSO Provider configurations

knolleary:
Part of #226 

This adds the frontend views for managing the SSO provider configurations.

This PR targets the `sso-config` branch that does all the backend work for provider configurations. I've split the frontend into its own PR for ease of review.

SSO SAML Provider configuration is hard. There are a small number of fields needed, but there is a lot of variation in how those fields are referred to be different identity providers.

<details>
This [doc from Okta](https://support.okta.com/help/s/article/okta-saml?language=en_US#:~:text=This%20is%20common,information%20they%27re%20providing) provides a good summary of the situation:

> This is common because there's unfortunately not much of a standard as far as SAML terminology is concerned. For example, some vendors may refer to the Assertion Consumer Service or Entity ID by a term unique to them. Because of this discrepancy in terminology, it can sometimes be a challenge to simply discover what information the vendor is asking for, or what to do with the information they're providing
</details>

I believe I've used the most consistent set of terms, but we'll need some good user docs - with worked examples for certain providers - to help the user. That PR will be coming tomorrow...

### DIFF
--- a/frontend/src/api/sso.js
+++ b/frontend/src/api/sso.js
@@ -1,0 +1,33 @@
+import client from './client'
+
+// Get list of active SSO providers
+const getProviders = async () => {
+    return client.get('/ee/sso/providers').then(res => {
+        return res.data
+    })
+}
+const getProvider = async (id) => {
+    return client.get(`/ee/sso/providers/${id}`).then(res => {
+        return res.data
+    })
+}
+const createProvider = async (options) => {
+    return client.post('/ee/sso/providers', options).then(res => {
+        return res.data
+    })
+}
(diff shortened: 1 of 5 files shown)


joepavitt:
Screenshot 2022-12-15 at 11 58 21

form validation doesn't require the indentity provider and public key fields - is this due to th chicken/egg scenario you mentioned @knolleary?

knolleary:
@joepavitt in part, also they are not all strictly required fields - it depends on what the Identity Provider needs you to set.

joepavitt (approved):

PR #1437, opened 2022-12-15
Increase default pagination limit

Pezmc:
Partially addresses https://github.com/flowforge/forge-ui-components/issues/101 and https://github.com/flowforge/flowforge/issues/1160

This is a shorter term fix and should be followed up by setting front-end side limits on any endpoints that we wish to paginate for **UX** reasons, that is to say, the server sets the default limit for technical/performance reasons, and the UI is responsible for setting the limit for any user experience or UI reason.

This effectively de-prioritises https://github.com/flowforge/flowforge/issues/1160 as significantly fewer users will be seeing pagination in the first place, but the long term solution is still to update our data tables to automatically fallback to server side search and sort when there is more than one page of results (https://github.com/flowforge/forge-ui-components/issues/101 and https://github.com/flowforge/flowforge/issues/1160).

### DIFF
--- a/docs/contribute/api-design.md
+++ b/docs/contribute/api-design.md
@@ -211,7 +211,7 @@ can be used to get the pagination options from the request, as well as specifyin
 default options:
 
 ```js
-const paginationOptions = app.getPaginationOptions(request, {limit: 30})
+const paginationOptions = app.getPaginationOptions(request, {limit: 1000})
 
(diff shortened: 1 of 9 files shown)


knolleary:
I'm not in a position to review this until next week. Please ask someone else in the team.

hardillb (approved):

Issue #1395, opened 2022-12-09
Update Docker/k8s docs to match local install header structure


joepavitt:
### Description

The Local install docs are easier to digest and navigate as they mostly use `<h2>` tags, whereas, currently the docker & k8s headers are `<h3>`. We should update them to make the documentations easier to navigate.

ZJvandeWeg:
@hardillb Any update on this issue? It's close to the release date, and it's still open.

PR #1438, opened 2022-12-15
Move File Storage section and fix header levels

### LINKED ISSUE #1400: Move "Using FF File Storage"
### Description

Currently, this section exists before the "First Run Setup", but cannot be actioned until after. As a result, we should move this section to after "First Run Setup" and make it a child section of this, or move it into `/docs/install/first-run/`

### Epic/Story

_No response_

hardillb:
Moves the File Storage set up to after First run.

Fixes #1400


hardillb:
Also part of #1395

joepavitt (approved):

PR #1442, opened 2022-12-15
Add a clear banner with a C2A when a teams subscription has expired

Pezmc:
Fixes https://github.com/flowforge/flowforge/issues/1422

The actual "click here to reactivate billing" work will be handled in the billing page improvements: https://github.com/flowforge/flowforge/issues/1421

This needs some basic E2E test coverage, I'll add as part of the billing page work in#1421 since right now there are no E2E billing tests.

### Admin
Screenshot 2022-12-15 at 16 02 57
Screenshot 2022-12-15 at 16 02 52

### Non-admin
Screenshot 2022-12-15 at 16 03 30

### DIFF
--- a/forge/ee/db/models/Subscription.js
+++ b/forge/ee/db/models/Subscription.js
@@ -39,8 +39,10 @@ module.exports = {
         const self = this
         return {
             instance: {
+                // Should this subscription be treated as active/usable
+                // Stripe states such as past_due and trialing are still active
                 isActive () {
-                    return this.status === STATUS.ACTIVE
+                    return !this.isCanceled()
                 },
                 isCanceled () {
                     return this.status === STATUS.CANCELED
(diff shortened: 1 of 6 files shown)


joepavitt:
The banner does show even before billing has been setup - not sure we want that?

joepavitt:
Even when filling out my billing details, I still get the warning that my subscription has expired:

Screenshot 2022-12-15 at 14 37 23

joepavitt:
Can we also left-align the text, make "Billing" bold and add a chevron far-right for when there is a clickable action to take (i.e. navigating to hte billing page) - I'd like the full notification clickable in this instance, rather than just the "Billing" text

Pezmc:
@joepavitt Ready for re-review

joepavitt (approved):
Happy with the UX & UI dev here. The project continued to run after changing the subscription to `canceled` in the DB when testing, it may be worth having a sanity check/API call that if we see the subscription canceled, then we automatically stop that project - that is out of the scope of this PR though, so happy to merge

PR #1451, opened 2022-12-16
docs: Explain how to use the Device agent Container

ZJvandeWeg:


docs/user/devices.md
```diff
@@ -16,6 +16,13 @@ command is on the path:
 sudo npm install -g @flowforge/flowforge-device-agent
 ```
 
+Or you can chose to run the Docker container. When you do, you'll need to mount
+the `device.yaml` obtained when [Registering the device](#register-the-device):
+
+```bash
+docker run --mount /path/to/device.yml:/opt/flowforge/device.yaml flowforge/device-agent:latest
```
hardillb, line 23:
```suggestion
docker run --mount /path/to/device.yml:/opt/flowforge/device.yaml -p 1880:1880 flowforge/device-agent:latest
```

PR #1477, opened 2022-12-21
Do not allow sso email domains to be registered with + emails

### LINKED ISSUE #1460: SSO with Plus Addresses
### Current Behavior

Create a brand new account using a plus addressed email (e.g. example+20221212@flowforge.com).

Assuming you're logged into the google account example@flowforge.com, you end up logged in as the `example@flowforge.com` account, not newly created example+20221212@flowforge.com account. 
This leaves you effectively locked out of the newly created account.

This isn't necessarily a true bug since as far as I'm aware google auth doesn't support plus addresses, but FF should prevent the user from ending up in this circumstance.

### Expected Behavior

Signed into the correct account
Or blocked at signup if SSO is enabled on a domain and trying to use a plus address

### Steps To Reproduce

Have SSO enabled for google workspace at a domain e.g. @flowforge.com
Create a brand new account using a plus addressed email (e.g. example+20221212@flowforge.com).
Verify the email address.
Try to log in with example+20221212@flowforge.com
Land in example@flowforge.com instead.

### Environment

- FlowForge version: v1.1.1
- Node.js version: v16.19.0
- npm version: n/a
- Platform/OS: Cloud
- Browser: chrome

knolleary:
Fixes #1460 

If a user attempts to create an account with a plus-address in their email (`foo+bar@example.com`) we reject the request with a message asking them to use the email their SSO provider identifies them as.

If a user does sign-up with a SSO-enabled email domain, we also now skip sending the email verification request - when they successfully login for the first time, we already set the `email_verified` flag on the account.

### DIFF
--- a/forge/routes/auth/index.js
+++ b/forge/routes/auth/index.js
@@ -308,6 +308,20 @@ module.exports = fp(async function (app, opts, done) {
             reply.code(400).send(resp)
             return
         }
+        let requireEmailVerification = true
+        if (app.config.features.enabled('sso') && request.body.email) {
+            if (app.sso.isSSOEnabledForEmail(request.body.email)) {
+                // This user is signing up with an SSO enabled email domain
+                // 1. validate they are not trying to use a plus-address (name+extra@domain)
+                if (/^.*\+.*@[^@]+$/.test(request.body.email)) {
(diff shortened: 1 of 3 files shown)


hardillb (approved):
While technically possible that the SSO provided email may include a `+` it is very unlikely, this should work.

PR #1482, opened 2022-12-22
Add upgrading docs for file server

knolleary:


knolleary:
Another updating coming imminently for the file serve docs....

knolleary:
Actually, lets get what's in this PR reviewed as-is, otherwise we'll hold up the release too long for a doc update we can push later.

hardillb (approved):

PR #1506, opened 2023-01-05
Allow download of snapshot package.json

hardillb:
## Description

This allows a user to download the package.json that represents the Node-RED version and all the installed nodes for a snapshot.

This can be used to create an offLine cache to be used with the Device Agent for instances where they can not connect out to npmjs.org

## Related Issue(s)

part of flowforge/flowforge-device-agent#45

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ frontend/dist-setup
 *.env.local
 package-lock.json
 coverage
+test/e2e/frontend/cypress/downloads
 test/e2e/frontend/cypress/videos
 test/e2e/frontend/cypress/screenshots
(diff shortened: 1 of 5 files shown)


hardillb:
@knolleary I'm not sure I like how I've done this, because it injects the modules into the response from `/api/v1/project/projectId/snapshot/snapshotId` which is normally only used for the UI

But it was the simplest way to do it without having to add a whole new endpoint

knolleary (changes requested):

knolleary (approved):

PR #1529, opened 2023-01-10
Add storage api routes for shared team library

knolleary:
## Description

Adds the storage api routes to support shared team libraries.

This is implemented as a custom [Library Store Plugin](https://nodered.org/docs/api/library/) that is setup by the FlowForge Launcher responsible for running Node-RED.

### API Updates

A pair of new API end-points are added alongside the other project storage apis - these are only registered if the platform is running with a license.

 - `POST /storage/:projectId/shared-library/:libraryId/:type`
 - `GET /storage/:projectId/shared-library/:libraryId/:type`

For the initial version, the `:libraryId` parameter must be the Project's Team Hashid, but in the future, could be used to support multiple team libraries (something that is out-of-scope for today).

### Database Updates

The library entries are stored in a new Database table - `StorageSharedLibrary`. This is very similar in structure to the existing `StorageLibrary` table used by the 'local' library in Node-RED. A separate table was needed as `StorageLibrary` uses the Project as a primary key - whereas we want the Team as the primary key for the Shared Library.

### Telemetry Updates

Adds two new metrics to the telemetry:

 - `platform.counts.libraryEntries` - number of entries in the 'local' library database table
 - `platform.counts.sharedLibraryEntries` - number of entries in the new shared library database table

Note that if running without an EE license, the latter metric will always be reported as 0.

## Related Issue(s)

Part of:
 - https://github.com/flowforge/flowforge/issues/237

Related PRs:
 - https://github.com/flowforge/flowforge-nr-storage/pull/27 - to add the Library Store plugin
 - https://github.com/flowforge/flowforge-nr-launcher/pull/93 - add the required library config to `settings.js`
 - TBD - `flowforge` to add the Storage API routes

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


knolleary:
Whilst implementing this, I found a bug with how the existing library endpoint handles multiple entries in the library. It would return duplicate listings.

I have fixed that via https://github.com/flowforge/flowforge/pull/1529/commits/e4dfdb9a6370dbf5d47cd17a811aa60723426dc1 and applied to both the existing local library routes and the new shared library routes.

I've added more extensive testing of the storage api to cover this scenario.

hardillb (approved):

Issue #1471, opened 2022-12-19
Provide several days trial


ZJvandeWeg:
### Description

For a customer of FlowForge Cloud it might be challenging to understand the value of our offering. When creating a team, the first thing they're greeted with is a checkout page. They're likely not aware what the value is, and might want to try before they buy.

For FlowForge Cloud there's an opportunity to allow a trial for a number of days. Stripe allows [`trial_period_days`](https://stripe.com/docs/api/subscriptions/create#create_subscription-trial_period_days) to be set. This change should allow anyone, even without coupon, to have a risk free onboarding experience.

The trial should work for:
* 1 small project
* ~once per team~ - For each users firs team

----

## Tracking

Summary of the comment thread below.

### MVP

- [x] Technical Implementation - https://github.com/flowforge/flowforge/pull/1544
  - [x] Add yaml config to configure a dollar amount to be applies to new subscriptions
  - [x] Apply a credit to the stripe customer when a user creates their first team (and if they are not a member of any team)
  - [x] Add current credit balance to team billing UI
  - [x] Add current credit balance to create project page UI
  - [x] Billing documentation updates on how to make use of this feature

### Follow ups
 - [x] https://github.com/flowforge/flowforge/issues/1552
 - [ ] https://github.com/flowforge/CloudProject/issues/115 (private repo)
 - [x] https://github.com/flowforge/flowforge/issues/1521
    - [x] #1548
        > This sets up your team for billing. You are only charged when creating a Project.
    - [x] #1549
        > As this is your first FlowForge team, we’re giving you free credit equivalent to a Small Project for one month’. This will be applied to your next Stripe invoice.
    - [x] #1550
    - [x] #1551
 - [x] https://github.com/flowforge/CloudProject/issues/118 (Private repo)
 - [x] https://github.com/flowforge/CloudProject/issues/119 (Private repo)
 - [ ] Marketing and website changes around free trials - was discussed in https://github.com/flowforge/flowforge/issues/1521
 - [x] Demo to team - Recording in Slack

Pezmc:
Ideally, I strongly believe that the customer should be able to start a free trial without having to go through the checkout flow in the first place, if we can mitigate any drawbacks.

[...]

PR #1538, opened 2023-01-11
Free Trials in the Form of Stripe Credit

Pezmc:
## Description

Work in progress PR  that adds free trials in the form of Stripe credit to new customers.

## Related Issue(s)

#1471 

## Checklist

    - [ ] Upgrade instructions
    - [x] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/ee/db/controllers/Subscription.js
+++ b/forge/ee/db/controllers/Subscription.js
@@ -19,5 +19,12 @@ module.exports = {
             subscription.destroy()
         }
         return null
+    },
+
+    userEligibleForFreeTrial: async function (app, user, newTeamAlreadyCreated = false) {
+        const teams = await app.db.models.Team.forUser(user)
+        const totalTeams = teams.length
+
+        return totalTeams <= newTeamAlreadyCreated ? 1 : 0
     }
 }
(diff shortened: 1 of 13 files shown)


forge/ee/db/controllers/Subscription.js
```diff
@@ -19,5 +19,12 @@ module.exports = {
             subscription.destroy()
         }
         return null
+    },
+
+    userEligibleForFreeTrial: async function (app, user, newTeamAlreadyCreated = false) {
+        const teams = await app.db.models.Team.forUser(user)
```
Pezmc, line 25: @knolleary @hardillb This feels like a sub-optimal way to check the number of teams a user is a member of, but I couldn't find nicer way to do the count given it's via the TeamMembership table, any ideas?

knolleary: I would add a `getTeamCount` instance finder on the `User` model, which in turn calls `TeamMember.countUserTeams` static finder on the `TeamMember` model.

knolleary: ... which now that I look at the changes made since this comment, I see you've done a slight variation of that which is just as good.

Issue #1232, opened 2022-11-16
Is possible to get a project status stuck in an intermediate state


Steve-Mcl:
### Current Behavior

Related to "Difficulty editing project settings when project state is intermediate" #1231 where the settings page was effectively unusable due to status refreshing...

It is possible to get a project stuck on an intermediate state e.g. not stopped/not suspended/not started.

##### Other notes...

* For the short time my environment was in this state (approx 10 mins) the `out.log` was littered with around 650 lines containing the following repeated pattern...
```
Loading project settings
Target state is 'running'
Starting Node-RED
/** snipped for brevity **/
Loading project settings
Target state is 'running'
Starting Node-RED
```

### Expected Behavior

Unsure - perhpas there should be a specific state like "Stalled"?

### Steps To Reproduce

1. Stop a project 
2. Edit the project settings and add a non-existing NPM package to the packages table
3. Save settings
4. Start the project 

### Environment

- FlowForge version: 1.0.3-git

Steve-Mcl:
@hardillb 

I would appreciate you testing the PRs for this issue in docker/k8s env if you can? 

#### Steps to test before this PR...
1. Stop a project
2. Edit the project settings and add a non-existing NPM package to the packages table
3. Save settings
4. Start the project
     1. Expect to see state stuck at "installing"


#### Steps to test after PR is applied...
1. Stop a project
2. Edit the project settings and add a non-existing NPM package to the packages table
3. Save settings
4. Start the project
     1. Expect to see state return to `stopped`
     2. Expect to see events updated in the "Recent Activity"
5. In the action menu, try to start/restart
     1. Expect to see state return to `stopped`
     2. Expect to see events updated in the "Recent Activity"

hardillb:
Works on K8s

hardillb:
Works on docker

hardillb:
@Steve-Mcl  Only problem is that the page doesn't stop polling when it enters Stopped state so you can't edit the settings again until you reload the page.

Steve-Mcl:
Tested in both staging an latest git ...

![Image](https://user-images.githubusercontent.com/44235289/213159409-9f0deef5-0ff7-46a7-af4f-8b33678399b1.png)

PR #1539, opened 2023-01-11
Improve stuck intermediate status of project when starting

Steve-Mcl:
## Description

Part of #1232

* Ensures known audit log entry "start-failed" to call the structured audit logger
* Permits body data in audit log entry made via API
* Introduces new icon for "loading" state (see other PR in launcher)

## Related Issue(s)

#1232 

## Checklist

    - [-] Upgrade instructions - NA
    - [-] Configuration details - NA
    - [-] Concepts - NA

## Labels

 - [-] Backport needed? -> add the `backport` label - NA
 - [-] Includes a DB migration? -> add the `area:migration` label - NA

### DIFF
--- a/forge/routes/logging/index.js
+++ b/forge/routes/logging/index.js
@@ -1,3 +1,5 @@
+const { getLoggers } = require('../../auditLog/project')
+
 /** Node-RED Audit Logging backend
  *
  * - /audit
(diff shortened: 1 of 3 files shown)


Steve-Mcl:
After struggling with the startup, correct order of audit logging and capturing the real state it was necessary to introduce a new "loading" state.   Typically this is never seen but if something stalls we will have a much clearer understanding of why.

That is because instead of blindly setting the start to "installing" we not have a "loading" state that differentiates it from **installing** and **starting** meaning we will know when it "stalls" if it due to installing or an issue loading settings.

Typical startups now go 
* stopped → loading → starting → started
* stopped → loading → installing → starting → started

chrome_31ZNz3hSRg

hardillb (approved):

PR #1555, opened 2023-01-13
Free Trials: Charges section on create project page

### LINKED ISSUE #1550: Charges Section on Create Project Page
image

Pezmc:
## Description

- Adds a charges section at the bottom of the create project page to clarify the new charges
- Specifically includes the current and monthly cost when confirming the charges*
- Takes into account and credit balance the customer may have on their account (from a free trial or prorated refund).

Screenshot 2023-01-13 at 15 09 40
Screenshot 2023-01-13 at 15 08 55

\* One caveat, the "now" price, isn't always 100% precise as it can't take into account pro-ration, but it's always larger than the amount actually charged (.e.g. if 50% of the way though the month, it'll still say the full amount e.g. $15 now, but stripe actually only charges $7.50)

## Related Issue(s)

Closes #1550
Part of #1471 

## Checklist

 - [x] Documentation has been updated - this does not have test coverage as there's currently no support for E2E tests with Billing set up
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/team/createProject.vue
+++ b/frontend/src/pages/team/createProject.vue
@@ -19,34 +19,38 @@
                 <div class="mb-8 text-sm text-gray-500">
                     <template v-if="!isCopyProject">Let's get your new Node-RED project setup in no time.</template>
                 </div>
-                <div v-if="subscription?.customer?.balance < 0" class="ff-banner ff-banner-info">
-                    You have a credit balance of {{ formatCurrency(Math.abs(subscription.customer.balance)) }} that will be applied to this project.
-                </div>
                 <div>
-                    <FormRow :error="errors.name" v-model="input.name">
-                        <template v-slot:default>Project Name</template>
-                        <template v-slot:append>
-                            <ff-button kind="secondary" @click="refreshName"><template v-slot:icon><RefreshIcon /></template></ff-button>
+                    <FormRow v-model="input.name" :error="errors.name">
+                        <template #default>Project Name</template>
+                        <template #description>
+                            Please note, currently, project names cannot be changed once a project is created
+                        </template>
+                        <template #append>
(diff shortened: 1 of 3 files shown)


joepavitt:
I'm also going to be a pain here and request E2E tests for the charges part. I know it's a pain, but this is a very important area we need to ensure is correct every time.

Pezmc:
>  I'm also going to be a pain here and request E2E tests for the charges part. I know it's a pain, but this is a very important area we need to ensure is correct every time.

I 100% agree, but as I noted in the issue description, we simply have zero scaffolding to support this at all. Our E2E tests run a web server without billing enabled, and there's currently no stubbing support for Stripe, but I'll re-investigate a way of doing it.

Pezmc:
> Just for my own understanding - why don't we charge for a full month from the point that the project is created?

If you set up billing today (i.e. team subscription), you're charged the full month when you create your project.

However, if you were to set up billing today, then wait half a month, you would only be charged until the end of the billing billing period i.e. 50% of the monthly cost, then next month is charged at full price again.

The [stripe prorations API](https://stripe.com/docs/billing/subscriptions/prorations) _can_ give estimates of how much will actually be charged to make this more precise in future.

knolleary:
> The [stripe prorations API](https://stripe.com/docs/billing/subscriptions/prorations#preview-proration) can give estimates of how much will actually be charged before a change is made to make this more precise in future, but that's definitely out of scope for this PR. Possible follow up @knolleary?

It depends where we choose to go with reworking the billing workflow. At some point, making lots of significant tweaks to something we plan to rework more fully becomes counter productive. So unless it brings significant customer value, compared to everything else, then I'm not sure it's worth looking at.

joepavitt (approved):

PR #1564, opened 2023-01-16
Guard against incomplete support/telemetry config

knolleary:
The change introduced by #1533 did not consider an existing configuration that has telemetry enabled, but does not have the newly added `support` config option. This leads to:

```
{
"statusCode": 500,
"error": "Internal Server Error",
"message": "Cannot read properties of undefined (reading 'enabled')"
}
```

when trying to access FlowForge.

This PR fixes it by properly guarding access to the new configuration options that may not exist.

### DIFF
--- a/forge/routes/ui/index.js
+++ b/forge/routes/ui/index.js
@@ -56,7 +56,7 @@ module.exports = async function (app) {
             </script>`
             }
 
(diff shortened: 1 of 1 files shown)


hardillb (approved):

knolleary:
Postgres test failed due to out of memory.

Given the nature of this change, moving ahead.

PR #1565, opened 2023-01-16
Limit Postgres to 3GB memory

hardillb:
## Description

A test to see if this fixes the OOM errors when testing against PostgreSQL

GH Action runners should have 7GB of RAM (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)

Limiting the PostgreSQL container to 3GB should leave enough to run the tests

## Related Issue(s)

The PostgresSQL tests now fail more often than they pass, always with a OOM error in NodeJS, this started before Christmas, but a re-run would normally pass.

If the PostgreSQL tests can run consistently then it somewhat defeats the point of running them.

## Checklist

 - [-] I have read the [contribution guidelines](https://github.com/flowforge/flowforge/blob/main/CONTRIBUTING.md)
    - [-] Upgrade instructions
    - [-] Configuration details
    - [-] Concepts

## Labels

### DIFF
--- a/.github/workflows/test-postgresql.yml
+++ b/.github/workflows/test-postgresql.yml
@@ -14,6 +14,7 @@ on:
       - 'test/unit/**'
       - 'test/system/**'
       - 'package.json'
+      - '.github/workflows/test-postgresql.yml'
 
 jobs:
   runner-job:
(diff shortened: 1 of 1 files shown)


knolleary:
Let's see how the test run goes on the PR before merging... 🤞🏻

knolleary:
... except this hasn't triggered the action because it doesn't touch the src code.

hardillb:
Unfortunately the only tests that will run are the link test, since we gate which tests run based on which files are changed.

Let me change the test to also run them when the tests themselves are changed

PR #1576, opened 2023-01-17
Team Library UI

joepavitt:
## Description

https://user-images.githubusercontent.com/99246719/212893069-2706ef88-f08e-428f-88f7-0f019c0e3aca.mov

View-only at the moment, basic navigation implemented.

## Related Issue(s)

https://github.com/flowforge/flowforge/issues/1368

## Checklist

    - [-] Upgrade instructions
    - [-] Configuration details
    - [-] Concepts

### DIFF
--- a/frontend/src/pages/team/Library.vue
+++ b/frontend/src/pages/team/Library.vue
@@ -1,40 +1,97 @@
 <template>
     <SectionTopMenu hero="Shared Library" help-header="FlowForge - Library" info="Shared repository to store common flows and nodes."></SectionTopMenu>
-    <ff-data-table :columns="columns"></ff-data-table>
-    {{ team }}
+    <div>
+        <span v-for="(crumb, $index) in breadcrumbs" :key="$index">{{ crumb.name }}</span>
+    </div>
+    <ff-data-table :columns="columns" :rows="rows" :rows-selectable="true" @row-selected="entrySelected"></ff-data-table>
 </template>
 
 <script>
+import { markRaw } from 'vue'
 import teamApi from '@/api/team'
 
 import SectionTopMenu from '@/components/SectionTopMenu'
+import TypeIcon from './components/LibraryEntryTypeIcon.vue'
 
 export default {
     name: 'SharedLibrary',
     props: ['team', 'teamMembership'],
     data () {
         return {
+            breadcrumbs: [{
+                name: 'Library'
+            }],
             columns: [{
+                key: 'type',
+                label: '',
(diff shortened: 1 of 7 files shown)


joepavitt:
```
<p>You can read more about <a href="https://nodered.org/docs/user-guide/editor/workspace/import-export" target="_blank">Import & Exporting Flows</a> and <a>Importing & Exporting Functions</a> in the Node-RED documentation</p>
```

Still looking for a link to the documentation of Node-RED that explains how to import/export functions in Node-RED.

joepavitt:
> Still looking for a link to the documentation of Node-RED that explains how to import/export functions in Node-RED.

Seems no such documentation exists, so will remove the reference i the dialog

joepavitt:
Not currently able to test the billable features in Cypress

knolleary:
Having played with it, the fact the 'files' get a mouse pointer when you hover over them, but they don't do anything on click feels off. Do we have a way to keep the normal cursor for non-clickable entries?

We need to capture some issues around future enhancements to this view. I don't think there is a lot needed, but one more iteration would help it feel a bit more complete.
 - Show a preview of entries on click. Something we pushed out of scope for this iteration.
    We have no way to show a rendered view of the flow, but providing access to the JSON would also be valuable to allow a user to copy an entry without having to open up a NR editor.
 - Add 'delete' option (will require backend api as well)
 - Add an 'updated-at' column - that info is not currently in the meta data returned by the api but can easily be added.

joepavitt:
> Having played with it, the fact the 'files' get a mouse pointer when you hover over them, but they don't do anything on click feels off. Do we have a way to keep the normal cursor for non-clickable entries?

That should be do-able

joepavitt:
> We need to capture some issues around future enhancements to this view. I don't think there is a lot needed, but one more iteration would help it feel a bit more complete.

Agreed, will record once this is finalised.

frontend/src/pages/team/Library.vue
```diff
@@ -0,0 +1,117 @@
...
+        })
+    },
+    methods: {
+        loadLibrary (parentDir) {
+            return teamApi.getTeamLibrary(this.team.id, parentDir).then((library) => {
+                return library
+            })
+        },
+        entrySelected (entry) {
+            console.log(entry)
```
knolleary, line 65:
```suggestion
```

knolleary (approved):
Removed one stray console.log, but otherwise, good for first iteration.

PR #1584, opened 2023-01-18
Audit Filtering UI

### LINKED ISSUE #1448: Add Filtering to Audit Log Screens
### Epic

#1166

### Description

**As a:** an admin user or team owner

**I want to:** be able to filter my Audit Log / Activity so that I can easily get a picture of when particular events took place 

**So that:** I can easily debug issues or backtrack when potentially breaking changes may have been made.

None

### Acceptance Criteria

**Completed in `feat-1468-audit-filtering`:**
- [x] Audit Log Component doesn't internally call API structure, but instead just accepts an Array of entires to render (so we can filter outside of the rendering component)
- [x] Implement a search bar that enables string querying over all of the available activity entries

**Todo:**:
- [ ] Pre-defined checkboxes for different event types & characteristics.

### Tasks

 - [x] #1489
 - [ ] Implement async API filtering/searching from the first task

joepavitt:
## Description

Adds a sidebar to each of the Audit Logs (Admin, Team & Project) that enables filtering by event type or user's name.

https://user-images.githubusercontent.com/99246719/213191763-7399c653-41bc-4738-800b-8368c91d969a.mov

E2E tests added to check filtering operates correctly too.

## Related Issue(s)

Closes #1448

## Checklist

    - [-] Upgrade instructions
    - [-] Configuration details
    - [-] Concepts

## Labels


frontend/src/components/Accordion.vue
```diff
@@ -7,7 +7,7 @@
                 <ChevronLeftIcon v-if="!disabled" class="ff-icon" />
             </div>
         </button>
-        <div ref="content" class="ff-accordion--content" :style="{'max-height': contentHeight}">
+        <div ref="content" class="ff-accordion--content">
```
joepavitt, line 10: this was giving warning in the console as we no longer use contentheight, so removed it here, as was flagging in the Audit Logs

frontend/src/components/audit-log/AuditEntryIcon.vue
```diff
@@ -37,7 +37,9 @@ const iconMap = {
     nodered: [
         'crashed',
         'stopped',
-        'settings.update'
+        'settings.update',
+        'flows.set',
+        'library.set'
```
joepavitt, line 42: Spotted a couple of new Node-RED events I hadn't seen before, so have added their verbose definition and icon accordingly.

frontend/src/components/audit-log/AuditLog.vue
```diff
@@ -25,46 +25,11 @@
...
-    'project.started': 'play',
-    'project.stopped': 'stop',
-    'project.created': 'create',
-    'project.duplicated': 'create',
-    'project.deleted': 'create',
-    'auth.login': 'user',
-    'auth.login.revoke': 'logout',
-    'flows.set': 'pencil'
-}
-
```
joepavitt, line 56: This should have all been removed with the last Audit Log work as we now use `AuditLogVerbose` to get descriptions

joepavitt:
There is some duplication with the introduction of filtering across each of the pages, and we may want to component-ise some of this at a later date, but due to the layout, and differences in the API call to get the users (at Admin level, we want _all_ users, at team/project, just members) it got more complicated than it needed to, so kept them separate for now.

knolleary:
A problem we have is the text shown on the page is set client side, whereas the search is done based on the database table contents - and they are different.

To find entries relating to a user logging in, the page shows entries that say:

> User Logged In
> User 'nick' has logged in.

But if I search for 'logged in' it doesn't find anything - I have to know to search for the event type which is 'account.login'.

A quick-ish solution would be to have a drop-down of the known event types as that is going to typically be the thing someone wants to look for. (maps to `event=<event-name>` in the query string).

joepavitt:
> A quick-ish solution would be to have a drop-down of the known event types as that is going to typically be the thing someone wants to look for. 

Yep, I'll sort that now

knolleary:
Looks good - will await the UI test run to confirm it passes, then merge and get onto staging.

PR #1588, opened 2023-01-18
docs: Add 1.3 update notes

knolleary:
## Description

Adds notes on upgrading to 1.3 - in particular that Team Library and FF Auth requires a project stack update.

@hardillb we need to review our stack creation docs as well - I think the k8s docs can be simplified as we now publish suitable container images ourselves.


hardillb:
I'll have a look as a separate PR

hardillb (approved):

Issue #1581, opened 2023-01-18
"Create Account" text updates


joepavitt:
### Description

It has been requested that we provide updates to the Create account page: https://github.com/flowforge/website/issues/333#issue-1535111693 that details the availability of a Free Trial.

Complexity here is that we only want this message showing on FF Cloud, and not on local install of FlowForge.

knolleary:
Proposal: Add `billing.createAccountBanner` to `flowforge.yml` that, if set, is displayed as a banner on the create account page.

We can then set that to whatever we want it to say. Could be plain text, or HTML if we want richer customisation.

Longer term, it might make sense to move out of `billing` and have a separate file of UI customisations that can be applied to individual instances... but this would be a quick first pass.

hardillb:
Wouldn't it be hung off the `/api/v1/settings` endpoint? from `forge/routes/api/settings.js`?

joepavitt:
<img width="811" alt="Screenshot 2023-01-18 at 17 13 27" src="https://user-images.githubusercontent.com/99246719/213249123-be11e6ad-06df-4f6e-a216-cf1cdbf8acf2.png">

joepavitt:
Agreed - can you create an issue for that on this repo please?

PR #1591, opened 2023-01-18
Add configurable banner on Create Account page (rework)

knolleary:
## Description

This is an alternative approach (compared with #1589)  for how we can inject custom text into the Create Account page.

This takes a slightly broader approach and considers that we may want to introduce other bits of customisable text in the future.

Whilst the immediate need is related to billing, and my original suggestion was to put this under the billing config, in hindsight there is nothing inherently related to billing in wanting to add text to the create account page.

This proposes we add a new `branding` section in `flowforge.yml` under which we can organise any bits of customisable text.

For #1581, this PR adds support for the following:

```yaml
branding:
  account:
    signUpTopBanner: Free donuts for all
```

The entire `branding` object is passed to the UI via the `/settings` object. It currently only includes it in the unauthenticated request to `/settings` (as we only need it for the create user page), but a future iteration will want to make that smarter. We may even want to move it to a separate api end point to load these customisations... but that's for another day.

The choice of `signUpTopBanner` was based on the discussion in https://github.com/flowforge/website/issues/339 where I can see us wanting `signUpSideBanner` fairly soon...

## Related Issue(s)

 - #1589  

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -1,6 +1,7 @@
 module.exports.init = async function (app) {
     // Set the billing feature flag
     app.config.features.register('billing', true, true)
+    app.config.features.register('createAccountBanner', app.config.billing.createAccountBanner, true)
 
     const stripe = require('stripe')(app.config.billing.stripe.key)
 
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -1,7 +1,6 @@
 module.exports.init = async function (app) {
     // Set the billing feature flag
     app.config.features.register('billing', true, true)
-    app.config.features.register('createAccountBanner', app.config.billing.createAccountBanner, true)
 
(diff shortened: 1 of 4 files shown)


joepavitt:
flagging for @hardillb as this'll require helm chart changes

joepavitt:
Is the nesting inside `account` necessary? Feels like over-engineering a tad for our use case atm? `branding.signUpTopBanner` is still clear enough as to it's target content?

joepavitt:
Tests failing atm - fix pending in https://github.com/flowforge/flowforge/pull/1592

knolleary:
> Is the nesting inside account necessary? Feels like over-engineering a tad for our use case atm?

I was trying to apply a little forethought on were this would go and we may appreciate a little organisation from the start, rather than have to move things around later.

PR #1592, opened 2023-01-19
Fix front end tests

knolleary:
Noticed a recent PR broke one of the frontend unit tests (`npm run test:unit:frontend`). The PR had only modified frontend files and the current GitHub Action setup meant those particular tests were not run against it automatically.

This PR fixes the test failure and also ensures any front-end change triggers the front-end unit tests and not just the e2e tests.

As it stands, this does mean we'll run those particular tests multiple times - but we should look at removing them from main build/test actions and leave them to the UI test actions.

### DIFF
--- a/.github/workflows/test-ui-e2e.yml
+++ b/.github/workflows/test-ui-e2e.yml
@@ -1,4 +1,4 @@
-name: UI E2E (Cypress) Tests
+name: UI Tests
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


joepavitt:
Can we change the workflow name to be "UI Tests", rather than Cypress/E2E specific? Then the two jobs can be:

- UI Unit Tests
- UI E2E (Cypress Tests)

joepavitt (approved):

Issue #595, opened 2022-05-17
Change project type


sammachin:
### Description

**As a:** Owner 

**I want to:** Change my project type

**So that:** So that I can use a stack with additional resources or features

2 paths to think about here, upgrading and downgrading, upgrade should be the simpler path but downgrading a project type may be that the project is using features which are not available in the lower tier.
For MVP only having upgrade would be a viable iteration.

A user can move a project to a different tier today by  copying the project to a new one on a different type but this means changing the name and URL of the project when may no be desirable.

sammachin:
@nick is this implemented in 0.7?

knolleary:
No - as described in #739, allowing a user to change the projecttype is not in scope of this initial work.

Steve-Mcl:
@knolleary.  @hardillb. 

Need a direction on this feature please.

I see this going 2 ways...

1. Upon selecting a different project type, the user is offered the choice of stacks
2. Upon selecting a different project type, the default stack is assumed (or is inhibited if no default set)

`1.` seems the more friendly way
`2.` presents us with an opportunity to encourage users to setup the defaults.

<br>

Regarding Sams comment about "only having upgrade would be a viable iteration" is this a necessity? And if so, what determines a smaller/larger stack?  Perhaps MVP is to permit either direction and the consequences are the users responsibility? (perhaps offered with a suitable warning)

ZJvandeWeg:
@Steve-Mcl is this going to ship? It would be great if everyone knew what 1.2 would include

Steve-Mcl:
I dont believe it will make the cut ZJ. 

I have started a branch but there is more than meets the eye.

Steve-Mcl:
Following discussion with Nick and Joe, this item is UI heavy and needs to be done right with consideration to the duplication of code and UI around the whole "Create Project, user billing changes etc). 
Instead of trying to rush this into 1.3 it is to be deferred once more. I have setup a meeting with Joe for next Friday (following launch retro) to get this right from the start.

robmarcer:
I've moved this to 1.4 as per Steve's post above, hope that's OK.

joepavitt:
Just spoke with Steve. Given @Pezmc's recent work on the Create Project page, and Steve's expertise not being FE, I think this would best suit Pez to conduct the front-end work required for this.

Steve-Mcl:
Following discussion with @joepavitt the approach should be as follows...

1. Work on frontend and backend should be split to suitable devs
[...]

PR #1610, opened 2023-01-25
Change Project Type Form

Pezmc:
## Description

Implements changing of project types as a "danger" action (this section likely to be renamed in future).

When a user changes project type, the project is suspended, reconfigured, then resumed.

Technically, this uses the same form as the create project form, with some tweaks in behaviour such as hiding the template fields.

### UI After UX Review

See comment thread for details.

Screenshot 2023-01-30 at 16 14 51

Screenshot 2023-01-30 at 17 08 18

### Original UI from PR

Screenshot 2023-01-25 at 16 17 36

Screenshot 2023-01-25 at 16 18 03

## Related Issue(s)

#595 

## Checklist

## Requirements

- [x] Extract shared logic between create/update/copy into a single component
- [x] During edit only permit the Confirm Changes when changes to the project have been made (i.e. dirty form)
- [x] Do not permit template to be updated
- [x] In the UI combine update stack and update project type
- [x] E2E test coverage of updating a projects stack or type
- [x] E2E test coverage of creating and copying projects (previously untested)


joepavitt:
@Pezmc just flagging a conflict, fairly sure we want what you have, but wanted to make sure as it's likely linked to the moving of the "you have credit" message

forge/routes/api/project.js
```diff
@@ -361,7 +361,55 @@ module.exports = async function (app) {
         }
     }, async (request, reply) => {
         let changed = false
-        if (request.body.stack) {
+        if (request.body.changeProjectType === true) {
```
knolleary, line 364: This end point is somewhat overloaded, but it does already let you set the ProjectType of the project **only if** the project doesn't already have a ProjectType set. This was added when we introduced ProjectTypes as a one-off task for users to set the type of their project. This is why the handling for that will bail out early if the Project already has a ProjectType set.

I'm wondering if, rather than introduce a `changeProjectType` flag when we don't have equivalents for the other things you can change, this logic should be moved down to the existing handling when `body.projectType` is set (line 434 of the proposed changes).

That block then has two branches - one for setting ProjectType initially, and one for changing ProjectType.

We should look at the data on production, but handling projects without a ProjectType is very much a legacy edge case that would only apply to FF instances that have been running since before ProjectTypes were introduced.

Steve-Mcl: > I'm wondering if, rather than introduce a changeProjectType flag when we don't have equivalents for the other things you can change, this logic should be moved down to the existing handling when body.projectType is set (line 434 of the proposed changes).

Essentially, there was the original "no project type" code that was rather implicit in the fact that if `body` supplied a `projectType`. 

With the introduction of Change Project Type, the `body` needs to contain both `projectType` AND `stack` in the `body` - so the explicit `changeProjectType` was introduced to make this 100% explicit as to intention rather than inferring intention from values in body. It is much more deliberate than being present or not.   

Also, since the first part of the logic would capture execution (because `body.stack` is needed in a project type change) we would have to have some `if (stack && !projectType)` logic in the first part of the handler (i.e. less explicit intention)

Pezmc:
I've resolved the conflict with main from #1613 and made a couple of minor tweaks to make the endpoint easier to follow.

I'm going to re-work that method in a follow up PR as it's become incredibly overloaded.

knolleary:
Can you summarize the public API changes this is making? (ie to the `/api/v1/projects/...` end point) I appreciate its an overloaded endpoint already, so it would be good to get something written down to outline what is being changed.

Pezmc:
@knolleary The only API change is adding a new flag `changeProjectDefinition` to the `/api/v1/projects` endpoint, which triggers a new path in the code (to be refactored away in follow up). When that flag is set, both projectType and stack are required, and the projects type and stack are changed, with a project restart in-between.

Pezmc:
For reference, the refactored method can be found in https://github.com/flowforge/flowforge/blob/125a4ba5d556797e37b0c6b6f6f8d604723c179c/forge/routes/api/project.js#L346-L595

knolleary:
Having tried this out, I see you've combined changing stack and changing type into one option.

Updating the stack is a far more routine maintenance action on the project than changing the type. That said, we did introduce versioning of stacks - and if there is a new version of the existing stack available, they get an 'Update Stack' button which gives them a one-click update without having to make any choices:

image


For the new button 'Change Project Settings' doesn't feel like the right text. As we discussed in slack, Project Settings means something else here. How about just labelling it as `Change Project Type` as that is what it's for.

Steve-Mcl:
> Updating the stack is a far more routine maintenance action on the project than changing the type. That said, we did introduce versioning of stacks - and if there is a new version of the existing stack available, they get an 'Update Stack' button which gives them a one-click update without having to make any choices

Hi @knolleary not sure if you are asking if the "Update Stack" is now part of the "Change Project Type" or if you are answering a previous question - but for reference, the "Update Stack" button is still there on the form: https://github.com/flowforge/flowforge/pull/1610/files#diff-a5605eb4d14a272712bacb8569e913d4c3f1885a374d594f77277d7d91d2656bR20-R29 

Apologies if I am misreading.

knolleary:
> ot sure if you are asking if the "Update Stack" is now part of the "Change Project Type" or if you are answering a previous question - but for reference, the "Update Stack" button is still there on the form:

I appreciate my comment is a little muddled because I changed tack half way through writing it.

My original point is the 'Change Stack' button has been replaced by this new 'Change Project Settings' button

image

This means changing a project's stack now forces the user through billing considerations, whereas it is only when they change the type will anything need to change with their billing.

Steve-Mcl:
> My original point is the 'Change Stack' button has been replaced by this new 'Change Project Settings' button
> 
>
> This means changing a project's stack now forces the user through billing considerations, whereas it is only when they change the type will anything need to change with their billing.

@knolleary

In my initial work, the change of stack did not show any change of billing. The whole section around billing was totally hidden unless the project type was being changed. I have yet to work through all the changes but I assume that is still the case.  I will update you on what I find

knolleary (approved):

PR #1619, opened 2023-01-26
Store project billing state when adding/removing subscription

### LINKED ISSUE #1616: Track project subscription status
### Description

If billing is enabled we currently add a project to the Team Subscription when it is created.

When the project is suspended, we remove from the Team Subscription. When it is restarted, we add it back.

There is currently a lot of assumption that these things stay in sync.

We have a couple stories that require, for different reasons, the ability to better track whether a project has been added to the subscription or not:

 - #1578 
 - #1598 

This task is to introduce a mechanism for doing this - rather than force a dependency between those two unrelated stories.

Rather than add a new column to the Projects table, we'll add this as a new ProjectSetting.

### Epic/Story

_No response_

Pezmc:
## Description

Stores the project's billing state whenever it is added to or removed from the billing subscription. 

## Related Issue(s)

Fixes #1616

## Checklist

### DIFF
--- a/forge/containers/wrapper.js
+++ b/forge/containers/wrapper.js
@@ -1,4 +1,3 @@
-
 class SubscriptionHandler {
     constructor (app) {
         this._app = app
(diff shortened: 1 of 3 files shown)


knolleary (approved):

knolleary:
Will let the tests run, then get this merged. 👍🏻

PR #1624, opened 2023-01-30
Remove unused export to JSON feature

Pezmc:
This was commented out as part of #547, it may be restored, but for now, let's remove the dead code, this PR can always be reverted down the line.


hardillb:
Following a comment from @knolleary while we were firefighting on Friday night I'm going to suggest we hold off for a moment on this.

knolleary:
For context - we found we had no way to take a complete backup of a project. This functionality would have been helpful at the time if it had existed.

As part of the lessons learnt from Friday, we may want to prioritise restoring this functionality.

An alternative is to add a Download option to snapshots - which would have its own set of benefits such as enabling 'offline' device agent. If we go that route, then we can and should strip out the unused code as per this PR.

The point being, let's not remove a bunch of code today that we may, within days, decide we need to restore.

Pezmc:
Is there an issue to follow for the possible restoration?

hardillb:
We are discussing Friday's incident here https://github.com/flowforge/CloudProject/issues/131

Actions yet to be determined.

Pezmc:
I'd strongly lean towards removing early, given restoring the feature is only a revert away. Unused code bloats the codebase and slows down work, however if it's literally likely to be restored this week, let's leave it for now, and re-visit this PR next week.

hardillb:
I'm going to say approve, we can pull it back from the history if we change our minds

hardillb:
I'm going to say approve, we can pull it back from the history if we change our minds

hardillb (approved):

PR #1631, opened 2023-01-31
Change project type API unit tests

Pezmc:
Unit tests of the API changes for #1610.

Best viewed with [?w=1](https://github.com/flowforge/flowforge/pull/1631/files?w=1) due to indent changes in the tests.

Pre-cursor to https://github.com/flowforge/flowforge/pull/1623

### DIFF
--- a/test/unit/forge/routes/api/project_spec.js
+++ b/test/unit/forge/routes/api/project_spec.js
@@ -755,222 +755,355 @@ describe('Project API', function () {
     })
 
     describe('Update Project', function () {
-        it('Cannot change project-type if already set', async function () {
-            const projectType = {
-                name: 'projectType2',
-                description: 'default project type',
-                active: true,
-                properties: { foo: 'bar' },
-                order: 2
-            }
-            const projectType2 = await app.db.models.ProjectType.create(projectType)
-            const response = await app.inject({
-                method: 'PUT',
-                url: `/api/v1/projects/${TestObjects.project1.id}`,
-                payload: {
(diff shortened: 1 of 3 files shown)


test/unit/forge/routes/api/project_spec.js
```diff
@@ -755,222 +755,410 @@ describe('Project API', function () {
...
-                TestObjects.tokens.project,
-                [{ id: 'node1' }],
-                { testCreds: 'abc' },
-                'key1',
-                {
-                    httpAdminRoot: '/test-red',
-                    env: [
-                        { name: 'one', value: 'a' },
-                        { name: 'two', value: 'b' }
+                response.statusCode.should.equal(200)
```
Pezmc, line 794: Unfortunately, I couldn't find a way to assert the project is suspended as the request only returns once the project has been restarted.

knolleary (approved):

test/unit/forge/routes/api/project_spec.js
knolleary: One approach would be to add some sort of internal logging to the stub driver that can be used to check what has been done to a given project. But not sure that's worth the effort at this point.

Pezmc: We could spy on the audit logs actually.

knolleary: I did wonder that, but couldn't recall if the audit logging is done at that level.

PR #1634, opened 2023-02-01
Refactor project put method to a single route

Pezmc:
## Description

Refactors the unwieldy Project update endpoint to have a single implementation for _updating_ a project. 

Three key objectives:
 - Easier to maintain code
 - Clearer overview of what is legacy
 - Improved failure state handling (transactions) - previously errors would leave projects in a half updated state

There are two subtle API changes as a result, the rest of the API behaves the same as previously (including the legacy setting of stack and type):
 - The `changeProjectDefinition` flag is no longer needed to change a projects time
 - Changing a projects type now returns early and completes the reboot of the project async

## Related Issue(s)

Replaces https://github.com/flowforge/flowforge/pull/1623
Follow up to https://github.com/flowforge/flowforge/pull/1631

## Checklist

### DIFF
--- a/forge/containers/stub/index.js
+++ b/forge/containers/stub/index.js
@@ -14,6 +14,7 @@ const forgeUtils = require('../../db/utils')
 
 module.exports = {
     START_DELAY: 500,
+    STOP_DELAY: 250,
     /**
      * Initialises this driver
      *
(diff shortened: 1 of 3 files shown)


Steve-Mcl (commented):
From my PoV, anything that helps us as devs on a growing code base is good. 
The re-ordering of code is definitely better (even if the PUT route is still rather large and could/should? be further refactored to encapsulate functionality) it is a good step.

The adoption may come down to effort vs benefit & from a consumer there may be no discernible improvements - however from a dev perspective:  I think the answer here is yes - considering things are only going to get more complicated as the code base grows.

You have my stamp of approval ;)

hardillb (approved):

knolleary (approved):

PR #1645, opened 2023-02-01
Reset team state after user deletes their last team

### LINKED ISSUE #1630: Deleting your only team, doesn't exit from the team UI
### Current Behavior

https://app.posthog.com/recordings/18604477d24127-0ea0ef21b0add68-40272c3d-1ea000-18604477d25954?t=1041

When a user goes to delete their only team, the UI still implies that the team exists, even after successfully removing the team. They can still see Logs, Projects, etc.

There are no "Members" listed, but the user can very clearly navigate around.

### Expected Behavior

If a user deletes their only team, they should be returned to the "Create Team" page they see when first creating an account. From here, they should not be able to access any of the previously created team features/UI.

### Steps To Reproduce

- Create a new user
- Add that user to a single team (or create a team as that user)
- Delete the team.

### Environment

- FlowForge version: 1.3.0
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Fixes #1630

### DIFF
--- a/frontend/src/store/account.js
+++ b/frontend/src/store/account.js
@@ -172,6 +172,7 @@ const actions = {
 
             if (teams.count === 0) {
                 state.commit('clearPending')
+                state.commit('setTeam', null)
                 if (/^\/team\//.test(router.currentRoute.value.path)) {
                     router.push({ name: 'Home' })
                 }


knolleary:
cypress tests failed for a seemingly completely unrelated issue. Rerunning to see if its repeatable.

Pezmc (approved):

Issue #1609, opened 2023-01-24
Async Team Slug Check


joepavitt:
### Description

We are getting a lot of users bouncing back when selecting the "Create Team" option. This is because a Team slug needs to be unique, which we do not explain prior to a user receiving the error/message to say that after they submit the form.

Instead, we should be perform an async API request to check a slug, once the input has been filled out. At that point, show the relevant message if required, otherwise, we can consider the form valid (subject to other fields being completed).

PR #1655, opened 2023-02-03
Team slug check

knolleary:
## Description

This adds an endpoint `POST /api/v1/teams/check-slug` that can be used to validate if a team slug is available for use or not.

Available returns status 200 `{status: 'okay'}`.
Unavailable returns status 409 `{ code: 'invalid_slug', error: 'Slug not available' }`

The Create Team and Update Team pages have been updated to use this endpoint and feedback to the user before they try using the slug.

## Related Issue(s)

#1609 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/api/team.js
+++ b/forge/routes/api/team.js
@@ -76,8 +76,8 @@ module.exports = async function (app) {
         }
     }, async (request, reply) => {
         const slug = request.body.slug.toLowerCase()
-        // Handle reserved names
-        if (slug === 'create') {
+        const reservedNames = ['create']
+        if (reservedNames.includes(slug)) {
             reply.code(409).send({ code: 'invalid_slug', error: 'Slug not available' })
             return
         }
--- a/forge/routes/api/team.js
+++ b/forge/routes/api/team.js
@@ -76,11 +76,16 @@ module.exports = async function (app) {
         }
     }, async (request, reply) => {
         const slug = request.body.slug.toLowerCase()
+        // Handle reserved names
+        if (slug === 'create') {
+            reply.code(409).send({ code: 'invalid_slug', error: 'Slug not available' })
+            return
+        }
         const existingCount = await app.db.models.Team.count({ where: { slug } })
         if (existingCount === 0) {
             reply.send({ status: 'okay' })
         } else {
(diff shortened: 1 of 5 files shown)


forge/routes/api/team.js
```diff
@@ -63,6 +63,32 @@ module.exports = async function (app) {
         }
     })
 
+    app.post('/check-slug', {
```
Pezmc, line 66: [nb] Not needed now, but I wonder if we should split the routes with a comment for member and collection endpoints.

(check-slug and `/` are collection routes, anything with `/:teamId/` is a member route)

forge/routes/api/team.js
```diff
@@ -63,6 +63,32 @@ module.exports = async function (app) {
...
+                required: ['slug'],
+                properties: {
+                    slug: { type: 'string' }
+                }
+            }
+        }
+    }, async (request, reply) => {
+        const slug = request.body.slug.toLowerCase()
+        // Handle reserved names
+        if (slug === 'create') {
```
Pezmc, line 80:
```suggestion
        const reservedNames = ['create']
        if (reservedNames.includes(slug)) {
```

frontend/src/pages/team/Settings/General.vue
```diff
@@ -140,6 +147,25 @@ export default {
 
         async fetchData () {
             this.cancelEditName()
+        },
+        checkSlug () {
+            if (this.pendingSlugCheck) {
+                clearTimeout(this.pendingSlugCheck)
+            }
+            this.pendingSlugCheck = setTimeout(() => {
```
Pezmc, line 155: [nb] There's now a debounce like this at least a couple of places now (Users/Genera, Users/Team), it might be worth extracting it into its own function, particularly as then it removes the extra attr in the Vue state.

Pezmc (approved):

frontend/src/pages/team/Settings/General.vue
knolleary: Raised as a future task

forge/routes/api/team.js
knolleary: Yup - something for another day, but not really worth tracking in an issue.

PR #1661, opened 2023-02-06
Stop HubSpot capturing unwanted data

### LINKED ISSUE #1660: Remove HubSpot form capture
### Story

_No response_

### Description

We seem to be capturing people who reset their FF Cloud password in HubSpot. We don't need to be doing this. I believe this is because we have some HubSpot javascript on our site that automatically captures any form completion. We have converted all the relevanant forms to HubSpot forms so this javascript is no longer needed.

hardillb:
## Related Issue(s)

fixes #1660

## Checklist

    - [-] Upgrade instructions
    - [-] Configuration details
    - [-] Concepts
    - [-] Issue/PR raised on `flowforge/helm` to update ConfigMap Template
    - [-] Issue/PR raised on `flowforge/CloudProject` to update values for Staging/Production

## Labels

 - [x] Backport needed? -> add the `backport` label

### DIFF
--- a/frontend/src/pages/account/Security/ChangePassword.vue
+++ b/frontend/src/pages/account/Security/ChangePassword.vue
@@ -10,6 +10,8 @@
         </ff-button>
         <div v-if="errors.password_change" class="ml-4 text-red-400 font-medium inline text-sm">{{errors.password_change}}</div>
         <div v-if="changeComplete" class="ml-4 font-medium inline text-md">Password changed</div>
+        <!-- https://community.hubspot.com/t5/Lead-Capture-Tools/Ignore-a-specific-non-Hubspot-form-from-being-collected/m-p/246957 -->
+        <span id="CollectedForms-1234"></span>
     </form>
 </template>
 


frontend/src/pages/account/Security/ChangePassword.vue
```diff
@@ -10,6 +10,8 @@
         </ff-button>
         <div v-if="errors.password_change" class="ml-4 text-red-400 font-medium inline text-sm">{{errors.password_change}}</div>
         <div v-if="changeComplete" class="ml-4 font-medium inline text-md">Password changed</div>
+        <!-- https://community.hubspot.com/t5/Lead-Capture-Tools/Ignore-a-specific-non-Hubspot-form-from-being-collected/m-p/246957 -->
+        <span id="CollectedForms-1234"></span>
```
Pezmc, line 14: I think the 1234 has to be swapped for our hubspot ID, which it looks like is currently stored in `support.frontend.hubspot.trackingcode`

Pezmc (changes requested):

frontend/src/pages/account/Security/ChangePassword.vue
hardillb: Doh, wasn't awake when wrote it this morning, will look for the right option to set it in page

hardillb:
Closing, having reviewed with Joe, this is not capturing anything that we need to worry about at this time.

PR #1679, opened 2023-02-07
Add "Last Seen" device badge

### LINKED ISSUE #1599: Improve communication of Device Last Seen and Status
Reference: https://github.com/flowforge/flowforge/issues/1116#issuecomment-1385580139 

Highlight the last-seen value of a device based on time since we last saw it.

- The eye is drawn to the "Running" pill, not the "Last Connected"
- We could move the connection into a pill too, red/yellow/green status depending on last connected status, which flags to the user.
- It's perfectly valid that the flows could be running, even though we haven't connected in a month, so it's difficult to flag otherwise.

For now we can use a fix set of time bands (TBD) for the colour of status.

joepavitt:
## Description

- Adds a new device status badge that shows a clearer picture of device status & connectivity.
- Change column header to "Last Known Status"

Screenshot 2023-02-07 at 22 34 52

**Categorisation:**

As implemented currently, shows as green if last seen within the last minute, otherwise red. Will likely want to add more text-based labels to cover more time periods beyond 1 hr ago.

| ~Last Seen~ | ~Colour~ |
| - | - |
| never | colourless |
| < 1 min | green |
| > 1 min | red |

updated table to: https://github.com/flowforge/flowforge/pull/1679#issuecomment-1423218275

## Related Issue(s)

Closes #1599

## Checklist

### DIFF
--- a/frontend/src/pages/device/components/DeviceLastSeenBadge.vue
+++ b/frontend/src/pages/device/components/DeviceLastSeenBadge.vue
@@ -10,7 +10,7 @@ import { ExclamationCircleIcon } from '@heroicons/vue/outline'
 
 export default {
     name: 'ProjectStatusBadge',
-    props: ['lastSeenAt'],
+    props: ['lastSeenAt', 'lastSeenSince'],
     computed: {
         since: function () {
             if (this.lastSeenAt) {
@@ -25,25 +25,22 @@ export default {
             }
         },
         status: function () {
+            // re-uses the status from last known status in order to get respective colour styling
             if (this.since < 0) {
-                return 'never'
-            } else if (this.since < 1) {
+                return 'never' // green
+            } else if (this.since < 1.5) {
                 return 'running'
+            } else if (this.since < 3) {
+                return 'safe' // yellow
             } else {
-                return 'error'
+                return 'error' // red
             }
         },
         label: function () {
             if (this.since < 0) {
                 return 'never'
-            } else if (this.since < 1) {
-                return 'less than a minute ago'
-            } else if (this.since < 2) {
-                return `${Math.floor(this.since)} minute ago`
-            } else if (this.since < 60) {
-                return `${Math.floor(this.since)} minutes ago`
             } else {
-                return 'over an hour ago'
+                return this.lastSeenSince
             }
         }
     },
--- a/frontend/src/pages/device/components/DeviceLastSeenBadge.vue
+++ b/frontend/src/pages/device/components/DeviceLastSeenBadge.vue
@@ -6,7 +6,7 @@
 </template>
 
 <script>
-import { ExclamationIcon, ExclamationCircleIcon, PlayIcon, StopIcon, DotsCircleHorizontalIcon, SupportIcon, CloudDownloadIcon, CloudUploadIcon } from '@heroicons/vue/outline'
+import { ExclamationCircleIcon } from '@heroicons/vue/outline'
 
 export default {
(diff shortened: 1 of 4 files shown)


Pezmc:
Are you able to pull some stats to get an idea of the current distribution of last-seen for devices?

joepavitt:
> Are you able to pull some stats to get an idea of the current distribution of last-seen for devices?

Not sure I would be able to - they're not too widely used right now on FF Cloud, although with the provisioning work that @Steve-Mcl has done, that should hopefully change.

They auto-ping every 30 seconds, so if they're connected, it'll be every 30 seconds that we hear from them

knolleary:
> They auto-ping every 30 seconds,

The default ping interval is 60 seconds, but the user can configure that to something else that makes sense for their scenario - and there is a bit of jitter applied so it isn't always exactly 60 seconds - I think it can be +/- 5 seconds by default.

If we assume most users stick to the defaults for now (until we address the broader solution outlined in the original issue), then I suggest:

| Last Seen | Colour |
| - | - |
| never | colourless |
| < 90 secs | green |
| 90 sec - 3 min | yellow |
| > 3 min | red |

But if you think green/red is sufficient, then combine yellow/red into one band.

Steve-Mcl:
### 1. no see long time
Supposing a customer has distant remote (often offline due to environment) devices.  e.g. 1 day unseen is perfectly ok and even expected due to "sleeping" over a weekend etc.  They are going to see red entries with `'over an hour ago'` - maybe we could expand it to "over x days ago" then to "over x months" ago.  

FMPoV, 
* Seeing  "over 1 days ago" on a sunday is OK
* Seeing  "over 4 days ago" on a monday is NG
* Seeing  "over 1 month ago" means I should probably delete this device.


### 2. sorting
Will having a variable "last seen" text break the sorting (or make it unpredictable)?
Does it sort on text or the underlying field value?

joepavitt:
> FMPoV

Agree there will be variation, i think as a first point of call, we want _something_ in, then as a second phase, we can make this configirable.

> Does it sort on text or the underlying field value?

It sorts on the `lastSeenAt` field, which is the UTC string of the date. Ideally we modify this. Will add that change in here.

joepavitt:
e.g:

Screenshot 2023-02-09 at 13 52 29

Steve-Mcl:
> > FMPoV
> 
> Agree there will be variation, i think as a first point of call, we want _something_ in, then as a second phase, we can make this configurable. I have changed the "Over x hrs" to be more descriptive

Just in case you missed my point (or more likely I wasnt clear) - I agree colour threshold setting can/should be a future iteration however I was mostly trying to convey this...

Only having 'over an hour ago' means I cannot judge if a device is simply sleeping or lost/broke/redundant.

Having additional "time frames" means I can see if a device is actually redundant or lost.  Something like...

<html>
<body>
<google-sheets-html-origin>

Last Seen | Colour | Displays
-- | -- | --
never | colourless | never
< 90 secs | green | 1 minute ago
90 sec - 3 min | yellow | 2 minutes ago
&gt;  3 min | red | `x` minutes ago
&gt;  1h | red | more than 1 hour ago
&gt;  2h | red | more than `x` hours ago
&gt;  24h | red | more than 1 day
&gt;  48h | red | more than `x` days ago
&gt; 1 month | red | more than 1 month ago


</body>
</html>

Food for thought hopefully Joe.

joepavitt:
> doh!

:)

Steve-Mcl (approved):
purrrfect

Issue #1597, opened 2023-01-19
Team Library 2.0


joepavitt:
### Description

**As a:** FlowForge Team Member

**I want to:** be able to share and manage common flows & functions that my team use

**So that:** we can easily spin up new instances of Node-RED, prevent duplicate effort in recreating commonly used nodes & patterns

https://github.com/flowforge/flowforge/pull/1576 saw the addition of a very basic Team Library interface. This story tracks the additional features that we didn't quite get into 1.3 Release, that we would like to see in the interface.

### Which customers would this be availble to

None

### Acceptance Criteria

- [x] #1657
- [x] We have no way to show a rendered view of the flow, but providing access to the JSON would also be valuable to allow a user to copy an entry without having to open up a NR editor.
    - Included in #1657 too
- [x] #1690
   - [x] https://github.com/flowforge/flowforge/pull/1704
- [x] #1691
   - [x] https://github.com/flowforge/flowforge/pull/1704
- [x] https://github.com/flowforge/flowforge/pull/1713

joepavitt:
> Add 'delete' option (will require backend api as well)

@knolleary still okay to add a Delete endpoint to the Storage API? If it's fairly straight forward, I don't mind attempting it.

joepavitt:
Pez taking over the last 2 bits

joepavitt:
@Pezmc is this done? what's the status?

Pezmc:
@joepavitt https://github.com/flowforge/flowforge/pull/1704 is open but has some UX bugs that I'm working on

PR #1690, opened 2023-02-10
Add delete route for shared library api

knolleary:
## Description

Adds support for deleting entries from the Shared Team Library

 - `DELETE /storage/library/:libraryId/:path`

Response:

 - `{ "status": "okay", "deleteCount": <how many things were deleted> }`

If `path` matches a single entry (ie a 'file') the file is deleted (`deleteCount: 1`)
If `path` matches a folder, all entries beneath that folder are deleted. (`deleteCount: n`)
If `path` doesn't match anything, it 404s.

An optional `type=<type>` parameter can be included to restrict the operation to library entries of a particular type. Whilst this isn't necessarily needed by the UI, it is consistent with the `GET` api that Node-RED requires to be able to filter by type.

## Related Issue(s)

#1597 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/test/unit/forge/ee/routes/sharedLibrary/index_spec.js
+++ b/test/unit/forge/ee/routes/sharedLibrary/index_spec.js
@@ -51,6 +51,20 @@ describe('Library Storage API', function () {
             })).json()
         }
 
+        async function deleteFromLibrary (libraryURL, name, type) {
+            let query = ''
+            if (type) {
+                query = `?type=${type}`
+            }
+            return (await app.inject({
+                method: 'DELETE',
+                url: `${libraryURL}${name}${query}`,
+                headers: {
+                    authorization: `Bearer ${tokens.token}`
+                }
+            })).json()
+        }
+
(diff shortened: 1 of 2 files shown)


joepavitt:
thanks for turning that round so quickly Nick - will review/merge once tests have run

test/unit/forge/ee/routes/sharedLibrary/index_spec.js
```diff
@@ -229,5 +243,103 @@ describe('Library Storage API', function () {
             const response2 = await addToLibrary(libraryURL, 'test/foo', 'functions')
             response2.statusCode.should.equal(400)
         })
+
+        it('Deletes an individual entry', async function () {
```
Pezmc, line 247: [nb] Could you add a test for the partial match fork?

Pezmc (approved):
I think an extra tests is needed, but otherwise 👍

Issue #1719, opened 2023-02-16
Remap Project-Level navigation.


joepavitt:
Epic: https://github.com/flowforge/flowforge/issues/1689

Where https://github.com/flowforge/flowforge/issues/1718 handles the moving of existing `/project` pages into the new `/instances` mapping, this Story handles the content of what will replace those pages at the `/project` level.

Whilst we are still using the same name "Project", the functionality of a Project is changing, as we are enabling multiple "Instances" per Project with #1689. This means that a Project will now purely be an organisational/folder-type concept, and not explicitly have any functionality directly bound to it with respect to Node-RED.

Top-level routes we are expecting to have on a project-by-project basis:

- [x] ~Overview~
    - ~We will still need a high-level insight into the instances, status, etc. within a given project. This would also act as the landing page for any given navigation to a project, for example from the higher-level "team"~
- [x] Instances - New landing page
    - An almost exact mirror of how we present this screen today. Unlikely to require much change, we want to view all instances mapped to a given project.
- [x] Audit Logs
    -  Needs a dropdown to select instance. Otherwise the same as today. We want to view all instances mapped to a given project.
- [x] Node Log
    -  Needs a dropdown to select instance. Otherwise the same as today. We want to view all instances mapped to a given project.
- [ ] #1734
    - Some discussion required here as to which settings remain at the new "Project" level, and which are moved down into Instance

Pezmc:
Overview, instances, audit log, node logs completed in #1792, re-working the settings tab will the tracked in #1734, so considering this done.

joepavitt:
verified in staging

Issue #1689, opened 2023-02-09
Multiple FlowForge Hosted Instances per Application


joepavitt:
### Description

<img width="1261" alt="Screenshot 2023-02-09 at 14 25 37" src="https://user-images.githubusercontent.com/99246719/217839945-94adcfe6-3f3e-4357-98c4-338806745ad4.png">

Two large ticket items we want to deliver to customers are:

- High Availability - https://github.com/flowforge/flowforge/issues/1678
- Staged Deployments & DevOps Pipeline - https://github.com/flowforge/flowforge/issues/1492
    - intermittent delivery in 1.4 - https://github.com/flowforge/flowforge/issues/1580

In both of these issues, to do them _properly_ we would need to support a FlowForge project being able to host multiple instances of Node-RED.

Conceptually, this diagram shows an evolution of the model:

<img width="2564" alt="Screenshot 2023-02-09 at 14 51 10" src="https://user-images.githubusercontent.com/99246719/217846681-8a78a487-0de7-45b7-bbea-57f554d41858.png">

This epic encompasses the discussion and design work to understand the scope of work required to support this, and then to detail the development tasks at hand to make that happen.

Source of this plan comes from https://github.com/flowforge/flowforge/issues/1450#issuecomment-1372564433, but to copy here for easy tracking/discussion:

**Required Changes:**
- Bare minimum requirement here is that each instance would need to have its own Settings page, as per our current "Device" page, for the reason you've raised (env vars)
- Audit Logging needs consideration here for each instance - quick fix would be to just log project.. events where we now have Instances, and with the recent introductions of the new Audit Log formatting, it makes it easy to distinguish between the source of the logs (i.e. which instance they belong to)

**Comments:**
- Under the covers here, we could just transfer everything we have for a "Project", and convert that to the functionality of an Instance, then on the front-end, we package those and arrange them into the new "Projects" concept, which actually keeps the Projects UX as things stand.
- Audit Log filtering becomes more valuable here, and relevant at the "Project" level, as we could filter by "Instance".
- Although this does break our audit logging concepts a bit, as we use project. there.

### UI Rework Task List

```[tasklist]
- [ ] **=== Needed for 1.5 ===**
- [x] flowforge/flowforge#1732
- [ ] flowforge/flowforge#1718
- [ ] flowforge/flowforge#1719
- [ ] https://github.com/flowforge/flowforge/issues/1802
- [ ] https://github.com/flowforge/flowforge/issues/1810
[...]

PR #1769, opened 2023-03-03
Application Instances: Simplify Project Navigation

Pezmc:
**Merge target** is the feature branch for application instances as part of #1689: [feat-1689-application-instances](https://github.com/flowforge/flowforge/tree/feat-1689-application-instances)

## Description

Screenshot 2023-03-03 at 17 58 43
Screenshot 2023-03-03 at 17 58 54

## Related Issue(s)

#1719


knolleary:
I'm not seeing the same thing as the screenshots.

When viewing an instance, I have the 'old' sidebar still - no navigation back to the project:

image


I'm pretty sure this isn't a local caching/rebuilding thing... although I can't 100% rule that out yet.

knolleary:
> I'm not seeing the same thing as the screenshots.

User error... ignore me.

---

I'm not convinced the navigation is quite right - still feeling 'lost' when viewing an instance.

A difficultly we have is that for users with a single instance in the app (ie every existing user) there is a 1-1 mapping of Application and Instance, and they could have the same name. That makes it tricky to know if you're seeing the application name or the instance name in the navigation. I know we aren't doing a full redesign of the navigation in this iteration, so we can move ahead as-is. But I would like to try to spare some cycles to iterate on this within this release.

knolleary (approved):

PR #1774, opened 2023-03-07
Remove unused dependency

Pezmc:
It seems like this dependency isn't used, if tests pass, we can remove it entirely rather than updating it (https://github.com/flowforge/flowforge/pull/1754)

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,6 @@
         "should": "^13.2.3",
         "should-sinon": "^0.0.6",
         "sinon": "^15.0.1",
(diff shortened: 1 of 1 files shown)


Steve-Mcl (commented):
It does appear to be unused - but Joe may use it from command line? 

Happy to remove if you are Pez - will leave for you to commit (or for @joepavitt to comment)

Pezmc:
Thanks! 

We can use `npx start-server-and-test` if it's need locally for CLI work!

PR #1813, opened 2023-03-13
Add Application migration and instances endpoint

knolleary:
This is a continuation of the work to introduce the Application concept into the runtime.

It targets the `feat-1735-application-model` branch which introduces the Application model and basic CRUD API.

This PR adds to that branch:

 - A migration that generates an Application object for every existing Project.
 - The `/api/v1/applications/:applicationId/instances` endpoint that lists the instances within an application
 - Updates the Project views to include the Application info

### DIFF
--- a/forge/db/models/Application.js
+++ b/forge/db/models/Application.js
@@ -7,11 +7,6 @@ const { DataTypes } = require('sequelize')
 module.exports = {
     name: 'Application',
     schema: {
-        id: {
-            type: DataTypes.INTEGER,
-            primaryKey: true,
-            autoIncrement: true
-        },
         name: { type: DataTypes.STRING, allowNull: false }
     },
     associations: function (M) {
--- a/forge/db/models/Project.js
+++ b/forge/db/models/Project.js
@@ -258,6 +258,10 @@ module.exports = {
                             model: M.Team,
                             include: [
                                 {
+                                    model: M.Application,
+                                    attributes: ['hashid', 'id', 'name', 'links']
+                                },
+                                {
                                     model: M.TeamMember,
                                     where: {
                                         UserId: user.id
@@ -286,6 +290,10 @@ module.exports = {
                             {
                                 model: M.Team,
                                 attributes: ['hashid', 'id', 'name', 'slug', 'links']
+                            },
+                            {
+                                model: M.Application,
+                                attributes: ['hashid', 'id', 'name', 'links']
                             },
                             {
                                 model: M.ProjectType,
@@ -320,6 +328,10 @@ module.exports = {
                                 model: M.Team,
                                 where: { id: teamId },
                                 attributes: ['hashid', 'id', 'name', 'slug', 'links']
+                            },
+                            {
+                                model: M.Application,
+                                attributes: ['hashid', 'id', 'name', 'links']
                             },
                             {
                                 model: M.ProjectType,
--- a/forge/db/models/Project.js
+++ b/forge/db/models/Project.js
@@ -320,6 +320,34 @@ module.exports = {
                         ]
                     })
                 },
+                byApplication: async (applicationHashId) => {
+                    const applicationId = M.Application.decodeHashid(applicationHashId)
+                    return this.findAll({
+                        include: [
+                            {
+                                model: M.Team,
+                                attributes: ['hashid', 'id', 'name', 'slug', 'links']
+                            },
+                            {
+                                model: M.Application,
+                                where: { id: applicationId },
+                                attributes: ['hashid', 'id', 'name', 'links']
+                            },
+                            {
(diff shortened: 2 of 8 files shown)


Pezmc:
@dependabot rebase

forge/db/models/Application.js
```diff
@@ -7,11 +7,6 @@ const { DataTypes } = require('sequelize')
 module.exports = {
     name: 'Application',
     schema: {
-        id: {
```
Pezmc, line 10: Does an application not need a primary key?

forge/db/views/Application.js
```diff
@@ -1,14 +1,18 @@
 module.exports = {
     application: function (app, application) {
-        const raw = application.toJSON()
-        const filtered = {
-            id: raw.hashid,
-            name: raw.name,
-            createdAt: raw.createdAt,
-            updatedAt: raw.updatedAt,
-            links: raw.links
+        if (application) {
```
Pezmc, line 3: Should this responsible for guarding against application being `null`? That feels like it belongs in whatever view is calling this view.

forge/db/views/Project.js
```diff
@@ -33,7 +33,14 @@ module.exports = {
 
         const settingsHostnameRow = proj.ProjectSettings?.find((projectSettingsRow) => projectSettingsRow.key === KEY_HOSTNAME)
         result.hostname = settingsHostnameRow?.value || ''
-
+        if (proj.Application) {
+            result.application = {
```
Pezmc, line 39: Should this also use `app.db.views.Application.application`?

forge/db/models/Project.js
```diff
@@ -320,6 +320,34 @@ module.exports = {
...
+                                where: { id: applicationId },
+                                attributes: ['hashid', 'id', 'name', 'links']
+                            },
+                            {
+                                model: M.ProjectType,
+                                attributes: ['hashid', 'id', 'name']
+                            }
+                            //,
+                            // {
+                            //     model: M.ProjectStack
```
Pezmc, line 342: ✂️

Pezmc (changes requested):
See comments above!

forge/db/models/Application.js
hardillb: Sequelize will automatically add a field called `id` if there isn't a field already marked as a primary key (with basically the same constraints as the one removed)

forge/db/models/Project.js
```diff
@@ -312,6 +320,34 @@ module.exports = {
...
+                            // },
+                            // {
+                            //     model: M.ProjectTemplate,
+                            //     attributes: ['hashid', 'id', 'name', 'links']
+                            // }
```
Pezmc, line 347:
```suggestion
```

forge/db/views/Project.js
```diff
@@ -12,28 +12,37 @@ module.exports = {
...
         }
 
         const settingsHostnameRow = proj.ProjectSettings?.find((projectSettingsRow) => projectSettingsRow.key === KEY_HOSTNAME)
         result.hostname = settingsHostnameRow?.value || ''
-
+        if (proj.Application) {
+            result.application = {
+                id: proj.Application.hashid,
+                name: proj.Application.name,
+                fred: true,
```
Pezmc, line 42:
```suggestion
```

forge/db/views/Application.js
knolleary: This was self-consistent with the Team view that does the same thing. I'm okay with it as-is.

forge/db/views/Project.js
knolleary: `app.db.views.Application.application` is intended to be the full view of the application.

Here we only want a summary - as we do in the later lines that add a team summary without deferring to a specific view.

Pezmc: This feels like something that should be centralised e.g. `app.db.views.Application.application(application, summary: true)` rather than spreading throughout the app

knolleary:
I have merged the suggested changes. A couple of the other comments regarding views I'm happy to leave as-is because the current code is relatively self-consistent with other views - and I don't want to get into a big refactoring of the views.

forge/db/views/Project.js
knolleary: I have pushed a change to introduce `applicationSummary` in line with the `teamSummary` view we already had. In doing so, I've hit the reason for why they weren't currently used. In *this* code path, `proj.Application` has already been converted to a vanilla object - it is no longer a database model object. The summary view was intended for use against the database model object.

I've updated the view to check if its a vanilla object or not and to do the required conversion.

It isn't perhaps the cleanest approach, but it works for now and shouldn't hold up progress on the broader Application work.

Pezmc (approved):

PR #1834, opened 2023-03-14
Relabel Project Types to Instance Types in UI

knolleary:
Part of #1689 

This wasn't originally in scope, but having started to write the doc updates for introducing Instances, the Project Type concept stuck out a lot.

This PR is a cosmetic update to the front-end to reference Instance Type instead of Project Type. In some components I've also updated internal variables to be `instance` rather than `project` where the changes are not too extensive. Some views need more detailed going through so I've elected to skip for now.


knolleary:
The final failing test is due to an update to cypress. @joepavitt has already merged the fix via #1835

knolleary:
Merged main into this branch and all tests are passing. Good for review and merge.

frontend/src/pages/admin/InstanceTypes/index.vue
```diff
@@ -0,0 +1,159 @@
+<template>
```
Pezmc, line 1: Assuming this file is not new and GitHub hasn't tracked the move

knolleary: Correct - I think the combination of rename *and* the number of changes of Project->Instance has exceeded what git thinks is a move and edit.

Pezmc (approved):
I've read through and checked all the changes, but can't guarantee it's 100%

PR #1846, opened 2023-03-16
Increase ESLint requirements for new files

Pezmc:
## Description

This PR increases our ESLint rule requirements in two ways:
 - Moving from [plugin:vue/vue3-essential](https://eslint.vuejs.org/rules/#priority-a-essential-error-prevention) to [plugin:vue/vue3-recommended](https://eslint.vuejs.org/rules/#priority-c-recommended-potentially-dangerous-patterns)
 - Enabling [plugin:import/recommended](https://github.com/import-js/eslint-plugin-import) which validates import paths and keeps the order of imports consistent

Importantly these new rules **only apply to new files**, that is to say, any file that doesn't meet the new higher standard, now has an override set in the ESLint. 

This means that new files are held to a higher standard, they're more consistent and more bugs and patterns likely to lead to errors are caught, while still allowing the old failing files to exist.

The significant majority of the new rules all support `eslint --fix`, which should be enabled in your editor [as per the handbook](https://github.com/flowforge/handbook/blob/main/development/how-we-work/contributing.md#linting) or run manually with `npm run lint:fix`.

## Related Issue(s)

None

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Pezmc:
[734d71a](https://github.com/flowforge/flowforge/pull/1846/commits/734d71ab03552f91e1730905a38b3a33e19fcbe9) has failures not seen locally, downgrading to a draft.

Pezmc:
@Steve-Mcl @knolleary @joepavitt This is ready for review!

Steve-Mcl (commented):
@Pezmc - loaded onto windows machine & tried creating new files etc - as expected, tighter linting rules were shown for new files (e.g. copying many of the existing VUE files showed numerous new warnings and errors) 👍 

1 Question I have is:
Most of the linting markers are only "warnings". E.g. enforcing order of directives makes for a familiar VUE file layout. Is this deliberate or should we be stricter?

image

Pezmc:
Thanks @Steve-Mcl, I'll be rebasing this branch on top of `main` just before it merges to ensure we don't have failures for files currently on other branches. 

I've left Vue.js style issues as warnings only rather than errors, that is to say, they don't cause problems, but ideally we'd fix them up. The majority of the warnings can be auto-fixed, but not all, so it felt best to leave it as a warning, rather than a hard requirement.

knolleary (commented):
My concern here is we now have over 1000 lines of exceptions hardcoded in the lint file. Is the intent to slowly chip away at those in the background when those files get touched for other work that is in plan?
The file says that list is autogenerated - how was that generated? Would be helpful to have that documented in the PR for future reference.

Pezmc:
@knolleary @Steve-Mcl  Would like to get this merged, did you have any follow up questions?

/cc: @joepavitt @hardillb

Steve-Mcl:
@Pezmc I do fully agree we should not simply `--fix` everything in one go however, we should (IMHO) actively progress the removal of exceptions to get the whole of the code base up to a higher standard. Some options/approaches might be to:
* have a rule whereby if you work on a file that is present in the list, you remove it from the list and fix up any lint issues as part of that work.
  * Perhaps we consider adding a checkbox or reminder to the git template?
* have a bunch of open issue that target sections of the code base so that this task does not get lost.

Otherwise, as stated before, in my testing, this was working as expected.  So long this doesnt just become technical debt & we have plans to whittle this down, I personally approve.

hardillb (approved):
I'll say yes to this, but I  think we should have a tax to pull the exceptions down.

e.g. an issue per release to remove one of the exception blocks

Pezmc:
In the interests of maintaining bias for action, I'm merging this PR, we can continue to finesse the best way to keep track of the list (see @Steve-Mcl's comment above) on this thread.

Issue #1859, opened 2023-03-20
Delelting instances from the instance list fails


hardillb:
### Current Behavior

If you try and delete an instance from the instance list (not the instance overview) using the entry in the kebab menu it fails and hangs at the spinner screen.

If you pick delete from the "Actions" drop down button on the overview page it works.

### Expected Behavior

Delete should work eveyrwhere

### Steps To Reproduce

- Create instance
- navigate to the instance list for a team
- select delete from kebab menu at end of row

### Environment

- FlowForge version: 1.5
- Node.js version: 16.x
- npm version: 8.x
- Platform/OS: K8s/Linux
- Browser: Chrome

knolleary:
Confirmed; doesn't look like any request is being made to the API when this option is selected.

Steve-Mcl:
confirmed on staging

PR #1879, opened 2023-03-23
fix deletion (call correct API function)

Steve-Mcl:
## Description

Fix deletion of instance from kebab menu 

NOTE: Upon deletion, the page refreshes and a 404 page is displayed (but the deletion is successful) 

FAO: @Pezmc - Upon deleting the instance, the page is refreshed causing `updateProject` to be called resulting an exception that causes the router push to `name: 'PageNotFound'`

I spent a fair while on this and even dragged @joepavitt in to assist with the VUE oddities. After another round of going in circles, we came to the conclusion this PR should simply address the issue at hand & the 404 error we are seeing will get resolved as a matter of course due to the imminent works on Applications/Instance.

## Related Issue(s)

#1859

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/project/index.vue
+++ b/frontend/src/pages/project/index.vue
@@ -264,7 +264,7 @@ export default {
         // TODO: Currently assumes 1:1 application to instance mapping
         deleteInstance () {
             this.loading.deleting = true
-            instanceApi.deleteProject(this.project.id).then(async () => {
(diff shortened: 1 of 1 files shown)


Pezmc:
I wasn't able to reproduce the 404 locally, but these pages will indeed be being re-worked.

Pezmc (approved):

Issue #1735, opened 2023-02-21
Introduce Application model to runtime


joepavitt:
### This Issue
This issue tracks *backend changes*, and front-end changes required to be inline with the latest backend.

UI only changes are being tracked in #1689.

### Description

Want to break this into a standalone task to track discussion and. final decision going forward as this is relevant for https://github.com/flowforge/flowforge/issues/1689

@ZJvandeWeg had proposed that we rename "Project" > "Application". My own thoughts are that I'm not against this, I understand the incentive behind it, and it would make the migration from Project > Application/Instance split easier than if we split Project > Project/Instance, where the new Project has a different meaning/scope to the original.

<img width="1159" alt="Screenshot 2023-02-13 at 20 00 52" src="https://user-images.githubusercontent.com/99246719/218562383-b720012e-e146-4e01-8bb1-db5b4940fc0c.png">

Diagram taken from: https://github.com/flowforge/flowforge/issues/1450#issuecomment-1428596043

### Epic/Story

#1450

Plan of action:

```[tasklist]
- [x] Introduce `Application` db model - this is a parent container of multiple `Projects`.
        `id` - use hashid like other things
        `name` - free form, no constraints
- [x] Create CRUD endpoints for working with Application
- [x] Application API end point to list Application instances (via #1813)
- [x] Create Migration to generate Application object for all existing Project objects (via #1813)
- [ ] https://github.com/flowforge/flowforge/issues/1863
- [ ] https://github.com/flowforge/flowforge/issues/1867
- [ ] https://github.com/flowforge/flowforge/issues/1868
- [x] Delete Instance - update UI to just delete instance, not application as well
- [ ] https://github.com/flowforge/flowforge/issues/1870
- [x] Delete Application - update UI to use Application endpoint for deleting
- [x] Team API end point to list Application instances
- [x] Renaming of Applications
- [x] Update top-level instances view (#1852) to use Team/Instance API
- [ ] https://github.com/flowforge/flowforge/issues/1911
- [ ] ----- review ------
- [x] Support changing Application name (API/Frontend)
```

knolleary:
Following a discussion around the backend changes needed for #1689, we have agreed, in the db model, to introduce a new `Application` object that sits about what we currently called `Projects`. This will allow us to introduce this collection object without forcing us to changing the meaning of Project throughout the code base at the same time.

Plan of action:

```[tasklist]
 - [ ] Introduce `Application` db model - this is a parent container of multiple `Projects`.
[...]

PR #1888, opened 2023-03-28
Application Model: Use applications and instances API's

Pezmc:
## Description

Hooks up the UI to use the real applications and instance API's

 - Create Application form creates real application _and_ instance objects
 - New create instance inside application form
 - Team > Overview - lists real applications
 - Applications > Overview - lists real applications
 - Application > Instances - lists real instances in an application

Due to the size of the overhaul, not _all_ previous functionality is hooked up just yet. Including:

 - Start/stop/restart of instances in Application list
 - Live reload of instance status in applications pages
 - View logs/activity logs on application level

Screenshot 2023-03-29 at 01 14 29
Screenshot 2023-03-29 at 01 14 20
Screenshot 2023-03-29 at 01 14 11
Screenshot 2023-03-29 at 01 14 06
Screenshot 2023-03-29 at 01 14 37

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/pages/instance/components/InstanceForm.vue
```diff
@@ -28,14 +43,13 @@
                 data-form="project-name"
             >
                 <template #default>
-                    Application Name
+                    Instance Domain
```
knolleary, line 46:
```suggestion
                    Instance Name
```

Lets keep calling it the name - domain doesn't feel right in this context. We can update the description text to reassert the connection between name and the sub-domain the instance is then accessed on... but I have to do that as a separate suggestion is it spans two different changes in the the diff...

frontend/src/pages/instance/components/InstanceForm.vue
```diff
@@ -28,14 +43,13 @@
...
                 </template>
                 <template
                     v-if="creatingNew"
                     #description
                 >
-                    This will be shared by the application and it's first instance<br>
-                    Please note, currently, names cannot be changed once created
+                    Please note, currently, domains cannot be changed once created
```
knolleary, line 52:
```suggestion
                    The instance name is used to access the editor so must be suitable for using in a url. It is not currently possible to rename the instance after it has been created.
```

frontend/src/pages/instance/components/InstanceForm.vue
Pezmc: @joepavitt Asked for this to be changed to domain for clarity, @joepavitt do you have further insights here?

joepavitt: > We currently happen to then use that name as part of the url to access it, but the identify of the thing is its name.

Currently, it's a domain, I want that clearly shown to the user that's it's the domain too.

Alternatively, if we do change this to "Name", then I vote we clearly include a "sample url" or something like that in the label/helper text of that field that shows it'll be their domain too.

Pezmc:
@knolleary @joepavitt Tests are finally green (lots of new coverage added too), and this is ready for review.

The PR is huge, so likely some click testing locally is going to give a better idea of how the branch is, as a reminder; there are several key features missing (listed in PR description).

There's a list of known follow ups in #1735

joepavitt:
just sanity checked this again with a fresh Netlify deploy, no edge function errors, but it is slow to load the first request after a deploy, but that seems linked to [how Netlify cache](https://answers.netlify.com/t/slow-connection-after-new-deploy-on-first-load-for-every-user/20625/6) rather than anything specific to Edge Functions

knolleary (approved):

joepavitt:
Screenshot 2023-03-31 at 16 23 07

we currently have `project` as a property on a `Device` - which I believe relates to the Instance. Is it possible to map the `Application` in too, so that we can show that on this view. I don't think it should be part of the `Device` db entry, that should stay bound to the Instance, but not sure fi there is a way, when doing the relative query for the API, to append the Application in too?

joepavitt:
> Instance Settings DevOps - Target Instance - as with 'Add Device to instance', the select box should group its entries by Application, and have the current application's instances listed first.

May leave this for now, as I suspect there will be some consideration to be made around restricting this to in-Application instances only?

All other items have been actioned in #1903

PR #1889, opened 2023-03-28
Ensure coupons are applied for all team subscriptions

### LINKED ISSUE #1788: Sign-up coupons not guaranteed to be applied following team trial
### Current Behavior

If a user reaches the create-account page with `?code=COUPON` in the url, we store that coupon in a cookie and apply it when they setup their billing on stripe (and then clear the cookie).

The cookie has a 7-day expiry. The rationale was, before we added team trials, that setting up stripe would be something they have to do very early on.

With the new Team Trial mode, this is not guaranteed to work:

 - A user is free to setup billing up to 30 days after account creation - even longer if they let the trial expire and then come back to it a week later. The cookie doesn't last that long.
 - Given this longer time frame, the user is more likely to log out, log back in, change browsers/computers etc. So we cannot assume the cookie is available.

Whilst team trials were originally intended as a replacement for offering a discount on sign-up (which is why they didn't get specific attention with the team trial work), there are cases where it would be useful to have both.

### Expected Behavior

To be able to apply a coupon when a user (eventually) sets up their billing based on the link they used to access the sign-up page.

### Steps To Reproduce

_No response_

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Fixes #1788 

## Description

The current coupon handling approach is broken. Accessing the sign-up page with `?code=BLAH` doesn't guarantee that coupon gets applied when they eventually checkout on stripe.

This PR does the following:

 - Introduces `UserBillingCode` model - this is a simple table mapping UserID to a coupon code
 - When a user signs up having accessed the page with `?code=BLAH`, we validate it is a valid code and then store the code in the `UserBillingCode` table. Note we store the code as-is - we do *not* store the stripe API id for the coupon.
 - Whenever that user creates a new Stripe checkout session we validate it is still an active/valid code and apply it. If the code is not valid, we don't try to apply it.
 - Stripped out the old cookie handling code that is no longer required. This includes some error handling when trying to apply invalid coupons - but given we now validate the coupon before trying to use it, we've removed the main failure case here.

As it stands, there is no ability to remove a user's coupon code. Given we validate it each time we try to use it, there is no issue of applying an invalid code and leaving them unable to checkout.

The most significant change in behaviour between this and previous coupon logic is that the coupon will be applied (if valid) *every time* the user sets up a billing session, not just the first time.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/migrations/20230328-01-add-userbillingcode-table.js
+++ b/forge/db/migrations/20230328-01-add-userbillingcode-table.js
@@ -1,0 +1,27 @@
+/**
+ * Add UserBillingCode table
+ */
+const { DataTypes } = require('sequelize')
+
+module.exports = {
+    up: async (context) => {
+        await context.createTable('UserBillingCodes', {
+            id: {
+                type: DataTypes.INTEGER,
+                primaryKey: true,
+                autoIncrement: true
+            },
+            code: { type: DataTypes.STRING },
+            createdAt: { type: DataTypes.DATE },
+            updatedAt: { type: DataTypes.DATE },
+            UserId: {
+                type: DataTypes.INTEGER,
+                references: { model: 'Users', key: 'id' },
+                onDelete: 'cascade',
+                onUpdate: 'cascade'
+            }
+        })
+    },
+    down: async (context) => {
+    }
+}
--- a/test/unit/forge/ee/lib/billing/index_spec.js
+++ b/test/unit/forge/ee/lib/billing/index_spec.js
@@ -183,7 +183,7 @@ describe('Billing', function () {
             })
         })
 
-        describe.only('billing codes', function () {
+        describe('billing codes', function () {
             it('creates a session with a user billing code', async function () {
                 app = await setup({
                     billing: {
--- a/test/unit/forge/ee/lib/billing/index_spec.js
+++ b/test/unit/forge/ee/lib/billing/index_spec.js
@@ -130,7 +130,7 @@ describe('Billing', function () {
                     await newTeam.addUser(user, { through: { role: Roles.Owner } })
                     should.equal(await app.db.controllers.Subscription.userEligibleForFreeTrialCredit(user, true), true)
 
-                    const result = await app.billing.createSubscriptionSession(newTeam, null, user)
+                    const result = await app.billing.createSubscriptionSession(newTeam, user)
 
                     result.should.have.property('subscription_data')
                     result.subscription_data.should.have.property('metadata')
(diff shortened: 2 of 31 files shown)


forge/db/migrations/20230328-01-add-userbillingcode-table.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Add UserBillingCode table
+ */
+const { DataTypes } = require('sequelize')
+
+module.exports = {
+    up: async (context) => {
+        await context.createTable('UserBillingCodes', {
+            id: {
```
knolleary, line 9: Ideally, this table would have `UserId` as the primary key - as each user can only have one code. But I couldn't get the Sequelize model to create the table in that way due to how it manages associations between models.

test/unit/forge/ee/lib/billing/index_spec.js
```diff
@@ -174,14 +174,39 @@ describe('Billing', function () {
...
+                const result = await app.billing.createSubscriptionSession(newTeam, user)
 
                 result.should.have.property('subscription_data')
                 result.subscription_data.should.have.property('metadata')
                 result.subscription_data.metadata.should.not.have.property('free_trial')
                 result.subscription_data.metadata.should.not.have.property('free_trial', true)
             })
         })
+
+        describe.only('billing codes', function () {
```
knolleary, line 186:
```suggestion
        describe('billing codes', function () {
```

knolleary:
Just hit a failing scenario where a user tries to setup a new subscription for an existing team. You cannot reapply the coupon for the new subscription.

knolleary:
> I'll check to see if we can detect this flag is set on the coupon and then automatically clear our record of it locally after then successfully setup stripe the first time.

We know when setting up a team subscription if we're actually setting up the subscription for an existing team (ie preview sub has been cancelled). In that case, we can check the flag on the coupon to know if its restricted to first time only and then withhold the coupon from the request if needed. This allows the coupon to stay associated with the user so that it gets applied if they create a new team.

Pezmc (approved):

forge/db/migrations/20230328-01-add-userbillingcode-table.js
knolleary: Right now we have to apply all migrations regardless - we don't have the concept of EE-only migrations. The issue being, a user could be running CE, it applies migrations (skipping this one), they then apply a license key to become EE... but the migration hasn't run.

PR #1892, opened 2023-03-29
Add support for past_due subscriptions

### LINKED ISSUE #1878: Improve reporting of failed payments
Currently if a payment fails we log the fact and take no further action. This can lead a user to think all is well.

As a quick iteration to improve this, I propose we;

1. record a count of payment failures per-team.
2. if the count is > 0, show a warning banner in the team view notifying the user:

image

## Technical details

1. Add 'paymentFailures' column to `Subscriptions`
2. When a payment failure event is received  increment the count for that subscription
3. When a payment succeeds event is received set count to `0`
4. Include this property in the billing information for a team
5. Add the banner at the team level as we do the TeamTrial banner.

We currently listen for `charge.failed` to log payment failures - but I'm not sure that's quite the one we want here. I think we should be using `invoice.paid` and `invoice.payment_failed` - but that needs some additional testing and validation.

knolleary:
Fixes #1878 

As I originally thought this was going to collide with the changes in #1889 I chose to based this work on the `fix-coupons` branch. In hindsight, I think the changes are somewhat more self-contained so don't strictly need to be based on that branch... but they both need to go in regardless.

## Description

This adds track of 'past_due' stripe subscriptions. This is the state they go into as soon as a payment has failed.

With this PR, we update our internal state to 'past_due', but we still treat that as 'active' as far as all the billing logic goes. This ensures no change in functionality at this point in time. We may choose to restrict behaviour in the future - but not as part of this PR.

This then allows us to add a banner notifying the user if there is an over due payment:

image

This also includes a related fix where, when a subscription is canceled and we suspend all projects, we weren't resetting the billing state. This prevented them being re-added to new subscription once setup.

## Related Issue(s)

#1878 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/containers/wrapper.js
+++ b/forge/containers/wrapper.js
@@ -63,6 +63,9 @@ class SubscriptionHandler {
 
         const subscription = await this.requireSubscription(project.Team)
         if (subscription.isCanceled()) {
+            // Mark this project state as 'not_billed' so that it will get
+            // added back to any future subscription
+            await this._app.billing.setProjectBillingState(project, this.BILLING_STATES.NOT_BILLED)
             this._app.log.warn(`Skipped removing project '${project.id}' from subscription for canceled subscription '${subscription.subscription}'`)
             return
         }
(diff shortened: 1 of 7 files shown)


Pezmc (commented):
Overall looks good! But the test failures are related, but this does need test coverage added!

knolleary:
Yup - looking at the tests now.

Pezmc (approved):

PR #1922, opened 2023-04-05
PostHog events & groups first pass

### LINKED ISSUE #1904: PostHog: Custom data & events review
### Epic

_No response_

### Description

Didn't quite warrant an epic, but there are multiple threads to this:

## Story 1:

**As a:** FlowForge Product Manager

**I want to:** get better insight into the engagement of our product and our product "stickiness"

**So that:** we can better understand what is keeping users in, and putting them off, particularly for Trial -> Paid customer conversions

## Story 2:

**As a:** FlowForge Product Manager

**I want to:** get better insight into how our products are used

**So that:** we can make better informed decisions on features to introduce/update/remove

## Story 3:

**As a:** Head of UX & Design at FlowForge

**I want to:** get better insight into which features are being used, and where users are spending their time within FlowForge

**So that:** we can make improvements on UX and upsell features, or remove them.

Other (See comments)

### Acceptance Criteria

- [ ] Detail which additional data points we want to be storing against a given `Person` in PostHog, e.g. "Date Joined FlowForge Cloud"
- [ ] Define full list of explicit custom events we want to track within PostHog. Currently we depend heavily on PostHog's autocapture, which is okay, and tells us when buttons are clicked, but not a guarantee that button clicks convert to instances being created, etc.
- [ ] Explore the use of [PostHog Groups](https://posthog.com/docs/product-analytics/group-analytics) to better understand how users are working with FlowForge as part of a team, rather than individual contributions, e.g. a user may never create an instance, but still be an active user of FlowForge.

joepavitt:
## Description

- New `product.js` service added to provide a centralised resource for calling PostHog. Also handles the case where no PostHog is configured or exists

Identify:
- Now when a user logs in, we also record a `ff-user: true` against the Person object, as currently we rely on `email` and `username` existing to infer this, which is reliable, but may not be guaranteed.

Events Added:
- `$ff-application-created`
- `$ff-application-deleted`
- `$ff-instance-created`
- `$ff-instance-deleted`
- `$ff-device-created`
- `$ff-device-deleted`
- `$ff-snapshot-created`
- `$ff-snapshot-deleted`
- `$ff-team-created`
    - Although, currently this is not triggered when a team is auto-created on sign-up as this is done server-side.
- `$ff-team-deleted`
- `$ff-invite-sent`
- `$ff-invite-removed`
- `$ff-invite-accepted`
- `$ff-invite-rejected`

Group Calls:
- `team` group is identified whenever the `getTeam` function is called, this then updates the PH group with number of instances, applications members, etc.
- `application` group is called upon instance creation, and currently only updated on deletion
- `instance` group is called upon instance creation, and currently only updated on deletion
- `device` group is called upon instance creation, and currently only updated on deletion

I have opted to architect this such that additional fields are required to pass to the services, e.g. team/application ids, then all of the PH logic is centralised to the services, rather than needing to be built on each page the functions/services are used.

## Related Issue(s)

Closes #1904

## Checklist

### DIFF
--- a/frontend/src/services/product.js
+++ b/frontend/src/services/product.js
@@ -1,0 +1,47 @@
+
+/**
+ *
+ * @param {String} userId - the unique identifier for the user
+ * @param {Object} set - an object containing any number of properties to bind with this user, these can be overridden
+ * @param {Object} setonce - an object containing any number of properties to bind with this user, these can never be overridden
+ */
+function identify (userId, set, setonce) {
+    window.posthog?.identify(userId, set, setonce)
+}
+
+/**
+ *
+ * @param {String} event - the name/identifier of the event
+ * @param {Object} properties - an object containing any number of properties to bind with this event, this can also contain
+ * a proeprty $set or $set_once which in turn can be an object to bind properties to the associated user
+ * @param {Object} groups - ties a given 'group' to the event. Optional keys: 'team', 'application', 'instance', 'device'
+ */
+function capture (event, properties, groups) {
+    if (!properties) {
+        properties = {}
+    }
+    if (groups) {
+        properties.$groups = groups
+    }
+    window.posthog?.capture(event, properties)
+}
+
+/**
+ *
+ * @param {String} type - team | application | instance | device
+ * @param {String} id - the associated id to this group type
+ * @param {Object} proeprties - a JSON object defining features for this group, all can be overriden
+ */
+function groupUpdate (type, id, properties) {
+    capture('$groupidentify', {
+        $group_type: type,
+        $group_key: id,
+        $group_set: properties
+    })
+}
+
+export default {
+    identify,
+    capture,
+    groupUpdate
+}
--- a/frontend/src/api/application.js
+++ b/frontend/src/api/application.js
@@ -1,9 +1,20 @@
+
+import product from '../services/product.js'
 import daysSince from '../utils/daysSince.js'
 
 import client from './client.js'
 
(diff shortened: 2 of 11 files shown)


frontend/src/services/product.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/**
+ *
+ * @param {String} type - team | application | instance | device
+ * @param {String} id - the associated id to this group type
+ * @param {Object} proeprties - a JSON object defining features for this group, all can be overriden
+ */
+function groupUpdate (type, id, properties) {
+    capture('$groupidentify', {
```
joepavitt, line 36: for context on this: https://github.com/PostHog/posthog-js/issues/600

currently, in the client-side JS library, we cannot call _just_ a `groupIdentify`, the client JS `group()` function currently sets a session variable such that _all_ future events are bound to that group, which we do not want. In speaking with the dev team at PostHog (see the linked Slack thread) they informed me that I can just do this instead.

Pezmc:
@joepavitt And I went over this PR on Friday. I'm happy for this to move forward.

joepavitt:
@Pezmc fyi - got a couple of little extras to add in given the Trial effectiveness analysis I was doing end of last week

joepavitt:
@MarianRaphael this is good to go and has sign off code-side, but want your eyes on the list above too to make sure no obvious ones missed.

RE: the project suspended, started, I'll be putting those in now

PR #1939, opened 2023-04-07
Refresh Instance state after Suspend and prevent redirect

### LINKED ISSUE #1938: Landed on the application overview after suspend instance, not the instance overview page (or staying on same instance page)

joepavitt:
## Description

No need to redirect to the "home" view after a suspension of an Instance, instead we now just reload the Instance such that we get the latest "state" to reflect "suspended"

## Related Issue(s)

Fixes #1938 

## Checklist


Pezmc:
I think https://github.com/flowforge/flowforge/pull/1926 fixes this as it adds the pending suspending state.

PR #1996, opened 2023-04-13
Redirect /project/xyz/* to /instance/xyz/*

knolleary:
## Description

Adds a redirect for all `/project/xyz` routes to `/instance/xyz` routes. This is ensures existing stacks that redirect back to `/project/xyz` are taken to the right place.


Steve-Mcl (approved):
tested locally 👍

PR #1998, opened 2023-04-13
Fix linting issues

knolleary:
## Description

Unclear how these passed the tests when introduced via #1992


joepavitt:
Screenshot 2023-04-13 at 12 21 54

which of these tasks are meant to be doing the linting checks?

joepavitt:
is this because the linting check _only_ runs when we also have server-side changes?

Issue #2004, opened 2023-04-13
Forage App unit tests take too long to run


knolleary:
### Description

Our [Default build and test](https://github.com/flowforge/flowforge/actions/workflows/build.yml) action is routinely taking 20 minutes.
Likewise our postgres tests - although they also suffer from more regularly OOM errors (#1960).

They don't take as long when running locally, but it is still many minutes.

This is only going to get worse as we write more tests (I'm told removing tests isn't the right answer....)

We need to change our approach.

Currently the typical pattern is for each individual test case to be run on a *clean* FlowForge instance. This means each test has to initialise the instance before it can be run. This setup time doesn't get reflected in the timings mocha provides for individual tests.

I've previously looked at ways to make this more efficient, for example reseting the database rather than recreating it. That had very limited success.

The main goal needs to be to minimise how many times the test suites recreate the forge app.

The general principle should be:

1. Each test suite (ie a _spec file) remains responsible for creating a forge app instance to test against.
    We must not end up needing to run the _spec files in a specific order, or being unable to run an individual _spec file because it assumes another has run before it.
3. Rather than recreating that app instance for each individual test, the tests should be able to run against the same instance.
    For many existing tests that will require them to be updated to use test-specific resource names - so it doesn't matter that a resource already exists from a previous test.
5. The tests must remain independent - even though they share a forge app instance. Minimise cases where TestB will only pass if TestA has run. That will make for brittle tests when adding new test cases in between etc.

Some tests will continue to require a clean app to test against - for example where specific platform-wide configurations are needed (eg billing config). The goal just needs to be to minimise how many times that's done.

knolleary:
Description added with a list of all spec files which can be ticked off as we chip away at them.

PR #2011, opened 2023-04-17
Use audit events to improve management of node module list

### LINKED ISSUE #2005: Cannot upgrade modules from inside Manage Palette
### Current Behavior

Tried to upgrade our posthog node-red module from `1.0.1` to `1.1.0`:

Screenshot 2023-04-14 at 07 44 00

Even after a restart, the version number did not change. I did manage to get it working by explicitly changing the version number in the Instance > Settings > Palette.

Even since then though, I still see the `module_not_allowed` errors, I see this twice because I attempted this twice. After the first attempt, this error only showed once.

### Expected Behavior

The modules should upgrade to the requested version upon restart.

### Steps To Reproduce

- Install a community module via "Manage Palette" in Node-RED
- Try to "upgrade" that module via the "Manage Palette" in Node-RED (you'll be told to restart)
- Restart the instance
- Check the version of the module inside "Manage Palette" - it has not changed and still offers the upgrade

### Environment

- FlowForge version: 1.6.0
- Node.js version: production
- npm version: production
- Platform/OS: Mac
- Browser: Chrome

knolleary:
Fixes #2005

## Description

The existing code to keep the platform's record of modules installed in an instance relied on the event triggered when Node-RED saves its node list.

That event isn't triggered when a module is upgraded - meaning the platform's record gets out of date.

The fix is to use the `nodes.install` audit event which does get triggered for the upgrade path.

I've also added support for the `nodes.remove` audit event to trigger removing modules from the list.

Whilst there, I've modified the test suite to only run setup once, rather than before each test. This only required updating the names of the projects the individual tests were creating to ensure uniqueness.

When running the tests locally, this reduced the runtime for the _spec file from 7s to 2s - a good step in the right direction for #2004

## Acceptance Criteria

On staging:
* [ ] Update an out-of-date module on Node-RED Palette manager
* [ ] Check that the list of installed modules on the instance settings in FF has updated.

## Checklist
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/controllers/Project.js
+++ b/forge/db/controllers/Project.js
@@ -1,5 +1,4 @@
 const crypto = require('crypto')
-const semver = require('semver')
 
 const { KEY_SETTINGS } = require('../models/ProjectSettings')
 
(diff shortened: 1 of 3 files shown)


Steve-Mcl (approved):
Looks sane.
Will check/update acceptance criteria to ensure specific manual operations are performed to test this on staging.

PR #2012, opened 2023-04-17
Rework app/db/controller tests to minimise app recreation

knolleary:
Part of #2004 
Rework app/db/controller tests to minimise app recreation.

Running locally, these changes reduce the time taken to run these tests from around 40s to 10s.


knolleary:
Hmm, genuine test failure with the postgres tests. Looks like its trying to reinitialise the forge app when we haven't closed it off properly. We don't see that with sqlite because it uses ephemeral in-memory databases.

knolleary:
Fix pushed

Steve-Mcl (approved):
Good stuff 👌

PR #2013, opened 2023-04-17
Reduce forge app reinitialisation in model tests

knolleary:
Part of #2004 

Reduces forge app reinit under `test/unit/forge/db/models/`

Not as productive as the controller test changes simply due to the smaller number of tests involved.

### DIFF
--- a/test/unit/forge/db/models/Device_spec.js
+++ b/test/unit/forge/db/models/Device_spec.js
@@ -5,14 +5,10 @@ describe('Device model', function () {
     // Use standard test data.
     let app
 
-    afterEach(async function () {
-        if (app) {
-            await app.close()
-            app = null
-        }
-    })
-
     describe('License limits', function () {
+        afterEach(async function () {
+            await app.close()
+        })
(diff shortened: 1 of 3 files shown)


knolleary:
I've rebased on `main` now that #2018 is merged as that was causing the tests to hang.

knolleary:
Fixed the postgres failure - hoping they run clean now

knolleary:
Postgres failed with a genuine error - not OOM. Must be a race condition as the normal tests passed. Will investigate

knolleary:
I've rebased on main again - this time picking up the latest updates on the spec file that was failing on postgres.

knolleary:
ok - postgres got past the previous point of failure then hit the OOM.

I think that's as good as we can get right now.

@Steve-Mcl all clear for you to review/merge

Steve-Mcl:
@knolleary "JavaScript heap out of memory" this time. Are you happy for me to proceed with review?

Steve-Mcl (approved):
👍

PR #2015, opened 2023-04-17
Update device/project api tests to minimise app reinitialisation

knolleary:
Updates the following two files to minimise how many times they reinitialise the forge app.
 - `test/unit/forge/routes/api/device_spec.js`
 - `test/unit/forge/routes/api/project_spec.js`

Running locally this has reduced the run time for these two suites from 1m to 20s.

It also fixes a bug discovered as part of the refactoring around retrieving a device's access token.


Steve-Mcl (approved):
👍

PR #2018, opened 2023-04-17
Ensure postoffice heartbeat interval is stopped on shutdown

knolleary:
## Description

I had noticed for a while that running individual tests locally would generally hang after the tests completed. I have also see occasional odd behaviour when running `npm run serve` around restarts hanging.

Finally tracked it down; the postoffice health-check interval wasn't been cleared on shutdown.

This fixes that.

### DIFF
--- a/forge/postoffice/index.js
+++ b/forge/postoffice/index.js
@@ -11,6 +11,11 @@ module.exports = fp(async function (app, _opts, next) {
     let poStartupCheck
     const isConfigured = EMAIL_ENABLED && (app.config.email.smtp || app.config.email.transport || app.config.email.ses)
     const mailDefaults = { from: app.config.email.from ? app.config.email.from : '"FlowForge Platform" <donotreply@flowforge.com>' }
+    app.addHook('onClose', async (_) => {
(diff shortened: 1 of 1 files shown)


forge/postoffice/index.js
```diff
@@ -11,6 +11,11 @@ module.exports = fp(async function (app, _opts, next) {
     let poStartupCheck
     const isConfigured = EMAIL_ENABLED && (app.config.email.smtp || app.config.email.transport || app.config.email.ses)
     const mailDefaults = { from: app.config.email.from ? app.config.email.from : '"FlowForge Platform" <donotreply@flowforge.com>' }
+    app.addHook('onClose', async (_) => {
+        if (poStartupCheck) {
+            clearInterval(poStartupCheck)
+        }
```
Steve-Mcl, line 17:
```suggestion
            clearInterval(poStartupCheck)
```

Only a suggestion (dont need to test `poStartupCheck` is "something" as `clearInterval` will not throw.

 
image

Steve-Mcl (approved):
1 Minor suggestion but happy to merge as is

PR #2023, opened 2023-04-19
Rework more tests to minimise app reinitialisation

knolleary:
Part of #2004 

Some more test refactoring to minimise reinitialisation of the forge app.

When running these tests locally, it reduces the runtime from 60s to 12s.

### DIFF
--- a/test/node_modules/flowforge-test-utils/forge/postoffice/localTransport.js
+++ b/test/node_modules/flowforge-test-utils/forge/postoffice/localTransport.js
@@ -19,6 +19,9 @@
     count() {
         return this.messages.length
     }
+    empty() {
+        this.messages = []
+    }
 }
 
 module.exports = {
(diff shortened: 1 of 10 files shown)


Steve-Mcl:
getting sooo much better. 6m for default test run.

Quick Question: Have you tried running the odd random, middle of the pack test on its own (using VSCode Mocha Test plugin or editing the file with `.only`)?

knolleary:
> Quick Question: Have you tried running the odd random, middle of the pack test on its own (using VSCode Mocha Test plugin or editing the file with .only)?

I haven't done that for every single test, but it is certainly one of the things I do when I need to isolate a test to figure out what its requirements are on the forge app to ensure it gets a clear start.

knolleary:
Odd that postgres is still getting a genuine test failure. I thought I had squashed that.

Steve-Mcl (approved):

PR #2025, opened 2023-04-20
Refactor more tests to minimise forge app initialisation

knolleary:
Part of #2004 

More tests refactored.

 - test/unit/forge/ee/routes/billing/index_spec.js
 - test/unit/forge/routes/api/application_spec.js
 - test/unit/forge/ee/routes/sharedLibrary/index_spec.js
 
Reduced 50s to 10s

### DIFF
--- a/test/unit/forge/ee/routes/billing/index_spec.js
+++ b/test/unit/forge/ee/routes/billing/index_spec.js
@@ -10,6 +10,9 @@ describe('Billing routes', function () {
 
     let app
     let stripe
+    let objectCount = 0
+    const generateName = (root = 'object') => `${root}-${objectCount++}`
+
     const TestObjects = {}
 
     async function getLog () {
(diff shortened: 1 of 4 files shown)


Steve-Mcl:
An odd failure in this one Nick. 

At a quick glance, it looks like an edge case rounding error?

knolleary:
@Steve-Mcl I've rerun and they have passed up to the point of memory failure.

The previous error must be a timing thing - but in a completely unrelated test. Not sure we've seen it much if at all. As it is unrelated to this PR, I'm happy to move forward as is.

Steve-Mcl:
> Not sure we've seen it much if at all. As it is unrelated to this PR, I'm happy to move forward as is.

Agreed and it is a fairly benign issue (what is one second dfif in a time duration string - in the grand scheme of things)

Happy to proceed.

Steve-Mcl (approved):

PR #2033, opened 2023-04-24
doc: Update security for instance

ZJvandeWeg:
It wasn't quite clear who would have access to the endpoint as it stated all users would.

### DIFF
--- a/docs/user/instance-settings.md
+++ b/docs/user/instance-settings.md
@@ -70,7 +70,8 @@ You can optionally enabled Basic Authentication, with a single hardcoded usernam
 and password.
 
 Alternatively you can require anyone accessing those routes to be logged into
-FlowForge.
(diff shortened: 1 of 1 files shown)


docs/user/instance-settings.md
```diff
@@ -70,7 +70,8 @@ You can optionally enabled Basic Authentication, with a single hardcoded usernam
 and password.
 
 Alternatively you can require anyone accessing those routes to be logged into
-FlowForge.
+FlowForge. The hosted pages or API endpoints are only available for users that
+have access to your team and the cloud instance.
```
ZJvandeWeg, line 74:
```suggestion
FlowForge. The hosted pages or API endpoints are only available for FlowForge users that
have access to the team on FlowForge and the cloud instance.
```

PR #2053, opened 2023-04-27
Add access token support to stats endpoint

### LINKED ISSUE #2043: Add api token scoped to just admin/stats end-point for system monitoring
### Description

**As an**: Admin user
**I want to**: enable scripted access to the stats admin endpoint, without providing full admin access to the platform
**So that**: I can securely monitor the system.

---

Our current ways to monitor the platform involve scanning logs, relying on stripe callbacks and other tricks that have to be setup manually.

The issue with the stripe callback notifications is that it only applies once billing is setup in a team - which is no longer something every user has to do when signing up to FF Cloud.

But more generally, it should be easier to monitor the current state of the platform.

We already have in place with the `/api/v1/admin/stats` end-point that can be used to monitor basic stats of the platform. However this endpoint is protected by admin level access.

**An Admin user should be able to create a long-lived access token that provides read-only access to the stats end-point and nothing else.**

We have an item in the backlog to support users creating API tokens - #14. That is a far more general solution (which we will need eventually), but this issue is a quick iteration to enable much better monitoring integration.

knolleary:
Closes #2043 

## Description

This PR adds the ability to monitor the platform from tools such prometheus.

It has ended up touching lots of files, but I have structured the commits to make it easier to review.

Highlevel description:

1. Enables token-based access to the admin stats endpoint
2. Fixes up how admin access to api endpoints is managed
3. Adds api endpoint to generate/delete a dedicated stats token
4. Updates Admin Settings to allow admin to enable/disable the token
5. Adds support for openmetrics-text format response in the stats end-point

Commits in order:

1. [Ensure all admin routes have explicit permissions declared](https://github.com/flowforge/flowforge/commit/4d81646fcb436d649693ef805a98b5c737fb16cc) - we previously used `verifyAdmin` as a catch-all preHandler to ensure a request comes from an admin. But to support tokens with reduced scope, we cannot do that. So I've added explicitly `needsPermission` preHandler to all admin routes. To go alongside this, I've added much better Unit Test coverage of `needsPermission` to cater for the growing number of scenarios.
2. [Add support for OpenMetrics format on /api/v1/admin/stats](https://github.com/flowforge/flowforge/commit/630fd95993d49748e9332137b2d8188abe25e1b4) - return the new format if the `accept` header requests it
3. [Add team billing info to stats](https://github.com/flowforge/flowforge/commit/fea897331cfe89fa1d619abb5ee2e133bc095abd) - include counts of subscriptions by state. Useful to get a sense of trial/non-trial etc
4. [Add api endpoints to manage platform stats token](https://github.com/flowforge/flowforge/commit/b68ea273964a980b191c24ab4b1c0d58ab820fd6) - basic endpoints to generate/delete the token, inc unit tests
5. [Make FormRow description field more consistent](https://github.com/flowforge/flowforge/commit/d193b8536372fb715d7cdd16902d7c47a2b3a1b6) - noticed inconsistent css for the 'description' field in FormRow
6. [Add platform stats option to Admin Settings page](https://github.com/flowforge/flowforge/commit/1850f3557182cae01061b8202d2add5790401655) - Adds UI to admin settings to manage the stats token
7. [Add docs on platform monitoring](https://github.com/flowforge/flowforge/commit/8847f4ef802e9b8d72af24959437501953d7545c) - doc updates

## Checklist

    - [ ] Upgrade instructions
    - [x] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/api/admin.js
+++ b/forge/routes/api/admin.js
@@ -83,7 +83,7 @@ module.exports = async function (app) {
      */
     function convertToOpenMetrics (stats) {
         const result = flattenObject('flowforge', stats)
-        const lines = Object.entries(result).map(m => `${m[0]} ${m[1]}`)
+        const lines = Object.entries(result).map(([key, value]) => `${key} ${value}`)
         return lines.join('\n') + '\n'
     }
 
--- a/forge/routes/api/admin.js
+++ b/forge/routes/api/admin.js
@@ -193,4 +193,13 @@ module.exports = async function (app) {
         const result = app.db.views.AuditLog.auditLog(logEntries)
         reply.send(result)
     })
+
+    app.post('/stats-token', { preHandler: app.needsPermission('platform:stats:token') }, async (request, reply) => {
+        const token = await app.db.controllers.AccessToken.generatePlatformStatisticsToken(request.session.User)
+        reply.send(token)
+    })
+    app.delete('/stats-token', { preHandler: app.needsPermission('platform:stats:token') }, async (request, reply) => {
+        await app.db.controllers.AccessToken.removePlatformStatisticsToken()
+        reply.send({ status: 'okay' })
+    })
 }
--- a/forge/routes/api/admin.js
+++ b/forge/routes/api/admin.js
@@ -1,8 +1,6 @@
 
 module.exports = async function (app) {
-    app.addHook('preHandler', app.verifyAdmin)
-
-    app.get('/stats', async (request, reply) => {
+    app.get('/stats', { preHandler: app.needsPermission('platform:stats') }, async (request, reply) => {
         const userCount = await app.db.models.User.count({ attributes: ['admin'], group: 'admin' })
         const projectStateCounts = await app.db.models.Project.count({ attributes: ['state'], group: 'state' })
         const license = await app.license.get() || app.license.defaults
@@ -30,15 +28,15 @@ module.exports = async function (app) {
             result.projectCount += projectState.count
             result.projectsByState[projectState.state] = projectState.count
         })
-
         reply.send(result)
     })
 
-    app.get('/license', async (request, reply) => {
+    app.get('/license', { preHandler: app.needsPermission('license:read') }, async (request, reply) => {
         reply.send(app.license.get() || {})
     })
 
     app.put('/license', {
+        preHandler: app.needsPermission('license:edit'),
         schema: {
             body: {
                 type: 'object',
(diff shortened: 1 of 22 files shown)


forge/routes/api/shared/users.js
```diff
@@ -38,6 +38,21 @@ module.exports = {
         /** @type {UserController} */
         const userController = app.db.controllers.User
 
+        if (!isAdmin) {
+            let resp
+            // Check if this is trying to modify anything on admin users can touch
```
Pezmc, line 43:
```suggestion
            // Check if this is trying to modify anything only admin users can touch
```

forge/routes/api/shared/users.js
```diff
@@ -38,6 +38,21 @@ module.exports = {
...
+            if (Object.hasOwn(request.body, 'email_verified') && user.email_verified !== request.body.email_verified) {
+                resp = { code: 'invalid_request', error: 'cannot verified own email' }
```
Pezmc, line 45:
```suggestion
                resp = { code: 'invalid_request', error: 'cannot verify own email' }
```

forge/routes/api/admin.js
```diff
@@ -30,15 +28,84 @@ module.exports = async function (app) {
...
+                result.teamsByBillingState[teamState.status] = teamState.count
+            })
+        }
+        return result
+    }
 
-        reply.send(result)
+    /**
+     * Converts a JSON Object to a flattened object with key names more aligned
+     * with OpemMetrics format
```
Pezmc, line 43:
```suggestion
     * with OpenMetrics format
```

forge/routes/api/admin.js
```diff
@@ -30,15 +28,84 @@ module.exports = async function (app) {
...
+        return result
+    }
+    /**
+     * Stringifies a JSON Object to OpenMetrics format
+     * @param {Object} stats the object to conver
+     * @returns a OpenMetrics formatted version of the object
+     */
+    function convertToOpenMetrics (stats) {
+        const result = flattenObject('flowforge', stats)
+        const lines = Object.entries(result).map(m => `${m[0]} ${m[1]}`)
```
Pezmc, line 86:
```suggestion
        const lines = Object.entries(result).map(([key, value]) => `${key} ${value}`)
```

frontend/src/pages/admin/Settings/General.vue
```diff
@@ -194,6 +244,23 @@ export default {
...
+        this.platformStatsToken = ''
+    },
+    watch: {
+        platformStatsTokenEnabled: function (v) {
+            if (this.platformStatsToken === null) {
+                // This is the initial setting of the value - ignore it
+                this.platformStatsToken = ''
+                return
+            }
+            if (v) {
```
Pezmc, line 258:
```suggestion
        platformStatsTokenEnabled: function (newValue) {
            if (this.platformStatsToken === null) {
                // This is the initial setting of the value - ignore it
                this.platformStatsToken = ''
                return
            }
            if (newValue) {
```

Pezmc (approved):

PR #2054, opened 2023-04-27
"Add Instance" button on /instances

### LINKED ISSUE #2052: "Add Instance" on /instances view
### Description

Screenshot 2023-04-26 at 08 54 26

is showing heavy traffic into /instances. Seems users are using this as their main point of entry, rather than /applications. To create a new instance right now, you'd need to select the application row. We see about 30% of traffic going from /applications to an application, so it does suggest people are using this functionality, but heavier traffic is going to /instances.

### Epic/Story

_No response_

joepavitt:
## Description

- "Add Instance" button put onto `/instances`
- New `/instances/create` page - which uses the `InstancesForm.vue`
- `InstanceForm` now has the option to pass `applications` array and `applicationSelection` boolean which will display the dropdown, with provided options. Needed to pass the bool, as it's valid to not pass `applications`, so that check on it's own to fall back ot the text input, wasn't enough.
- Adds E2E tests for instance creation & checking the reidrect.

## Related Issue(s)

Closes #2052

## Checklist

### DIFF
--- a/frontend/src/pages/team/createInstance.vue
+++ b/frontend/src/pages/team/createInstance.vue
@@ -127,7 +127,6 @@ export default {
 
             try {
                 const instance = await this.createInstance(applicationId, instanceFields, copyParts)
-                console.log(instance)
 
                 await this.$store.dispatch('account/refreshTeam')
 
--- a/frontend/src/pages/team/createInstance.vue
+++ b/frontend/src/pages/team/createInstance.vue
@@ -1,0 +1,162 @@
+<template>
+    <Teleport
+        v-if="mounted"
+        to="#platform-sidenav"
+    >
+        <SideNavigation>
+            <template #options>
+                <a @click="$router.back()">
+                    <nav-item
+                        :icon="icons.chevronLeft"
+                        label="Back"
+                    />
+                </a>
+            </template>
+        </SideNavigation>
+    </Teleport>
+    <main>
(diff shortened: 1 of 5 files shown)


frontend/src/pages/team/createInstance.vue
```diff
@@ -0,0 +1,161 @@
...
+        this.mounted = true
+    },
+    methods: {
+        async handleFormSubmit (formData, copyParts) {
+            this.loading = true
+
+            // Drop applicationId & applicationName from the payload
+            const { applicationId, applicationName, ...instanceFields } = formData
+
+            // if we have applicationName, a new application needs to be created first
```
Pezmc, line 126: What's this referencing?

test/e2e/frontend/cypress/tests/instances.spec.js
```diff
@@ -28,7 +37,7 @@ describe('FlowForge - Instances', () => {
         cy.wait('@getInstanceTypes')
     })
 
-    it('can be navigated to from team page', () => {
+    it.skip('can be navigated to from team page', () => {
```
Pezmc, line 40: To be removed?

joepavitt: woops - forgot to unskip them - thanks

Pezmc (approved):

Issue #1937, opened 2023-04-07
Devices page re-design to give a dashboard


joepavitt:

joepavitt:
First pass:

<img width="1282" alt="Screenshot 2023-04-13 at 14 33 18" src="https://user-images.githubusercontent.com/99246719/231775062-770158c6-65b4-4556-a7b8-5f0ae31a34d5.png">

<img width="1293" alt="Screenshot 2023-04-13 at 14 33 30" src="https://user-images.githubusercontent.com/99246719/231775130-892e0c53-1e3b-4c9f-9ed6-4fa49a62668b.png">

joepavitt:
@MarianRaphael fyi - first pass of this, _should_ be a fairly quick development effort if we want to squeeze a first iteration into 1.7

joepavitt:
Easy follow-on then becomes having filters for last known running status too, and possibly even deployed snapshots.

joepavitt:
Quick iteration with two bars, and the option to "hide" a given visualisation if they're too tall.

<img width="1314" alt="Screenshot 2023-04-13 at 14 39 53" src="https://user-images.githubusercontent.com/99246719/231776892-e2a0d167-784e-43a2-9090-42538cbb41bf.png">

<img width="1318" alt="Screenshot 2023-04-13 at 14 39 48" src="https://user-images.githubusercontent.com/99246719/231776886-403cc94c-205c-42aa-a443-b164aba9da05.png">

joepavitt:
Have opened #2113 to cover the API modifications. Can close this issue, as  for up to 1,000 devices, this is feature complete.

PR #2058, opened 2023-04-27
Add "Last Seen" chart for filtering Devices

joepavitt:
## Description

- Moves some of the Device Status bucketing logic into a service, as this is now shared across multiple screens. This contains logic to bucket to a css class & label pair.
- Adds a new "DeviceStatusBar" component which is driven by a `property` and list of all `devices`.

https://user-images.githubusercontent.com/99246719/235135164-5c2b9611-b244-4181-8234-a6d8d634e28c.mov

### Considerations:

- Driven by the existing devices API call, which will hit device pagination, may need separate API call. Currently, we are dependent upon having the device id's in order to set the filter at the top level. This is because (in the example of `status` this is not a property on the `device` object directly, but computed within the filtering)

## Related Issue(s)

First part of #1937

## Checklist

 - [x] E2E Tests

### DIFF
--- a/frontend/src/components/DevicesBrowser.vue
+++ b/frontend/src/components/DevicesBrowser.vue
@@ -16,7 +16,7 @@
             message="Deleting Device..."
         />
         <template v-else>
-            <DevicesStatusBar :devices="Array.from(devices.values())" property="lastSeenSince" @filter-selected="applyFilter" />
+            <DevicesStatusBar :devices="Array.from(devices.values())" property="status" @filter-selected="applyFilter" />
             <ff-data-table
                 v-if="devices.size > 0"
                 data-el="devices-browser"
--- a/frontend/src/components/DevicesBrowser.vue
+++ b/frontend/src/components/DevicesBrowser.vue
@@ -16,7 +16,8 @@
             message="Deleting Device..."
         />
         <template v-else>
-            <DevicesStatusBar :devices="Array.from(devices.values())" property="status" @filter-selected="applyFilter" />
+            <DevicesStatusBar data-el="devicestatus-lastseen" label="Last Seen" :devices="Array.from(devices.values())" property="lastseen" :filter="filter" @filter-selected="applyFilter" />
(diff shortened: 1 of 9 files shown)


frontend/src/components/DevicesBrowser.vue
```diff
@@ -317,6 +323,17 @@ export default {
 
             return columns
         },
+        filteredDevices () {
+            let filteredDevices = []
+            if (!this.filter) {
+                filteredDevices = Array.from(this.devices.values())
+            } else {
+                filteredDevices = Array.from(this.devices.values()).filter((d) => {
+                    return this.filter.devices.includes(d.id)
```
Pezmc, line 332: By converting the map to an array, then looping over the array searching `this.filter.devices` each time for each ID, this will end up `N^2`.

Given this.devices is already a `Map`, something like the below will be significantly faster:

```javascript
this.filter.devices.forEach((device) {
  filteredDevices.push(this.devices.get(device.id))
})
```

frontend/src/components/DevicesBrowser.vue
```diff
@@ -342,6 +359,9 @@ export default {
         clearInterval(this.checkInterval)
     },
     methods: {
+        applyFilter (filter) {
+            this.filter = filter
```
Pezmc, line 363: I found the spec of this `filter` object unclear, and had to dig into `DeviceStatusBar` to understand how it works. 

Does the DevicesBrowser need the `this.property`, `bucket` on top of devices? Or could the logic for those two properties remain within the two DeviceStatusBars, making the event something like:

```
applyFilter (devices) {
            // logic to filter this.devices to only include devices
}
```

Pezmc (changes requested):

frontend/src/components/DevicesBrowser.vue
joepavitt: > Or could the logic for those two properties remain within the two DeviceStatusBars

The reason for this was for when a user switches between two filters, the original filter needs to be reset. The parent (`DevicesBrowser`) just acts as a broker to communicate the `property` and `bucket` data between the different `DeviceStatusBar` components

I will add comments accordingly to make this more explicit

Pezmc (approved):

PR #2072, opened 2023-05-02
Add instructions to run agent as a service on a Pi

robmarcer:
## Description

Added details of how to run the FlowForge device agent as a service on a Raspberry Pi

### DIFF
--- a/docs/user/devices.md
+++ b/docs/user/devices.md
@@ -215,3 +215,41 @@ device, and then copy the modules on to your device.
 <br>
 
 [^global-install]: Starting the agent via the command `flowforge-device-agent` assumes it was installed as a global npm module and your path is properly configured to pick that up.
+
+## Running the device agent as a service on a Raspberry Pi
+
+Can can run the device agent as a service, this means it can run in the background and be enabled to automatically start on boot.
+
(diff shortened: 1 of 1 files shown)


docs/user/devices.md
```diff
@@ -215,3 +215,41 @@ device, and then copy the modules on to your device.
...
+
+Can can run the device agent as a service, this means it can run in the background and be enabled to automatically start on boot.
+
+### Creating a Service File
+
+The first step is creating the systemd unit file for your service. You can start by creating a new file in the /etc/systemd/system directory with a .service file extension:
+
+```sudo nano /etc/systemd/system/flowforge-device-agent.service```
+
+In the service file, you'll need to define the following parameters:
```
hardillb, line 229: Are the parameters missing here?

hardillb (commented):
Mostly looks good, just thing there is the bit missing in as flagged

docs/user/devices.md
robmarcer: Ah sorry, I get you. I've removed the parameters as they were duplicated but I didn't remove the title. Fixing now.

hardillb (approved):

PR #2118, opened 2023-05-10
Add deviceEditor feature flag in place of license check

knolleary:
## Description

The existing license check only works for admin users as regular uses don't get visibility of that setting.

This PR switches over to a deviceEditor feature flag which is how we manage access to licensable features elsewhere in the UI.


Steve-Mcl (approved):
good catch

PR #2121, opened 2023-05-10
Do not enable device editor if comms are unavailable

knolleary:
## Description

If running without comms (ie no broker config), but with a license applied, FF doesn't start up as the deviceEditor feature assumes comms will be available.

We guard against that on other EE features so should do so here.


Pezmc (approved):
Thanks, this fixes my local issue

PR #2149, opened 2023-05-15
Docs: Add additional SSO config notes on signed responses

knolleary:
## Description

With FF 1.7, the underlying SAML library now defaults to the more secure configuration of requiring both SAML assertions and responses to be signed.

This updates the docs for google/azure SSO to highlight where this option needs to be enabled in their SAML configuration.


joepavitt (approved):
approved subject to tests passing

Issue #1646, opened 2023-02-02
Autosaving Env Vars


joepavitt:
### Current Behavior

Not _technically_ a bug, but an annoying feature nonetheless. When defining Environment Variables on a project, you have to hit both the "+" _and_ the "Save Settings".

I'm assuming this is linked to the fact we could delete an env var by mistake, and ensuring the "Save Settings" is clicked backs this up, but it can be unintuitive when adding environment variables.

### Expected Behavior

I propose one of two alternatives:

1. (preferred) Auto save/update when a key is added or deleted to that table, with confirmation alerts. Where deleting should probably provide an "Are you sure?" dialog warning.
2. Disable the "Save Settings" button until changes have been made, to make it clearer to the user that the button must be clicked

### Steps To Reproduce

https://user-images.githubusercontent.com/99246719/216337871-faebb214-0ca3-4ee7-a80d-7e42cb86a9e5.mov

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

joepavitt:
It's a fairly quick win, so I could squeeze this into 1.4 as part of the work I'm doing with pseudo-deployments, but want to check for any gotchas I may have not considered?

MarianRaphael:
Idea: Popup/Hint asking do you want to restart

MarianRaphael:
Verified on staging

PR #2175, opened 2023-05-23
Improve env variable editing uxp

Steve-Mcl:
## Description

* Ensure save button is disabled/enabled accordingly
* Prompt user upon navigation if there are unsaved changes
* update existing toast on instance edit - to inform user of when changes will be applied (as per @MarianRaphael request)
* add toast to device env var editing (to match instance env var editing) - to inform user of success and when the changes will be applied (as per @MarianRaphael request)

## Related Issue(s)

#1646 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/layouts/Platform.vue
+++ b/frontend/src/layouts/Platform.vue
@@ -15,7 +15,7 @@
                                        :countdown="a.countdown || 3000" @close="clear($index)"></ff-notification-toast>
             </TransitionGroup>
             <interview-popup v-if="interview?.enabled" :flag="interview.flag" :payload="interview.payload"></interview-popup>
-            <ff-dialog ref="dialog" data-el="platform-dialog" :header="dialog.header" :kind="dialog.kind" :disable-primary="dialog.disablePrimary" :confirm-label="dialog.confirmLabel" @cancel="clearDialog" @confirm="dialog.onConfirm">
+            <ff-dialog ref="dialog" data-el="platform-dialog" :header="dialog.header" :kind="dialog.kind" :disable-primary="dialog.disablePrimary" :confirm-label="dialog.confirmLabel" @cancel="clearDialog(true)" @confirm="dialog.onConfirm">
                 <p v-if="dialog.text">{{ dialog.text }}</p>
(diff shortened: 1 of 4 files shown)


Steve-Mcl:
@MarianRaphael requested your review in absence of Joe and Pez. Feel free to re-assign.

MarianRaphael (approved):
Looks good to me, but let's wait for Pez's review

frontend/src/layouts/Platform.vue
```diff
@@ -103,15 +104,20 @@ export default {
                 this.dialog.disablePrimary = msg.disablePrimary
             }
             this.dialog.onConfirm = onConfirm
+            this.dialog.onCancel = onCancel
         },
-        clearDialog () {
+        clearDialog (cancelled) {
```
Pezmc, line 109: `cancelled` always seems to be set to true, I'd suggest removing it and updating the card to only if onCancel is defined.

Steve-Mcl: > cancelled always seems to be set to true

That is correct however, `clearDialog(false)` should be called upon `onConfirm` (IMO) - so this parameter is present for "correctness".  The change to call `clearDialog in `onConfirm` is out of scope tho at this point.

frontend/src/layouts/Platform.vue
```diff
@@ -103,15 +104,20 @@ export default {
...
+            if (cancelled && this.dialog.onCancel) {
+                this.dialog.onCancel()
+            }
```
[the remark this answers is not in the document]
Steve-Mcl:
```suggestion
        clearDialog (cancelled) {
            if (cancelled) {
                this.dialog.onCancel?.()
            }
```

Counter proposal. Keeping `cancelled` and using optional chaining. (see previous comment about keeping `cancelled` param)

frontend/src/layouts/Platform.vue
Pezmc: I think this falls under YAGNI and would suggest dropping it for now given it's not in use, but not going to block.

PR #2193, opened 2023-05-30
Experiment: Mass minor upgrade of all packages

Pezmc:
Minor upgrade of all packages.

Of note:

 - Minor version only, if upstream packages are following semver correctly this should be backward compatible
 - Due to lack of package lock, in most cases, we're likely running these versions already


Pezmc:
Rebased onto the latest copy of main

Pezmc:
@knolleary Given the tests have passed, some rudimentary click testing locally of areas of the codebase that these packages are used hasn't brought up issues, that the next release isn't for a few weeks giving us plenty of time to test on staging, and we can trust most packages to stick to sever (minor version bump only); this PR is almost as "safe" as it will get; and I propose we merge this PR as is. 

Once merged and deployed, we can highlight to the team at the dev meeting to keep an eye out for any odd behaviour on staging or locally in the mean time to get extra QA.

knolleary (approved):

PR #2198, opened 2023-05-30
Handle missing stripe subscription when viewing billing

knolleary:
## Description

If a Stripe subscription is cancelled from within Stripe, we won't know about it. In theory the stripe callbacks will have notified us, but when developing locally, it is entirely possible your dev instance wasn't running when it happens.

This is a bit of an edge case, but once you hit it, you're stuck.

The only way to fix it is to create a new subscription for the team. However, the existing logic around the billing page couldn't handle this scenario - it fails to load because of the 404 from stripe under the covers.

This PR catches the error from stripe and reports it back in our format rather than Stripe's.

The Billing page handles the 404 and reverts back to the 'you need to setup billing' view if it is hit. This gives the user a path to creating a new subscription.

This new subscription will *not* have any of the existing instances listed. This is a gap we've identified previously - and this PR doesn't attempt to address it.

### DIFF
--- a/forge/ee/routes/billing/index.js
+++ b/forge/ee/routes/billing/index.js
@@ -282,34 +282,40 @@ module.exports = async function (app) {
             return response.code(404).send({ code: 'not_found', error: 'Team does not have a subscription' })
         }
 
-        const stripeSubscriptionPromise = stripe.subscriptions.retrieve(
-            sub.subscription,
-            {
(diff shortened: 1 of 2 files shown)


Pezmc:
This was _way_ easier to review with [?w=1](https://github.com/flowforge/flowforge/pull/2198/files?w=1)

PR #2205, opened 2023-05-31
Update the Pipeline UX for editing/deleting

joepavitt:
## Description

After the developer demo, we had some feedback on the UX around editing/deleting Pipelines. This addresses those concerns.

- "Edit" icon for changing name with Save/Cancel options
- Trash/Delete Icon always visible
- "Add Stage" always visible
- No more cog icon

https://github.com/flowforge/flowforge/assets/99246719/82e28aa0-89f6-4f71-85f9-788c894282db

fyi @Steve-Mcl 

## Related Issue(s)

Part of https://github.com/flowforge/flowforge/issues/2124

## Checklist

### DIFF
--- a/frontend/src/components/pipelines/PipelineRow.vue
+++ b/frontend/src/components/pipelines/PipelineRow.vue
@@ -56,7 +56,7 @@
 </template>
 
 <script>
-import { ChevronRightIcon, CogIcon, PencilAltIcon, TrashIcon } from '@heroicons/vue/outline'
+import { ChevronRightIcon, PencilAltIcon, TrashIcon } from '@heroicons/vue/outline'
 
 import ApplicationAPI from '../../api/application.js'
 import Alerts from '../../services/alerts.js'
@@ -68,7 +68,6 @@ export default {
     name: 'PipelineRow',
     components: {
         ChevronRightIcon,
-        CogIcon,
         PencilAltIcon,
         TrashIcon,
         PipelineStage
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
oh, me likey.

one very minor point - the mouse miles to the save/cancel buttons is a bit much (IMO) & the use of text buttons instead of icon buttons is a wee bit inconsistent. 

Otherwise, very very slick.

I will give you a moment to respond to the above and then do a proper review.

frontend/src/components/pipelines/PipelineRow.vue
```diff
@@ -85,7 +90,8 @@ export default {
     data () {
         const pipeline = this.pipeline
         return {
-            editing: false,
+            editingName: false,
+            pipelineName: pipeline.name,
```
Pezmc, line 94: Could this go under `input` as per our usual pattern? It threw me slightly that pipeline was here twice.
```suggestion
            input: {
                pipelineName: pipeline.name,
            },
```

joepavitt:
> the mouse miles to the save/cancel buttons is a bit much (IMO) 

Fair point, although for now, going to leave as it is.

> the use of text buttons instead of icon buttons is a wee bit inconsistent.

I do understand the hesitation here. Having the buttons in the Pipeline action banner felt very congested, so have gone for icons for now. Can consider how this design pattern persists across other areas of FlowForge too.

joepavitt:
Also fixed the edit stage functionality being lost given the cog was removed. This is now always shown by default

Pezmc (approved):
👍 but will not merge to give @Steve-Mcl a moment to reply

Steve-Mcl:
I understand the points and think it is very slick guys.

🚀

PR #2207, opened 2023-05-31
Add endpoint for setting ha configuration on an instance

knolleary:
## Description

This adds an EE only end point for updating an instance's HA configuration:

 - `GET /api/v1/projects/:projectId/ha` - returns the current HA config
 - `PUT /api/v1/projects/:projectId/ha` - updates the current HA config
 - `DELETE /api/v1/projects/:projectId/ha` - deletes the current HA config

The response from GET is:

 - if HA is enabled on the instance: `{ replicas: 2 }`
 - if HA is *not* enabled: `{ }`

The update end point (`PUT`...) expects an object of the form: `{ replicas: 2}`. It will *only* accept a value of `2` for the replica count in this iteration.

### DIFF
--- a/forge/db/models/Application.js
+++ b/forge/db/models/Application.js
@@ -2,9 +2,9 @@
  * An application definition
  * @namespace forge.db.models.Application
  */
-const { DataTypes } = require('sequelize')
+const { DataTypes, Op } = require('sequelize')
 
-const { KEY_SETTINGS } = require('./ProjectSettings')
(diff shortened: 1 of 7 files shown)


hardillb:
Test failures are the same blank lines as with the patch release.

hardillb (approved):
Approved, but needs opening blank lines removing from files to pass tests

knolleary:
I'll rebase both branches

PR #2224, opened 2023-06-05
Disable useless-return-rule

Pezmc:
There are lint failures in main after #2192 was merged. Some of these changes are from https://github.com/flowforge/flowforge/pull/2207, but some have been in main for months, it's not clear why they didn't fail in #2192.


hardillb:
Postgres test failed with OOM

Issue #2156, opened 2023-05-19
Instance Load Balancing


MarianRaphael:
## Epic
https://github.com/flowforge/flowforge/issues/1678

## Description
Allow horizontal scaling of n Node-RED instances in a FlowForge environment, distribute incoming (HTTP) requests between instances in a round-robin manner.

### Constraints
- Editor disabled 
- Flows have to be stateless or store all states in a separate storage layer, we have to inform the user about this fact and provide them with examples of how to achieve it.

### User Story
As a FlowForge customer, I want to leverage instance load balancing.
This will allow me to run business-critical processes within Node-RED and ensure that they are always available and can handle increasing workloads.

### Assumption
- Flows for business critical processes will not be developed in the target system / instance. 
- It is assumed that we can effectively communicate the significance of stateless functions to the user and ensure that they understand the concept and use it correctly.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

---

```[tasklist]
### Backend Tasks
- [x] HA: Add HA EE feature flag
- [x] HA: Create multiple replicas of instance based on ha setting
- [x] HA: Add HA flag to create instance endpoint
- [x] HA: Add endpoint for modifying HA settings
- [x] HA: Add HA flag to instance view
- [ ] https://github.com/flowforge/flowforge/issues/2260
- [ ] https://github.com/flowforge/flowforge-nr-launcher/issues/121
- [ ] https://github.com/flowforge/flowforge/issues/2261
- [ ] https://github.com/flowforge/flowforge-nr-project-nodes/issues/28
- [ ] https://github.com/flowforge/flowforge/pull/2225
- [ ] https://github.com/flowforge/flowforge/issues/2262
```
```[tasklist]
### UX Tasks
- [x] HA UX: Add option in Instance Settings to enable HA
- [x] HA UX: Add HA indicator on Instance overview page
- [x] HA UX: Disable editor button if HA is enabled
```
```[tasklist]
### Documentation Tasks
- [x] HA: Add HA feature description to docs
- [x] HA: Guidance on creating HA-capable flows
```

knolleary:
We need to decide how HA (in whatever form) is exposed to the user as a choice, and how that relates to billing. Whilst there's a lot of business decisions wrapped in that which could evolve, it has a specific impact on how we choose to implement it from the start. I want to make sure our initial iteration is pointed in the right direction.

[...]

PR #2225, opened 2023-06-05
Add shared subscriptions to broker ACL for project nodes

knolleary:
## Description

Part of #2156 

See also https://github.com/flowforge/flowforge-nr-project-nodes/issues/28

Adds support for using shared subscription topics on certain project-node topics.

Given we need to support both shared and non-shared versions of the topics, rather than copy/paste the rules, this adds a flag to indicate a topic can be handled as either plain or shared.

The share group id is required to the project/instance id.

It also adds a new rule for the response topic, to provide the ability for individual replicas to have a randomly generated component of the response topic.

## Checklist

### DIFF
--- a/forge/comms/aclManager.js
+++ b/forge/comms/aclManager.js
@@ -66,6 +66,8 @@ module.exports = function (app) {
         }
     }
 
+    const SHARED_SUB = /^\$share\/([^/]+)\/(.*)$/
+
     const ACLS = {
         forge_platform: {
             sub: [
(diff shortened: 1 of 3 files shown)


knolleary:
@Steve-Mcl ready for review - tested with updated projects nodes. Will have a PR with those updates shortly.

knolleary:
@Steve-Mcl this side is good to review and merge so we can get onto staging. The outstanding work is in the project nodes.

Steve-Mcl:
Will do as my next task (aprox 10 mins)

PR #2307, opened 2023-06-16
Guard against a container not existing and still allow removal

Pezmc:
## Description

I ran against this issue locally where removal failed half way through (removed from docker but not FlowForge).
This then left my account in a state where I could not delete the instance as it had already been removed from docker.

It's _unlikely_ this would happen in production, but adding a guard just in case doesn't hurt!

## Related Issue(s)

None

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/containers/stub/index.js
+++ b/forge/containers/stub/index.js
@@ -156,7 +156,7 @@ module.exports = {
                 }, 250)
             })
         } else {
-            throw new Error({ error: project.id + ' not found' })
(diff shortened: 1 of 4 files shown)


hardillb (approved):

knolleary (approved):
Not sure why this hasn't been merged already - doing so

PR #2318, opened 2023-06-20
Improve error handling and UI for launcher "offline"

### LINKED ISSUE #1083: Instance Logs page doesn't handle errors well
### Current Behavior

If the project launcher has crashed, then the polling request to get the latest logs gets an `ECONNREFUSED` error.

If the user is viewing the project logs page, the page is polling the backend for those logs - and the underlying ECONNREFUSED is getting passed back to the user... each time it polls

image

### Expected Behavior

Better reporting of the error to the user without an explosion of notifications

 - We should spot the ECONNREFUSED error and translate it to something more consumable when returning it to the frontend
 - Pages that poll the API need a better strategy for reporting errors on each poll. Multiple toast notifications aren't pleasant.

### Steps To Reproduce

Manually kill the nr-launcher process for a project whilst viewing its logs page

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

Steve-Mcl:
closes #1083 

## Description

* Curates the error responses from the launcher to provide better info to the frontend (prevents the default technical "catch all" Error 500 toast notification)
* Makes the toast message more friendly and ONLY displays it once
* Adds a friendly / less alarming banner above the log panel to inform users that the logs are currently unavailable
* Stylises the log panel to indicate it may be stale
* Adds a timer `mixin` for controlling polling
  * This was something discussed in a meeting some time back. This component can be (should be) reused wherever polling is required as it can be easily controlled and automatically stops when component is unmounted etc.
 

https://github.com/flowforge/flowforge/assets/44235289/92d607e8-6bcc-4b9e-9d37-cc3dc427749a

## Related Issue(s)

#1083

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -820,7 +820,9 @@ module.exports = async function (app) {
      *  - returns most recent 30 entries
      *  - ?cursor= can be used to set the 'most recent log entry' to query from
      *  - ?limit= can be used to modify how many entries to return
+     * @name /api/v1/projects/:id/logs // << Added an 's'
      * @name /api/v1/projects/:id/log
+     * @name /api/v1/projects/:id/logs // << Added an 's'
      * @memberof forge.routes.api.project
      */
     app.get('/:projectId/logs', {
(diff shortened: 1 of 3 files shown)


forge/routes/api/project.js
```diff
@@ -820,7 +820,9 @@ module.exports = async function (app) {
      *  - returns most recent 30 entries
      *  - ?cursor= can be used to set the 'most recent log entry' to query from
      *  - ?limit= can be used to modify how many entries to return
+     * @name /api/v1/projects/:id/logs // << Added an 's'
      * @name /api/v1/projects/:id/log
+     * @name /api/v1/projects/:id/logs // << Added an 's'
```
Pezmc, line 825: Should this be named three times?

Steve-Mcl: nope. that is a VSCode bug I was recording and reporting - it slipped by.

Steve-Mcl, line 825:
```suggestion
     * @name /api/v1/projects/:id/logs
```

Steve-Mcl: fixed in [1c04d76](https://github.com/flowforge/flowforge/pull/2318/commits/1c04d76aed5b420d471914fec2f2528286d15135)

Pezmc (approved):
The timer mixin is a good way to handle this, though currently a bit complex for our needs, I also note it doesn't have any sort of backing off build in, which I believe we should have by default to reduce the risk of self DDOS, but I can add that down the line, so happy to move forward!

Steve-Mcl:
> though currently a bit complex for our needs

I understand your point however the ugly work to make polling work (and stop working when you leave a page) in vue is just awful & we need a way to simplify this and have a common approach IMO.  So while the timer logic (in the mixin) is slightly more complex (internally), implementing timers (on a page) that dont get "left running" and are easy to start/stop on demand outweighs that (I hope).  

Are you going to merge or shall I Pez?

Steve-Mcl:
> I also note it doesn't have any sort of backing off build in, which I believe we should have by default to reduce the risk of self DDOS

TBF, neither does the random implementations of `setInterval` we currently do 🤷 

We could add an `options` with `.timeout` to the API calls to ensure we poll < timer set interval?

Pezmc:
@Steve-Mcl There's a best practice setTimeout implementation as part of instance status polling, that _was_ the standard I wanted our codebase to reach. I think we can merge that with this new timer library and have the best of both worlds: https://github.com/flowforge/flowforge/blob/main/frontend/src/components/InstanceStatusPolling.vue#L1 (this is admittedly in my head and not documented anywhere given I'm _often_ the one doing frontend nowadays).

PR #2327, opened 2023-06-22
fix(docker): Update devices.md

biancode:
fix for typo in command for Docker and add docker-compose example


hardillb (approved):

hardillb:
@biancode thanks

Issue #1556, opened 2023-01-13
API Swagger Documentation


joepavitt:
### Description

As a developer in the UI, it is very difficult to navigate and understand our available APIs, as I have to rely on exploring source code and tests to find the relevant API routes available, and their respective options.

By deploying Swagger docs, we would be making our API readability and usability considerably better.

knolleary:
Related/Duplicate #1364

knolleary:
https://github.com/fastify/fastify-swagger will auto-generate api docs based on the routes defined on the fastify app.

But for the doc to be useful, we need to update our route definitions to include all the schema meta-data. That'll be a good thing because it means we aren't maintaining a separate doc.

 - That plugin is design to serve up the swagger doc from the platform directly. I don't think that is quite what we want if the goal is to have an API Docs section on the website (which is what I think our ultimate aim should be). So we'd need to figure out a workflow that can use that plugin to generate the docs and capture it for publishing elsewhere.
 - We'd only want to document certain routes - not those that exist in support of serving the frontend and its resources.

I'll have a play.

ZJvandeWeg:
@knolleary You might have found it already, but there's a static mode; https://github.com/fastify/fastify-swagger#modes

ZJvandeWeg:
Requested by customers:
1. https://app-eu1.hubspot.com/contacts/26586079/company/6697969132
2. https://app-eu1.hubspot.com/contacts/26586079/company/7381916362
3. https://app-eu1.hubspot.com/contacts/26586079/company/7697298655

sumitshinde-84:
I'm interested in working on this issue, @knolleary @joepavitt . Could you please assign it to me?

knolleary:
Hi @sumitshinde-84.

Thanks for the interest, but this issue is already assigned and being actively worked on at the moment.

MarianRaphael:
1.9 Release

PR #2349, opened 2023-06-27
Refector application collection endpoints to not return full Instance objects

knolleary:
## Description

A pair of end points that return lists of Applications and their instances was including the *full* instance object. This is was far more information than was needed.

This PR changes those endpoints to run an `InstanceSummary` view which is the minimum information needed for list views.

Affected endpoints:
 - `/api/v1/application/:id/instances`
 - `/api/v1/teams/:teamId/applications`

I have verified all of the users of these endpoints in the UI work with these summaries.

Work to document these changes will happen as part of the api doc work in #1556 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/views/Project.js
+++ b/forge/db/views/Project.js
@@ -93,7 +93,6 @@ module.exports = {
     },
     instancesSummaryList: function (app, instancesArray) {
         return instancesArray.map((instance) => {
-            // Full settings are not
             const result = app.db.views.Project.projectSummary(instance)
             if (!result.url) {
                 delete result.url
--- a/forge/db/views/Project.js
+++ b/forge/db/views/Project.js
@@ -80,7 +80,7 @@ module.exports = {
         return result
     },
     instancesList: async function (app, instancesArray) {
-        return await Promise.all(instancesArray.map(async (instance) => {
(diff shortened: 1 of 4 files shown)


Pezmc:
I was worried this would cause troubles in the UI with the cached objects not being full objects any more (with regards to startus, last deploy, etc...), but the work to add an explicit status endpoint seems to have prevented that from being a problem.

forge/db/views/Project.js
```diff
@@ -91,14 +91,38 @@ module.exports = {
             return result
         }))
     },
+    instancesSummaryList: function (app, instancesArray) {
+        return instancesArray.map((instance) => {
+            // Full settings are not
```
Pezmc, line 96: [nb] Not sure what this comment is trying to say!

Pezmc (approved):

forge/db/views/Project.js
knolleary, line 96:
```suggestion
```

knolleary: Comment removed

PR #2351, opened 2023-06-27
Add api schemas and swagger ui support

knolleary:
Part of #1556 

Raised as a Draft to get some early feedback on the general approach.

## Description

This adds the initial Swagger UI and adds schemas to the authentication, user and teams routes.

 - The Swagger doc is served from `/api/json`
 - The Swagger UI is browsable (without authentication) at `/api/` (the trailing slash is required... may tidy that up)

This starts to rework how the `forge/db/views` are defined - so that we can define view schemas alongside the code that creates them. The ultimate goal will be to allow fastify do more work using the schema, but that is out of scope for this work.

I've removed `app.db.views.User.shortProfile` and replaced it with the existing `app.db.views.User.userSummary` - trying to maximise view reuse.

---

In terms of testing, this doesn't add any new testable functionality other than the Swagger generation. The main risk is it results in a field being stripped from a response that was previously there, and it's an edge case not covered by the existing tests. I'm doing various bits of UI click testing in addition to exercise the routes as well.

Note that the changes in #2349 will need taking into account as well once that is reviewed and merged.

---

API Files updated (under `forge/routes/api`):

 - [x] admin.js (#2370)
 - [x] application.js (this pr)
 - [x] device.js (#2357)
 - [ ] deviceLive.js - used by devices to communicate with the platform. Not an external API, so lower priority to document here
 - [x] project.js (#2356)
 - [x] projectActions.js (#2356)
 - [x] projectDevices.js (#2356)
 - [x] projectSnapshots.js (#2356)
 - [x] projectType.js (#2369)
 - [x] settings.js (#2370)
 - [x] stack.js (#2370)
 - [x] team.js (this pr)
 - [x] teamDevices.js (#2366)
 - [x] teamInvitations.js (this pr)
 - [x] teamMembers.js (this pr)
 - [x] teamType.js (this pr)
 - [x] template.js (#2369)
 - [x] user.js (this pr)
 - [x] userInvitations.js (#2369)
 - [x] users.js (this pr)


knolleary:
Looks like I've tripped over failing to update package-log.json...

Pezmc:
@knolleary Let's frame it as you tested the validation was working!

knolleary:
Final update - have removed the 'Try it out' button from swagger ui as the auth side is a bit funky at this stage.

Also hidden all of the non-api 'authentication' routes.

Finally, figured out how to order the tags so have done so.

@Pezmc all good to review (although you have done most of it via the child-PRs)

knolleary:
Oh - one more thing... worked out how to document the auth schemes for the routes, but the swagger UI experience isn't ideal IMHO. So have left it in place, but commented out as it isn't entirely obvious to figure out.

Pezmc:
Yes, realistically I'm not able to go through this with a fine comb, so have tested locally and given a quick skim over.

Pezmc (approved):

PR #2355, opened 2023-06-28
Fix instance flow update dialog

### LINKED ISSUE #2200: Duplicate labels in Instance Import dialog
### Current Behavior

image

### Expected Behavior

Only say it once.

### Steps To Reproduce

1. go to Instance overview page
2. click 'Import Instance' button
3. observe the duplicate labels

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Closes #2200

## Description

Fixes the duplicate labels in the Instance Import dialog. Whilst fixing, also spotted the dialog calls the wrong endpoint; also fixed.

For bonus points, I've added a UI test to cover this dialog and verify it passes the provided values to the API. I feel so proud of myself for figuring out a little bit more of Cypress.

## Checklist


frontend/src/pages/instance/components/ImportInstanceComponents.vue
```diff
@@ -99,15 +99,9 @@ export default {
...
-                    const creds = JSON.parse(reader.result)
-                    if (creds.$) {
-                        // Good Start
-                        this.parts.credentials = reader.result
-                        this.errors.creds = ''
-                    } else {
-                        this.parts.credentials = undefined
-                        this.errors.creds = 'Does not look like a creds file'
-                    }
+                    JSON.parse(reader.result)
```
Pezmc, line 102: This parse seems to be dangling. Is it just to catch it _not_ being JSON?

Pezmc (approved):

frontend/src/pages/instance/components/ImportInstanceComponents.vue
knolleary: Correct. Previously the code was parsing and looking for the `$` property. That is a sensible check for it being an encrypted creds file, however users may have unencrypted files which we should allow. So I reduced it to just check its valid JSON.

PR #2378, opened 2023-06-30
Add "What brings you to FF?" to sign-up form

### LINKED ISSUE #2362: "What brings you to FlowForge?" question at sign-up
### Epic

_No response_

### Description

**As a:** FlowForge Admin

**I want to:** better understand why a user has joined FlowForge

**So that:** we can better segment users and enhance our understanding of what features are being used, better understanding of churn and most importantly understanding what the user expectations would likely be.

None

### Acceptance Criteria

- [ ] On sign-up to FlowForge Cloud users will be asked "What brings you to FlowForge"
    - Learning Node-RED
    - Personal Use
    - Business Operations? (need a better term here?) for work? 
- [ ] Is this data stored in the `User` table or elsewhere? Posthog maybe?
    - Either way, we should definitely include it in _at least_ Posthog, as analytics are useful there.
- [ ] How is this configurable for OS versions

joepavitt:
## Description

Adds a new question "What brings you to FlowForge?" to the sign up form. This will likely need to be configurable at at the admin level too, and have a DB migration script (assuming we're storing this as part of the User object in the DB, which seems reasonable?)

### Posthog-only storage of data
- [x] Question added to UI
- [x] Include value in "Person" object for PostHog analysis

### FlowForge storage of data
Work started here: https://github.com/flowforge/flowforge/tree/2362-db-storage
- [ ] DB Migration Script 
- [ ] Admin Setting Added
- [ ] Store value in DB when form submitted

## Related Issue(s)

- Blocked by https://github.com/flowforge/forge-ui-components/pull/145
- Closes #2362

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/account/Create.vue
+++ b/frontend/src/pages/account/Create.vue
@@ -11,24 +11,24 @@
                v-html="settings['branding:account:signUpTopBanner']"></p>
             <div>
                 <label>Username</label>
-                <ff-text-input ref="signup-username" label="username" :error="errors.username" v-model="input.username" />
+                <ff-text-input ref="signup-username" data-form="signup-username" label="username" :error="errors.username" v-model="input.username" />
                 <span class="ff-error-inline">{{ errors.username }}</span>
                 <label>Full Name</label>
-                <ff-text-input ref="signup-fullname" label="Full Name" :error="errors.name" v-model="input.name" />
+                <ff-text-input ref="signup-fullname" data-form="signup-fullname" label="Full Name" :error="errors.name" v-model="input.name" />
                 <span class="ff-error-inline">{{ errors.name }}</span>
                 <label>E-Mail Address</label>
-                <ff-text-input ref="signup-email" label="E-Mail Address" :error="errors.email" v-model="input.email" />
+                <ff-text-input ref="signup-email" data-form="signup-email" label="E-Mail Address" :error="errors.email" v-model="input.email" />
                 <span class="ff-error-inline">{{ errors.email }}</span>
                 <label>Password</label>
-                <ff-text-input ref="signup-password" label="password" :error="errors.password" v-model="input.password" type="password"/>
+                <ff-text-input ref="signup-password" data-form="signup-password" label="password" :error="errors.password" v-model="input.password" type="password"/>
                 <span class="ff-error-inline">{{ errors.password }}</span>
             </div>
             <div class="pt-3" v-if="askJoinReason">
                 <ff-radio-group label="What brings you to FlowForge?" v-model="input.join_reason" orientation="grid"
-                                :options="[{label: 'Business Needs', value: 'business'}, {label: 'Personal Use', value: 'personal'}, {label: 'Educational Use', value: 'education'}, {label: 'Other', value: 'other'}]" />
+                                data-form="signup-join-reason" :options="reasons" />
             </div>
             <div v-if="settings['user:tcs-required']">
-                <ff-checkbox v-model="input.tcs_accepted">
+                <ff-checkbox v-model="input.tcs_accepted" data-form="signup-accept-tcs">
                     I accept the <a target="_blank" :href="settings['user:tcs-url']">FlowForge Terms &amp; Conditions.</a>
                 </ff-checkbox>
             </div>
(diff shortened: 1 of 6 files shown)


joepavitt:
Question included on the site, with (currently) 3 options - can we think of any other options here @MarianRaphael? Want them vague enough that they're not overwhelmed with options, but defined enough to provide us value in customer segmentation.

Screenshot 2023-06-30 at 16 32 25

joepavitt:
@knolleary options welcome/required on the storage of the answer too, is the `Users` table reasonable?

hardillb:
Aware this is still draft, but would this show on even a local install at the moment if sign up is enabled?

joepavitt:
> even a local install at the moment if sign up is enabled?

right now, yes, but that's my open checklist item about `Admin Setting Added`. Not my expertise in that part of the code base, so was hoping for an assist there

joepavitt:
default would be that this is disabled, and I only expect us to have this in FF Cloud.

MarianRaphael:
@joepavitt, having the default set as disabled is a great idea (Self hosted vs. FF Cloud). The four categories are a good choice. I might suggest changing the wording from "Learning Node-RED" to “Educational Use” in order to more clearly delineate the categories and prevent potential overlap. But it's ultimately up to you to decide.

Steve-Mcl:
@joepavitt I am not sure the FF database is the _right_ place to store this?

Should this not be something passed to posthog or another service intended for meta data?

joepavitt:
@knolleary If I recall, we had the conversation about where this data is stored, and agreed it had to go in FF database? I was happy just to throw it into PostHog, and nowhere else. Input please.

knolleary:
Personally, I don't think it fits in the database as its a one-off piece of information that we *never* need to refer to again in the platform. As I understand it, the requirement is to get this information into Hubspot. We need to identify the path it takes to get there. Lets have a chat with @robmarcer as he owns the bit that syncs up hubspot afaik.

robmarcer:
Currently, any data which flows from FF Cloud to Hubspot has to go via the database. The system checks the database for specific user actions every X minutes (including new sign ups) and creates or updates the corresponding record in Hubspot as needed.

Therefore, if we want to add the users' reply to this question to their record in Hubspot without developing an alternative path we will need the answers to be in the database.

I assume that the marketing and sales functions of FlowForge (@iskerrett @ZJvandeWeg) would like to have this data in Hubspot although I don't see that stated in this issue as a specific requirement.

I appreciate that storing the data in the database forever (well at least until a user is deleted) is not ideal but I suspect it's also the path of least resistance to achieve the ACs.

joepavitt:
Just to size the dev work required here ahead of the release on Thursday:
 - To include in PostHog = done
 - Requires:
     - DB Migration Script (Written)
     - User/Admin configuration to define whether or not to show the question (not written)
         - Requires changes to the settings API, settings DB? and models.

joepavitt:
I would propose, for 1.9.0, we push as we are (PH storage), and then utilise the PH [`/persons` API ](https://posthog.com/docs/api/persons#get-api-projects-project_id-persons) to populate HubSpot in a similar fashion for now given the additional work required by Thursday if we want this into our DB.

As part of 1.10.0, we could then come up with a non-ph dependent solution which makes DB retrieval an option (subject to agreement on _how_ it's stored in said DB)

knolleary:
Looks fine - just needs the new ui-components to be published and package/package-lock updated to pick it up.

knolleary (approved):

ZJvandeWeg:
It'd rather have some data from Thursday onwards than nothing, so I recommend we move forward with this as is and find a definitive solution next release

knolleary:
postgres tests appear to have got stuck - have cancelled and kicked off again

frontend/src/pages/account/Create.vue
```diff
@@ -91,8 +102,12 @@ export default {
             return (this.input.email && !this.errors.email) &&
                    (this.input.username && !this.errors.username) &&
                    this.input.password.length >= 8 &&
+                   (this.input.join_reason) &&
```
knolleary, line 105: This line breaks sign-up for instances without posthog, as `join_reason` is never set to anything so the sign-up button is never enabled.

knolleary: In fact, it is broken for all instances. Even with posthog enabled, the button never becomes enabled.

PR #2386, opened 2023-07-03
Update changelog for 1.8.1 release

knolleary:
Update changelog/package version for 1.8.1 maintenance release

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@flowforge/flowforge",
-    "version": "1.8.0",
+    "version": "1.8.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@flowforge/flowforge",
-    "version": "1.8.0",
+    "version": "1.8.1",
```
hardillb, line 3: We already have a 1.8.1 for the nr-launcher and helm, do we want to go to 1.8.2?

knolleary: ok - will make it so

hardillb (approved):

PR #2390, opened 2023-07-03
Add schema and views to pipeline routes

knolleary:
## Description

Adds schemas and routes to the pipeline routes.

Also speeds up the pipeline tests from 24s -> 3s.

Having done this work, I have some opinions on some API rework we should do before documenting - there is some amount of inconsistency between accessing a pipeline via `/api/v1/applications/:id/pipelines/:id` and `/api/v1/pipelines:id`.

I will be doing that as a separate PR to minimise merge conflicts with the pipeline work @Pezmc is working on.

### DIFF
--- a/forge/ee/db/views/PipelineStage.js
+++ b/forge/ee/db/views/PipelineStage.js
@@ -41,7 +41,7 @@ module.exports = function (app) {
         }
     })
 
-    async function stageList (app, stages) {
+    async function stageList (stages) {
         // Must ensure the stages are listed in the correct order
         const stagesById = {}
         const backReferences = {}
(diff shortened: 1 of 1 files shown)


knolleary:
I'm going to hold this out to prevent any delays on #2399. Once #2399 is stable, I can redo the changes in this PR over the top.

Pezmc:
I've rebased this onto main, and resolved the latest merge conflicts.

Given I reviewed the original changes, and just made more myself, if someone could please carefully review the above once the tests are green, that would be appreciated. cc/ @hardillb @Steve-Mcl

hardillb:
We have a test failure, so need to work out what's failed

PR #2391, opened 2023-07-03
Reorganise pipeline api structure

knolleary:
## Description

Whilst documenting the pipeline apis, I found some general inconsistencies with how we approach objects in the API. Primarily, hanging *some* apis under `/api/v1/applications/:id/pipelines` and others under `/api/v1/pipelines/:id` looked out of place once they had been documented and sat next to each other in the docs.

In summary, this is the existing api:

pipelines-api-pre

This is how it looks after this PR:

pipelines-api-post

To summarise:

 - All pipeline operations now live under `/api/v1/pipelines/...`
 - Listing the pipelines that belong to an application stays where it was - `/api/v1/applications/:id/pipelines` - as that is consistent with other apis for listing resources owned by the application
 - Creating a pipeline (`PUT /api/v1/pipelines`) now expects an `applicationId` property in the body (where it was previously one of the url params)

I have also reordered the routes in the file so they appear in a more logic order in the swagger output.

Unit tests updated to handle changed routes. UI updated as well. Although, as I type this, I realise I haven't run the `e2e` tests locally... will do so next.

### DIFF
--- a/forge/ee/routes/pipeline/index.js
+++ b/forge/ee/routes/pipeline/index.js
@@ -5,13 +5,11 @@ const { Roles } = require('../../../lib/roles.js')
 
 module.exports = async function (app) {
     registerPermissions({
-        'pipeline:view': { description: 'View a pipeline', role: Roles.Member },
+        'pipeline:read': { description: 'View a pipeline', role: Roles.Member },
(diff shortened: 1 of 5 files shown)


Pezmc:
I also moved these routes around as part of the pipeline! Looks good to me!

Pezmc (approved):
Great!

PR #2399, opened 2023-07-04
Snapshots in Pipelines

Pezmc:
- Re-order methods in pipeline file
- Extract snapshot creation logic into a service
- Deployment of pipeline stages using snapshots

### DIFF
--- a/forge/ee/routes/pipeline/index.js
+++ b/forge/ee/routes/pipeline/index.js
@@ -1,7 +1,10 @@
+const { restart } = require('nodemon')
 const { ValidationError } = require('sequelize')
 
 const { registerPermissions } = require('../../../lib/permissions')
 const { Roles } = require('../../../lib/roles.js')
+
+const { createSnapshot, copySnapshot } = require('../../../services/snapshots')
 
 module.exports = async function (app) {
     registerPermissions({
@@ -120,7 +123,7 @@ module.exports = async function (app) {
 
         const stages = await pipeline.stages()
         if (stages.length > 0) {
-            // delete stages too
+        // delete stages too
             for (let i = 0; i < stages.length; i++) {
                 stages[i].destroy()
             }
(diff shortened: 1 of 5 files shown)


hardillb:
As mentioned when discussing:

- What happens with credentials as part of the deploy, they should probably be copied over along with the settings and flows. Credential values can be pulled from Env Vars so if done right should "work"

forge/services/snapshots.js
```diff
@@ -0,0 +1,49 @@
...
+        await setSnapShotAsTarget(app, snapShot, instance, user)
+    }
+    return snapShot
+}
+
+module.exports.copySnapshot = async (app, snapshot, toInstance, setAsTarget = true) => {
+    const { settings, flows, name, description } = snapshot.toJSON()
+    const snapshotToCopyProps = { settings, flows, name, description }
+
+    const newSnapshot = await await app.db.models.ProjectSnapshot.create(
```
hardillb, line 40:
```suggestion
    const newSnapshot =  await app.db.models.ProjectSnapshot.create(
```

forge/ee/routes/pipeline/index.js
```diff
@@ -1,8 +1,11 @@
+const { restart } = require('nodemon')
```
hardillb, line 1:
```suggestion
```

hardillb (commented):
Looks like credentials are getting copied if I've read this right

hardillb:
Notes from testing

- Ticking the "deploy to device" box when creating a step doesn't appear to set the flag in the db, editing the step does set the flag
- Not seeing the flow be copied to the second stage
- Not seeing the flow being sent to the device (after flag fixed by editing stage)
- Last Deployed time appears to never be updated (even with page refresh)

hardillb:
@Pezmc will do

knolleary:
This is all looking good - although we have identified a blocker with the current implementation that will require more time than available to get this in to 1.9 tomorrow.

When promoting between stages, the snapshot is created at the source, copied to target and then applied - all good. However, the snapshot overwrites the target's env var values - where as it should be merging them.

So the step to create the target's copy of the snapshot needs to merge in the target's env vars at that point in time.


A secondary issue I've just spotted, but needs a bit more diagnosis work; I think the credentials object needs to be re-encrypted for the target instance's key, otherwise it cannot decrypt properly.

Pezmc:
Rebased onto main to resolve the merge conflicts

Pezmc:
@knolleary @hardillb  Theoretically this is ready for review and to be tested locally, but the tests are hanging for a reason I haven't fully explained yet. Please validate it's behaving as you'd expect locally while I get to the bottom of the hanging.

knolleary:
Looking good to me locally.

Created a flow using a credential-type env var.

Deployed through the pipeline to another instance, which had access to the env var value.

Confirmed the stored versions (ie encrypted) of the credentials were different - so re-encryption had occurred. 

Just want to figure out why the test runs are running away in GH Actions.

Pezmc:
@knolleary Still hanging and I really can't explain why... the tests should all have a 5000ms timeout set on them...

knolleary:
It feels like the step of the action is just never starting, so not getting as far as have any tests to timeout.

Pezmc:
@hardillb @knolleary Tests fixed, PR ready for re-review.

knolleary (approved):

PR #2400, opened 2023-07-04
Rename device/provisioningToken project field to instance in api

knolleary:
## Description

Renames the `project` field in both the Device object and the ProvisioningToken object to `instance` in the public API.

Internally, it is still called `project` in the database.


knolleary:
Investigaging test failures

Steve-Mcl (approved):

Steve-Mcl:
Approved pending test fails

Issue #2397, opened 2023-07-04
Swagger UI fixes


knolleary:
Having got the swagger UI on to staging, some final tasks have been identified:

1. Revert #2396 and identify a suitable path to load the FF logo from
2. Some validation errors being reported - only just seen them as they appear in a small footnote right at the bottom of the swagger page:

```
{"messages":["attribute paths.'/api/v1/projects/{instanceId}/snapshots/'(post).requestBody.content.'application/json'.schema.items is missing","paths.'/api/v1/teams/{teamId}/audit-log'. Declared path parameter teamId needs to be defined as a path parameter in path or operation level","paths.'/api/v1/teams/{teamId}/members/{userId}'. Declared path parameter teamId needs to be defined as a path parameter in path or operation level","paths.'/api/v1/teams/{teamId}/members/{userId}'. Declared path parameter teamId needs to be defined as a path parameter in path or operation level","paths.'/api/v1/projects/{instanceId}'. Declared path parameter instanceId needs to be defined as a path parameter in path or operation level","paths.'/api/v1/templates/{templateId}'. Declared path parameter templateId needs to be defined as a path parameter in path or operation level","paths.'/api/v1/project-types/{instanceTypeId}'. Declared path parameter instanceTypeId needs to be defined as a path parameter in path or operation level"],"schemaValidationMessages":[]}
```

knolleary:
Fixed via #2402 and #2403

MarianRaphael:
Verified on staging

PR #2402, opened 2023-07-04
Fix swagger validation errors

knolleary:
Part of #2397 

## Description

Fixes the validation errors identified in #2397 

The swagger doc ui includes a validation icon at the bottom. However it doesn't do this when on localhost as the external validator cannot access localhost.

But this raises a separate issue - for self-hosted instances, having the page ping out to an external service not controlled by the custom (or flowforge) may be a cause for concern. So I have disabled the validation in the UI.

I'm looking for a way to validate the generated openapi spec as part of a unit test - that doesn't rely on the external service to do it. The only module I've found does schema validation (is the json structured properly) but *not* spec validation (does it have all the `params` it should). We already know the json will be well formed as it is generated by fastify-swagger - we need to know if it is spec compliant. I'll keep looking, but lets get this fixes in now.

### DIFF
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -89,7 +89,7 @@ module.exports = async function (app) {
             params: {
                 type: 'object',
                 properties: {
-                    teamId: { type: 'string' }
+                    instanceId: { type: 'string' }
                 }
             },
             response: {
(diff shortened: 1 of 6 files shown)


Steve-Mcl (commented):
Look good from a code PoV however the OP states "So I have disabled the validation in the UI" but I dont see that in the code?

knolleary:
@Steve-Mcl good point - it is included in #2403 here - https://github.com/flowforge/flowforge/pull/2403/files#diff-4ec45305cd8a764dea00207f0dc8dc8d2f607ec9c0a3ba55d95c6b6fe2ed91d2R77

Steve-Mcl (approved):

Issue #2206, opened 2023-05-31
Remove Confirmation Button for Additional Charges


MarianRaphael:
### Description

To streamline our purchasing experience and reduce the frequency of price mentions, it has been proposed to remove the button that confirms additional charges. Our aim is to create a more seamless, efficient, and user-friendly purchasing process.

![image](https://github.com/flowforge/flowforge/assets/73583313/445d7283-23bc-4954-b7b6-216fa2ffd667)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #2404, opened 2023-07-04
remove billing confirmation

Steve-Mcl:
## Description

* Removes the confirmation Check box from instances AND devices
* Moves additional costs to the costs table
* Several bug fixes
   * cost table not hidden when no project type selected
   * sub text under the section headers were randomly shown/not shown
   * Use the formatCurreny mixin for devices - same as instances

### Before
https://github.com/flowforge/flowforge/assets/44235289/cc1bf475-d078-4a14-a03d-9e790560f29d

#### Create Application
https://github.com/flowforge/flowforge/assets/44235289/10de6fa1-68f5-482c-a26e-399ad98445c5

#### Create Instance

https://github.com/flowforge/flowforge/assets/44235289/149a300c-3ebf-48f5-b6f6-28daf38d29ea

#### Add Device
image

## Related Issue(s)

#2206 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/team/Devices/dialogs/TeamDeviceCreateDialog.vue
+++ b/frontend/src/pages/team/Devices/dialogs/TeamDeviceCreateDialog.vue
@@ -8,12 +8,9 @@
             <form class="space-y-6 mt-2">
                 <FormRow v-model="input.name" data-form="device-name" :error="errors.name" :disabled="editDisabled">Name</FormRow>
                 <FormRow v-model="input.type" data-form="device-type" :error="errors.type" :disabled="editDisabled">Type</FormRow>
-                <FormRow v-if="deviceIsBillable" id="billing-confirmation" v-model="input.billingConfirmation" type="checkbox">
-                    Confirm additional charges
-                    <template #description>
-                        {{ billingDescription }}
-                    </template>
-                </FormRow>
+                <div v-if="billingDescription">
+                    Additional charges: {{ billingDescription }}
+                </div>
             </form>
         </template>
     </ff-dialog>
@@ -26,6 +23,7 @@ import devicesApi from '../../../../api/devices.js'
 import teamApi from '../../../../api/team.js'
 
 import FormRow from '../../../../components/FormRow.vue'
+import formatCurrency from '../../../../mixins/Currency.js'
 import alerts from '../../../../services/alerts.js'
 
 export default {
@@ -33,6 +31,7 @@ export default {
     components: {
         FormRow
     },
+    mixins: [formatCurrency],
     props: {
         team: {
             type: Object,
@@ -66,8 +65,7 @@ export default {
             totalDeviceCount: 0,
             input: {
                 name: '',
-                type: '',
-                billingConfirmation: false
+                type: ''
             },
             errors: {},
             editDisabled: false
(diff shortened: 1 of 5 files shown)


joepavitt:
Why in the "Create Instance" example is there a "Payable Now" row that isn't in "Create Application". Not  a fan of that being a row in that table as it seems like _another_ item that you're paying for.

I think a little acknowledgement as per the existing description of the checkbox, where the existing checkbox is, suffices. Otherwise, all good.

Steve-Mcl:
> Why in the "Create Instance" example is there a "Payable Now" row that isn't in "Create Application".

It is only there when there is an amount to pay now that differs the monthly cost.

i.e. If you have $5 credit and monthly cost is $15, it tells you that payable now is $10

In essence, Create Application and Create Instance behave exactly the same (it is the same form)


---

> I think a little acknowledgement as per the existing description of the checkbox, where the existing checkbox is, suffices.

Not sure what you mean exactly Joe?

joepavitt:
> Not sure what you mean exactly Joe?

Was thinking something like:

Screenshot 2023-07-04 at 17 48 28

Not convinced by the "pay now" row unfortunately. It's natural to read those rows and just mentally add up all items in that far column, which isn't reflective of the amount they're due to pay.

Alternatively, if we want to stick with the table format, we go with:

Screenshot 2023-07-04 at 17 54 06

maybe even without the "due now":

Screenshot 2023-07-04 at 17 55 19

Steve-Mcl:
Updated to ...

image


Hoping this is a good compromise?

@joepavitt if you can respond great - otherwise I will push the review to a.n.other.

Ta.

joepavitt:
Works for me, thanks Steve

Steve-Mcl:
Temporarily in draft due to bug found - shows odd value when credit is greater than cost.

MarianRaphael:
I would suggest changing the "Additional charges: ..." label to bold. Furthermore, consider renaming it to simply "Price: ...". When creating an Instance, it's abundantly clear that this step incurs charges. I want to ensure that the price label for Devices cannot be overlooked. 
Other than that, it looks good.

Steve-Mcl:
@MarianRaphael this would have been good information much earlier than now :( 

Can you provide a mock up please - same me re-doing this again?

Steve-Mcl:
PS @MarianRaphael 

> I would suggest changing the "Additional charges: ..." label to bold. Furthermore, consider renaming it to simply "Price: ...". When creating an Instance, 

The "Additional Charges" is original text on the Device Creation only.

The Instance / Application & Instance now looks like this (as of the PR status right now)

image

Please confirm if this is OK or if you want further changes.

MarianRaphael:
@Steve-Mcl Sorry, my comment was solely about the "devices," not the "instances." The purchasing process for instances seems great!

frontend/src/pages/team/Devices/dialogs/TeamDeviceCreateDialog.vue
```diff
@@ -8,12 +8,9 @@
...
                 <FormRow v-model="input.name" data-form="device-name" :error="errors.name" :disabled="editDisabled">Name</FormRow>
                 <FormRow v-model="input.type" data-form="device-type" :error="errors.type" :disabled="editDisabled">Type</FormRow>
-                <FormRow v-if="deviceIsBillable" id="billing-confirmation" v-model="input.billingConfirmation" type="checkbox">
-                    Confirm additional charges
-                    <template #description>
-                        {{ billingDescription }}
-                    </template>
-                </FormRow>
+                <div v-if="billingDescription">
+                    Additional charges: {{ billingDescription }}
```
MarianRaphael, line 12:
```suggestion
                   <b>Additional charges: {{ billingDescription }}</b>
```

MarianRaphael (approved):

PR #2406, opened 2023-07-04
Ensure Team Billing link is present on initial page load

### LINKED ISSUE #2398: Billing team menu item missing on first page load
### Current Behavior

On staging, the `Billing` entry in the sidebar is not present when you first load a team page. As soon as you click on any menu option to go to a different view, the billing entry appears.

### Expected Behavior

Billing entry to be there from the start

### Steps To Reproduce

On staging, load a team page. See billing is missing. Click to navigate to another section - see billing appear.

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Fixes #2398

## Description

When loading any team page, the Billing link was missing. As soon as you navigated to another team view, the billing link would appear.

Having added a bunch of debug, I could see the `checkFeatures` call being made, and `this.navigation.admin` was being updated to add the billing link. However the rendered view wasn't changing in response. As soon as you click around, it rerenders the sidebar and displays the link correctly.

Exactly why vue isn't refreshing the component when one of its computed props changes, I'm not entirely sure. And why it has started happening now, when it doesn't happen on production is also unknown.

The fix is to do the `this.features.billing` check in the original computed property function - it was already doing that for the shared library. That means we don't do any mutation of the computed property outside of the function.

Have click tested this locally.

### DIFF
--- a/frontend/src/components/SideNavigationTeamOptions.vue
+++ b/frontend/src/components/SideNavigationTeamOptions.vue
@@ -69,7 +69,7 @@ export default {
             return (this.$slots['nested-menu'] && this.loaded) || this.closeNested
         },
         navigation () {
-            return {
+            const result = {
                 general: [{
                     label: 'Applications',
                     to: '/applications',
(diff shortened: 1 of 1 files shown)


frontend/src/components/SideNavigationTeamOptions.vue
```diff
@@ -164,17 +171,6 @@ export default {
                 }
             }
         },
-        checkFeatures () {
-            if (this.features.billing) {
-                // insert billing in second slot of admin
-                this.navigation.admin.splice(1, 0, {
```
Pezmc, line 170: I belive Vue doesn't always handle deep observations, i.e. this.navigation changing will trigger a re-render, but this.navigation.admin mutating will not

Pezmc (approved):

PR #2414, opened 2023-07-05
Fix sign-up form when posthog not enabled

knolleary:
## Description

Without posthog enabled, the sign-up button was never becoming enabled.

This was a regression introduced in #2378.

This fixes it, plus adds better test coverage of the sign-up form to ensure the button becomes enabled when the form is filled in under various conditions.

## Checklist


Steve-Mcl (approved):
Makes sense. (waiting test completion before merge)

joepavitt:
apologies

PR #2426, opened 2023-07-06
Update node version statement in devices docs

knolleary:
## Description

Updating to sync with the Device Agent support. Decided to keep the top line saying we support 16, but with a note about 14 support for older hardware. We don't want to encourage Node 14 use if it can be avoided.

### DIFF
--- a/docs/user/devices.md
+++ b/docs/user/devices.md
@@ -11,7 +11,7 @@ This guide explains how to get starting with the Devices feature.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


knolleary:
Unfortunately this PR now needs reworking as the device docs have been restructured. @Steve-Mcl please hold off reviewing until I get the conflict resolved.

Steve-Mcl:
Can we close this Nick?

PR #2437, opened 2023-07-07
Add codecov reporting

knolleary:
## Description

This modifies our builds and npm actions to run code coverage reporting.

 - Main build action:
    - Removes the frontend unit tests as they are also run in the e2e action - we don't want to be reporting on them twice
    - Adds upload to codecov with the 'backend' tag applied
 - E2E build - updated
    - Adds upload to codecov with the 'frontend' tag applide
 - Updates vitest and adds the istanbul coverage reporter - needed for compatibility with the nyc reports.
 - Updates the `cover:*` actions to generating their coverage reports in a consistent location and renames them so they are individually identifiable. (nyc lets you pick the dir it'll write the json to, but no the name of the json file - so we have to do this dance to get the files from separate runs into the same dir without overwriting each other)


knolleary:
Merging this PR will get a baseline of coverage stats into codecov. From then on, subsequent PRs should get a coverage report added.

Pezmc (approved):

PR #2461, opened 2023-07-11
Lint fix chore

hardillb:
## Description

Fix up lint for rule:

- "vue/prop-name-casing": "off"
- "vue/v-on-event-hyphenation": "off"

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -537,16 +537,6 @@
         },
         {
             "rules": {
-                "vue/prop-name-casing": "off"
-            },
-            "files": [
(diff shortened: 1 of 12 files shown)


knolleary:
I'm trying to figure out how to tell codecov/patch to not complain like it is doing.

joepavitt (approved):
from a Vue-perspective, happy. Although not sure what we're doing on the codecov front, so will approve, but won't merge.

hardillb:
Can we ignore the codevoc/patch failure and merge this?

PR #2465, opened 2023-07-11
Fix documentation links

joepavitt:
## Description

Fix the errors flagged in https://github.com/flowforge/website/pull/953#issuecomment-1630783979

For the device agent link, this was a broken/404 error, for the others, they were all valid links to `#` anchors that existed, so the only thing I could think of that caused them to trip was the trailing `/` after the `#`, so have removed those.

## Related Issue(s)

https://github.com/flowforge/website/pull/953#issuecomment-1630783979

## Checklist

## Labels

 - [x] Backport needed? -> add the `backport` label


joepavitt:
split from `maintenance` not `main`

PR #2469, opened 2023-07-11
Fix yaml -> yml references in device docs

knolleary:
## Description

Some instances of `device.yaml` slipped in.


joepavitt (approved):
Blame Co-Pilot. Thanks for spotting.

Issue #2294, opened 2023-06-15
Allow Team Library access from Dev Mode devices


griemenschneider:
### Description

As a FF developer that exclusively uses devices I would like my team to utilize the flows library functionality offered by FlowForge.

Currently, that option is missing from the menu:

![image](https://github.com/flowforge/flowforge/assets/135891848/8765e982-b403-462f-9dde-42f9b1b82a66)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

hardillb:
Notes on current situation:

Libraries are provided by the storage plugin, currently the device-agent uses the standard filesystem storage plugin so has no access to the shared team library.

griemenschneider:
@hardillb Can this be resolved by adding "@flowforge/nr-storage" to the device's package.json or will it conflict with the device-agent filesystem plugin?

hardillb:
@griemenschneider no, this would require a new storage plugin to be written that mixes loading flows from local storage and libraries remotely.

hardillb:
There are also security considerations required here with token scopes for devices not having access to the library endpoints.

knolleary:
Just to clarify what is needed here.

The Shared Library is provided by a *library plugin* and not the storage module. It happens to be the case that both of these things are currently packaged together in the `nr-launcher` module that is only used in the hosted instances.

The required tasks are:

 - Make the existing library plugin available to the device agent
    - either done by pulling the plugin out of nr-launcher and into its own npm package (which goes against out attempt to consolidate the modules we publish)
    - or use some sort of git submodule approach to pull the files into the device agent when it is build and published.
 - Add the additional configuration needed to the device agent settings file
 - Add the necessary scopes to the device agent token

The main issue is the security side. We minimise the scope of what a device token is able to do on the platform as the devices exist outside the platform security boundary. If the device is given full read/write access to the library the risk is anyone with physical access to the device will be able to extract the token and have that same access. That *might* be an acceptable risk for some users, but not necessarily all. That is the piece we need to make sure we have a good position on.

griemenschneider:
[...]

PR #2472, opened 2023-07-12
Allow device in developer mode to access team library

Steve-Mcl:
## Description

NOTE: This is part 1 of 2 PRs to enable the feature described in #2294 

1. passes a boolean "licensed" flag to device in status updates
   1. device uses this to determine if the settings file should include the code to load the plugin
2. removes the `if (request.teamMembership) {` test from the pre-handlers to ensure `needsPermission` is called.
   1. Device uses project session token to authenticate (i.e. `user` is not present)
3. Update the implicit token scopes to ensure access is permitted
4. adds tests
   1. `Add to Library from Devices`
   2. `Prevents Library access from Device in different team`
   3. `Device state` -> `unlicensed` -> `gets live state`
   4. `Device state` -> `licensed` -> `gets live state`
4. updates tests
   1. various tests updated to ensure presence of `licensed` flag in state

NOTE: Docs do not mention devices _cannot access the team library_ so no docs have been updated at this point

## Related Issue(s)

#2294 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/routes/auth/permissions.js
```diff
@@ -7,13 +7,19 @@ const { Roles } = require('../../lib/roles.js')
...
     ],
     project: [
         'user:read',
         'project:flows:view',
         'project:flows:edit',
-        'team:projects:list'
+        'team:projects:list',
+        'library:entry:create',
+        'library:entry:list',
+        'library:entry:delete'
```
knolleary, line 22: Were these changes necessary? We aren't changing any behaviour for the managed instances that use project-scope tokens. They access the shared library using a user-token based on who is logged in. The project-scope token should not have these permissions.

In a future iteration we will want to update the device editor to use user-scoped tokens for this.

knolleary (changes requested):
I think the changes to the Project implicit scope should be reverted - unless I've missed a reason for them to have been changed.

Steve-Mcl:
> I think the changes to the Project implicit scope should be reverted - unless I've missed a reason for them to have been changed.

They were required - but I admit now to thinking there may be a different issue to solve. 

I also replied to the suggestion [here](https://github.com/flowforge/flowforge/pull/2472#discussion_r1274758896)

I will feedback shortly.

Steve-Mcl:
ok, so this took me a while to get back to it - my apologies.

### From a project/instance

It is the team lib plug in that makes the request to team lib. 
https://github.com/flowforge/flowforge-nr-launcher/blob/main/lib/storage/libraryPlugin.js#L21-L26 and it uses a bearer token

Before these modifications, the plugin got a free ride through the `preHander` [here ](https://github.com/flowforge/flowforge/pull/2472/files#diff-06f5b933a416234b5ab9be70426962fcd6c0e1e4c1af7e9aa2fd34d3b994f531L42-L44)and [here](https://github.com/flowforge/flowforge/pull/2472/files#diff-06f5b933a416234b5ab9be70426962fcd6c0e1e4c1af7e9aa2fd34d3b994f531L109-L111) because the `needsPermission` was wrapped in a test of `if (request.teamMembership) { ... }`

With these mods, the `needsPermission` MUST be called & therefore project types _needed_ implicit permissions (which is better than the free ride it was/is getting now IMO).

Does that make sense?

Steve-Mcl:
@hardillb in absence of Nick, can I lean on you for a review please?

Happy to discuss the issue Nick raised and the thoughts behind his comments (they related to a future where user tokens with specific, limited access instead bearer tokens) for RBACs 

As this PR stands, to permit the device access - it was necessary to ensure `needsPermission` was called for the routes and this lead to discovering projects/instances were never actually checking permissions. This then leads to the need, in this iteration to add the implicit permissions for both the device and the project.

If any of this is unclear, I am more than happy to step through it with you.

Ta.

hardillb:
First pass looks ok, Can we discuss it in the morning, just to make sure I've got everything?

hardillb (approved):
Merge once `library:entry:delete` checked

Steve-Mcl:
> Merge once `library:entry:delete` checked

Checked & fails due to invalid token. 

While the end point is seemingly open to use (when a valid bearer is present), the plug in never generates a call to node-red for deletion (there is no code to support it) so the request is not made via NR client to NR server through to forge.

For completeness and piece of mind, I am going to delete the implicit permission regardless.

Will run a batch of manual checks to ensure that doesn't break anything.

Steve-Mcl:
@hardillb tests on local all good with the implicit scope for deletion removed (though I did find a different UI bug!)

Could you verify/merge pls?

hardillb (approved):

PR #2515, opened 2023-07-21
Handle node-red log messages with undefined payloads

knolleary:
## Description

Seen in production. Not clear how a log message could end up with this property - will investigate separately. But better to ensure it's handled in UI.

### DIFF
--- a/frontend/src/pages/instance/components/InstanceLogs.vue
+++ b/frontend/src/pages/instance/components/InstanceLogs.vue
@@ -117,7 +117,7 @@ export default {
                     entries.log.forEach(l => {
                         const d = new Date(parseInt(l.ts.substring(0, l.ts.length - 4)))
                         l.date = `${d.toLocaleDateString()} ${d.toLocaleTimeString()}`
-                        if (l.msg === undefined) {
+                        if (typeof l.msg === `undefined`) {
                             l.msg = 'undefined'
                         } else if (typeof l.msg !== 'string') {
                             l.msg = JSON.stringify(l.msg)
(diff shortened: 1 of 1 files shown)


frontend/src/pages/instance/components/InstanceLogs.vue
```diff
@@ -117,7 +117,9 @@ export default {
                     entries.log.forEach(l => {
                         const d = new Date(parseInt(l.ts.substring(0, l.ts.length - 4)))
                         l.date = `${d.toLocaleDateString()} ${d.toLocaleTimeString()}`
-                        if (typeof l.msg !== 'string') {
+                        if (l.msg === undefined) {
```
Steve-Mcl, line 120:
```suggestion
                        if (typeof l.msg === `undefined`) {
```

undefined is mutable

Steve-Mcl (approved):

frontend/src/pages/instance/components/InstanceLogs.vue
```diff
@@ -117,7 +117,9 @@ export default {
                     entries.log.forEach(l => {
                         const d = new Date(parseInt(l.ts.substring(0, l.ts.length - 4)))
                         l.date = `${d.toLocaleDateString()} ${d.toLocaleTimeString()}`
-                        if (typeof l.msg !== 'string') {
+                        if (typeof l.msg === `undefined`) {
```
knolleary, line 120:
```suggestion
                        if (typeof l.msg === 'undefined') {
```

PR #2549, opened 2023-07-31
Update flowforge.yml to match docs

hardillb:
## Description

Update the example ratelimit period to match docs

### DIFF
--- a/etc/flowforge.yml
+++ b/etc/flowforge.yml
@@ -104,7 +104,7 @@ rate_limits:
   # global: true
   # timeWindow: 60000
   # max: 1000
-  # maxAnonymous: 10 # optional
+  # maxAnonymous: 100 # optional (defaults to value of max if not set)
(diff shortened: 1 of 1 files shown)


etc/flowforge.yml
```diff
@@ -102,7 +102,7 @@ telemetry:
 rate_limits:
   enabled: false
   # global: true
-  # timeWindow: '1 minute'
+  # timeWindow: 60000
   # max: 1000
   # maxAnonymous: 10 # optional
```
hardillb, line 107:
```suggestion
  # maxAnonymous: 100 # optional (defaults to value of max if not set)
```

Steve-Mcl (approved):

Steve-Mcl:
@hardillb. approved. However I will not merge any of these - I will leave merging for you to coordinate appropriately.

PR #2561, opened 2023-08-04
Merge forge-ui-components into core

joepavitt:
## Description

- Pulls forge-ui-components into core FlowForge, and updates the relevant imports within `main.js`.
- Future item to still work out how to render/run all of the ui-components docs in an easy to access way, but for now, functionally, this does what we need.

## Checklist

### DIFF
--- a/frontend/src/components/StatusBadge.vue
+++ b/frontend/src/components/StatusBadge.vue
@@ -25,6 +25,7 @@ import {
     ExclamationCircleIcon,
     ExclamationIcon,
     PlayIcon,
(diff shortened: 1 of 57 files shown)


Pezmc:
Changes themselves look good. Have you thought about still having forge-ui-components as a separate package within this same repo (mono-repo style)? That may be advantages as it would be easy to split out again later, and/or publish independently, while still solving the issue you're trying to fix.

joepavitt:
No longer required as work was done in #2570

PR #2574, opened 2023-08-09
Add basic Team Type docs

knolleary:
## Description

This adds very basic TeamType docs - mostly only relevant for self-hosted users. The key information being the need to enable which InstanceTypes are available to a given TeamType.


Steve-Mcl (approved):
approved pending team types review

PR #2575, opened 2023-08-09
Fix device running or reporting old snapshot id/name

Steve-Mcl:
## Description

This PR addresses several issues found while debugging the reported problem.

NOTE: None of the work here involved messing with the device files. The state of the device and its files were all a result of actions taken in the forge application.  I feel this is important to point out because as it stands, it is possible to get the device to crash and into odd states.

The issues encountered...

### problem 1 - device running snapshot of deleted instance
Device active/target snapshot get populated in checkin code
device reports it is on old snapshot & that matches 
checkin code did not check that the activeSnapshots project matched the device current project

### problem 2 - snapshot table is not cleaned - holds snapshots from deleted instances
When device checks in, its reported snapshot is found and the device objects active/target snapshot are populated meaning the device passes its checkin & does not instruct an update.

NOTE: No work was done to delete the old snapshots, only to mitigate the problem. A separate issue will be raised detailing thoughts and care points around this.  

### problem 3
Even if we force a DeviceUpdate, it just passes the exact same settings due to the snapshot table having old entries

### problem 4
Once we do get the device to clean up, switching between dev/autonomous mode caused a crash due to missing files
```
[AGENT] 09/08/2023 09:47:27 [info] Cleaning instance directory
[AGENT] 09/08/2023 09:47:50 [info] Enabling developer mode
[AGENT] 09/08/2023 09:47:51 [info] Enabling remote editor access
[AGENT] 09/08/2023 09:47:51 [info] No running Node-RED instance, not starting editor
[AGENT] 09/08/2023 09:47:56 [info] Enabling remote editor access
[AGENT] 09/08/2023 09:47:56 [info] No running Node-RED instance, not starting editor
[Error: ENOENT: no such file or directory, open 'C:\opt\flowforge-device\project\flows.json'] {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'open',
  path: 'C:\\opt\\flowforge-device\\project\\flows.json'
}
```

### problem 5
when the agent determines a new snapshot is available, we dont instruct it to grab new settings.
Since some env vars are computed by the platform (e.g. FF_SNAPSHOT_ID/NAME) we must also instruct
the agent to refresh its settings.

#### 1: [ensure device is not using orphaned snapshot](https://github.com/flowforge/flowforge/commit/c516bd3f8b044f2b57e3aa348d55c95ca67fd56f)

When checking the status of the device, grab the full row from ProjectSnapshots and verify the associated project is both present and matches the device. If not, the snapshot is considered to be an orphan or mismatched and a device update is requested.

#### 2: [ensure platform doesnt send ophaned snapshot > dev](https://github.com/flowforge/flowforge/commit/8b2427c535cab08f4baaeff2ca2e7d86a1c6ea3a)

When sending an update command to device, check the targetSnapshot has a project AND it matches the projectId that owns the device. If this fails, set the snapshot id in the payload update to null. This will signal the device to clear its current snapshot.

This can occur when an instance (with a member device and target snapshot) is deleted. The ProjectId field is cleared but the snapshot remains in the database. 

### ~TODO~ Unit Tests
Adds tests to `test/unit/forge/db/controllers/Device_spec.js`
- Device controller
    -  sendDeviceUpdateCommand
        - sends update command to device
        - does not send orphaned snapshot in update command to device

### Manual Tests

1. delete instance - expect device to clear its config

* On next poll, agent reports correct change  👍 
```
[AGENT] 09/08/2023 15:17:39 [info] Stopping Node-RED
[AGENT] 09/08/2023 15:17:39 [info] Stopped Node-RED
[AGENT] 09/08/2023 15:17:39 [info] Cleaning instance directory
```
* Device status on forge app shows Application and Instace _Unassigned_ 👍 
* Last Known Status shows `stopped` 👍 

2. delete active snapshot - expect device to clear its config

* On next poll, agent reports correct change  👍 
```
[AGENT] 09/08/2023 15:13:52 [info] Stopping Node-RED
[AGENT] 09/08/2023 15:13:52 [info] Stopped Node-RED
[AGENT] 09/08/2023 15:13:52 [info] Cleaning instance directory
```

## Related Issue(s)

https://github.com/flowforge/flowforge-device-agent/issues/132

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/test/unit/forge/db/controllers/Device_spec.js
+++ b/test/unit/forge/db/controllers/Device_spec.js
@@ -1,13 +1,85 @@
+const sinon = require('sinon')
 const should = require('should') // eslint-disable-line
+const snapshotServices = require('../../../../../forge/services/snapshots.js')
+const TestModelFactory = require('../../../../lib/TestModelFactory.js')
 const setup = require('../setup')
 
 describe('Device controller', function () {
-    // Use standard test data.
     let app
-    before(async function () {
-        app = await setup()
-    })
+    /** @type {TestModelFactory} */ let factory
+    const TestObjects = {}
 
+    async function setupCE () {
+        app = await setup()
+        factory = new TestModelFactory(app)
+        await setupTestObjects()
+    }
+
+    async function login (username, password) {
+        const response = await app.inject({
+            method: 'POST',
+            url: '/account/login',
+            payload: { username, password, remember: false }
+        })
+        response.cookies.should.have.length(1)
+        response.cookies[0].should.have.property('name', 'sid')
+        TestObjects.tokens[username] = response.cookies[0].value
+    }
+
+    async function setupTestObjects () {
+        // alice : admin
+        // ATeam ( alice  (owner) )
+
+        // Alice is created in setup()
+        const userAlice = await app.db.models.User.byUsername('alice')
+        // ATeam is created in setup()
+        const ATeam = await app.db.models.Team.byName('ATeam')
+        // Alice is set as ATeam owner in setup()
+
+        // generate a template
+        const template1 = await factory.createProjectTemplate({
+            name: 'template1',
+            settings: {
+                httpAdminRoot: '',
+                codeEditor: '',
+                palette: {
+                    modules: []
+                }
+            }
+        }, userAlice)
+        const projectType1 = await factory.createProjectType({
+            name: 'projectType1',
+            description: 'default project type',
+            properties: { foo: 'bar' }
+        })
+        const stack1 = await factory.createStack({ name: 'stack1' }, projectType1)
+        const application1 = await factory.createApplication({ name: 'application-1' }, ATeam)
+        const instance1 = await factory.createInstance(
+            { name: 'project1' },
+            application1,
+            stack1,
+            template1,
+            projectType1,
+            { start: false }
+        )
+
+        // Set up TestObjects
+        TestObjects.alice = userAlice
+        TestObjects.team = ATeam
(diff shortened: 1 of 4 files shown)


test/unit/forge/db/controllers/Device_spec.js
```diff
@@ -1,13 +1,85 @@
+const sinon = require('sinon')
 const should = require('should') // eslint-disable-line
+const snapshotServices = require('../../../../../forge/services/snapshots.js')
+const TestModelFactory = require('../../../../lib/TestModelFactory.js')
 const setup = require('../setup')
 
-describe('Device controller', function () {
-    // Use standard test data.
+describe.only('Device controller', function () {
```
Steve-Mcl, line 7:
```suggestion
describe('Device controller', function () {
```

Steve-Mcl: remove .only!

hardillb (approved):

hardillb:
Before I merge these do we want to add the backport label so we can include it in next weeks point release?

Steve-Mcl:
> Before I merge these do we want to add the backport label so we can include it in next weeks point release?

Happy either way Ben. 

Is this a good candidate for next weeks point release?

hardillb:
I think the same comments as the device agent apply, this was something only I had hit so far, so no great rush...

I'll add the flag, but not merge the backport PR

PR #2584, opened 2023-08-14
Personal access tokens

### LINKED ISSUE #14: Add Personal Access Token Support
- [ ] Add database model for Personal Access Tokens (PATs)
 - [ ] Define REST API interface to allow user to generate/delete tokens
 - [ ] Add auth middleware to handle PAT in place of session cookie
 - [ ] Create UI view to list user's tokens
 - [ ] Create UI view to create a new token
```[tasklist]
### User Stories
- [ ] https://github.com/flowforge/flowforge/issues/42
- [ ] https://github.com/flowforge/flowforge/issues/41
- [ ] https://github.com/flowforge/flowforge/issues/2482
```

hardillb:
fixes #14 
## Description

Adds support for creating a Personal Access Token.

Adds a new entry to the User Settings -> Security tab

Allows multiple tokens with different scopes.

## Related Issue(s)

#14

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
OK, I don't understand the UI Tests failure.

If you get a second @joepavitt can you have a look (and feel free to critique the whole UI, there is a feature demo video in the channel)

joepavitt:
Just had a little look at the logs. These are all of the ui-components. Have you rebased since we merged?

Can only think that's a cause...

hardillb:
Rebased and still failing the same way

joepavitt:
@hardillb noticed the latest tests are all passing? What changed?

hardillb:
@joepavitt The actual failure was a VS code auto complete thing.

It had injected:

```
<script>
...
export default ({
...
})
</script>
```

Note the extra none curly braces round the export bit

joepavitt:
so is this "ready for review"?

hardillb:
I add more tests... and code coverage goes down

joepavitt:
Have modified the dialog slightly to improve spacing in layout, but also make it more explicit styling on the disabled state & make the checkbox say "Add Expiry Date", rather than "never":

Screenshot 2023-08-18 at 10 13 15

Screenshot 2023-08-18 at 10 13 19

joepavitt:
Am also going to add E2E tests

joepavitt (approved):
Approved front-end, E2E's also added. Leaving server-side eyes to Nick

forge/db/migrations/20230808-01-add-access-token-name.js
```diff
@@ -0,0 +1,14 @@
+/**
```
knolleary, line 1: This migration will need to be renamed. We have already merged `20230814-01-add-description-to-applications` so this file will need a later date.

forge/routes/api/settings.js
```diff
@@ -75,6 +76,7 @@ module.exports = async function (app) {
                     }
                 })
                 response['platform:stats:token'] = app.settings.get('platform:stats:token')
+                response['platform:auth:permissions'] = Permissions
```
knolleary, line 79: Given we aren't including custom scopes for this iteration we don't need to expose the full permissions table.

frontend/src/pages/account/Security/dialogs/TokenDialog.vue
```diff
@@ -0,0 +1,145 @@
...
+        }
+    },
+    mounted () {
+        this.getScopes()
+    },
+    methods: {
+        getScopes: async function () {
+            const temp = await settings.getSettings()
+            this.scopes = Object.keys(temp['platform:auth:permissions'])
+        },
```
knolleary, line 104:
```suggestion
    methods: {
```
As per other comment - not including scopes in this iteration, so suggest removing this code as we may want a different approach in the future.

frontend/src/pages/account/components/ExpiryCell.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+    <div class="flex items-center">
+        <div>{{ expires }}</div>
+    </div>
+</template>
+<script>
+export default {
+    name: 'ExpiryCell',
+    props: ['expiresAt'],
```
knolleary, line 9: Linting warning:

`Prop "expiresAt" should define at least its type`

forge/routes/api/user.js
```diff
@@ -189,4 +189,145 @@ module.exports = async function (app) {
             reply.code(400).send(resp)
         }
     })
+
+    /**
+     * Get Personal Access Tokens
+     * /api/v1/user/pat
+     */
+    app.get('/pat', {
```
knolleary, line 197: Sorry to say, I think the api end point should be `/api/v1/user/tokens`...

knolleary (changes requested):

frontend/src/pages/account/Security/dialogs/TokenDialog.vue
```diff
@@ -0,0 +1,137 @@
+<template>
+    <ff-dialog ref="dialog" data-el="add-token-dialog" header="Add Token" :confirm-label="token ? 'Save' : 'Create'" :disablePrimary="disableConfirm" @confirm="confirm()">
```
knolleary, line 2:
```suggestion
    <ff-dialog ref="dialog" data-el="add-token-dialog" :header="(token ? 'Edit' : 'Add') + ' Token'" :confirm-label="token ? 'Save' : 'Create'" :disablePrimary="disableConfirm" @confirm="confirm()">
```

Update title to reflect add/edit mode

frontend/src/pages/account/Security/dialogs/TokenDialog.vue
```diff
@@ -0,0 +1,137 @@
...
+        <template #default>
+            <form class="space-y-4" @submit.prevent>
+                <FormRow v-model="input.name" data-form="token-name" :disabled="edit">
+                    Name
+                    <template #description>Token name</template>
```
knolleary, line 7:
```suggestion
                <FormRow v-model="input.name" data-form="token-name" :disabled="edit">
                    Token name
```

Feels redundant to have both the label and description set to much the same - suggest collapsing down to just the label.

hardillb:
argh, I assume removing the "Name" label has broken the cypress tests

knolleary:
@hardillb I eyeballed it and didn't spot a reason for that change to require a UI test change... but I'll dig in.

knolleary:
@hardillb Don't think this is a related failure. The tests pass locally; unless I run them repeatedly then sometimes they fail.

knolleary:
Got it - the tests use the count of selectable table rows as a proxy for how many tokens there are. By making them not selectable, it isn't finding them.

The fiddly bit is the 'No data found' row needs to be accounted for.... working on it

knolleary:
@joepavitt I'm modifying it to be more consistent with the Device Provisioning Tokens table we already have:

When there are no tokens, rather than show the default 'no data found' message, it has a custom element:

image

knolleary:
Have pushed an update so the token table now has a consistent appearance with the provisioning token table:

image

 - Added SectionHeader with help text
 - Changed button to `+ Add Token`
 - Changed the 'empty state' to be a custom message - which then also makes the tests easier to handle the empty state (no longer needing to check for `selectable`)

Issue #2021, opened 2023-04-18
Add support for sending email via HubSpot


knolleary:
### Description

We integrate with HubSpot to track interactions with customers of FlowForge Cloud. That includes any emails sent to the customer.

This is separate from the emails the FF platform sends itself; such as the initial welcome email.

We would like to enable the ability to send all customer emails via hubspot (if so configured) so we can track the interactions and assess the effectiveness of the emails.

**Update 7/7/23**: read this comment for the current plan of action: https://github.com/flowforge/flowforge/issues/2021#issuecomment-1624451446

---
**Original description**

This would be an optional configuration via `flowforge.yml`.

Open questions
 - Does this apply to all emails? If the goal is consistency in appearance, then it should - unless we also do #1967 to let us apply more custom templates for our platform-sent emails to achieve consistency between the two.
 - Presumably Hubspot requires the contact to exist already. That implies we need to generate the hubspot contact as part of the sign-up process as well....
 - What does this mean for the 'email change' email we just added - that gets sent to their new address to verify its real? Hubspot won't know about that new address...
 - What configuration is needed for the platform to know how to send a particular email template to a particular user?
 - There are 'transactional' emails around email verification/forgot password etc that need custom content such as link urls - is that possible to achieve via hubspot api?

robmarcer:
@knolleary This feels like it's very close to conversations I've heard over the past few weeks with @iskerrett. I wonder if the goals are not what was intended, or maybe it's just unrelated but sounds similar?

My take is as follows.

Our goal is to move all emails which are not integral to using FF over to Hubspot. By integral, I mean emails such as email verify, password reset, invite a team member. It could be just those three but we'd need to audit all the emails to work out what we feel is 'integral'. 

Everything else would go via Hubspot and we can use our existing integration with Hubspot to trigger those emails. This allows us to iterate on the content, timing and targeting of those emails rapidly as part of trying to do better 'customer success'. 

This would also draw a good line between what FlowForge self hosting users will need to for the platform to function and what we want to be able to send to make FlowForge Cloud successful.

If I'm on the right tack I think the steps would be:

1. Identify all emails being send via FF Cloud.
2. Decide which are 'integral' to all FF installs and which are customer success driven.
3. Set up all the existing non-integral emails to be sent via our existing Hubspot integration.
4. Remove the triggering of those emails via FF's configured SMTP connection.

I would deal with step 3.

[...]

PR #2594, opened 2023-08-15
Removes triggering of emails to trial users

robmarcer:
## Description

I have removed the code which triggers emails to trial teams. The plan is to replace these with emails sent from Hubspot using the existing database integration from FF Cloud to Hubspot.

## Related Issue(s)

[](https://github.com/flowforge/CloudProject/issues/175)

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -338,16 +338,6 @@ module.exports.init = async function (app) {
                     )
                     if (teamTrialInstanceTypeId) {
                         const trialProjectType = await app.db.models.ProjectType.byId(teamTrialInstanceTypeId)
-                        await app.postoffice.send(
-                            user,
-                            'TrialTeamCreated',
-                            {
-                                username: user.name,
-                                teamName: team.name,
-                                trialDuration: teamTrialDuration,
-                                trialProjectTypeName: trialProjectType.name
-                            }
-                        )
                     }
                 }
             }
--- a/forge/ee/lib/billing/trialTask.js
+++ b/forge/ee/lib/billing/trialTask.js
@@ -18,14 +18,10 @@ module.exports.init = function (app) {
                 await subscription.Team.reload({ include: [app.db.models.TeamType] })
                 // The subscription has been setup on Stripe.
                 await app.billing.endTeamTrial(subscription.Team)
-                await sendTrialEmail(subscription.Team, 'TrialTeamEnded')
             } else {
                 app.log.info(`Team ${subscription.Team.hashid} ending trial - suspending instances`)
                 // Stripe not configured - suspend the lot
                 await suspendAllProjects(subscription.Team)
-                await sendTrialEmail(subscription.Team, 'TrialTeamSuspended', {
-                    teamSettingsURL: `${app.config.base_url}/team/${subscription.Team.slug}/billing`
-                })
             }
 
             // We have dealt with this team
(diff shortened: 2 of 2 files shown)


robmarcer:
@knolleary it was suggested I had a go at the edits I think need to be made to stop FF from sending the 4 trial emails. Sorry if I've butchered our code!

forge/ee/lib/billing/index.js
```diff
@@ -338,16 +338,6 @@ module.exports.init = async function (app) {
                     )
                     if (teamTrialInstanceTypeId) {
                         const trialProjectType = await app.db.models.ProjectType.byId(teamTrialInstanceTypeId)
```
Steve-Mcl, line 340:
```suggestion
                        await app.db.models.ProjectType.byId(teamTrialInstanceTypeId)
```

forge/ee/lib/billing/trialTask.js
```diff
@@ -59,12 +55,7 @@ module.exports.init = function (app) {
         for (const subscription of pendingEmailSubscriptions) {
             const endingInDurationDays = Math.ceil((subscription.trialEndsAt - Date.now()) / ONE_DAY)
             const endingInDuration = endingInDurationDays + ' day' + ((endingInDurationDays !== 1) ? 's' : '')
-            const billingUrl = `${app.config.base_url}/team/${subscription.Team.slug}/billing`
-            await sendTrialEmail(subscription.Team, 'TrialTeamReminder', {
-                endingInDuration,
-                billingSetup: subscription.isActive(),
-                billingUrl
-            })
+            const billingUrl = `${app.config.base_url}/team/${subscription.Team.slug}/billing`
```
Steve-Mcl, line 58:
```suggestion
```

This line is now redundant

Steve-Mcl (changes requested):
Some individual comments left.  Also, `function sendTrialEmail` should be removed in its entirety.

forge/ee/lib/billing/index.js
Steve-Mcl: since the code removed leaves an unused variable, the variable should be deleted too Rob.

forge/ee/lib/billing/trialTask.js
Steve-Mcl: As above, since the code removed leaves an unused variable, the variable should be deleted too.

knolleary:
The intention was to add an admin flag to disable these emails rather than delete them entirely.

We have generally avoided hardcoding in FlowForge Cloud specific choices in how the platform runs - based on the idea that we allow people to self-host the platform - and we don't require them to setup any sort of hubspot integration.

I appreciate the billing feature is one that we're likely to be the only people who make use of, but that hasn't been explicitly stated anywhere. If we remove these emails, then the platform will not be compliant with VISA (and others) rules around trial notifications, *unless* additional actions are taken outside of the platform. We will then need to document those requirements for platform admins. Alternatively, we decide the billing feature is one that is only available on FF Cloud and not one we would ever license for a 3rd party to run themselves.

Thoughts @MarianRaphael ?

MarianRaphael:
Absolutely agree, @knolleary . Removing is only one part of the story. A 'disable option' via the admin panel (or config option) should be the right way. Nevertheless, I appreciate @robmarcer  'Bias for Action'. In 1.12, I have capacity planned for further refinements on this issue/PR. (In 1.11 there are too many other strategic issues.)

Pezmc:
@MarianRaphael @robmarcer Should this PR be closed out?

knolleary:
This PR is for https://github.com/flowforge/flowforge/issues/2021 which is planned for this release, but not picked up yet.

This branch could be the starting point, or it may make sense to start clean. Depends on who picks up #2021 - currently assigned to Steve, but aware he has a lot on his plate.

MarianRaphael:
If this https://github.com/flowforge/flowforge/issues/2021 will be implemented through another PR, it's fine to close this PR.

PR #2601, opened 2023-08-16
Add projectComms feature flag to instance settings

knolleary:
## Description

This adds a flag to the TeamType to control whether the Project Nodes feature is available.

This feature isn't exposed in the frontend at all - so the only frontend change is the admin ui for enabling the feature flag for the team type.


Pezmc (approved):

frontend/src/pages/admin/TeamTypes/dialogs/TeamTypeEditDialog.vue
```diff
@@ -142,6 +142,14 @@ export default {
                         this.input.properties.trial.instanceType = '_'
                     }
                     this.input.order = '' + (teamType.order || 0)
+
+                    // Apply default feature values if undefined
+                    if (this.input.properties.features['shared-library'] === undefined) {
+                        this.input.properties.features['shared-library'] = true
+                    }
+                    if (this.input.properties.features.projectComms === undefined) {
```
Pezmc, line 150: The two different syntaxes (square brackets vs dot notation) threw me slightly here. Why are the two features not using the same format? (snakeCase vs kebab-case)

knolleary: Because the underlying feature flags were originally added with this inconsistency.

PR #2604, opened 2023-08-16
Add HA feature flag to team

knolleary:
## Description

This adds a flag to the TeamType to control whether the HA feature is available.

In the UI, if a team does not have access to HA they see this:

image

As with #2599, the link to upgrade your team doesn't yet go anywhere.

Unlike the other feature flags, no changes are needed in the launcher as we can use this the flag to inhibit the existing `ha` config object from being passed down to the launcher.

### DIFF
--- a/forge/ee/routes/ha/index.js
+++ b/forge/ee/routes/ha/index.js
@@ -14,6 +14,11 @@ module.exports = async function (app) {
                         reply.code(404).send({ code: 'not_found', error: 'Not Found' })
                         return
                     }
+                    const teamType = await request.project.Team.getTeamType()
+                    if (!teamType.getFeatureProperty('ha', true)) {
+                        reply.code(404).send({ code: 'not_found', error: 'Not Found' })
+                        return // eslint-disable-line no-useless-return
+                    }
                     if (request.session.User) {
                         request.teamMembership = await request.session.User.getTeamMembership(request.project.Team.id)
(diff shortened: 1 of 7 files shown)


forge/ee/routes/ha/index.js
```diff
@@ -14,6 +14,11 @@ module.exports = async function (app) {
                         reply.code(404).send({ code: 'not_found', error: 'Not Found' })
                         return
                     }
+                    const teamType = await request.project.Team.getTeamType()
+                    if (!teamType.getFeatureProperty('ha', true)) {
+                        reply.code(404).send({ code: 'not_found', error: 'Not Found' })
```
Pezmc, line 19: 403 feels like a better fit here

frontend/src/pages/instance/Settings/HighAvailability.vue
```diff
@@ -66,8 +70,13 @@ export default {
         }
     },
     computed: {
+        ...mapState('account', ['team', 'features']),
         isHA () {
             return this.instance?.ha?.replicas !== undefined
+        },
+        haFeatureAvailable () {
+            const flag = this.team.type.properties.features?.ha
+            return flag === undefined || flag
```
Pezmc, line 79: I believe this could be `flag ?? true`, but it's perhaps easier to read as it's written right now

Pezmc (approved):
Approved but with two suggested improvements

PR #2605, opened 2023-08-17
Allow admin to change sso_enabled users email

knolleary:
## Description

This allows an admin user to modify the email of an sso_enabled user.

As part of doing this, it will clear the sso_enabled flag on the user - as their new email may no longer be covered by an existing sso configuration. The flag will get re-set the next time they log in via sso with the new email.

## Checklist

### DIFF
--- a/test/unit/forge/routes/api/users_spec.js
+++ b/test/unit/forge/routes/api/users_spec.js
@@ -26,6 +26,7 @@ describe('Users API', async function () {
         // elvis <-- this user doesn't have email_verified
 
         // fred <-- this user only gets created in the 'delete' tests. Do not add elsewhere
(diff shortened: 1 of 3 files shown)


test/unit/forge/routes/api/users_spec.js
```diff
@@ -331,6 +332,25 @@ describe('Users API', async function () {
                 const result = response.json()
                 result.should.have.property('error')
             })
+            it.only('admin can modify sso_enabled user email', async function () {
```
hardillb, line 335:
```suggestion
            it('admin can modify sso_enabled user email', async function () {
```

hardillb (commented):
`only` needs removing

hardillb (approved):

PR #2611, opened 2023-08-17
Add team-type feature flag on ff-auth for http endpoints

knolleary:
## Description

This adds a flag to the TeamType to control whether the FF Team auth option is available.

In the UI, if a team does not have access to the option they see this:

image

Notes:

 - Includes a fix ff-radio-group to address https://github.com/flowforge/forge-ui-components/issues/117 that is required for this feature
 - The banner has been updated so we can include the feature name in the text of the banner. Helpful when it applies to one option on the page, rather than the whole page. Not 100% on using the banner like this, but it will do for now.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/banners/FeatureUnavailableToTeam.vue
+++ b/frontend/src/components/banners/FeatureUnavailableToTeam.vue
@@ -5,7 +5,7 @@
     >
         <SparklesIcon class="ff-icon mr-2" style="stroke-width: 1px;" />
         <div>
-            This feature is not available for your current Team. Please <a class="ff-link" href="#" target="_blank" rel="noopener noreferrer">upgrade</a> your Team in order to use it.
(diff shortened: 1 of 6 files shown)


frontend/src/components/banners/FeatureUnavailableToTeam.vue
```diff
@@ -5,7 +5,7 @@
     >
         <SparklesIcon class="ff-icon mr-2" style="stroke-width: 1px;" />
         <div>
-            This feature is not available for your current Team. Please <a class="ff-link" href="#" target="_blank" rel="noopener noreferrer">upgrade</a> your Team in order to use it.
+            {{ featureName }} is not available for your current Team. Please <a class="ff-link" href="#" target="_blank" rel="noopener noreferrer">upgrade</a> your Team in order to use it.
```
Pezmc, line 8: As per previous PR, we should drop the `a` or add links before this is merged to main

PR #2621, opened 2023-08-20
Allow a device to be assigned to application

Steve-Mcl:
## Description

Permit a device to be assigned to an application.

NOTE: Raised in draft to permit other engineers to test - before writing a whole bunch of tests and updating docs.

Notable:
1. when a device is added to an application, it is offered a fake snapshot with ID '0', computed (relevant) Env Vars and a modules object containing `"node-red": "latest"`.  This permits the device to start

#### Application level device implementation:
 - Developer Mode / Remote Access ✅
 - Device Logs ✅
 - Team Library ✅
 - Theme, device name and NR menu links ✅
 - Adding device in application view adds device to application ✅
 - Newly added device in application view updates status when config is applied to device and it starts up ✅
 - Upload Snapshot ❌
 - Download Snapshot ❌
 - Project Links ❌
 - Grouping and settings ❌
- Module cache for device at application level  ❌

#### TODO (this iteration):
 - [x] tests
 - [x] docs
 - [ ] manual testing combinations - to ensure there are no issues when running older platform with newer agent and vice versa
  
#### TODO (next iteration)
 - Provisioning Tokens for auto add to application
 - Module cache for device at application level
 - Support snapshots (upload/download)
 - Support grouping
 - Group settings (env/modules/etc)
 - Project Links (Device links? Concept?)

## Related Issue(s)

[Devices to be assigned to an Application, without being assigned to an Instance#2619](https://github.com/flowforge/flowforge/issues/2619)

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


docs/user/team/README.md
```diff
@@ -64,6 +64,7 @@ table summaries what actions are available to the different roles.
 | • Modify Device Settings             | ✓     | -      | -      | -              |
 | • Modify Environment Variables       | ✓     | ✓      | -      | -              |
 | • Assign to/Remove from Instance     | ✓     | -      | -      | -              |
+| • Assign to/Remove from Application  | ✓     | -      | -      | -              |
 | • Delete Device                      | ✓     | -      | -      | -              |
```
knolleary, line 68:
```suggestion
| • Modify Environment Variables       | ✓     | ✓      | -      | -              |
| • Assign to/Remove from Application  | ✓     | -      | -      | -              |
| • Assign to/Remove from Instance     | ✓     | -      | -      | -              |
| • Delete Device                      | ✓     | -      | -      | -              |
```

Think application logically comes above Instance.

forge/comms/aclManager.js
```diff
@@ -91,7 +103,10 @@ module.exports = function (app) {
                 { topic: /^ff\/v1\/[^/]+\/d\/[^/]+\/command$/ },
                 // Send commands to all project-assigned devices
                 // - ff/v1/+/p/+/command
-                { topic: /^ff\/v1\/[^/]+\/p\/[^/]+\/command$/ }
+                { topic: /^ff\/v1\/[^/]+\/p\/[^/]+\/command$/ },
+                // Send commands to all application-assigned devices
+                // - ff/v1/+/p/+/command
```
knolleary, line 108:
```suggestion
                // - ff/v1/+/a/+/command
```

forge/db/controllers/Device.js
```diff
@@ -75,11 +85,24 @@ module.exports = {
             return { name, value: value || '', platform: true } // add `platform` flag for UI
         }
         const result = []
+        let snapshotId
+        let snapshotName
+
+        if (device.ownerType === 'application') {
+            snapshotId = '0' // '0' is temporary value to indicate that the device should start in application mode with starter flows
+            snapshotName = 'None'
```
knolleary, line 93: I wonder if we should set them blank rather than these non-blank values. I appreciate why you set it to `0` for the internal usage between platform and agent, but in the flows, I feel like a blank value would be clearer so we don't have to explain it away otherwise.

forge/db/migrations/20230816-01-add-ApplicationId-to-Devices.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable no-unused-vars */
```
knolleary, line 1: This file needs to be renamed to ensure it is the most recent migration in the list.

At the time of commenting, the most recent migration is `20230818-01-add-access-token-name.js`

forge/routes/api/application.js
```diff
@@ -276,9 +276,49 @@ module.exports = async function (app) {
...
+     * Get a list of devices owned by this application
+     * @name /api/v1/applications/:id/devices
+     * @static
+     * @memberof forge.routes.api.application
+     */
+    app.get('/:applicationId/devices', {
+        preHandler: app.needsPermission('team:device:list'),
+        schema: {
+            summary: 'Get a list of all devices in an application',
+            tags: ['Application Devices'],
```
knolleary, line 289:
```suggestion
            tags: ['Applications'],
```

Not sure a separate section is needed - keeping it consistent with the /instances end point for eg

forge/routes/api/device.js
```diff
@@ -334,36 +335,52 @@ module.exports = async function (app) {
...
+                    device.mode = 'autonomous'
+                    await device.save()
+                }
+
+                if (device.Application !== null) {
+                    // const oldApplication = device.Application
+                    // unassign from application
+                    await device.setApplication(null)
+                    await commonUpdates()
+                    // TODO
```
knolleary, line 360: Are these genuine todos or left overs?

forge/routes/api/device.js
```diff
@@ -653,5 +693,13 @@ module.exports = async function (app) {
         device.targetSnapshotId = deviceSettings?.targetSnapshot
         return true
     }
+    async function assignDeviceToApplication (device, application) {
+        await device.setApplication(application)
+        // Set the target snapshot to match the project's one
+        // TODO
```
knolleary, line 699: I assume this means 'todo later once we figure out how to handle application settings (if at all)' rather than needed for 1st iteration?

forge/routes/api/deviceLive.js
```diff
@@ -74,6 +84,36 @@ module.exports = async function (app) {
...
+                        { id: 'FFCHA00000000001', type: 'change', z: 'FFF0000000000001', name: 'Get Env Vars', rules: [{ t: 'set', p: 'payload', pt: 'msg', to: '{}', tot: 'json' }, { t: 'set', p: 'payload.device', pt: 'msg', to: 'FF_DEVICE_NAME', tot: 'env' }, { t: 'set', p: 'payload.application', pt: 'msg', to: 'FF_APPLICATION_NAME', tot: 'env' }], action: '', reg: false, x: 320, y: 160, wires: [['FFDBG00000000001']] },
+                        { id: 'FFDBG00000000001', type: 'debug', z: 'FFF0000000000001', name: 'Info', active: true, tosidebar: true, console: true, tostatus: true, complete: 'payload', targetType: 'msg', statusVal: 'payload', statusType: 'auto', x: 490, y: 160 }
+                    ],
+                    modules: {
+                        'node-red': 'latest' // TODO: get this from the "somewhere!?!?" - this is where TAGs might work well.
+                    },
+                    env: {
+                        FF_SNAPSHOT_ID: '0',
+                        FF_SNAPSHOT_NAME: 'None',
```
knolleary, line 107:
```suggestion
                        FF_SNAPSHOT_ID: '0',
                        FF_SNAPSHOT_NAME: 'None',
```

Same comment as elsewhere - I wonder if, when exposing in the flow, these should be blank rather than placeholders. Not sure.

forge/db/migrations/20230816-01-add-ApplicationId-to-Devices.js
```diff
@@ -0,0 +1,23 @@
...
+     * upgrade database
+     * @param {QueryInterface} context Sequelize.QueryInterface
+     */
+    up: async (context) => {
+        await context.addColumn('Devices', 'ApplicationId', {
+            type: DataTypes.INTEGER,
+            references: { model: 'Applications', key: 'id' },
+            onDelete: 'set null',
+            onUpdate: 'cascade'
+        })
```
knolleary, line 19: Does the migration need to also populate the `ApplicationId` column for all existing devices?

frontend/src/pages/device/dialogs/ModeChoiceDialog.vue
```diff
@@ -5,7 +5,7 @@
                 <template v-if="unsupportedVersion">
                     <p>
                         Developer Mode requires Device Agent v0.8.0 or later.
-                        Please update your Device Agent to the latest version.
+                        Please ensure this device has checked in at least once and that it is running the latest version of Device Agent.
```
knolleary, line 8: I understand the intent, my concern is the user doesn't know how to make the device check in. Does the following help?

```suggestion
                        Please ensure the Device Agent is up to date and connected to the platform.
```

knolleary (changes requested):
Looks good overall. Some comments and a couple bits that need updating - particularly around the migration.

forge/db/migrations/20230816-01-add-ApplicationId-to-Devices.js
Steve-Mcl: No. As per existing devices linked to an instance, the `project` object has its own relationship to Application - this remains.

The presence of this property is to signify a device is lined to an application not a project.

forge/db/controllers/Device.js
Steve-Mcl: This is _very_ temporary Nick. 

Happy to use a name or something other than 'None'

Are you proposing using `Blank` or `Something-else` as the name?

Are you also proposing using something other than `0` for the ID?  I was torn between somthing like `DEFAULT` ot `APPLICATION-SNAPSHOT` but ultimately, the temporary nature of this, I went with '0'

forge/routes/api/application.js
Steve-Mcl: Should this then not be `['Devices']` as it returns devices?

Not sure now that I understood the relevance of tags 🤔

forge/routes/api/device.js
Steve-Mcl: Future TODOs - in line with other TODOs already in surrounding tests.

TBH, they can be removed as, strictly speaking,  they are really separate unit test related to audit logging.

frontend/src/pages/device/dialogs/ModeChoiceDialog.vue
Steve-Mcl: better :)

forge/db/migrations/20230816-01-add-ApplicationId-to-Devices.js
Steve-Mcl: @knolleary resolved in [9f9b53f](https://github.com/flowforge/flowforge/pull/2621/commits/9f9b53f46ef545d82102e3ff9974ec6725ca456f)

forge/db/controllers/Device.js
knolleary: In terms of the env vars on the device, I was meaning literally blank - ''.

forge/routes/api/application.js
knolleary: The tags determine where it shows up in the swagger doc output - https://app.flowforge.com/api/static/index.html

This is an application based API so should be under applications.

forge/db/migrations/20230816-01-add-ApplicationId-to-Devices.js
knolleary: Does this mean the `/applications/<id>/devices` endpoint only returns the devices assigned to just the application? I was working on the assumption it returned the superset of devices assigned to the application directly and devices assigned to instances within the application?

forge/db/controllers/Device.js
Steve-Mcl: hmmm. Would have to double check the logic across platform and device. 

For example, `'0' is used as a catch point in the device logic to know that A) the snapshot is _something_ and B) it is a _temporary/starter_snapshot for the device.  Existing logic for checking there IS a snapshot then works as is (because `.id` of '0' is _something_ and it signifies an application starter snapshot)

I do suspect this would cause some inconsistencies as it stands right now.

Happy to change - or move forward as is.  Please advise.

forge/db/migrations/20230816-01-add-ApplicationId-to-Devices.js
Steve-Mcl: Apologies if that has been lost in translation. The current implementation is literal i.e. "Allow a Device to be assigned to Application" as per my understanding and interpretation. 

As it happens, I am not totally sold on your counter proposal - if feels odd somehow?  Perhaps with a rejig of the UI that might feel right but in the current form, it more _feels_ right to me. 

And of course, as this all changes again when grouping is introduced, I/you might feel different about one vs other? And anyhow, the design stage should route this out and I will be sure to ask for clarification to minimise confusion.

forge/db/controllers/Device.js
knolleary: I would have chosen to leave them blank as there is no running snapshot and I don't think the scenario of an existing device using logic based on this env var wanting to migrate to this new mode instantly (without snapshots in place) is really enough of a concern.

But I'm not going to push this.

forge/db/migrations/20230816-01-add-ApplicationId-to-Devices.js
knolleary: It's a matter of perspective. Instances are owned by an Application. If a Device is assigned to an instance then it is 'owned' by the parent Application. So a view of devices in the application would include those that have been further assigned to individual instances. But regardless, as I said, we can proceed with what you have and I'll write up a proposed set of changes next week we can discuss in the broader context.

forge/db/migrations/20230816-01-add-ApplicationId-to-Devices.js
knolleary: Not resolved - we now how two migration files :) Need to delete the old one.

knolleary (changes requested):

forge/db/migrations/20230816-01-add-ApplicationId-to-Devices.js
Steve-Mcl: oh, odd, i did a rename of file and (wrongly) expected that to remove the old one. 

Will fix now.

knolleary:
I may have missed it, but is there any guarding against assigning an older device agent to an application? This is only going to work with agents that are running with https://github.com/flowforge/flowforge-device-agent/pull/161

forge/db/migrations/20230816-01-add-ApplicationId-to-Devices.js
Steve-Mcl: Double odd

image

Steve-Mcl: Doh! 👐 

image

Steve-Mcl:
> I may have missed it, but is there any guarding against assigning an older device agent to an application

Yeah, there is an agent version check.  And the reason for updating the vue page comment about "Device needs to check in" - so that it knows it is permitted to assign the device to an application.

let me double verify - 5 mins.

Steve-Mcl:
> > I may have missed it, but is there any guarding against assigning an older device agent to an application
> 
> Yeah, there is an agent version check. And the reason for updating the vue page comment about "Device needs to check in" - so that it knows it is permitted to assign the device to an application.
> 
> let me double verify - 5 mins.

Urg, seems I lost that bit of work. 

please hold off until I find / merge / re-add.

knolleary:
Hmm, I can no longer find any trace of the conversation thread in this PR related to the `applications/id/devices` api... which is going to make it hard to refer back to.

Adding a note that, whilst reviewing Device Agent changes, I can see lots of assumptions based on `application` only being set if assigned to an application and not an instance. So any future change that sees `application` set if assigned to either the app or an instance within the application will be hard to retrofit.

forge/db/migrations/20230816-01-add-ApplicationId-to-Devices.js
knolleary: Looks good - resolving this thread.

Steve-Mcl:
@knolleary pretty sure this is now ready for re-review. Had some issue with merge conflict after fixing tests locally & pushing  but this should all be resolved now.

Steve-Mcl:
small buglet found while testing the changes requested in the partner PR on the device agent.

Resolving this now & pushing update to agent imminently.

frontend/src/pages/team/Devices/index.vue
```diff
@@ -3,16 +3,16 @@
...
                 </template>
                 <template #helptext>
-                    <p>FlowFuse can be used to manage instances of Node-RED running on remote devices.</p>
-                    <p>Each device must run the <a href="https://flowfuse.com/docs/user/devices/" target="_blank">FlowFuse Device Agent</a>, which connects back to the platform to receive updates.</p>
-                    <p>Devices are registered to a Team, and assigned to an Instance within an Application.</p>
-                    <p>Flows can then be deployed remotely to the devices as an Instance Snapshot.</p>
+                    <p>FlowForge can be used to manage instances of Node-RED running on remote devices.</p>
+                    <p>Each device must run the <a href="https://flowforge.com/docs/user/devices/" target="_blank">FlowForge Device Agent</a>, which connects back to the platform to receive updates.</p>
+                    <p>Devices are registered to a Team, and assigned to an Application or an Instance.</p>
+                    <p>Flows can then be deployed remotely to the devices as a Snapshot.</p>
```
knolleary, line 15:
```suggestion
                    <p>FlowFuse can be used to manage instances of Node-RED running on remote devices.</p>
                    <p>Each device must run the <a href="https://flowfuse.com/docs/user/devices/" target="_blank">FlowFuse Device Agent</a>, which connects back to the platform to receive updates.</p>
                    <p>Devices are registered to a Team, and assigned to an Application or an Instance.</p>
                    <p>Flows can then be deployed remotely to the devices as a Snapshot.</p>
```

knolleary (approved):

PR #2634, opened 2023-08-23
Support for changing team type

### LINKED ISSUE #2532: TeamTypes: enable upgrade path
### Description

Part of #2359 

We currently do not provide a way for changing team types. This is a requirement before we can launch the new team types in FF Cloud.

This will require an api endpoint to apply the change. This will need to check the change is valid. There are a number of scenarios (typically when 'downgrading') where the change will not be valid. For example:

 - A Team is exceeding a limit (users/devices/instances of a type)
 - A Team is using a feature that is not available in the target type. This could get quite hard to detect. For example, the limits on context/file storage - I don't think we have a way to query the current usage.

If billing is enabled, all of the stripe price/product ids need switching over. These include:
 - the Team billing item
 - the Device billing item
 - the billing items for each instance type

Note that if a team is in trial mode, the Team billing item may be using the trial price/product - this is an edge case that will need handling.

Once the API endpoint is in place, the UI needs implementing. We already have simple 'teamtype' choice when the user creates a new team. This likely needs some addition UX improvement and rework, but gives us a starting point.

### Epic/Story

2359

```[tasklist]
### Tasks
- [x] Create API endpoint to modify team type
- [x] Create UX for modifying Team type
```

knolleary:
## Description

Closes #2532 

This adds the ability to change the Team Type.

```[tasklist]
### Tasks
- [x] Update api to support modifying team type - CE only
- [x] Add audit log entry type for changing team type
- [x] Add Change Type UI - CE only
- [x] Add EE/Billing logic around changing type
- [x] Update Change Type UI to give better billing change feedback
- [ ] Unit Test coverage of API changes
```

### DIFF
--- a/forge/db/models/Team.js
+++ b/forge/db/models/Team.js
@@ -354,6 +354,81 @@ module.exports = {
                  */
                 checkInstanceStartAllowed: async function (instance) {
                     return true
+                },
+
+                /**
+                 * Checks whether the team type can be modified to the requested type.
+                 * The team must be within any limits of the target type.
+                 */
+                checkTeamTypeUpdateAllowed: async function (teamType) {
+                    // console.log(teamType.properties)
+                    await this.ensureTeamTypeExists()
+
+                    // Check the following limits:
+                    // - User count
+                    // - Device count
+                    // - Instance Type counts
+                    const errors = []
+
+                    const currentMemberCount = this.get('memberCount')
+                    const targetMemberLimit = teamType.getProperty('users.limit', -1)
+                    if (targetMemberLimit !== -1 && targetMemberLimit < currentMemberCount) {
+                        errors.push({
+                            code: 'member_limit_reached',
+                            error: 'Member limit reached',
+                            limit: targetMemberLimit,
+                            count: currentMemberCount
+                        })
+                    }
+
+                    const currentDeviceCount = await this.deviceCount()
+                    const targetDeviceLimit = teamType.getProperty('devices.limit', -1)
+                    if (targetDeviceLimit !== -1 && targetDeviceLimit < currentDeviceCount) {
+                        errors.push({
+                            code: 'device_limit_reached',
+                            error: 'Device limit reached',
+                            limit: targetDeviceLimit,
+                            count: currentDeviceCount
+                        })
+                    }
+
+                    const currentInstanceCountsByType = await this.instanceCountByType()
+                    // console.log(currentInstanceCountsByType)
+                    const targetInstanceLimits = {}
+                    for (const instanceType of Object.keys(currentInstanceCountsByType)) {
+                        // console.log('checking', instanceType, teamType.properties.instances[instanceType])
+                        if (!teamType.getInstanceTypeProperty(instanceType, 'active', false)) {
+                            targetInstanceLimits[instanceType] = 0
+                        } else {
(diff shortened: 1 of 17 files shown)


knolleary:
I've added a bit of test coverage for the base model.

The existing test coverage of the `/api/v1/teams` end points is pretty bad. Will look to address that early next week.

Will also work on additional EE test coverage for the billing side.

forge/db/models/Team.js
```diff
@@ -354,6 +354,87 @@ module.exports = {
...
                 checkInstanceStartAllowed: async function (instance) {
                     return true
+                },
+
+                /**
+                 * Checks whether the team type can be modified to the requested type.
+                 * The team must be within any limits of the target type.
+                 */
+                checkTeamTypeUpdateAllowed: async function (teamType) {
+                    // console.log(teamType.properties)
```
Pezmc, line 364: ✂️

forge/db/models/Team.js
```diff
@@ -354,6 +354,87 @@ module.exports = {
...
+                            limit: targetDeviceLimit,
+                            count: currentDeviceCount
+                        })
+                    }
+
+                    const currentInstanceCountsByType = await this.instanceCountByType()
+                    // console.log(currentInstanceCountsByType)
+                    const targetInstanceLimits = {}
+                    for (const instanceType of Object.keys(currentInstanceCountsByType)) {
+                        // console.log('checking', instanceType, teamType.properties.instances[instanceType])
```
Pezmc, line 399: ✂️  Lots of this over the file, I won't comment on each one

forge/db/models/Team.js
```diff
@@ -354,6 +354,87 @@ module.exports = {
...
+                    // }
+                },
+                /**
+                 * Update the team type.
+                 *
+                 * When running with EE, this function is replaced via ee/lib/billing/Team.js
+                 * to add additional checks related to billing
+                 */
+                updateTeamType: async function (teamType) {
+                    await this.setTeamType(teamType)
```
Pezmc, line 435: I'm slightly surprised that this method (updateTeamType) doesn't call `checkTeamTypeUpdateAllowed` and raise an exception if the change can't be made. Feels like it opens up the potential of team type being updated to an invalid state if `checkTeamTypeUpdateAllowed` is forgotten to be called first.

frontend/src/pages/team/Settings/General.vue
```diff
@@ -12,7 +12,10 @@
         <FormRow v-model="input.teamType" type="uneditable">
             <template #default>Type</template>
             <template #description>
-                <div v-if="editing">You cannot currently change the type of team</div>
+                <template v-if="editing">
+                    Click <router-link :to="{name: 'TeamChangeType'}">here</router-link> to change the team type
+                </template>
+                <!-- <ff-button v-if="editing" size="small" :to="{name: 'TeamChangeType'}">Change</ff-button> -->
```
Pezmc, line 18: ✂️

Pezmc (approved):
Some nitpicks, but overall this looks good!

forge/db/models/Team.js
knolleary: All tidied up

frontend/src/pages/team/Settings/General.vue
knolleary, line 18:
```suggestion
```

knolleary:
Still some EE test coverage to add - but moving forward to get this onto staging today to allow parallel testing

PR #2670, opened 2023-08-29
Add banner to DeviceBrowser when device limit reached

knolleary:
## Description

This is part of the Team Types work. It adds the standard banner to the DeviceBrowser component and disabled the 'add device' button if the device limit has been reached.

Except... it doesn't quite work yet. Given the work around Device pagination, the DeviceBrowser is getting a bunch of rework in parallel. The current code doesn't appear to have the 'total number of devices in this team' baked into it. If I go about adding that, it'll insta-clash badly with the pagination work. So this is a placeholder PR to get the bulk of this work in - and the final piece can be updated when the pagination work lands.

Alternatively, if the pagination work is delayed, it will be a quick task to add the right logic here and get this merged and deal with the clashes later.

### DIFF
--- a/forge/db/models/Team.js
+++ b/forge/db/models/Team.js
@@ -104,6 +104,15 @@ module.exports = {
                                         "members"."TeamId" = "Team"."id"
                                     )`),
                                     'memberCount'
+                                ],
+                                [
+                                    literal(`(
+                                        SELECT COUNT(*)
+                                        FROM "Devices" AS "devices"
+                                        WHERE
+                                        "devices"."TeamId" = "Team"."id"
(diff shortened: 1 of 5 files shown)


Pezmc:
My branch has total number of devices in a Vue data prop, so indeed will be easy to grab!

Pezmc:
Can be rebased on top of https://github.com/flowforge/flowforge/pull/2677 and use `allDeviceStatuses.size()`

Pezmc (approved):
Approved on release call.

knolleary:
Reviewed the additional changes from @Pezmc - happy to merge

PR #2686, opened 2023-08-31
Disable filestore by default

hardillb:
## Description

As this is a licensed feature it is already diabled by default, but the localfs installer does not include the file-server. So When you add a license to the localfs build it enables persistent context and fails to start instance.

With this off by default it will require file-server to be installed and started and configured.

Will look at bundling file-server in installer, but this is best safe quick option

## Related Issue(s)

Mentioned on NR Slack
https://node-red.slack.com/archives/C02V3BN77GU/p1693347012101579

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/docs/install/local/README.md
+++ b/docs/install/local/README.md
@@ -233,6 +233,25 @@ you can use a pre-built docker image that provides everything needed.
 
     This will map the `1883`/`1884` ports to the host machine so they can be accessed outside of the container. If you already have an MQTT broker running on port 1883, then you'll need to modify the `-p` options to use a different set of ports. For example: `-p 9883:1883 -p 9884:1884`.
 
+## File Server
+
+By default the FlowFuse File Server component is disabled as it is a licensed feature. If you provide a license you can start the File Server with the following command:
+
+```bash
+sudo service flowforge-file start
+```
+
+You can then uncomment the following section in the `/opt/flowforge/etc/flowforge.yml` file
+
+```yaml
+#################################################
+# File Server config                            #
+#################################################
+
+fileStore:
+  url: http://localhost:3001
+
+```
 
 ## Upgrade
 
--- a/etc/flowforge.yml
+++ b/etc/flowforge.yml
@@ -71,8 +71,8 @@ driver:
 # File Server config                            #
 #################################################
 
-fileStore:
-  url: http://localhost:3001
+# fileStore:
+#   url: http://localhost:3001
 
 
 #################################################


hardillb:
I need to work out what to put in the documentation for this.

Steve-Mcl:
> I need to work out what to put in the documentation for this.

will docs update be part of the PR Ben?

hardillb:
Doh, knew I'd forgotten something, will look at it on Monday.

hardillb:
@Steve-Mcl ready

docs/install/local/README.md
```diff
@@ -233,6 +233,25 @@ you can use a pre-built docker image that provides everything needed.
...
+
+```yaml
+#################################################
+# File Server config                            #
+#################################################
+
+fileStore:
+  url: http://localhost:3001
+
+```
```
Steve-Mcl, line 254: Ben, should these not be reversed? (i.e. setting YAML BEFORE launching service)


**Unfortunately, due to multiple code fences, I cannot do this as a suggestion (it messes up the code suggestion)**

Here is my proposal:

<pre>
By default the FlowFuse File Server component is disabled as it is a licensed feature.
If you provide a license you can start the File Server with the following instructions:

Uncomment the `fileStore` keys in the `/opt/flowforge/etc/flowforge.yml` file:

```yaml
#################################################
# File Server config                            #
#################################################

fileStore:
  url: http://localhost:3001
```

Then you can start the File Server with the command:

```bash
sudo service flowforge-file start
```
</pre>

hardillb: NO!

hardillb: Because you have to start the file-server before you can try and use it

Steve-Mcl: Ah, soory, this wasnt clear at all.

Now you have emphatically stated NO I thought about it (and read it again) and realised that the yml file you are talking about is NOT for THIS application but actually for the FlowFuse application.  That makes sense now, but I am suspect that will not be clear to others either.

Question: Should we also state that FlowFuse application will need a restart after modifying `/opt/flowforge/etc/flowforge.yml`?  As it stands, this implies simply updating the yml is enough.


> Because you have to start the file-server before you can try and use it

Should that be included in docs?


> If you provide a license you can start the File Server with the following command

That reads like "If i provide a license to the File Server it will work". Should we not be clear about (or link to) the licensing? 

---

I got tripped up, imagine those reading this for the first time 🤷‍♂️

hardillb: You already have to restart the forge app after applying a license to enable the features so it's covered.

Steve-Mcl: ok

Steve-Mcl (approved):

PR #2701, opened 2023-09-05
Chore: Update some of the attribute order lint

hardillb:
## Description

Reduce lint exceptions for argument order.


Pezmc (approved):
I skimmed over this and caught nothing, likely this was done with autofix so I'm not going to scrutinise further.

PR #2708, opened 2023-09-07
Change default device command timeout to 10s

knolleary:
## Description

This PR increases the command timeout for devices from 5s to 10s.

We have seen a case where a command is taking _just_ over 5s to respond due to a device on a slow network connection.

Increasing this timeout should give plenty of headroom for slow devices without impacting the user experience too much in the cases where a device is unresponsive and genuinely needs to be timed out and feedback to the user.

### DIFF
--- a/forge/comms/devices.js
+++ b/forge/comms/devices.js
@@ -6,7 +6,7 @@
 const { v4: uuidv4 } = require('uuid')
 
 const noop = () => {}
-const DEFAULT_TIMEOUT = 5000
+const DEFAULT_TIMEOUT = 10000
 
 // declare command and response monitor types (and freeze them)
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):
👍 

I assume the failed PG test is the memory thing - will leave for you to judge and merge

knolleary:
Given the comment in the device agent side of this, there maybe an additional change to make on the `No tunnel` error code. Will make it so this morning.

https://github.com/flowforge/flowforge-device-agent/pull/169#pullrequestreview-1615564105

knolleary:
Have raised a separate PR for modifying the error code to keep things clean.

Merging this PR to update the timeout.

PR #2711, opened 2023-09-08
Update websocket close codes to proper range

knolleary:
## Description

Partner PR to #2708 and https://github.com/flowforge/flowforge-device-agent/pull/169

In reviewing our WS close codes, I found we were using reserved codes from the `100x` range. Application specific codes should be in the unreserved space `4000+`.

As such, I've modified *some* of the return codes here that are relevant to the issue we're fixing to use distinct `400x` level codes.

I'll raise a follow up task to fully review and update the remaining. We have no code (other than newly added by the above linked PR) that cares about the actual code value, so we can change these without issue.


Steve-Mcl (approved):

Steve-Mcl:
Will merge depending on tests passing

knolleary:
Fixing tests....

Steve-Mcl (approved):

Steve-Mcl:
Failed PG test is `heap limit Allocation failed` - happy to accept the non PG test passes

PR #2716, opened 2023-09-08
Move pg and fastify routes to runtime dependencies

Pezmc:
## Description

Previously these were erroneously listed in dev dependencies, which we use for _build time_ not run time.

## Related Issue(s)

n/a

## Checklist

 - [ ] Suitable unit/system level tests have been added and they pass - NA
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
>Previously these were erroneously listed in dev dependencies, which we use for build time not run time.

@Pezmc Is that sentence upside down?

I see you have moved `pg` and `@fastify/routes` to `dependencies` but claim "we use for build time not run time"

It may just be grammatical or my (mis)understanding!

Also, curious as to how this has worked up until now? Unless everone is doing `npm install` instead of `npm install --omit=dev` or `npm install --production`

Pezmc:
Yep confusing words there, I meant that currently, for FlowFuse:

- Dependencies = Used at run time
- Dev Dependencies = Used at build time or for development

PG and Fastify/routes are used a run time, so should be in the that section.

I believe, because we _build_ on the sever, we always do full installs. It's open to interpretation as to whether build dependencies should live in dev or not (I think I lean towards not but we are where we are).

Steve-Mcl (approved):

PR #2720, opened 2023-09-08
Update sequelize 

Pezmc:
As part of updating moment-timezone above 0.5.34

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
                 "pg": "^8.11.3",
                 "pino-pretty": "^10.0.0",
                 "semver": "~7.5.4",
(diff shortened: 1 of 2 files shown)


Steve-Mcl (approved):

Steve-Mcl:
bah - conflicts @Pezmc

Pezmc:
On it

Pezmc:
Rebased on top of `main`

Steve-Mcl:
@Pezmc I realise this is the familiar "limit Allocation failed" error, but as this is related to DB stuff, I think we should try to get at leasy 1 pass on PG tests?

PR #2734, opened 2023-09-11
Pin reusable workflow to commit SHA

ppawlowski:
## Description

Pin reusable workflow to commit SHA instead of branch name.

## Related Issue(s)

[220](https://github.com/flowforge/CloudProject/issues/220)

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -25,7 +25,7 @@ jobs:
       ( github.event.workflow_run.conclusion == 'success' && github.ref == 'refs/heads/main' ) ||
       ( github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/main' ) ||
       github.event_name == 'schedule'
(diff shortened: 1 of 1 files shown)


Pezmc:
@ppawlowski Is there a reason to use a SHA rather than a Git tag? I could see you instead using a version like v0.0.5? Perhaps this isn't possible with GH workflows?

ppawlowski:
> @ppawlowski Mostly out of curiosity, is there a reason to use a SHA rather than a Git tag? I could see you instead using a version like v0.0.5? Perhaps this isn't possible with GH workflows?

Definitely, tags are more readable and in the future I would like to create automatic tagging/releasing for our GitHub reusable workflows and use tags to reference them instead of commit SHA.
However, since I am introducing breaking changes into existing workflows, and would like to have control over when each pipeline is going to use a new approach, I switched to SHA for now.

knolleary (approved):

PR #2736, opened 2023-09-12
Further Postgres memory limit increase

Pezmc:
## Description

Possible fix for the latest round of OOM errors in our progress tests.

- Instead npm memory limit to 6gb
- Configure vitest to ignore coverage directory

## Related Issue(s)

https://github.com/flowforge/flowforge/issues/1960

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/.github/workflows/test-postgresql.yml
+++ b/.github/workflows/test-postgresql.yml
@@ -57,5 +57,5 @@ jobs:
           FF_TEST_DB_POSTGRES_USER: postgres
           FF_TEST_DB_POSTGRES_PASSWORD: secret
           FF_TEST_DB_POSTGRES_DATABASE: flowforge
-          NODE_OPTIONS: "--max_old_space_size=4096"
+          NODE_OPTIONS: "--max-old-space-size=6144"
         run: npm run test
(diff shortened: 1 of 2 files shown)


Steve-Mcl (approved):
discussed the exclude option added with Pez. I understand there is a chance, due to vitest being a JS "thing" that due to the size of the coverage/** dir, it may impact old-space.

happy to merge.

.github/workflows/test-postgresql.yml
```diff
@@ -57,5 +57,5 @@ jobs:
           FF_TEST_DB_POSTGRES_USER: postgres
           FF_TEST_DB_POSTGRES_PASSWORD: secret
           FF_TEST_DB_POSTGRES_DATABASE: flowforge
-          NODE_OPTIONS: "--max_old_space_size=4096"
+          NODE_OPTIONS: "--max-old-space-size=6144"
```
knolleary, line 60: AIUI this controls when GC kicks in, rather than being about allocating more memory to the process. The previous lower value would mean GC is kicking in earlier and reclaiming memory.

Pezmc: As to why the process crashes with the GC runs 🤷

PR #2743, opened 2023-09-13
Fix create TeamType dialog

knolleary:
## Description

The Create TeamType dialog wasn't behaving for two reasons:

 - the event handler on the button to open the dialog was passing through the MouseEvent which was being misinterpreted as a TeamType to be edited
 - the dialog was looking accessing properties of an undefined on the blank TeamType object


knolleary:
Workingon some tests for this view now... will be fairly minimal, but better than we have.

knolleary:
Simple tests for create/edit/delete added.

PR #2758, opened 2023-09-14
Add invoice proration option to TeamType

### LINKED ISSUE #2746: Add option for monthly invoices rather than triggering invoices on every billing change
### Description

We currently ask stripe to generate new invoices with every billing interaction. That can lead to many emails and receipts - which is inconvenient to users.

There is a requirement to move to monthly invoicing. This issue is to track the work needed to achieve that.

First task is to investigate the Stripe api to understand how the existing invoicing behaviour is being triggered - and to identify how to modify that behaviour.

Need to clarify if this should be a whole-sale change to how our billing operates, or only for certain TeamTypes. If the latter, will need a flag adding on the TeamType to indicate the desired billing mechanism.

### Epic/Story

_No response_

knolleary:
Fixes #2746 

## Description

This adds a new option to the TeamType to set the invoice proration behaviour.

image

The default option is the existing behaviour - triggering an invoice with every change made.

The new option, 'Add proration items to monthly invoice', will mean the team invoice is still set monthly, but will list all of the changes made through the month and the protated charges.

This also addresses an issue where deleting an Instance would not remove it properly from billing.
The unit tests have been updated to match how the relevant parts of the code are called to ensure they properly match the real behaviour - specifically, not deleting the Project database object before calling `billing.removeProject`.

## Checklist

### DIFF
--- a/forge/containers/wrapper.js
+++ b/forge/containers/wrapper.js
@@ -150,14 +150,14 @@ module.exports = {
      * @returns {Promise} Resolves when the project has been stopped
      */
     remove: async (project) => {
-        if (project.state !== 'suspended') {
+        if (this._driver.remove) {
+            await this._driver.remove(project)
+        }
+        if (!project.state !== 'suspended') {
             // Only updated billing if the project isn't already suspended
             if (this._isBillingEnabled()) {
                 await this._subscriptionHandler.removeProject(project)
             }
-        }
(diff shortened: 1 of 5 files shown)


Pezmc (approved):
I've been through this here and locally and nothing has stood out, but there should be some extra validation on staging before we deploy.

PR #2772, opened 2023-09-18
Chore: Remove all linting exceptions from admin pages

knolleary:
## Description

Removes all linting exceptions from admin pages.

Most done with auto-fixing.

A bit of work done to add proper types/defaults to `props` properties. Spent some time clicking around the relevant admin pages and verified I'd got the types right. There are still some lingering vue warnings about prop types in other components (specifically `FormRow` when it wants a boolean but gets a null) - but we'll address that separately.


Pezmc:
@knolleary Is it feasible to spit this into separate commits? I'm finding it hard near-on impossible to review the manual and the auto-fix changes at the same time

knolleary:
@Pezmc I'll redo... ignore for now

knolleary:
@Pezmc  have force pushed the same changes split over 5 commits. I've tried to keep the commits 'of a kind' to help reviewing.

knolleary:
- [Remove exceptions from eslintrc](https://github.com/flowforge/flowforge/pull/2772/commits/f2246a64dc0aeb67d90d239d7196f064dea00e00)

 - [Remove linting exceptions from inside <template> sections](https://github.com/flowforge/flowforge/pull/2772/commits/623057a1aca00daa0b0f4c11205ad26aed4e120f) - all auto-fixed reordering and whitespace

 - [Remove unused Permissions vue](https://github.com/flowforge/flowforge/pull/2772/commits/b14fe1112212987edeae99f17a83657b01e41c2e) - an unused component

 - [Reorder vue component properties](https://github.com/flowforge/flowforge/pull/2772/commits/241d39b96ef8788cb9276680535837cfd2aa246b) - getting the component properties in the right order

 - [Ensure props have types and defaults](https://github.com/flowforge/flowforge/pull/2772/commits/43927368aea05b476472ed1d2b5bfb2550e95c9b) - fixing up the `props` property

There are a couple cases where the `props` property is modified in the penultimate commit, or a reorder is done in the last commit... just a result of the changes being too close to separate out. Tried to minimise it.

Pezmc:
That'll do, for next time around, I think the changes I would like to see separate is anything done manually, the auto-fix can be mostly trusted, but the manual work is more likely to have mistakes.

The way I've done it previously is remove the disables one by one, running auto-fix and committing in between, then doing any manual fix as a final step.

Pezmc (approved):

PR #2788, opened 2023-09-20
Chore: remove all eslint self-closing tag exceptions

hardillb:
## Description

Remove all the html-self-close-tag exceptions

Also re-ordered a couple of vue-in-order-components

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -205,41 +205,9 @@
         },
         {
             "rules": {
-                "vue/html-self-closing": "off"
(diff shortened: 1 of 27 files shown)


Pezmc:
> Also re-ordered a couple of vue-in-order-components

It's much easier to review if only a single eslint rule is fixed per commit. In this case it's not too long and looks to all be auto-fix, so 👍

Pezmc (approved):

PR #2836, opened 2023-09-27
Add option not to create NR instance when creating Application

### LINKED ISSUE #1986: Feature: Introduce a way to create an application without an instance
Right now, it's only possible to create an application with a first instance. 

Creating an application without a first instance would make managing applications easier long term.

knolleary:
Closes #1986 

## Description

Adds a checkbox to the Create Application page to optionally *not* create a Node-RED instance. By default the box is checked.

The InstanceForm component gets reused in a few places. The checkbox is only shown in the Create Application context.

image

image

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/instance/components/InstanceForm.vue
+++ b/frontend/src/pages/instance/components/InstanceForm.vue
@@ -5,7 +5,7 @@
         @submit.prevent="$emit('on-submit', input, copyParts)"
     >
         <SectionTopMenu
-            :hero="creatingNew ? (creatingApplication ? 'Create a new Application & Instance' : 'Create Instance') : 'Update Instance'"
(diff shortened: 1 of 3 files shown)


joepavitt (changes requested):
Think this may need a little "info" and dialog to explain a little more context to the user should they need it. The language makes it feel a little out of place, and adding more language would make it distracting. 

In the dialog, we could have something like "This will create an Instance of Node-RED that will be managed by FlowForge in your new Application."

knolleary:
I saw that in the original design, it felt a little redundant as its many of the same words as the option title... but have added this:

image

joepavitt (approved):
Approved subject to tests

PR #2818, opened 2023-09-25
Push Snapshots Though Devops Pipelines

Pezmc:
## Description

Marked draft as I'd like some UI feedback from @joepavitt (can be completed in a follow up though!).
Edit: And test failures

Key features:
 - A stage can now be configured on deploy to either create a new snapshot, always copy the existing snapshot, or prompt 
 - Deploy snapshot names and descriptions are more intelligent and persist the original details along the pipeline stages
 - User is prompted to select a snapshot on deploy if set as such

### Edit Pipeline Stage
Screenshot 2023-09-25 at 16 40 20

### Improved Names
Screenshot 2023-09-25 at 16 34 28

### Select Snapshot to Deploy
Screenshot 2023-09-25 at 16 33 51

## Related Issue(s)

Fixes #2651, https://github.com/flowforge/flowforge/issues/2600

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/flowforge/flowforge/blob/main/CONTRIBUTING.md)
 - [x] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `flowforge/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `flowforge/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Backport needed? -> add the `backport` label
 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #2838, opened 2023-09-27
Add api docs to pipeline stage deploy end point

knolleary:
Adds api docs to the pipeline stage deploy end point

### DIFF
--- a/forge/ee/routes/pipeline/index.js
+++ b/forge/ee/routes/pipeline/index.js
@@ -438,7 +438,35 @@ module.exports = async function (app) {
      * Approach depends on stage.action
      */
     app.put('/pipelines/:pipelineId/stages/:stageId/deploy', {
-        preHandler: app.needsPermission('pipeline:edit')
+        preHandler: app.needsPermission('pipeline:edit'),
+        schema: {
+            summary: 'Triggers a pipeline stage',
+            tags: ['Pipelines'],
+            params: {
+                type: 'object',
+                properties: {
+                    pipelineId: { type: 'string' },
+                    stageId: { type: 'string' }
+                }
+            },
+            body: {
+                type: 'object',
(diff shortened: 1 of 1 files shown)


forge/ee/routes/pipeline/index.js
```diff
@@ -438,7 +438,35 @@ module.exports = async function (app) {
...
+                type: 'object',
+                properties: {
+                    pipelineId: { type: 'string' },
+                    stageId: { type: 'string' }
+                }
+            },
+            body: {
+                type: 'object',
+                properties: {
+                    sourceSnapshotId: { type: 'string', description: 'The snapshot to deploy if the stage action is set to "prompt"' }
```
Pezmc, line 455: This is optional, does it need to be flagged as such?

knolleary: Properties are optional by default. Otherwise, we would set the `required` flag on them

Pezmc:
This will need to be rebased to see the test fixes. Will merge after #2818.

Pezmc (approved):

knolleary:
Ah yes - running with `-b` locally means it only tells me about the first failure 🤦🏻 ... that mystery solved. Still investigating.

knolleary:
Tests should be fixed. Whilst the schema defaulted to `body.sourceSnapshotId` being optional, it still expected a request body to be present, even if empty.

knolleary:
Setting `body.required = false` seems to really upset fastify and causes issues in completely unrelated routes.

Have changed the approach to specifying the type of body as `['object', 'null']` which appears to keep everyone happy.

knolleary:
Tests related to this PR all passing... failures related to issues on main persist.

Once they are resolved, will rebase and force push

PR #2844, opened 2023-09-28
Fix Tests: Guard against getOwners including a null

Pezmc:
## Description

See: https://github.com/flowforge/flowforge/pull/2839

In the [failing test](https://github.com/flowforge/flowforge/blob/b621af4e195b42122904ede629eea66715d2bdcb/test/unit/forge/routes/api/user_spec.js#L613-L627) we're destroying the user bob, but not the team membership (that's done by a cascade). That leads to a race condition where `Team.getOwners()` can return null as the team membership still exists, but points to a `null` user.

A simple guard fixes this. Though it is a state that _should_ not be possible.

https://github.com/flowforge/flowforge/blob/b621af4e195b42122904ede629eea66715d2bdcb/test/unit/forge/routes/api/user_spec.js#L613-L616

## Related Issue(s)

Replaces https://github.com/flowforge/flowforge/pull/2839

### DIFF
--- a/forge/db/models/Team.js
+++ b/forge/db/models/Team.js
@@ -260,7 +260,11 @@ module.exports = {
                         TeamId: this.id,
                         role: Roles.Owner
                     }
-                    return (await M.TeamMember.findAll({ where, include: M.User })).map(tm => tm.User)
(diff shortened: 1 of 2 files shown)


knolleary (approved):

knolleary:
Merge on green

knolleary:
All green. Merging

PR #2856, opened 2023-09-29
Fix Pipeline Issue where "Create New Snapshot" results in error

joepavitt:
## Description

- Fixes regressions preventing Instance to Instance pushing within a pipeline when a "Create New Snapshot" option is configured.
- Also fixed use of 'error' on the Alert type to be 'warning'.

## Related Issue(s)

https://github.com/FlowFuse/flowforge/pull/2818

## Checklist


knolleary (approved):
Merging once tests green

PR #2865, opened 2023-10-02
Report Correct Config File Path

Pezmc:
- Guard against frontend or sentry being undefined
- Fix bad config file logging
- Improved check


forge/config/index.js
```diff
@@ -131,7 +132,7 @@ module.exports = {
         app.log.info(`FlowFuse v${config.version}`)
         app.log.info(`FlowFuse running with NodeJS ${process.version}`)
         app.log.info(`FlowFuse Data Directory: ${process.env.FLOWFORGE_HOME}`)
-        if (!opts.config) {
+        if (config.configFile) {
             app.log.info(`Config File: ${config.configFile}`)
```
Pezmc, line 136: Previously just always logging undefined as config.configFile was undefined

knolleary (approved):

PR #2877, opened 2023-10-04
Add description of Agent and FF Cloud to docs index

knolleary:
## Description

Improve docs index by including a one liner description of the device agent and FF Cloud.

Also fixes a link into the FF Cloud docs because the redirects on `/docs/cloud` -> `/docs/cloud/introduction` isn't working for some reason (which I'll look at separately)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -17,18 +17,25 @@ Node-RED application development and delivery.
 
 ## Device Agent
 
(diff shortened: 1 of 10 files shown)


knolleary:
This also now fixes a bunch of readme redirects

joepavitt (approved):

PR #2889, opened 2023-10-04
metrics: Expose `/metrics` with Prometheus data

ZJvandeWeg:
To get requests per seconds data for FlowFuse a new package is added that leverages fastify to get data for the Node.JS runtime and request data.

For example (not exhaustive):
1. Request per endpoint with timing
2. GC timings
3. Memory utilization

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/docs/admin/telemetry.md
+++ b/docs/admin/telemetry.md
@@ -135,6 +135,7 @@ telemetry:
   backend:
     sentry:
       dsn: <dsn-key>
+    prometheus: true
 ```
 #### Telemetry During Build
 
--- a/forge/forge.js
+++ b/forge/forge.js
@@ -78,8 +78,8 @@ module.exports = async (options = {}) => {
         logger: loggerConfig
     })
 
-  	const metricsPlugin = require('fastify-metrics');
-    await server.register(metricsPlugin, { endpoint: '/metrics' });
+    const metricsPlugin = require('fastify-metrics')
+    await server.register(metricsPlugin, { endpoint: '/metrics' })
 
     if (runtimeConfig.telemetry.backend?.sentry?.dsn) {
         server.register(require('@immobiliarelabs/fastify-sentry'), {
--- a/forge/forge.js
+++ b/forge/forge.js
@@ -78,8 +78,10 @@ module.exports = async (options = {}) => {
         logger: loggerConfig
     })
 
-    const metricsPlugin = require('fastify-metrics')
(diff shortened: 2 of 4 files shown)


hardillb:
- Do we know what the performance overhead of this is? 
- It should be optional behind a configuration flag

ZJvandeWeg:
> Do we know what the performance overhead of this is?

Stats are captured in memory, the biggest impact is on memory and additional updates per request, GC cycle, etc. Than there's a txt page generated upon request. This page renders the data from memory. In practise while I was at GitLab we've measured this and found most request completed in 5ms. That's not to say we should poll the endpoint every second. Expect 15s or even 30s is fine.

hardillb:
This is breaking most of the tests

ppawlowski:
With default metrics provided by the library we will have information about nodejs vm parameters. Information about requests handled by the application requires defining a custom counter, as described [here](https://squaredup.com/blog/instrument-node-with-prometheus/)

ZJvandeWeg:
@hardillb I've added the requested config gate around it, that does make the test pass as you'd expect. Though I'm not sure why I'd failed in the first place.

Do you have any pointers?

ZJvandeWeg:
Identified so far: 

1. Double registering of metrics shouldn't really occur on production as the server is only started once. Not multiple times like in the tests. The package imported has chosen not to make the registration idempotent sadly. But again, if you start the server once, it works.

2. A `MaxListenersExceededWarning`, which was set to 11. I'm not sure if our code was at 10 already and I pushed it over the edge, but yes, this exposes another listener.

hardillb:
It will be because the test harness creates and disposes of the forge "Object" before and after a number of tests due to needing to test in different states that persist once set up.

It looks like what ever hooks that modules is registering in NodeJS can only be done once per process (at least with a given name) so when a second forge object is created it tried to re-register and failed.

Having it behind a flag that defaults to off means it doesn't get setup so the tests pass

hardillb (approved):

hardillb:
This will need issues raising against both the CloudProject and the Helm project to get the `flowforge.yml` flag deployed

ZJvandeWeg:
@hardillb Can we backport this so it's deployed next week?

hardillb:
As the PR didn't have the backport label before it was merged it will need a new PR targeting the `maintenance` branch

ZJvandeWeg:
K, cool. I’ll cherry pick.

PR #2894, opened 2023-10-05
backport: Prometheus for the app

ZJvandeWeg:
Combined backport of:
1. https://github.com/FlowFuse/flowfuse/pull/2889
2. https://github.com/FlowFuse/flowfuse/pull/2893

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

### DIFF
--- a/forge/forge.js
+++ b/forge/forge.js
@@ -76,7 +76,7 @@ module.exports = async (options = {}) => {
         logger: loggerConfig
     })
 
-    if (runtimeConfig.telemetry.backend?.prometheus) {
+    if (runtimeConfig.telemetry.backend?.prometheus?.enabled) {
         const metricsPlugin = require('fastify-metrics')
         await server.register(metricsPlugin, { endpoint: '/metrics' })
     }
--- a/forge/forge.js
+++ b/forge/forge.js
@@ -76,8 +76,8 @@ module.exports = async (options = {}) => {
         logger: loggerConfig
     })
 
-  	const metricsPlugin = require('fastify-metrics');
-    await server.register(metricsPlugin, { endpoint: '/metrics' });
+    const metricsPlugin = require('fastify-metrics')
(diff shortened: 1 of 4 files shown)


forge/forge.js
```diff
@@ -75,6 +75,20 @@ module.exports = async (options = {}) => {
...
+        await server.register(metricsPlugin, { endpoint: '/metrics' })
+    }
+
+    if (runtimeConfig.telemetry.backend?.sentry?.dsn) {
+        server.register(require('@immobiliarelabs/fastify-sentry'), {
+            dsn: runtimeConfig.telemetry.backend.sentry.dsn,
+            environment: process.env.NODE_ENV,
+            release: `flowforge@${runtimeConfig.version}`
+        })
+    }
```
hardillb, line 90: Did you mean to include the `sentry` stuff in here? That has more changes than just this and the dependency is not in the `package.json` included here.

```suggestion
```

hardillb (commented):
Need to check the sentry inclusion.

forge/forge.js
knolleary: We are validating the sentry work in staging before deciding whether to adopt in production as well (which is why it didn't get backported automatically).

Given this is guarded on `runtimeConfig.telemetry.backend?.sentry?.dsn` - which should not get set in production then this should be okay as is.

knolleary: Ah - but the linting won't like the `require` if not also in package.json.

Let me discuss with @Pezmc if there's any reasons at this point not to get sentry setup on production as well.

ZJvandeWeg: @hardillb Didn't spot this in the diff, usually cherry-pick doesn't show one. As the sentry changes created a conflict, I figured it was already in the maintenance branch.

I can manually remove it for now, to make it easier for this PR to go through without any additional scope inflation.

hardillb:
Going to close this given Nicks announcement

PR #2899, opened 2023-10-05
Remove not null constraint on ProjectSnapshot.ProjectId column

knolleary:
## Description

The `ProjectSnapshot.ProjectId` column has got a `not null` constraint applied. This is preventing Device/Application snapshots from being created in *some* environments.

This constraint is not set in all environments.

 - Local sqlite, clean db - no constraint
 - Local sqlite, existing db that has run through migrations - constaint applied
 - Staging Postgres - no constraint
 - Production Postgres - constraint

I have tested this migration against:

 - sqlite with the constraint already present
 - sqlite without the constraint
 - postgres with the constraint
 - postgres without the constraint

### DIFF
--- a/forge/db/migrations/20231005-01-update-projectSnapshot-constraint.js
+++ b/forge/db/migrations/20231005-01-update-projectSnapshot-constraint.js
@@ -7,19 +7,21 @@ module.exports = {
     up: async (context) => {
         const dialect = context.sequelize.options.dialect
         if (dialect === 'sqlite') {
-            // Disable foreign_keys so that when sequelize does its sqlite-specific
-            // copy/drop workaround for renaming columns, we don't cause cascade
-            // delete on any other tables
-            await context.sequelize.query('PRAGMA foreign_keys = 0;')
-        }
-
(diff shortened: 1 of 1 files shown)


forge/db/migrations/20231005-01-update-projectSnapshot-constraint.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+const { DataTypes } = require('sequelize')
+
+module.exports = {
+    up: async (context) => {
+        const dialect = context.sequelize.options.dialect
+        if (dialect === 'sqlite') {
+            // Disable foreign_keys so that when sequelize does its sqlite-specific
+            // copy/drop workaround for renaming columns, we don't cause cascade
+            // delete on the Teams table
```
knolleary, line 12:
```suggestion
            // delete on any other tables
```

knolleary:
Issue found with sqlite constraint removal... working onit.

hardillb (approved):

Steve-Mcl (approved):

Issue #2874, opened 2023-10-03
Create UI for selecting Flow Blueprints when creating instance


knolleary:
Part of #2828 MVP

Design proposals by @joepavitt in https://github.com/FlowFuse/flowfuse/issues/2828#issuecomment-1735335432

This will require the CRUD api to be delivered in #2871 

The `InstanceForm` component will need to be updated to allow the FlowTemplate selection to be made.

MarianRaphael:
@Yndira-FlowForge first iteration

- Very simple selection while creating an instance
- Default selection should be "Blank" Node-RED

Yndira-FlowForge:
Hi, I've spoken with @MarianRaphael, and we looked at the screen for creating an instance. We agreed on having the default-blank template already selected, and the option to choose a different template, perhaps with a button that opens a modal window displaying a list of templates.

As I examined all the elements and options we already have in place, trying to determine the best location for this feature, I noticed that we already have something labeled "template" when creating an instance. The "default" option is already preselected. Initially, I thought that maybe @Pezmc had already started working on the smallest possible iteration. However, after discussing it with him, I understood that this is a different type of template.

<img width="1181" alt="Screenshot 2023-10-10 at 16 55 16" src="https://github.com/FlowFuse/flowfuse/assets/129537638/d84cb3f9-7a5d-4873-a977-801b6b120c11">

<img width="1184" alt="Screenshot 2023-10-10 at 16 55 28" src="https://github.com/FlowFuse/flowfuse/assets/129537638/9a2decdc-390d-4d81-b5bf-8a186467cc3b">

Should we consider renaming these options? Having two items labeled "template" can be confusing. While one would be referred to as a 'Flow Template,' are we planning to keep the other one simply as 'template'? I don't see a description explaining what it means...

knolleary:
In production, we only have one (Project) Template, which causes this row to be hidden as there is no choice for the user. In staging we have two templates - hence why it shows that row.

For the majority of FF Cloud users, the existing 'Template' terminology isn't going to be something they have had to learn about. So I think we do have some ability to relabel if we see fit.

knolleary:
@Pezmc I've added a couple curl commands [here](https://github.com/FlowFuse/flowfuse/issues/2871#issuecomment-1755829289) for listing and injecting templates into the database.

Yndira-FlowForge:
@MarianRaphael I know we discussed a pop-up, but how about this for the smallest possible iteration?

[...]

PR #2922, opened 2023-10-11
Flow Blueprints: Select a flow blueprint during instance creation

Pezmc:
## Description

Screenshot 2023-10-11 at 09 46 35

## Related Issue(s)

For https://github.com/FlowFuse/flowfuse/issues/2874, this is the quick win, to be followed up with UI improvements down the line.

## Checklist

 - [ ] Suitable unit/system level tests have been added and they pass - **Draft as EE UI coverage is needed**
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/api/flowBlueprints.js
+++ b/frontend/src/api/flowBlueprints.js
@@ -1,0 +1,20 @@
+import paginateUrl from '../utils/paginateUrl.js'
+
+import client from './client.js'
+
+/**
+ * Get all flow blueprints (formerly flow templates) from the backend
+ * @param {string} state - 'active' (default), 'all' or 'inactive'
+ * @param {G} cursor
+ * @param {number} limit
+ * @returns [{ id:number, active:boolean, name:string, description:string, createdAt:string, updatedAt:string }]
+ */
+const getFlowBlueprints = async (state = 'active', cursor, limit) => {
+    const url = paginateUrl('/api/v1/flow-templates', cursor, limit, null, { filter: state })
+    return client.get(url).then(res => {
+        return res.data.templates
+    })
+}
+export default {
+    getFlowBlueprints
+}
--- a/frontend/src/pages/instance/Settings/ChangeInstanceType.vue
+++ b/frontend/src/pages/instance/Settings/ChangeInstanceType.vue
@@ -5,7 +5,7 @@
             v-else :instance="instanceDetails || instance"
             :team="team"
             :billing-enabled="!!features.billing"
-            :flow-blueprints-enabled="!!features.flowBlueprints"
+            :flow-blueprints-enabled="false"
             @on-submit="changeInstanceDefinition"
         />
     </div>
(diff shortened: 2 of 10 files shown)


knolleary:
I think the 'Blank Template' should be a hardcoded entry in the front-end, rather than rely on it having been created first in the database.

frontend/src/pages/instance/components/InstanceForm.vue
```diff
@@ -204,11 +205,41 @@
                         v-model="copyParts"
                     />
                 </template>
+                <div v-else-if="creatingNew && flowBlueprintsEnabled && flowBlueprints.length > 1">
```
knolleary, line 208:
```suggestion
                <div v-else-if="creatingNew && flowBlueprintsEnabled && flowBlueprints.length > 0">
```

Think this was an off-by-one error. I only had one legit blueprint in my database, so nothing was being shown.

frontend/src/pages/instance/Settings/ChangeInstanceType.vue
```diff
@@ -1,7 +1,13 @@
 <template>
     <div class="w-full max-w-4xl" data-el="change-project">
         <ff-loading v-if="saving" message="Updating Instance..." />
-        <InstanceForm v-else :instance="instanceDetails || instance" :team="team" :billingEnabled="!!features.billing" @on-submit="changeInstanceDefinition" />
+        <InstanceForm
+            v-else :instance="instanceDetails || instance"
+            :team="team"
+            :billing-enabled="!!features.billing"
+            :flow-blueprints-enabled="!!features.flowBlueprints"
```
knolleary, line 8: I don't think in this context (changing an instance type) we ever want to offer the blueprints. I see that is guarded against inside InstanceForm, but would it be more explicit to hardcode this to false here?
```suggestion
            :flow-blueprints-enabled="false"
```

knolleary (changes requested):

frontend/src/api/flowBlueprints.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {number} limit
+ * @returns [{ id:number, active:boolean, name:string, description:string, createdAt:string, updatedAt:string }]
+ */
+const getFlowBlueprints = async (state = 'active', cursor, limit) => {
+    const url = paginateUrl('/api/v1/flow-templates', cursor, limit, null, { filter: state })
+    return client.get(url).then(res => {
+        return res.data.templates
+    })
+}
+export default {
```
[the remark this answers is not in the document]
Pezmc: I don't think we need this, I already had markdown support, note the fullstops and italics in the screenshot.

Pezmc:
Test coverage will be added as a follow up, marking ready to review so this can go to staging!

knolleary (approved):

PR #2928, opened 2023-10-11
Apply a flowTemplate if specified on Instance Create

### LINKED ISSUE #2873: Update Create Instance API to access optional FlowBlueprint id
Part of #2828 MVP

The 'create instance' API endpoint needs to be updated to allow an optional FlowTemplate id to also be provided.

When provided, this should be used to pre-populate the instance with the flows from the template.

For MVP there is no access restriction to particular FlowTemplates. That may come in a future iteration.

 - Add `flowBlueprintId` as optional string parameter to `POST /api/v1/projects` endpoint.

knolleary:
Merge target is: #2922 

Closes #2873

## Description

Allows `flowTemplate` to be specified on the instance create API and applies the template's flows/credentials/modules to the instance on creation.

Will be ready to merge once some tests are added

### DIFF
--- a/forge/db/controllers/Project.js
+++ b/forge/db/controllers/Project.js
@@ -299,7 +299,7 @@ module.exports = {
      * @param {ProjectType} type
      * @param {ProjectStack} stack
      * @param {ProjectTemplate} template
-     * @param {{name: string, ha: {}, sourceProject: Project, sourceProjectOptions: {}}} properties Props of the project to create
+     * @param {{name: string, ha: {}, sourceProject: Project, sourceProjectOptions: {}, flowTemplate: FlowTemplate}} properties Props of the project to create
      * @returns
      */
     create: async function (
@@ -314,7 +314,8 @@ module.exports = {
             name = '',
             ha = null,
             sourceProject = null,
-            sourceProjectOptions = {}
+            sourceProjectOptions = {},
+            flowTemplate = null
         } = {}
     ) {
         if (!user) {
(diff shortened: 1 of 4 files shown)


forge/routes/api/project.js
```diff
@@ -128,6 +128,7 @@ module.exports = async function (app) {
                     applicationId: { type: 'string' },
                     projectType: { type: 'string' },
                     stack: { type: 'string' },
+                    flowTemplate: { type: 'string' },
```
Pezmc, line 131: The frontend currently calls this `flowTemplateId` since it's an ID for a template rather than a flowTemplate itself

knolleary: Yeah, I paused on this. The existing properties `projectType`, `stack` and `template` are all also ids but not labelled as such. I erred on local consistency... but can be swayed to introduce `flowTemplateId`.

Pezmc: I’d like our convention going forward to be use the word ID if it’s an ID. It’s much more explicit about what is expected.

knolleary: I'm okay with that. Will updated

knolleary:
Ready to review - tests added and `flowTemplate` changed to `flowTemplateId`

knolleary:
Need to agree on flowTemplateId and flowBlueprintId... we are currently using both.

Will update this PR for the latter.

Pezmc:
As per slack, I'd support renaming the backend to use the new flow blueprint terminology, but if it's a time sync, having the front use `flowBlueprint` and the back use `flowTemplate` is not the end of the world. In which case I'd expect the backend to use `flowTemplateId` to be consistent with itself and the frontend to handle the mapping.

forge/db/controllers/Project.js
```diff
@@ -625,6 +633,30 @@ module.exports = {
...
             await project.updateSetting('settings', currentProjectSettings)
         }
+    },
+    /**
+     * Applies the contents of a FlowTemplate to an Instance. We assume this
+     * is being invoked when the Instance is being created - so it will overwrite
+     * any existing flows and will merge any modules with those provided by the ProjectTemplate.
+     * @param {*} app
+     * @param {*} instance
+     * @param {*} flowTemplate
```
Pezmc, line 643:
```suggestion
     * @param {FlowTemplate} flowTemplate
```

Pezmc (approved):

PR #2931, opened 2023-10-11
Pin fastify to 4.23.0 due to HEAD bug

knolleary:
## Description

Fastify 4.24.0 has introduced a regression (due to https://github.com/fastify/fastify/pull/5078 I think) that causes the forge app to fail to start with the error:

```
Failed to start: Error: Method 'HEAD' already declared for route '/api/v1/devices/:/editor/proxy/*' with constraints '{}'
```

This reverts the module to the last working version - and pins it for now. We can unpin once the issue is confirmed and fix released.

I'll be raising an issue against fastify tomorrow - just need to produce a clear reproduction for them - will link back here for reference.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,7 +35,7 @@
                 "bcrypt": "^5.1.0",
                 "cronosjs": "^1.7.1",
                 "dotenv": "^16.3.1",
-                "fastify": "^4.18.0",
+                "fastify": "4.23.2",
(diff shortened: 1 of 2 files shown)


knolleary:
Upstream issue raised - https://github.com/fastify/fastify/issues/5089

Pezmc (approved):

PR #2935, opened 2023-10-12
Unpin fastify

knolleary:
## Description

Follow-up to #2931 , this updates our route handler to avoid the issue with fastify 4.24.0.

In summary, by default, whenever you register a GET route, fastify will add a default HEAD route to match. If you don't want it to do that, you either set `exposeHeadRoute: false` or you register the HEAD route first.

In this instance, I've gone for the more explicit fix - setting the `exposeHeadRoute` property.

They have identified a bug where you should be able to register a HEAD route even if it matches one of these default HEAD routes - but we don't need to wait for that fix by using the setting above.

### DIFF
--- a/forge/ee/routes/deviceEditor/index.js
+++ b/forge/ee/routes/deviceEditor/index.js
@@ -192,6 +192,9 @@ module.exports = async function (app) {
     app.route({
         method: 'GET', // only GET is permitted for WS
         url: '/proxy/*',
+        // By default, fastify adds a HEAD route for each GET route. Given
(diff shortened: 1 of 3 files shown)


Pezmc (approved):

Pezmc:
@knolleary Should we add an npm run start to our tests? Why didn't CI fail?

knolleary:
I believe the tests would have failed if they were run against 4.24.0.

The question is how did CI result in 4.24.0 being included when it wasn't in package-lock.

Issue #2828, opened 2023-09-26
Blueprint Library


MarianRaphael:
### Description

The Template Library is a concept analogous to the FlowFuse Team Library. It allows a FlowFuse Admin to share flow.json files with all teams. This would be a unidirectional channel, as users of teams do not have the ability to upload any flows (for that purpose, they can use the Team Library). As a result, the FlowFuse Admin can share a set of standard flows and examples with all FlowFuse teams.

```[tasklist]
### MVP - shipped in 1.12.4
- [ ] https://github.com/FlowFuse/flowfuse/issues/2871
- [ ] https://github.com/FlowFuse/flowfuse/issues/2873
- [ ] https://github.com/FlowFuse/flowfuse/issues/2874
- [ ] https://github.com/FlowFuse/flowfuse/issues/2872
```

```[tasklist]
### UX Improvements
- [ ] https://github.com/FlowFuse/flowfuse/issues/2948
- [ ] https://github.com/FlowFuse/flowfuse/issues/2945
- [ ] https://github.com/FlowFuse/flowfuse/issues/3266
- [ ] https://github.com/FlowFuse/flowfuse/issues/3265
- [ ] https://github.com/FlowFuse/flowfuse/issues/3598
```

```[tasklist]
### Future: Enable Shared Library in Node-RED
- [ ] Create Node-RED Shared Library endpoint for FlowTemplates
- [ ] Add NR Shared Library config to NR Launcher
- [ ] Add NR Shared Library config to Device Agent
```

## Epic

https://github.com/flowforge/flowforge/issues/2814

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

MarianRaphael:
Usage of existing templates instead of "Team Library" concept

joepavitt:
Complexity of templates living at the Library level is that some of the them will have third-party nodes, loading those flows via the Library, would mean Users then have to _know_ (and search/install) the third-party dependencies

joepavitt:
When first hearing this idea, my thought was this would be a step in the "New Instance" workflow, you choose from a set of Flow Templates (which would be broken into sections, e.g. "Blank", "Getting Started with Node-RED", "Manufacturing"

joepavitt:
[...]

PR #2938, opened 2023-10-12
Flow Blueprints mvp tidy

knolleary:
## Description

This applies some final tidy up to the first iteration of Flow Blueprints work - #2828

I've split it into separate commits

 1. Include `instance-blueprint` in the payload sent to posthog when creating an instance
 2. Updates the Blank template wording:
    image
 3. Updates all external apis to use `flow blueprint` terminology rather than flow template.
     - CRUD end-points moved to `/api/v1/flow-blueprints` and scheme updated
     - Property on Instance Create api renamed to `flowBlueprintId`
     - It does *not* modify the model/database table - that felt a step too far for today.

### DIFF
--- a/frontend/src/api/instances.js
+++ b/frontend/src/api/instances.js
@@ -17,7 +17,8 @@ const create = async (options) => {
         const props = {
             'created-at': res.data.createdAt,
             'instance-stack': res.data.stack.id,
-            'instance-template': res.data.template.id
+            'instance-template': res.data.template.id,
+            'instance-blueprint': options.flowTemplateId
         }
         product.capture('$ff-instance-created', props, {
             team: res.data.team.id,
--- a/frontend/src/api/instances.js
+++ b/frontend/src/api/instances.js
@@ -9,16 +9,15 @@ import paginateUrl from '../utils/paginateUrl.js'
 import client from './client.js'
 
 const create = async (options) => {
-    // Backend calls these flow templates, UI refers to it as flow blueprints
-    options.flowTemplateId = options.flowBlueprintId
-    delete options.flowBlueprintId
-
+    if (options.flowBlueprintId === '') {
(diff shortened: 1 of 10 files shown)


frontend/src/api/instances.js
```diff
@@ -9,15 +9,15 @@ import paginateUrl from '../utils/paginateUrl.js'
 import client from './client.js'
 
 const create = async (options) => {
-    // Backend calls these flow templates, UI refers to it as flow blueprints
-    options.flowTemplateId = options.flowBlueprintId
-    delete options.flowBlueprintId
-
+    if (options.flowBlueprintId === '') {
+        delete options.flowBlueprintId
```
Pezmc, line 13: Does the backend not handle receiving ''?  I think it should!

frontend/src/pages/instance/components/InstanceForm.vue
```diff
@@ -11,7 +11,7 @@
         <!-- Form title -->
         <div class="mb-8 text-sm text-gray-500">
             <template v-if="creatingNew">
-                <template v-if="applicationSelection">
+                <template v-if="!creatingApplication || applicationSelection">
```
Pezmc, line 14: Is this change intended?

frontend/src/api/instances.js
knolleary: It does handle it. I just dislike sending blank properties if they are meant to be ignored.

frontend/src/pages/instance/components/InstanceForm.vue
knolleary: Yes - albeit slightly off-topic and it slipped in.

I noticed when we're on the 'Create Instance' page (so no Application creation/selection to be done), we were showing the 'Applications are used to ...' explainer. This fixes that to show the header related to creating an instance.

PR #2940, opened 2023-10-12
Guard against refreshing an application page when route is changing

### LINKED ISSUE #2929: Page Not Found 404
### Current Behavior

1. Select an application
2. Attempt to go back to applications via the three lines. I call it a hamburger and then click applications.
3. Displays 404

### Expected Behavior

I expect it to return to Applications Page.

### Steps To Reproduce

See Current behavior

### Environment

- FlowForge version: v1.12.3
- Node.js version:
- npm version:
- Platform/OS: Windows
- Browser: Chrome

I can not provide an initial effort estimate

knolleary:
Fixes #2929 #2934

## Description

When navigating from an application route, such as `/application/j5KRXxRx8y/instances` to a team route, such as `/team/dev/applications`, a watch on the `id` param of the `/application/...` page is triggered to reload the 'new' application. However as its moving away from the application view, there is no id param - causing it to try loading urls with `undefined` in.

This is a quick fix to only try updating the application if the `id` param is present.

I suspect there's a more vue way of doing this given the component is being unmounted so we shouldn't be watching the param any more.

### DIFF
--- a/frontend/src/pages/application/index.vue
+++ b/frontend/src/pages/application/index.vue
@@ -147,46 +147,48 @@ export default {
     methods: {
         async updateApplication () {
             const applicationId = this.$route.params.id
-            try {
-                this.applicationInstances = []
-                const applicationPromise = ApplicationApi.getApplication(applicationId)
-                const instancesPromise = ApplicationApi.getApplicationInstances(applicationId) // To-do needs to be enriched with instance state
-                const devicesPromise = ApplicationApi.getApplicationDevices(applicationId)
+            if (applicationId) {
+                try {
+                    this.applicationInstances = []
+                    const applicationPromise = ApplicationApi.getApplication(applicationId)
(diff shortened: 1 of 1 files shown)


Pezmc:
@knolleary Where is the watch happening? We can likely add an explicit unwatch as part of the unmount

knolleary:
Same file I've touched in the PR - https://github.com/FlowFuse/flowfuse/blob/748a3a1ed5225322191bbca717decd602c80be49/frontend/src/pages/application/index.vue#L137-L142

Pezmc:
The watch is being removed on unmount, but the component hasn't been unmounted yet when we _start_ the navigation. After speaking to @knolleary I've proposed an alternative for review @joepavitt.

knolleary:
Happy with that. I cannot approve as the original raiser of the PR.

Pezmc (approved):

PR #2946, opened 2023-10-13
Fix navigating away from instance page

knolleary:
## Description

This applies the same fix from #2940 but to the Instance page.


Pezmc (approved):

knolleary:
Merging once UI green

Issue #2905, opened 2023-10-06
Improve UX to subscribe to starter tier.


iskerrett:
### Description

The current UX for the someone to subscribe to the Starter tier is not obvious.  The current workflow is as follows:
1. Create an account that starts a free trial.
2. Go to Team Billings or go to Team Setting > Edit Team Setting > click on change team type that leads to the same Team Billings page
3. Click on Setup Payment Details
4. The Stripe page then shows $0 for the Trail Starter Team.

There are a number of issues with this UX.
1. It doesn't appear to be possible for someone to setup a $15/month charge for the Starter tier before the trial period ends. We should be encouraging people to start paying for FF Cloud as soon as possible. 
2. Team Billings should really be just Billings. The Starter tier is aimed as individuals so they won't be thinking in terms of Teams.  In fact, I don't think we should be talking about different billing tiers/packages and not be using Teams for the term. 
3. On the billing page, it would seem we should list the 3 tiers so people select which on they want to purchase before they enter their credit card details. I suggest we should always be showing the monthly cost before we ask for someone credit card number.

The ideal work flow to subscribe to the start tier.
1. Create an account that starts a free trial.
2. Click on Billing, to get a list of different pricing tiers/packages.  The starter tier being one of these listed.
3. Use selected Starter tier.
4. Clicks Setup Payment details.
5. Stripe page shows $15/month and request credit card info.
6. After they credit card info is entered, the user exists trial mode and has the starter tier.

As a bonus, it would be great to have a url that we can embed into an email that say 'Upgrade to Starter Now' that will take the user to the Billing page.

```[tasklist]
### Tasks
- [ ] https://github.com/FlowFuse/flowfuse/pull/2963
- [ ] https://github.com/FlowFuse/flowfuse/issues/2970
- [ ] https://github.com/FlowFuse/flowfuse/issues/3105
```

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

MarianRaphael:
First iterartion: 
- End Trial, if a user provides cc information
- 15 $ for Starter Package would be displayed in Stripe

Next Iteration:
- Selection Screen for the tier

knolleary:
See also #2904 which is related.

knolleary:
#2328 contains designs relevant to this. Still need to scope down a bit, but gives a better UX for the process around first setting up billing and leaving trial mode.

knolleary:
[...]

PR #2963, opened 2023-10-16
Exit trial when billing setup

knolleary:
Part of #2905 

## Description

This modifies the billing flow around trials.

With this change, when a user is in Trial mode and clicks through to setup billing, Stripe will now show the full team plan details, rather than the $0 trial plan. Once they complete billing setup, their team is no longer in trial mode and they are paying for the team as expected.

The net result is we will no longer consider being in Trial mode with Billing setup as a valid state. However it is possible we will have a small number of teams already in that state when this change goes live. So I've left in place the code to handle them - but marked it up as being removable in 1.14 (ie, at least 2 weeks after this change hits production).

Along the way I've tidied up some bits of the language presented to the user, however this is likely to be replaced with the next iteration of #2905 that will allow them to pick a team type before setting up billing.

### DIFF
--- a/forge/ee/db/controllers/Subscription.js
+++ b/forge/ee/db/controllers/Subscription.js
@@ -5,6 +5,10 @@ module.exports = {
         if (existingSub) {
             existingSub.customer = customer
             existingSub.subscription = subscription
+            if (existingSub.status === app.db.models.Subscription.STATUS.TRIAL) {
+                existingSub.trialEndsAt = null
+                existingSub.trialStatus = app.db.models.Subscription.TRIAL_STATUS.ENDED
+            }
             existingSub.status = app.db.models.Subscription.STATUS.ACTIVE
             await existingSub.save()
 
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -40,16 +40,11 @@ module.exports.init = async function (app) {
     return {
         createSubscriptionSession: async (team, user = null) => {
             const billingIds = await team.getTeamBillingIds()
-            let teamPrice = billingIds.price
+            const teamPrice = billingIds.price
 
             // Use existing Stripe customer
             const existingLocalSubscription = await team.getSubscription()
-            if (existingLocalSubscription && existingLocalSubscription.isTrial()) {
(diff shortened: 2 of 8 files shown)


forge/ee/db/controllers/Subscription.js
```diff
@@ -5,6 +5,10 @@ module.exports = {
         if (existingSub) {
             existingSub.customer = customer
             existingSub.subscription = subscription
+            if (existingSub.status === app.db.models.Subscription.STATUS.TRIAL) {
+                existingSub.trialEndsAt = null
+                existingSub.trialStatus = app.db.models.Subscription.TRIAL_STATUS.ENDED
+            }
```
knolleary, line 11: This is handling the 'subscription created' callback from stripe - at which point we clear the trial status of the local subscription object.

forge/ee/lib/billing/index.js
```diff
@@ -40,16 +40,11 @@ module.exports.init = async function (app) {
...
+            const teamPrice = billingIds.price
 
             // Use existing Stripe customer
             const existingLocalSubscription = await team.getSubscription()
-            if (existingLocalSubscription && existingLocalSubscription.isTrial()) {
-                // Currently in trial mode. Check for trial billing ids
-                if (billingIds.trialPrice) {
-                    teamPrice = billingIds.trialPrice
-                }
-            }
```
knolleary, line 52: When initiating the stripe session, we will no longer apply the 'trial price'. This means stripe will show the true cost of the subscription - and we don't stay in trial mode after creating the subscription.

test/unit/forge/ee/lib/billing/trialTask_spec.js
```diff
@@ -121,143 +117,6 @@ describe('Billing - Trial Housekeeper Task', function () {
         // Testing the specific body text is going to be too brittle
     })
 
-    it('adds trial projects to billing the team trial has ended', async function () {
```
knolleary, line 124: This scenario is no longer supported, so removing the tests

Pezmc (approved):

Pezmc (approved):

PR #2965, opened 2023-10-17
Ignore subscription events if we don't know the subscription

### LINKED ISSUE #2962: Handling of customer.subscription.deleted for subscription not tied to the platform
### Current Behavior

Whenever we receive a Stripe callback we use this code to parse the contents and retrieve our local `Subscription` and `Team` objects.

```
async function parseSubscriptionEvent (/** @type {StripeEvent} */ event) {
        const stripeSubscriptionId = event.data.object.id
        const stripeCustomerId = event.data.object.customer
        const subscription = await app.db.models.Subscription.byCustomerId(stripeCustomerId)
        const team = subscription?.Team

        logStripeEvent(event, team, subscription, null, stripeCustomerId)

        return {
            stripeSubscriptionId, stripeCustomerId, subscription, team
        }
    }
```

Note that we get `stripeSubscriptionId` from the event. We then use `byCustomerId` to lookup the Subscription object - as we operate on the assumption each Customer has a single Subscription.

We do not validate the `subscription` we find matches the `stripeSubscriptionId` from the event.

If a customer has multiple subscriptions - possible if someone has manually created a subscription in the stripe dashboard - then any operation performed on those subs will be interpreted as an operation on the platform-owned subscription.

For example, deleting a 'extra' subscription in the stripe dashboard will lead us to suspending all team instances as we think the main sub has been deleted/canceled.

We need to validate the local subscription against the subscription id in the event before proceeding.

Note this function is used for multiple cases - including the initial creation of the subscription - where there may not be a local subscription yet.

### Expected Behavior

Operations are not performed on the local subscription when other subscriptions owned by the same customer are touched in the stripe dashboard.

### Steps To Reproduce

1. Setup a Team with a subscription and a running instance.
2. In Stripe dashboard, manually create a second subscription for the customer
3. Cancel that second subscription
4. Observe the event has been triggered to suspend that team's instances

### Environment

- FlowForge version: 1.12.4
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Fixes #2962 

## Description

If a customer has multiple subscriptions in Stripe (because some have been manually created), then any operation on those subscriptions was being mishandled as on operation on the platform-owned subscription.

This PR fixes that by validating the local subscription ID against the id in the stripe event.

### DIFF
--- a/forge/ee/routes/billing/index.js
+++ b/forge/ee/routes/billing/index.js
@@ -16,7 +16,17 @@ module.exports = async function (app) {
     function logStripeEvent (/** @type {StripeEvent} */ event, team, subscription, teamId = null, stripeCustomerId = null) {
         const intro = `Stripe ${event.type} event ${event.data.object.id} from ${stripeCustomerId} received for`
         if (team) {
-            app.log.info(`${intro} team '${team.hashid}'`)
+            if (subscription ||
+                event.type === 'checkout.session.completed' ||
+                event.type === 'checkout.session.expired'
+            ) {
+                app.log.info(`${intro} team '${team.hashid}'`)
+            } else {
+                // A subscription event for a subscription we don't know about
+                // for this known customer. This can happen if additional
+                // subscriptions were created manually within Stripe for this customer
+                app.log.warn(`${intro} team '${team.hashid}' for unknown subscription`)
+            }
         } else if (teamId) {
             app.log.error(`${intro} unknown team by team ID '${teamId}'`)
         } else if (subscription) {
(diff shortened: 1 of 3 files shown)


forge/ee/routes/billing/index.js
```diff
@@ -16,7 +16,17 @@ module.exports = async function (app) {
     function logStripeEvent (/** @type {StripeEvent} */ event, team, subscription, teamId = null, stripeCustomerId = null) {
         const intro = `Stripe ${event.type} event ${event.data.object.id} from ${stripeCustomerId} received for`
         if (team) {
-            app.log.info(`${intro} team '${team.hashid}'`)
+            if (subscription ||
+                event.type === 'checkout.session.completed' ||
```
Pezmc, line 20: I think this if statement should be other way around to make it clearer what we're checking, currently it feels backwards:

`/^customer.subscription/.test(event.type)`

Pezmc: Solved in [6412a0c](https://github.com/FlowFuse/flowfuse/pull/2965/commits/6412a0c1db373d06df61eef67ec6f674db854715) by setting an explicit flag

Pezmc (approved):

Pezmc (approved):

PR #2966, opened 2023-10-17
Restrict custom catalogs and npmrc to licensed team types

### LINKED ISSUE #2950: Fix licensing of custom catalog feature
### Description

When #2526 was implemented it did not get scoped to the right license level. We need to ensure the custom npmrc and catalog list is only available to:
 - EE licensed instances
 - FF Cloud - Team tier or higher

```[tasklist]
### Tasks
- [ ] Add filter to these settings in the runtime if not EE licensed
- [ ] Add 'customCatalogs' feature flag to platform - based on EE license
- [ ] Add 'customCatalogs' feature flag to TeamType object
- [ ] In Template view, hide the relevant fields if feature flag is not set for either platform or current TeamType.
``` 

### Epic/Story

_No response_

knolleary:
Fixes #2950

## Description

This PR moves the custom catalog and npmrc settings to be behind a licensed feature flag.

Furthermore, it is added as a TeamType feature flag so only select teams types within a licensed instance will have access to the feature.

---

This is currently UI only. Getting that side raised whilst I add some backend handling for this to match along with tests.

### DIFF
--- a/frontend/src/pages/instance/Settings/Palette.vue
+++ b/frontend/src/pages/instance/Settings/Palette.vue
@@ -1,8 +1,8 @@
 <template>
     <form class="space-y-6">
         <TemplateSettingsPalette v-model="editable" :editTemplate="false" />
-        <TemplateSectionCatalogue v-model="editable" :editTemplate="false" :readOnly="!catalogueEditable" :project="project" />
-        <TemplateSectionNPM v-model="editable" :editTemplate="false" :readOnly="!npmEditable" :project="project" />
+        <TemplateSectionCatalogue v-if="catalogFeatureEnabledForTeam" v-model="editable" :editTemplate="false" :readOnly="!catalogueEditable" :project="project" />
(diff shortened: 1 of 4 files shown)


frontend/src/pages/instance/Settings/Palette.vue
```diff
@@ -72,7 +72,14 @@ export default {
...
     computed: {
-        ...mapState('account', ['team', 'teamMembership']),
+        ...mapState('account', ['team', 'teamMembership', 'features']),
+        catalogFeatureEnabledForTeam () {
+            if (!this.features.customCatalogs) {
+                return false
+            }
+            const flag = this.team.type.properties.features?.customCatalogs
+            return flag === undefined || flag
+        },
```
knolleary, line 82: This is the instance-specific view of settings, so we have to check both the platform feature flag and the current team type features.

hardillb:
Looks good, will review in concert with the back end

hardillb (approved):

Issue #2755, opened 2023-09-14
Resource utilization alerts in Auditlog for Node-RED Instances


MarianRaphael:
### Description

Add a feature to the Auditlog that displays alerts when the resources (CPU, Memory) reach 80%, 90%, and 100% utilization for each Node-RED instance. This feature is aimed at assisting users in decision-making about the need for resource upgrades and identifying the root cause of instance crashes.

## Epic 
https://github.com/flowforge/flowforge/issues/223

## User Story
As a FlowFuse User, 
I want to be informed about the resource utilization of each Node-RED instance 
so that I can make informed decisions about upgrades and troubleshoot issues proactively.

## Acceptance Criteria

- [ ] The Auditlog should display a clear and concise alert message when CPU or Memory usage reaches 80%, 90%, and 100%
- [ ] If CPU or Memory utilization jumps to 100% and multiple alerts (80%, 90%, 100%) are triggered within the last 3 seconds, only the message with the highest utilization percentage (100%) should be sent.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

knolleary:
I'm increasing the sizing for this. The requirement here is all about presenting the information to the user. But first we have to do the work to be able to monitor and capture that information in any meaningful way for each of the drivers - localfs/docker/k8s. The solutions will be different for each driver.

MarianRaphael:
Discussion during product sync: priority on Kubernetes

hardillb:
This should give us memory usage and a start on CPU metrics (but we may need to scale them to match the stack limits

hardillb:
OK, so I've exposed the CPU and Memory constraints to the nr-launcher but we need to decide when to trigger alerts.

I suggest that if the instance is above 75% of the limit for more than X samples then it should add a message to the audit log. 

The question is what the X should be (samples are currently every 5 seconds).

Also when should we reset and resend the alert?

Do we only send it once or if the usage drop bellow for Y samples then goes up again? I don't want to flood the audit log.

@knolleary @MarianRaphael  comments?

hardillb:
I'm suggesting averaging cpu and memory over the last minute (12 samples) for the trigger and reporting once until that average falls under.

The trigger point is 75% of the limits

ppawlowski:
5 seconds is quite often IMO. We should increase this to 15-30 seconds. 
[...]

PR #2968, opened 2023-10-18
Add Resource Audit Log entries

hardillb:
part of #2755 
## Description

Add the Audit Log formating

## Related Issue(s)

#2755 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/audit-log/AuditEntryVerbose.vue
+++ b/frontend/src/components/audit-log/AuditEntryVerbose.vue
@@ -503,12 +503,12 @@
 
     <template v-else-if="entry.event === 'resource.cpu'">
         <label>Instance High CPU usage</label>
-        <span>Instance has spent more than {{ entry.body.interval / 60 }} minutes at more than {{ entry.body.threshold }}% of CPU limit</span>
+        <span>Instance has spent more than {{ Math.floor(entry.body.interval / 60) }} minutes at more than {{ entry.body.threshold }}% of CPU limit</span>
     </template>
 
     <template v-else-if="entry.event === 'resource.memory'">
         <label>Instance High Memory usage</label>
-        <span>Instance has spent more than {{ entry.body.interval / 60 }} minutes at more than {{ entry.body.threshold }}% of Memory limit</span>
+        <span>Instance has spent more than {{ Math.floor(entry.body.interval / 60) }} minutes at more than {{ entry.body.threshold }}% of Memory limit</span>
     </template>
 
     <!-- Catch All -->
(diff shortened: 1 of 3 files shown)


frontend/src/components/audit-log/AuditEntryVerbose.vue
```diff
@@ -501,6 +501,16 @@
         <span>Nodes have been removed via the "Manage Palette" option inside Node-RED</span>
     </template>
 
+    <template v-else-if="entry.event === 'resource.cpu'">
+        <label>Instance High CPU usage</label>
+        <span>Instance has spent more than 5 minutes at more than 75% of CPU limit</span>
```
knolleary, line 506: I'm slightly concerned about having these numbers hardcoded here. If we find we want to tune the values, we're then stuck in how it gets reported to the user.

Should the audit log event include this information and we use it to generate the full message?

hardillb: Sorted, at the moment it defaults to minutes, do we want to use `moment.humanize`?

knolleary: I think until we have real experience of the results of this, it's hard to say. This'll do for now.

frontend/src/components/audit-log/AuditEntryVerbose.vue
```diff
@@ -501,6 +501,16 @@
         <span>Nodes have been removed via the "Manage Palette" option inside Node-RED</span>
     </template>
 
+    <template v-else-if="entry.event === 'resource.cpu'">
+        <label>Instance High CPU usage</label>
+        <span>Instance has spent more than {{ entry.body.interval / 60 }} minutes at more than {{ entry.body.threshold }}% of CPU limit</span>
```
knolleary, line 506:
```suggestion
        <span>Instance has spent more than {{ Math.floor(entry.body.interval / 60) }} minutes at more than {{ entry.body.threshold }}% of CPU limit</span>
```

frontend/src/components/audit-log/AuditEntryVerbose.vue
```diff
@@ -501,6 +501,16 @@
...
+    </template>
+
+    <template v-else-if="entry.event === 'resource.memory'">
+        <label>Instance High Memory usage</label>
+        <span>Instance has spent more than {{ entry.body.interval / 60 }} minutes at more than {{ entry.body.threshold }}% of Memory limit</span>
```
knolleary, line 511:
```suggestion
        <span>Instance has spent more than {{ Math.floor(entry.body.interval / 60) }} minutes at more than {{ entry.body.threshold }}% of Memory limit</span>
```

knolleary (approved):
Approved with a minor tweak - want to avoid decimal places when presenting this value to the user

PR #2973, opened 2023-10-19
Chore: Throw for unhandled exceptions in tests

Pezmc:
## Description

I discovered that have unhandled promise rejections being thrown in production meaning that some error states are being silently dropped without our knowledge.

Adding this guard to our tests should stop them slipping through the net while we work on more long term fixes.

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/2972

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Backport needed? -> add the `backport` label
 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #2978, opened 2023-10-19
Prevent error when device is not found

Pezmc:
## Description

Previously the following line was throwing if `device` wasn't defined, but the error was silently being dropped https://github.com/FlowFuse/flowfuse/blob/a8261d8cc5b325e868e7f330d5c39213405aab04/forge/comms/devices.js#L68-L70

Rather than adding a guard, I added a virtual attribute directly to the model, and then I took the opportunity to reactor all the `device.ownerType === 'application'` checks throughout our code (which were slightly inconsistent).

## Related Issue(s)

Fixes https://github.com/FlowFuse/flowfuse/issues/2972

Merge target is: https://github.com/FlowFuse/flowfuse/pull/2973

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/comms/devices.js
+++ b/forge/comms/devices.js
@@ -109,6 +109,11 @@ class DeviceCommsHandler {
                 }
             } catch (err) {
                 // Not a JSON payload - ignore
+                if (err instanceof SyntaxError) {
+                    return
+                }
+
+                throw err
             }
         }
     }
--- a/forge/comms/devices.js
+++ b/forge/comms/devices.js
@@ -66,7 +66,6 @@ class DeviceCommsHandler {
         if (status.id && status.status) {
             const deviceId = status.id
             const device = await this.app.db.models.Device.byId(deviceId)
-            const isApplicationOwned = device.ownerType === 'application' && device.Application?.id
             if (!device) {
                 // TODO: log invalid device
                 return
@@ -95,7 +94,7 @@ class DeviceCommsHandler {
                     if (payload.snapshot !== (targetSnapshot?.hashid || null)) {
                         // The Snapshot is incorrect
                         sendUpdateCommand = true
-                    } else if (targetSnapshot && !isApplicationOwned && payload.project !== (targetSnapshot?.ProjectId || null)) {
+                    } else if (targetSnapshot && !device.isApplicationOwned && payload.project !== (targetSnapshot?.ProjectId || null)) {
                         // The project the device is reporting it belongs to does not match the target Snapshot parent project
                         sendUpdateCommand = true
                     }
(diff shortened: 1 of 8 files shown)


forge/comms/devices.js
```diff
@@ -66,7 +66,6 @@ class DeviceCommsHandler {
         if (status.id && status.status) {
             const deviceId = status.id
             const device = await this.app.db.models.Device.byId(deviceId)
-            const isApplicationOwned = device.ownerType === 'application' && device.Application?.id
```
Pezmc, line 69: This was the line causing the error when `device` was undefined

forge/comms/devices.js
```diff
@@ -110,6 +109,11 @@ class DeviceCommsHandler {
                 }
             } catch (err) {
                 // Not a JSON payload - ignore
+                if (err instanceof SyntaxError) {
```
Pezmc, line 112: Was previously swallowing all exceptions, even those unrelated to JSON parsing.

forge/db/controllers/Device.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
             // Check the snapshot is one we recognise
             const snapshotId = app.db.models.ProjectSnapshot.decodeHashid(state.snapshot)
             // hashid.decode returns an array of values, not the raw value.
-            if (snapshotId.length > 0) {
+            if (snapshotId?.length > 0) {
```
Pezmc, line 20: If `state.snapshot` was not found in the DB this was raising an error (that was previously silently dropped)

forge/db/models/Device.js
```diff
@@ -26,10 +26,16 @@ module.exports = {
         mode: { type: DataTypes.STRING, allowNull: true, defaultValue: 'autonomous' },
         /** @type {'instance'|'application'|null} a virtual column that signifies the parent type e.g. `"instance"`, `"application"` */
         ownerType: {
-            type: DataTypes.VIRTUAL,
+            type: DataTypes.VIRTUAL(DataTypes.ENUM('instance', 'application', null)),
```
Pezmc, line 29: While in the area I added types to this virtual column

test/unit/forge/comms/devices_spec.js
```diff
@@ -203,4 +203,55 @@ describe('DeviceCommsHandler', function () {
             sockets[1].received().should.have.length(0)
         })
     })
+
+    describe('Device Status', function () {
```
Pezmc, line 207: This method previously had no test coverage, added some red/green coverage.

knolleary:
Tests failed as we had a test that was passing in a fake Device object into `insertPlatformSpecificEnvVars` which didn't have the new `isApplicationOwned` property on it.

Passing in fake objects isn't ideal in the tests for just this sort of reason - but not sinking time into that.

knolleary (approved):

PR #2996, opened 2023-10-23
feat: enable custom password and generate access token during initialization for ff-admin

AllanOricil:
## Goal

Enable users to programmatically manage their FF instance resources using the Rest API without requiring them to manually complete the setup, or create an admin user through the UI.

## Description

With these changes:
- `ff-admin` is created during start up.
- `ff-admin` password can be set using `FF_ADMIN_PASSWORD` env variable instead of being random.
- `ff-admin` personal access token can be created at start up using a new flow forge config called `create_admin_access_token`.
- when `create_admin = true` the setup is done automatically

I decided to move the `ff-admin` user creation from `./forge/routes/setup/index.js` to `./forge/forge.js` because the `ff-admin` creation no longer happens after accessing the `/setup` route, but when a forge server is started for the first time with no users.

## BEFORE
Screenshot 2023-10-25 at 16 31 16

## AFTER

image

## Related Issue(s)

N/A 
we discussed it on slack

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


AllanOricil:
@hardillb I've fixed all lint issues: 

image

knolleary:
I'm not sure we want this to be the default behaviour for a new install - which your changes to `etc/flowforge.yml` imply they would be.

You second screenshot also shows some of the fields it logs are mixed up - refresh_token and expires at. 

We also need to think about your prosed moving of this logic to the main start up rather than the setup route and what consequences that has. I'm not going to rush this into today's release, but once we resolve things, it can get into the next maintenance release.

AllanOricil:
@knolleary just pushed changes to fix what you mentioned

AllanOricil:
I've rebased it

AllanOricil:
@knolleary @hardillb is there a problem with these changes?

knolleary:
Sorry @AllanOricil, Ben and I have both been travelling this week. We'll get this PR reviewed properly asap.

hardillb:
Just looking this over, if we have both a admin created and a admin API token created, do we want to run `await app.settings.set('setup:initialised', true)` to disable the setup wizard?

This would only if both those options are enabled as we do not expect the user to access the web interface to complete the setup?

Otherwise it looks OK, will need to check it still applies cleanly

AllanOricil:
@hardillb 

One of the steps of the setup allows people to add a license. Is there another way to input it later?

image

hardillb:
Yes an admin can enter a license under the admin settings page

AllanOricil:
I'm going to add a condition to avoid running that piece of code you shared @hardillb Is that OK?

hardillb:
@AllanOricil I think we need to have an internal discussion about this first. It's a quick change we can add on if needed.

hardillb:
OK, having discussed this, calling `app.settings.set('setup:initialised', true)` is not going to work.

What will be better is to enable the token based auth for the last step of the setup wizard, which generates the default stacks and generates some other internal values that are needed.

This would mean that the first API call after the startup would be to complete setup properly which will in turn disable the setup wizard.

AllanOricil:
could you move the user creation call to the right place or show the line where it has to be moved to?

AllanOricil:
I think I know where to do it. I will finish it this weekend.

hardillb:
> This would mean that the first API call after the startup would be to complete setup properly which will in turn disable the setup wizard.

By this we mean after start up for the first time (when the user and API token are generated) the first API call should be to `/setup/finish` which creates the default instance type, stack and template and team type 

https://github.com/FlowFuse/flowfuse/blob/2004f93287e69712235317225d00cbcd8bd88de1/forge/routes/setup/index.js#L67

Which looking at the code doesn't need any authentication so there shouldn't need to be a code change

forge/forge.js
```diff
@@ -19,6 +21,38 @@ const settings = require('./settings')
...
+const generatePassword = () => {
+    const charList = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz~!@-#$'
+    return Array.from(crypto.randomFillSync(new Uint32Array(8))).map(x => charList[x % charList.length]).join('')
+}
+
+async function createAdminAccessToken (server, userId) {
+    const { token, expiresAt, refreshToken } = await server.db.controllers.AccessToken.createTokenForUser(userId, null, ['all'], true)
+    server.log.info(`[SETUP] token: ${token}`)
+    server.log.info(`[SETUP] refresh token: ${refreshToken}`)
+    server.log.info(`[SETUP] expires at: ${expiresAt}`)
```
knolleary, line 33:
```suggestion
    const { token, expiresAt, refreshToken } = await server.db.controllers.AccessToken.createPersonalAccessToken(userId, '', null, 'Admin Access Token')
    server.log.info(`[SETUP] token: ${token}`)
```

AllanOricil: done

AllanOricil:
> > This would mean that the first API call after the startup would be to complete setup properly which will in turn disable the setup wizard.
> 
> By this we mean after start up for the first time (when the user and API token are generated) the first API call should be to `/setup/finish` which creates the default instance type, stack and template and team type
> 
> https://github.com/FlowFuse/flowfuse/blob/2004f93287e69712235317225d00cbcd8bd88de1/forge/routes/setup/index.js#L67
> 
> Which looking at the code doesn't need any authentication so there shouldn't need to be a code change

this is done as well @hardillb

forge/setup.js
AllanOricil, line 1: I had to create this new file to avoid circular dependency. Both forge.js and routes/setup/setup.js need this logic.

AllanOricil:
Looking forward to becoming a contributor. The first Brazilian to contribute? Can I add it to me resume? If u give me a medal I can maybe solve some open issues for you haha I'm kidding. I do it mostly for curiosity and to learn.

test/system/002-setup-admin_spec.js
```diff
@@ -8,47 +11,32 @@ describe('Default Admin User', function () {
     // inbox - a local transport we can use to capture email without an SMTP server
     const inbox = new LocalTransport()
 
-    before(async function () {
```
hardillb, line 11: Why have you removed the `before` function? this is needed to reset the test environment.

hardillb:
This is failing tests, most likely because of the removed `before()` function.

Also did you rebase on main before the last update?

Thanks.

test/system/002-setup-admin_spec.js
AllanOricil: It is because I'm creating flowfuse inside each test method with different configs. I could not find another way.

hardillb: If you need to totally reset the environment between each test then the `before()` and `after()` functions should be renamed to `beforeEach()` and `afterEach()`

AllanOricil:
> This is failing tests, most likely because of the removed `before()` function.
> 
> Also did you rebase on main before the last update?
> 
> Thanks.

I merged main to the branch. Let me rebase it, instead.

test/system/002-setup-admin_spec.js
```diff
@@ -8,47 +11,32 @@ describe('Default Admin User', function () {
...
+        },
+        create_admin: true
+    }
+
+    async function createFlowForgeApp (config) {
+        return await FF_UTIL.setupApp({ ...defaultConfig, ...config })
+    }
 
     after(function () {
         return forge.close()
```
hardillb, line 35:
```suggestion
        return await forge.close()
```

This is probably the problem, you need to wait for the close to finish before the next test

test/system/002-setup-admin_spec.js
hardillb: OK, this can probably stay as it is, but the await in the `after()` will be needed

test/system/002-setup-admin_spec.js
hardillb: Ahh, this will also need a change to line 34 as well (which I can't do as a suggestion)

It will need to be:

```
    after( async function () {
        return await forge.close()
}
```

AllanOricil:
@hardillb try again. I changed after to afterEach, since every test has its own forge instance. I have also added the missing "async" keyword. 

image
image

AllanOricil:
I'm excited and emotive right now  🥺

forge/forge.js
```diff
@@ -15,9 +17,40 @@ const license = require('./licensing')
...
+        email_verified: true,
+        password,
+        admin: true,
+        password_expired: true
+    })
+    server.log.info('[SETUP] Created default Admin User')
+    server.log.info('[SETUP] username: ff-admin')
+    server.log.info(`[SETUP] password: ${password}`)
+
+    if (server.config.create_admin_access_token) await createAdminAccessToken(server, userId)
```
hardillb, line 51:
```suggestion
    if (server.config.create_admin_access_token) {
        await createAdminAccessToken(server, userId)
    }
```

Just a style thing

forge/forge.js
```diff
@@ -15,9 +17,40 @@ const license = require('./licensing')
...
+    return Array.from(crypto.randomFillSync(new Uint32Array(8))).map(x => charList[x % charList.length]).join('')
+}
+
+async function createAdminAccessToken (server, userId) {
+    const { token } = await server.db.controllers.AccessToken.createPersonalAccessToken(userId, '', null, 'Admin Access Token')
+    server.log.info(`[SETUP] token: ${token}`)
+}
+
+async function createAdminUser (server) {
+    if (await server.db.models.User.count() !== 0) return
```
hardillb, line 35:
```suggestion
    if (await server.db.models.User.count() !== 0) {
       return
    }
```

Just a style thing

hardillb (changes requested):
Looks good, one minor style thing, I prefer `if` blocks to to have curly braces as it makes them more obvious and easier to read.

And we could also do with another line adding to the `docs/install/configuration.md` file to describe the `create_admin_access_token` option under `create_admin` entry

AllanOricil:
Anything else master @hardillb ?

AllanOricil:
is it possible to exclude docs from codecov analysis? It failed because coverage decreased 0.04% after I added a new line in one of the docs

AllanOricil:
It is https://docs.codecov.com/docs/ignoring-paths

You can do it on the project config in codecov's UI, or in a codecov.yml file in the project

hardillb:
The docs are not covered by the codecov stuff, don't worry about it.

We will get this wrapped up this week, just been busy today

hardillb (approved):

hardillb:
@knolleary I'm happy with this if you want to give it a quick look.

AllanOricil:
@hardillb @knolleary I think we need a definition of a super admin because ff-admin can be deleted by any other admin. If an admin deleted ff-admin, then we lose control over the instance.

image

hardillb:
This is not related to this PR, but it is not possible to delete the last admin, so it is not possible to have no admin and  you should only promote trusted users to be an admin

AllanOricil:
🤨

PR #3003, opened 2023-10-25
Update built-in module list to include newly scoped packages

knolleary:
This updates the list of built-in modules we need to filter from the view.

As this list appeared in three different places, I've pulled them into one place that can be reused.


knolleary:
@Pezmc my first iteration of this just updated the lists in place. I then thought 'What would Pez say' and decided to pull it into a single place. Thank you for your review comment in my head before you ever saw the code.

hardillb (approved):

PR #3012, opened 2023-10-26
Fix recursion when adding mutating palette catalogue urls

Pezmc:
## Description

- Only mutate this.editable.changed.settings.palette_catalogue once per handling of the watch 
- Replace object duplication with a computed property - unrelated clean up

This is a quick fix, long term, we should look into removing the change tracking from the editable object.
Mutating an object inside its own watch handler is bad practice and should be avoided to prevent accidental recursion and improve performance.
 
## Related Issue(s)

No issue yet.

## Checklist

 - [ ] Suitable unit/system level tests have been added and they pass - **We should add test coverage**
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/admin/Template/index.vue
+++ b/frontend/src/pages/admin/Template/index.vue
@@ -125,18 +125,17 @@ export default {
                             changed = changed || pmChanges.changed
                             errors = errors || pmChanges.errors
                         } else if (field === 'palette_catalogue') {
-                            this.editable.changed.settings.palette_catalogue = false
-                            if (this.original.settings.palette_catalogue.length !== this.editable.settings.palette_catalogue.length) {
-                                this.editable.changed.settings.palette_catalogue = true
-                            } else {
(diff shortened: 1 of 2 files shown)


frontend/src/pages/admin/Template/index.vue
```diff
@@ -125,18 +125,17 @@ export default {
                             changed = changed || pmChanges.changed
                             errors = errors || pmChanges.errors
                         } else if (field === 'palette_catalogue') {
-                            this.editable.changed.settings.palette_catalogue = false
```
Pezmc, line 128: Previously this was set to false, then toggled to true if the palette_catelogue has changed, this meant that after the first watch handler had fired and was watching `this.editable.changed` too, the object would _always_ mutate.

Setting it only once means it does not mutate.

Pezmc:
As soon as the tests pass, we should get this onto staging to verify the fix.

knolleary (approved):

knolleary:
Merging once UI Tests passing

hardillb (approved):

Issue #2919, opened 2023-10-10
Allow new users to be created on first SSO logon


hardillb:
### Description

**As a** FlowFuse Admin 
**I want** to enable new users to be created on first login via SSO
**So that** I do not need to pre-create new users in FlowFuse with a matching email address

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

knolleary:
A very high level description of how this could work and what open questions we'd need to address:

 - Add an option to the SSO config option to enable user-provisioning
 - When a user logs in for the first time, we need to see what profile information we are able to get via SAML to populate the user object in our database. I suspect this could end up being fairly provider-specific...
 - There is then the decision of what team to put them in. If the 'auto-create team on sign-up' option is enabled (as we do in FF Cloud), we *could* treat this the same. However, if an org has setup SSO with us, it's unlikely they want each user to be put in a new Trial team. This suggests the SSO config needs to include nominating a default Team that new users should be added to.
 - Related to the above is #2829 which is about getting a user's team membership via the SAML config. There are lots of open questions about how that would work.

hardillb:
Just a thought I forgot to add, we may not need to enable this on ff-cloud, it may be of more use to self hosting deployments.

knolleary:
Complete via https://github.com/FlowFuse/flowfuse/pull/4207

Issue #2832, opened 2023-09-27
OAuth mapping with FlowFuse roles


zackwasli:
### Description

FlowFuse should map their RBAC roles (Owner, Member, Viewer, Dashboard) to OAuth so that permissions can be set via SSO.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

knolleary:
The technical challenge here is establishing a way to map from a customer's SAML configuration to something we understand in FlowFuse.

A user can be a member of multiple teams, each with a different role. So this would need to be able to determine their role in every team they should have access to.

There are other scenarios to consider - such as if a user creates a new team.

Before committing to this item, it would be useful to identify some potential users of such a feature so we can discuss more specific requirements with them.

johnebgood:
I'm looking into possibly customizing FlowFuse for integration with Azure Active Directory (AD) SSO (Now Azure Entra ID). I think having these SAML mapping features will open up the use of FlowFuse/Node-RED to many larger enterprises. 

I might be able to assist with describing features and mappings as I've integrated Azure AD with quite a few enterprise products. The most robust and easiest to configure that I've encountered is Splunk. I'll go through the mappings and interfaces for Splunk and then maybe someone who is more familiar with FlowFuse can walk me through where to start adding integrations for custom SAML endpoints and configuraitons to achieve similar results. Where I reference Azure AD groups these could also be LDAP roles or groups; I'm not as familiar with LDAP. Basically the SAML response contains a list of group memberships and FlowFuse would handle the team/capability/access mappings. One added benefit with utilizing Azure AD is having MFA/2FA available instantly. I've also integrated with Ping Federate, the FlowFuse group mapping interface could be done it a way to integrate with any SAML response as Splunk has done.

Example Splunk -> Azure AD integration:

1: Create access/capability groups in Azure Entra ID Active Directory (AD) The SAML response contains the Object Id's to map to Splunk roles.

![image](https://github.com/FlowFuse/flowfuse/assets/663211/101565b7-ba2c-4748-820c-ba1f6738aa5e)

2: Add AD users, they can be local accounts to the AD enviornment or invited users, so any existing Microsoft, Hotmail, live.com or other email can be used.

![image](https://github.com/FlowFuse/flowfuse/assets/663211/12b0884e-e6ca-4b14-9b5e-092e1ec72955)

[...]

PR #3017, opened 2023-10-27
Draft Pull Request: Added SSO/SAML group mapping options

johnebgood:
## Description
Draft PR for feedback.

Testing changes and requesting feedback for features to map SAML groups to teams, admin and user logins. Also the ability to default the login page to SSO provider and auto create users on initial sucessful login attempt from the SSO provider. This is my 2nd PR ever so apologies for my ignorance on coding and git best practices. If anyone could provide guidance on anything I should adjust in code, naming conventions etc. I would very much appreciate it.

Additions:

SSO Provider Settings additional fields:
image

Team Settings additional fields:
image

Added DB columns:

    // These were added but then realized that I was storing the values on the options field and they're not used. 😄 
    await context.addColumn("SAMLProviders", "allowAllSSO")
    await context.addColumn("SAMLProviders", "samlUserGroup")
    await context.addColumn("SAMLProviders", "samlAdminGroup")
    await context.addColumn("SAMLProviders", "defaultLogin")
    await context.addColumn("SAMLProviders", "autoCreateUser")

    // Added to pass SAML groups on the user object from ee/routes/sso/auth.js to Session.js
    await context.addColumn('Users', 'saml_groups')

    // Team group mappings
    await context.addColumn("Teams", "samlGroupOwner")
    await context.addColumn("Teams", "samlGroupMember")
    await context.addColumn("Teams", "samlGroupViewer")
    await context.addColumn("Teams", "samlGroupDashboard")

    // Added to track if a user was added to a team with SAML mapping. On SSO login all SAML mapped teams are removed 
    // and added fresh from the list of returned groups. This will allow for non-SAML mappings of teams that will persist.
    addColumn("TeamMembers", "samlAdded")
          
## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/2832
https://github.com/FlowFuse/flowfuse/issues/2919

## Checklist

   I have no idea how to make unit tests

### DIFF
--- a/forge/auditLog/user.js
+++ b/forge/auditLog/user.js
@@ -14,6 +14,9 @@ module.exports = {
             async login (actionedBy, error, user) {
                 await log('account.login', actionedBy, generateBody({ error, user: (error ? user : null) }))
             },
+            async samlAdminLogin (actionedBy, error, user) {
+                await log('account.saml-admin-login', actionedBy, generateBody({ error, user: error ? user : null }))
+            },
             async forgotPassword (actionedBy, error, user) {
                 await log('account.forgot-password', actionedBy, generateBody({ error, user }))
             },
--- a/forge/db/controllers/Session.js
+++ b/forge/db/controllers/Session.js
@@ -19,23 +19,28 @@ module.exports = {
         if (user && !user.suspended) {
             if (user.sso_enabled) {
                 try {
-                    const tmpSamlGroups = new Set(JSON.parse(user.saml_groups))
+                    const tmpSamlGroups = JSON.parse(user.saml_groups)
                     const allTeams = await app.db.models.Team.getAll()
                     const samlAssignedTeams = await user.getSAMLAssignedTeams()
 
                     // Remove all SAML assigned teams
-                    samlAssignedTeams.forEach(async (team) => {
-                        await app.db.controllers.Team.removeUser(team.Team, user)
-                    })
+                    for (const team of samlAssignedTeams) {
+                        app.log.info('Removing user ' + user.email + ' from team ' + team.Team.name)
+                        if (await app.db.controllers.Team.removeUser(team.Team, user)) {
+                            app.log.info('User was removed')
+                        } else {
+                            app.log.info('User was not removed')
+                        }
+                    }
 
-                    allTeams.teams.forEach(async (team, index) => {
+                    for (const team of allTeams.teams) {
                         let owner = false
                         let member = false
                         let viewer = false
                         let dashboard = false
 
                         // First iterate to determine if the user is a member of multiple groups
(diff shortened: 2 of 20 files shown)


knolleary:
Thanks for the PR @johnebgood  - I was hoping to join one of your Twitch live streams this week, but timezones didn't work in my favour :)

This PR is covering a lot of ground and a number of different items from our backlog, so it may take some time for us to work through all the details. We hadn't done any design work on our side yet for any of this, so we need to do some catch-up to properly understand what is being proposed here.

Part of that is better understanding on our part how SAML groups can be managed across the different SAML providers we already have customer using.

A large part of it is the UX of how any such mappings are configured. We have to keep in mind how this all appears to our FF Cloud platform users. Your screenshots show some level of configuration is needed in the individual Teams. I'm not sure that is suitable for how we manage our FF Cloud platform; most users aren't SSO enabled and presenting all of these options to them in their teams would be confusing.

I'll have a play with your PR over the next few days and see where we get to.

johnebgood:
Thanks for the feedback @knolleary, I completely understand that the way I did it is probably not suitable for all SSO providers or for release in the current state, I mainly wanted to hack in some features and get to know how the application works and test out some possibilities to communicate what I would like to see for the end result. The issue thread is probably not the best place for me to be hashing out what I'm working on either. I'm going to keep this draft PR open as I make adjustments and move stuff around, I would appreciate any feedback as I've been out of web development for 18 years. 😅

knolleary:
Hi @johnebgood 

Wanted to update you on what we've been doing in this space.

With the 2.0 release tomorrow we are introducing the ability to manage team membership via SAML groups. We've taken an approach that is somewhat different to yours, but is more suited to the FF Cloud use cases we have.


You can see the details in the PR: https://github.com/FlowFuse/flowfuse/pull/3291

To summarise the approach:

 - An SSO Configuration has a new option to enable team membership management
 - If enabled, it can be applied to the whole platform, or to select teams.
 - If 'selected teams' is chosen, the admin provides a list of the teams that are 'in scope' of the membership management

We then rely on a naming convention of the groups: `ff-<team slug>-<role>`.

This relies on the Identity Provider giving us the group names. I appreciate Microsoft Entra provides ids by default, but it can be configured to return the group names instead. This avoids the need to manage id/mappings on the platform side.

This does not cover:
 - user provisioning - we still require a user to be created on the platform first
 - managing admins - it almost got included, but was beyond what was needed for the immediate customer requirement so I pushed it back

Would be interested to hear your feedback on this.

hardillb:
@johnebgood I'm going to close this as we have now delivered our SAML approach and we are not going to merge this and it is getting more and more out of sync with the current main branch.

Thanks again for sharing this.

PR #3025, opened 2023-10-27
Avoid unnecessary retrieval of full snapshot contents where possible

### LINKED ISSUE #2980: N+1 Query: Project snapshot in /api/comms/auth/acl
Sentry Issue: [FLOWFUSE-BACKEND-P](https://flowfuse.sentry.io/issues/4551514656/?referrer=github_integration)
|  |  |
| ------------- | --------------- |
| **Offending Spans** | db - SELECT "id", "name", "description", "setti... 

```sql
UPDATE Devices SET lastSeenAt=CURRENT_TIMESTAMP,activeSnapshotId=$1,updatedAt=$2
WHERE id = $3
```
Repeating Spans (24)	
```sql
SELECT id, name, description, settings, flows, createdAt, updatedAt, ProjectId, DeviceId, UserId
FROM ProjectSnapshots AS ProjectSnapshot
WHERE ProjectSnapshot.id IN (
  1486
)
LIMIT 1;
```

knolleary:
Fixes #2980 

## Description

Whenever a device checks in, we validate the snapshot it reports is still one it is allowed to be using. This only requires checking the owner of the snapshot - it doesn't require the full contents of the snapshot.

Given how often devices check-in, we want to be mindful of not doing any more work than is needed.

This PR updates the `ProjectSnapshot.byId` function to accept options to turn off retrieval of the flows and settings objects.

It then updates any callers to that function that don't require the full object.

I did spot a couple places where the full object is retrieved in a preHandler - but the flows are only needed for one of the routes. I have documented that in the code - but didn't want to overdo the refactoring at this point.

### DIFF
--- a/forge/db/models/ProjectSnapshot.js
+++ b/forge/db/models/ProjectSnapshot.js
@@ -71,11 +71,10 @@ module.exports = {
                             exclude: toExclude
                         }
                     }
-                    const r = await self.findOne({
+                    return self.findOne({
                         where: { id },
                         attributes
                     })
-                    return r
                 },
                 forProject: async (projectId, pagination = {}) => {
                     const limit = parseInt(pagination.limit) || 1000
--- a/forge/db/models/ProjectSnapshot.js
+++ b/forge/db/models/ProjectSnapshot.js
@@ -51,15 +51,31 @@ module.exports = {
         const self = this
         return {
             static: {
-                byId: async function (id) {
-                    // This returns the full snapshot - including settings and flows
-                    // This should _only_ be used when getting all of that information
+                byId: async function (id, { includeFlows = true, includeSettings = true } = {}) {
+                    // By default, this returns the full snapshot - including settings and flows
+                    // This should _only_ be used when getting all of that information.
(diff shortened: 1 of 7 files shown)


forge/db/models/ProjectSnapshot.js
```diff
@@ -51,15 +51,31 @@ module.exports = {
...
+                    if (!includeSettings) {
+                        toExclude.push('settings')
+                    }
+                    let attributes = null
+                    if (toExclude.length > 0) {
+                        attributes = {
+                            exclude: toExclude
+                        }
+                    }
+                    const r = await self.findOne({
```
Pezmc, line 74: This can return directly, no need for r

Pezmc (approved):

forge/db/models/ProjectSnapshot.js
Pezmc: I added it in [85bf808](https://github.com/FlowFuse/flowfuse/pull/3025/commits/85bf808b8d71bc4581c7bc75ae75cfbff7ee2d44)

Pezmc (approved):

PR #3026, opened 2023-10-30
Prevent snapshots of devices without applications / instances

Pezmc:
- Red test coverage of creating snapshots of instance devices
- Guard against application owned devices
- Disable the create snapshot button if device does not have an application

## Description

Previously the create snapshot dialog was still available on the Device > Snapshots page if a device was not assigned to an application or instance.

This PR adds API level guards for creating instance snapshots of devices that are assigned to an application or not assigned to an instance and adds a simple UI guard for it too. The developer mode tab is not affected as that is only visible when the device is assigned to a project.

Screenshot 2023-10-30 at 13 15 18

Fixes https://github.com/FlowFuse/flowfuse/issues/2927

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/2927

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/controllers/ProjectSnapshot.js
+++ b/forge/db/controllers/ProjectSnapshot.js
@@ -76,12 +76,12 @@ module.exports = {
             ProjectId: project.id,
             UserId: user.id
         }
-        if (deviceConfig.flows) {
+        if (deviceConfig?.flows) {
             const projectSecret = await project.getCredentialSecret()
             snapshotOptions.flows.flows = deviceConfig.flows
             snapshotOptions.flows.credentials = app.db.controllers.Project.exportCredentials(deviceConfig.credentials || {}, device.credentialSecret, projectSecret)
         }
-        if (deviceConfig.package?.modules) {
+        if (deviceConfig?.package?.modules) {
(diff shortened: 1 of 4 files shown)


knolleary:
We want to be able to create snapshots of Application-owned devices. That is expected to work.

The only guard I believe we need is if the Device is completely unassigned.

Pezmc:
I believe that uses a completely separate endpoint. We have three endpoints that can create snapshots. Let me check when I’m at my laptop.

Pezmc:
Yes, we currently have:

- `POST /api/v1/devices/:deviceId/snapshot` - the one from this PR, creates snapshots for devices owned by an **instance**
- `POST /api/v1/devices/:deviceId/snapshots` - creates snapshot for a device owned by an application
- `POST /api/v1/projects/:projectId/snapshots` - creates a snapshot _of_ an instance

knolleary (approved):

knolleary:
Ooops - just after approving I though to check one more thing... and you merged just as I found an issue.

PR #3062, opened 2023-11-09
Enable CSP in report only mode

hardillb:
part of https://github.com/FlowFuse/security/issues/30
## Description

Enables Content-Security-Policy in report only mode to allow us to asses the if the allow lists are are correct for both staging and production 

## Related Issue(s)

https://github.com/FlowFuse/security/issues/30

## Checklist

    - [ ] Upgrade instructions
    - [x] Configuration details
    - [ ] Concepts
 - [x] Changes `flowforge.yml`?
    - [x] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template (https://github.com/FlowFuse/helm/pull/212)
    - [x] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production (https://github.com/FlowFuse/CloudProject/issues/288)

## Labels


hardillb:
Looking to add some tests

hardillb:
Also added solution to https://github.com/FlowFuse/security/issues/33

This will enable StrictTransportSecurity header is `base_url` starts with `https://` (this is the same test as used to build instance URLs.

forge/forge.js
```diff
@@ -153,14 +153,70 @@ module.exports = async (options = {}) => {
...
+
+            if (runtimeConfig.content_security_policy.report_only) {
+                contentSecurityPolicy.reportOnly = true
+                if (runtimeConfig.content_security_policy.report_uri) {
+                    contentSecurityPolicy.directives['report-uri'] = runtimeConfig.content_security_policy.report_uri
+                }
+            }
+
+            if (runtimeConfig.telemetry.frontend?.plausible?.domain) {
+                contentSecurityPolicy.directives['script-src'].push('plausible.io')
```
knolleary, line 185: Only slight concern is this assumes `directives['script-src']` exists and is an array. That'll be the case if the config hasn't provided a custom `directives` object and we've set it up ourselves (line 159 above). But if a custom object has been provided but doesn't contain everything we expect, this will fail.

hardillb: I'll add some tests that `script-src` exists and if not add it with the values (but if it doesn't exist in the user provided values it's going to break everything anyway).

knolleary (approved):

PR #3066, opened 2023-11-10
Allow admin to disable MFA for a user

knolleary:
## Description

Allows an admin user to disable MFA on a users account.

In absence of providing users recovery codes, this will be one path for a user to get back into their account via a support request.

### DIFF
--- a/frontend/src/pages/admin/Users/General.vue
+++ b/frontend/src/pages/admin/Users/General.vue
@@ -56,8 +56,10 @@ export default {
             nextCursor: null,
             columns: [
                 { label: 'User', class: ['flex-grow'], key: 'name', component: { is: markRaw(UserCell) }, sortable: true },
-                { label: 'Password Expired', class: ['w-56', 'text-center'], key: 'password_expired', sortable: true },
-                { label: 'Email Verified', class: ['w-56', 'text-center'], key: 'email_verified', sortable: true },
+                { label: 'Password Expired', class: ['w-32', 'text-center'], key: 'password_expired', sortable: true },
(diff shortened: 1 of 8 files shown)


frontend/src/pages/admin/Users/General.vue
```diff
@@ -56,8 +56,10 @@ export default {
             nextCursor: null,
             columns: [
                 { label: 'User', class: ['flex-grow'], key: 'name', component: { is: markRaw(UserCell) }, sortable: true },
-                { label: 'Password Expired', class: ['w-56', 'text-center'], key: 'password_expired', sortable: true },
-                { label: 'Email Verified', class: ['w-56', 'text-center'], key: 'email_verified', sortable: true },
+                { label: 'Password Expired', class: ['w-32', 'text-center'], key: 'password_expired', sortable: true },
+                { label: 'Email Verified', class: ['w-32', 'text-center'], key: 'email_verified', sortable: true },
+                { label: 'SSO', class: ['w-32', 'text-center'], key: 'sso_enabled', sortable: true },
```
Pezmc, line 61: Where is `sso_enabled` coming from?

Pezmc (approved):

frontend/src/pages/admin/Users/General.vue
knolleary: It's a pre-existing flag on the user object. Thought it useful to add to what we expose here... although it's getting to a point where this table needs a bit of redesign.

knolleary:
This can be merged once #3065 has been reviewed. I'm holding off merging it now to avoid complicating the 3065 review.

Issue #3018, opened 2023-10-27
Project Nodes unavailable on a Device assigned to an Application


knolleary:
### Current Behavior

When a Device is assigned to an Application, not an instance, the Project nodes are unavailable.

This is because the Project Nodes expect to be operating within the context of an instance in order to establish the proper ACL checks.

This is not a regression - this has been a limitation since the ability to assigned devices to Applications was introduced. However it needs to be addressed as users expect to use the Project Nodes regardless.

### Expected Behavior

The Project Nodes should be available to Application-assigned devices

### Steps To Reproduce

 - Assign device to Application
 - Put in developer mode
 - open editor - note lack of Project nodes

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Steve-Mcl:
@knolleary 

What I have working thus far:

1. App Assigned Device can broadcast (and other instances / devices) can receive them
2. App Assigned Device can send direct to an instance (that instance can of course be an instance assigned device) 
3. Can perform link-call to an instance 


https://github.com/FlowFuse/flowfuse/assets/44235289/d52a3ca3-c6f4-4143-9415-e3e8352f13e1




Missing functionality:
1. An instance or other device CANNOT send direct to an App Assigned Device
2. An instance or other device CANNOT link-call to an App Assigned Device


Having had a bit of a chew on getting to this point (dev env issues, debugging issues, etc) I must apologise for forgetting the exact details of our chat last Monday. That leads to the question:

Is it enough in this iteration to support the "working set" (as listed in the first 3 points) - OR - is it desirable to get the "missing functionality" implemented too?

---

CC @MarianRaphael ^

knolleary:
The goal is a minimal set of functionality that starts filling the gap caused by the absence of the project nodes in an App-assigned device.

Lets get what you've done so far into a PR so we can start reviewing. We need to make sure the MQTT topic structures are appropriate and your video doesn't show anything about how these new options are exposed in the UI.

We will want to be able to messaging directed at the Device, but that can be a second iteration.

Steve-Mcl:
> We need to make sure the MQTT topic structures are appropriate

[...]

PR #3074, opened 2023-11-14
Add app device access verification to ACL manager

Steve-Mcl:
part of #3018

## Description

This pull request adds device access verification to the ACL manager, allowing for devices assigned to an application to use the nr-project-nodes 

The changes include new functions for checking device access to projects and applications, as well as updates to existing functions for verifying device access to topics.
### TODO: 
- Explore possible tests
- Update docs
- (Next iteration) support device/instance --> to device direct messaging

#3018

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/comms/aclManager.js
+++ b/forge/comms/aclManager.js
@@ -25,10 +25,20 @@ module.exports = function (app) {
 
             // Do the simple team id check
             if (requestParts[1] !== ids[1]) {
-                return false
+                return false // not in this team
             }
+
+            // check to see if the device is assigned to an instance?
             const assignedProject = await app.db.models.Device.getDeviceProjectId(ids[2])
-            return !!assignedProject
+            if (assignedProject) {
+                return true
+            }
+            // check to see if the device is assigned to an application
+            const assignedApplication = await app.db.models.Device.getDeviceApplicationId(ids[2])
+            if (assignedApplication) {
+                return true
+            }
+            return false
         },
         checkDeviceAssignedToProject: async function (requestParts, ids) {
             // requestParts = [ _ , <teamid>, <projectid> ]
@@ -64,12 +74,16 @@ module.exports = function (app) {
             }
             // Get the project this device is assigned to
             const assignedProject = await app.db.models.Device.getDeviceProjectId(ids[2])
-            if (!assignedProject) {
-                return false
-            }
-            if (assignedProject === requestParts[2]) {
-                // Access the project we're assigned to - all good
-                return true
+            if (assignedProject) {
+                if (assignedProject === requestParts[2]) {
+                    // Access the project we're assigned to - all good
+                    return true
+                }
+            } else {
+                const assignedApplication = await app.db.models.Device.getDeviceApplicationId(ids[2])
+                if (!assignedApplication) {
+                    return false // Device is not assigned to an application or a project - no access
+                }
             }
 
             // Need to check if this project is in the same team.
@@ -192,6 +206,10 @@ module.exports = function (app) {
                             break
                         } else if (aclList[i].verify && verifyFunctions[aclList[i].verify]) {
                             allowed = await verifyFunctions[aclList[i].verify](m, usernameParts)
+                            // ↓ Useful for debugging ↓
+                            // if (allowed !== true) {
+                            //     console.log('DENIED!', topic, aclList[i])
+                            // }
                         } else {
                             allowed = true
                         }
--- a/forge/ee/lib/projectComms/index.js
+++ b/forge/ee/lib/projectComms/index.js
@@ -62,41 +62,59 @@ module.exports.init = function (app) {
             'sub',
             { topic: /^ff\/v1\/([^/]+)\/p\/([^/]+)\/out\/[^/]+($|\/.*$)/, verify: 'checkDeviceIsAssigned' }
         )
-        // Receive messages sent to this project
+        // Receive messages sent to this device
         // - ff/v1/<team>/p/<project>/in/+/#
         app.comms.aclManager.addACL(
             'device',
             'sub',
-            { topic: /^ff\/v1\/([^/]+)\/p\/([^/]+)\/in\/[^/]+($|\/.*$)$/, verify: 'checkDeviceAssignedToProject' }
+            { topic: /^ff\/v1\/([^/]+)\/p\/([^/]+)\/in\/[^/]+($|\/.*$)$/, verify: 'checkDeviceCanAccessProject' }
         )
-        // Receive link-call response messages sent to this project
+        // Receive link-call response messages sent to this device (instance owned)
         // - ff/v1/<team>/p/<project>/res/+/#
         app.comms.aclManager.addACL(
             'device',
             'sub',
-            { topic: /^ff\/v1\/([^/]+)\/p\/([^/]+)\/res\/[^/]+($|\/.*$)$/, verify: 'checkDeviceAssignedToProject' }
+            { topic: /^ff\/v1\/([^/]+)\/p\/([^/]{11,})\/res\/[^/]+($|\/.*$)$/, verify: 'checkDeviceCanAccessProject' }
+        )
+        // Receive link-call response messages sent to this device (app owned)
+        // - ff/v1/<team>/p/<appid>/res/+/#
+        app.comms.aclManager.addACL(
+            'device',
+            'sub',
+            { topic: /^ff\/v1\/([^/]+)\/p\/([^/]{10})\/res\/[^/]+($|\/.*$)$/, verify: 'checkDeviceAssignedToApplication' }
         )
 
-        // Send message to other project
-        // - ff/v1/<team>/p/+/in/+/#
+        // Send message to specific project
+        // A device ID is exactly 10 chars long, so any topic with an ID longer than that
+        // is a project instance topic
+        // At this time, we don't support sending messages direct to an app owned device
+        // - ff/v1/<team>/p/<project>/in/+/#
         app.comms.aclManager.addACL(
             'device',
             'pub',
-            { topic: /^ff\/v1\/([^/]+)\/p\/([^/]+)\/in\/[^/]+($|\/.*$)/, verify: 'checkDeviceCanAccessProject' }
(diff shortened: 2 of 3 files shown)


forge/ee/lib/projectComms/index.js
```diff
@@ -62,41 +62,59 @@ module.exports.init = function (app) {
...
+        app.comms.aclManager.addACL(
+            'device',
+            'sub',
+            { topic: /^ff\/v1\/([^/]+)\/p\/([^/]{10})\/res\/[^/]+($|\/.*$)$/, verify: 'checkDeviceAssignedToApplication' }
         )
 
-        // Send message to other project
-        // - ff/v1/<team>/p/+/in/+/#
+        // Send message to specific project
+        // A device ID is exactly 10 chars long, so any topic with an ID longer than that
```
knolleary, line 88: It's a bit hard to figure out due to the way the git diff is showing the changes - but I'd expect any future support for a publish targeting a device to be on a `/d/` topic, not a `/p/` topic - so not sure this reliance on the length of this topic component is something we should rely on.

It would be helpful to a have a summary of the topics/rules being introduced in the PR description for ease of reference. I haven't got to the other related PRs to know if you've documented them elsewhere, but as this is the PR dealing with changing the platform ACLs, it would be handy to have here.

docs/user/projectnodes.md
```diff
@@ -9,25 +9,27 @@ very quick and easy to securely send and receive messages between different
 instances in a team.
```
knolleary, line 9: We should add a line to say the nodes are only available on the Team/Enterprise plans

knolleary (changes requested):

docs/user/projectnodes.md
Steve-Mcl: Updated in [6437fb5](https://github.com/FlowFuse/flowfuse/pull/3074/commits/6437fb5f91422d7e1ff3edc8d42eb891b53b35cf)

Adds the line "The project nodes are only available in the Team and Enterprise tiers of FlowFuse."

forge/comms/aclManager.js
```diff
@@ -64,12 +74,16 @@ module.exports = function (app) {
...
-            if (assignedProject === requestParts[2]) {
-                // Access the project we're assigned to - all good
-                return true
+            if (assignedProject) {
+                if (assignedProject === requestParts[2]) {
+                    // Access the project we're assigned to - all good
+                    return true
+                }
+            } else {
+                const assignedApplication = await app.db.models.Device.getDeviceApplicationId(ids[2])
```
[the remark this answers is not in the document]
Steve-Mcl: Resolved in [eba6e63](https://github.com/FlowFuse/flowfuse/pull/3074/commits/eba6e6371bd4682fd223e5e2c8db6127bea3bcac)

forge/comms/aclManager.js
```diff
@@ -25,10 +25,20 @@ module.exports = function (app) {
...
             }
+
+            // check to see if the device is assigned to an instance?
             const assignedProject = await app.db.models.Device.getDeviceProjectId(ids[2])
-            return !!assignedProject
+            if (assignedProject) {
+                return true
+            }
+            // check to see if the device is assigned to an application
+            const assignedApplication = await app.db.models.Device.getDeviceApplicationId(ids[2])
```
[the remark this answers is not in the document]
Steve-Mcl: Resolved in [eba6e63](https://github.com/FlowFuse/flowfuse/pull/3074/commits/eba6e6371bd4682fd223e5e2c8db6127bea3bcac)

forge/ee/lib/projectComms/index.js
Steve-Mcl: It was an absolute melon twister Nick 😄 

I will attempt to answer the your questions in a meaningful manner (I am also working on suitable documentation for future reference)


>  but I'd expect any future support for a publish targeting a device to be on a /d/ topic, not a /p/ topic 

Agreed and considered and will likely be how it is handled for instance/device → device, but at this point in time, it is not supported and is in effect interlocked by insisting on the topic being 11 or more characters.  I suspect I will remove these once I have a working schema for instance/device → device



However, it was necessary to differentiate a PUB originating from an instance device vs PUB originating from an app assigned device in some scenarios and therefore an alternate verification routine was needed

Consider these

## "Send message to specific project"
### instance assigned device publishes direct to a project
1. The device has access to a `projectId` and can form the topic `ff/v1/7N152GxG2p/p/23d79df8-183c-4104-aa97-8915e1897326/in/a/b` 
    * this is the topic it uses to PUB direct to project `23d79df8-183c-4104-aa97-8915e1897326`
2. The receiving projects' Project-In node subscribes to `a/b` which equates to `ff/v1/7N152GxG2p/p/23d79df8-183c-4104-aa97-8915e1897326/in/a/b`

### application assigned device publishes direct to a project
1. As the Project In Node will subscribe to `ff/v1/7N152GxG2p/p/23d79df8-183c-4104-aa97-8915e1897326/in/a/b` the device must also publish on this topic

In summary - the `/p/` remains even though it is a device publishing


## "Broadcast a message"
### instance assigned device
1. The device has access to a `projectId` and can form the topic `ff/v1/7N152GxG2p/p/23d79df8-183c-4104-aa97-8915e1897326/out/a/b` 
    * this is the topic it uses to PUB broadcasts `23d79df8-183c-4104-aa97-8915e1897326`

### Project In subscription
3. A receiving projects' Project-In node subscribes to `a/b` which equates to `ff/v1/7N152GxG2p/p/+/out/a/b`
4. Therefore a device (both instance and app owned) Project-In node must also subscribe to `ff/v1/7N152GxG2p/p/+/out/a/b`

### application assigned device
1. The device does not have a `projectId`, only its own `deviceId` or the `applicationId`
2. Since the code already knows which device is which, I supplant the `projectId` with the  `applicationId`
5. This means we get the topic  `ff/v1/7N152GxG2p/p/1234567890/out/a/b`  (where `1234567890` is the app id)
6. This means we cannot use the same verification function since we are not verifying a project ID but an application ID

In summary - the `/p/` remains even though it is an application device publishing

To get over this hump, I chose to leave the Project In node as-is and differentiate the verification logic path by the `/ID/` provided.


The alternative solution I considered was to modify the Project-In node to either subscribe to both `ff/v1/7N152GxG2p/p/...` and `ff/v1/7N152GxG2p/d/...` OR subscribe to `ff/v1/7N152GxG2p/+/...`


---

Tomorrow, I hope to have a clearer picture of the direct instance/device → device and a full summary of current existing condition, this PR condition and our final stance.  

I think it may be prudent to hang fire and evaluate this before any merges?


---

PS: I appreciate that ↑ wasn't entirely as clear as I had hoped!  Imagine how my head was noodled doing the work ;)

Steve-Mcl:
@knolleary 

I have got all tests passing and managed to format a response that doesnt mess up regex (harder than you might imagine 😆 

Anyhow, I have one remaining reservation and that is, as noted before in the review comment, this PR is publishing app-owned-devices broadcasts out on `xxx/p/xxx` - and it works, and this is how it needs to be so that project-in nodes on an instance can subscribe to the broadcasts when the target is "all instances"

As discussed before, this was to minimise multiple subs or using a wildcard.

I am now thinking should we make separate entries in the `project-in` node dropdown.  e.g:

[ All instances ]
[ All devices  ]
[ instances 1 ]
[ instances 2 ]
[ instances n ]


This would permit us the change the topic ACL to `xxx/d/xxx` and remove the `{11,}` length check.

Steve-Mcl:
Hi @knolleary 

I have updated the topics and ACLs to use the `app:` prefix 

I will update the OP with the new topic/ACLs for reference shortly.

Steve-Mcl:
@knolleary - topic prefix adjusted as per discussion.

Had an odd Postrgre only tests fail (re-running it)

knolleary (approved):

knolleary:
Approved pending the postgres test passing

Steve-Mcl:
@knolleary I think I clicked Request Approval milliseconds after you clicked "Approved" - can you do it again please?

knolleary:
@Steve-Mcl Postgres test failed again. Need to understand that.

knolleary:
Postgres failure is can be reproduced locally:

```
[2023-11-22T14:14:08.735Z] ERROR: invalid input syntax for type uuid: "rbXm6nmA62" {"reqId":"req-11","req":{"method":"POST","url":"/api/comms/auth/acl","hostname":"localhost:80","remoteAddress":"127.0.0.1"},"res":{"statusCode":500,"request":{"url":"/api/comms/auth/acl","method":"POST","remoteAddress":"127.0.0.1"}}}
    err: {
      "type": "DatabaseError",
      "message": "invalid input syntax for type uuid: \"rbXm6nmA62\"",
      "stack":
          Error
              at Query.run (/Users/nol/code/ff/dev-env/packages/flowfuse/node_modules/sequelize/lib/dialects/postgres/query.js:50:25)
              at /Users/nol/code/ff/dev-env/packages/flowfuse/node_modules/sequelize/lib/sequelize.js:315:28
              at async PostgresQueryInterface.select (/Users/nol/code/ff/dev-env/packages/flowfuse/node_modules/sequelize/lib/dialects/abstract/query-interface.js:407:12)
              at async Function.findAll (/Users/nol/code/ff/dev-env/packages/flowfuse/node_modules/sequelize/lib/model.js:1140:21)
              at async Function.findOne (/Users/nol/code/ff/dev-env/packages/flowfuse/node_modules/sequelize/lib/model.js:1240:12)
              at async Function.getProjectTeamId (/Users/nol/code/ff/dev-env/packages/flowfuse/forge/db/models/Project.js:459:37)
              at async Object.checkDeviceCanAccessProject (/Users/nol/code/ff/dev-env/packages/flowfuse/forge/comms/aclManager.js:90:35)
              at async Object.verify (/Users/nol/code/ff/dev-env/packages/flowfuse/forge/comms/aclManager.js:209:39)
              at async Object.<anonymous> (/Users/nol/code/ff/dev-env/packages/flowfuse/forge/comms/authRoutes.js:54:25)
```

knolleary:
The failing test is because it's checking a device cannot broadcase if it doesn't provide the `dev:` prefix. This leds the code to assumes its a project id - which postgres rejects because it doens't look like a uuid - unlike sqlite that just returns no results.

knolleary (approved):

Issue #2970, opened 2023-10-18
Update Setup Billing page to offer Team Type selection


knolleary:
Part of #2905

The purpose here is to improve the UX around conversion from Trial to paid user.

When a user signs up, they get a Starter Team in Trial mode. They are directed to the Team Billing page where they are invited to go add their details in stripe. With the work delivered in #2963, this results in their trial ending.

This issue is to update the Team Billing page for a Trial Mode user to offer a selection of the available Team Types before going to stripe. This would allow a Trial mode user to go straight to the Team/Enterprise tier, rather than it being a two-stage process (set up billing, *then* change team type).

I've copied in below input from @joepavitt on the design for this particular aspect.

The Description of the team tiers is currently optimised for the tile selection display which is limited in space. We could stick with that for a first iteration, or if we want the larger scale display, we need to consider how that impacts the existing 'change team type' view.

---

_The following originally posted by @joepavitt in https://github.com/FlowFuse/flowfuse/issues/2328#issuecomment-1624032062_

Quick thoughts based on https://github.com/flowforge/flowforge/issues/2328#issuecomment-1612963139 (note the empty state graphic will be updated to reflect the page, just put that in as a placeholder), where the rows in the plan options would be driven by what they're using right now in their Team, e.g. in this example we have 2 x Team Members, 1 x Instance and 2 x Devices.

"Upgrade to _Tier_" options would then go to Stripe.

### Figma Prototype

https://www.figma.com/proto/rMB2w5ZZUtTC0nxbwLYKJY/FlowForge-Platform?page-id=0%3A1&type=design&node-id=2493-18259&viewport=-7845%2C-5052%2C0.74&scaling=min-zoom&starting-point-node-id=2493%3A18259&show-proto-sidebar=1&mode=design&t=DhF8dJvDse2TG530-1

### Static Screenshots:

<img width="1705" alt="Screenshot 2023-07-06 at 18 13 55" src="https://github.com/flowforge/flowforge/assets/99246719/6937f37f-5cd2-487f-9b85-c80684fb545f">

<img width="1682" alt="Screenshot 2023-07-06 at 18 14 01" src="https://github.com/flowforge/flowforge/assets/99246719/e1d382fc-56c1-45d5-9f6d-94b02f3954a1">

knolleary:
Via #3092 the billing setup workflow has been updated. However the team type selection via still uses the original TileSelection component which doesn't match the ideal designs.

[...]

PR #3092, opened 2023-11-17
Update billing setup page

knolleary:
Part of #2970

## Description

Updates the billing setup UX to allow choice of TeamType before going to stripe.

The existing code assumed we only needed to setup the subscription with the Team plan item - originally on the basis this was done before any instances/devices were created, and then on the basis that it was done for a team in trial mode.

This PR updates that code to ensure the initial subscription object contains all of the appropriate line-items for the desired team type.

1. A user starts with a Starter team in trial mode
2. They go to billing, which directs them to select what type of team they want - Starter/Team/Enterprise
3. We create a subscription for the selected team type and direct user to Stripe checkout that shows all the appropriate details
4. Eventually the user checksout and Stripe notifies us via webhook it has happened. At this point, we update the Team's type to reflect what they chose. We cannot do it before this point because if they abandon the stripe checkout process we need to keep their team as it was before they started the billing setup.

```[tasklist]
### Remaining work before full review
 - [x] UI Test fixes
 - [x] Tests for the new subscription setup flow and stripe callback handling
 - [ ] (possible) Improve the Team Type selection UX - the `ff-tile-selection` is a much more compact design then the concept designs done. Will see if there's a quick iteration of improvement to be had
 - [x] Any necessary doc updates
```


knolleary:
To help get this over line, I've tagged in @hardillb and @joepavitt to help review.

@joepavitt can you take a look at the TileSelection changes specifically to make sure you're happy with them? (and any other frontend stuff that catches your eye)

@hardillb can you go through the backend logic - ping if you want to step through together.

hardillb (approved):
I've been through all the Backend and Unit test code and it all looks good to me.

joepavitt:
If. click away to a different team type, I can't then click back to the original selection, there is no "back" or "cancel" option either?

Screenshot 2023-11-22 at 14 18 54

joepavitt (approved):

knolleary:
@joepavitt via last commit, have added a cancel button that does a simple `$router.back()` to take you back to wherever you came from.

Also turned off the disabling of the current team-type - if you have that one selected, the confirm button is disabled.

image

joepavitt:
That's much better - thanks Nick

forge/ee/lib/billing/index.js
```diff
@@ -70,6 +82,50 @@ module.exports.init = async function (app) {
...
 
+            // Need to ensure the subscription contains all of the expected items
+            // to correlate with the teamType. This includes:
+            // - Team Plan item - already added above
+            // - Device item
+            // - An item for each instance type
+
+            // Check if a device item is required
+            const deviceBillingIds = await teamType.getDeviceBillingIds()
+            if (deviceBillingIds.product) {
```
Pezmc, line 93: I'm hesitant this this is near-duplicating the logic of `updateTeamDeviceCount` (and below) `updateTeamInstanceCount`. I'd much prefer we have one single place like a BillingService that handles all of this kind of logic, but appreciate that release is tomorrow so don't want to block on refactoring...

Pezmc: Given that `updateTeamInstanceCount` and `updateTeamInstanceCount` both by outputting line items needed for the subscription, I can see a reasonable route where we have a single helper class that calculates the appropriate line items for a subscription, then we update/push them as required.

https://github.com/FlowFuse/flowfuse/blob/0500dc63c7d745dbfc508d84a76dea2cdb1fb83b/forge/ee/lib/billing/index.js#L269

https://github.com/FlowFuse/flowfuse/blob/0500dc63c7d745dbfc508d84a76dea2cdb1fb83b/forge/ee/lib/billing/index.js#L170

Pezmc: This can be a follow up though, so I propose we open an issue and unblock this PR.

frontend/src/api/billing.js
```diff
@@ -13,8 +13,12 @@ const getSubscriptionInfo = async (teamId) => {
 }
 
 // Create a new subscription for a team
-const createSubscription = async (teamId) => {
-    return client.post('/ee/billing/teams/' + teamId).then(res => {
+const createSubscription = async (teamId, teamTypeId) => {
```
Pezmc, line 16: Given `teamTypeId` is optional, it should be set to undefined (even though redundant) to notate this:

```suggestion
const createSubscription = async (teamId, teamTypeId = undefined) => {
```

forge/ee/lib/billing/index.js
knolleary: Agreed - the logic is close but not quite identical and didn't want to fall into the hole of trying to pull out the common parts.

PR #3138, opened 2023-11-28
Allow a trial team to be put into unmanaged mode to bypass billing

knolleary:
## Description

This PR allows a trial team to be put into 'unmanaged' mode. Once in this mode, the team no longer interacts directly with Stripe - on the assumption that all billing is being handled manually.

Once in 'unmanaged' mode, a team type cannot be changed, but it is free to create instances/devices as needed.

For this iteration, the Billing page notifies the user they cannot manage their subscription and should contact FF Support for help. The Change Team Type page does similar.

Under the covers, this adds a new state to the Subscription object - `unmanaged` (inc migration for postgres enum).

The API end point for putting a Trial Team into this mode is only accessible by Admin users.

The Team Settings - Danger page has an Admin Only Tools section (only shown when viewing as Admin user on a Trial Team):

image

The Setup manual billing button opens a dialog to confirm - and select the desired TeamType

image

### DIFF
--- a/forge/ee/db/models/Subscription.js
+++ b/forge/ee/db/models/Subscription.js
@@ -12,7 +12,10 @@ const STATUS = {
     CANCELED: 'canceled',
     PAST_DUE: 'past_due',
     // Local only status, not from Stripe
-    TRIAL: 'trial'
+    TRIAL: 'trial',
+    // Local only - means this team's subscription on stripeon strip is not
+    // managed by the platform
+    UNMANAGED: 'unmanaged'
 }
 
 Object.freeze(STATUS)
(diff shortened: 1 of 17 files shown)


forge/ee/db/models/Subscription.js
```diff
@@ -12,7 +12,10 @@ const STATUS = {
     CANCELED: 'canceled',
     PAST_DUE: 'past_due',
     // Local only status, not from Stripe
-    TRIAL: 'trial'
+    TRIAL: 'trial',
+    // Local only - means this team's subscription on stripeon strip is not
```
Pezmc, line 16:
```suggestion
    // Local only - means this team's subscription on stripe is not
```

knolleary:
UI Tests failing consistent for a seemingly unrelated error. Will investigate.

Pezmc (approved):
Thank you for the bump, 👍

Issue #3034, opened 2023-11-01
Set a canonical URL for app.flowfuse.com 


ZJvandeWeg:
### Description

Just for a bit of an SEO boost, and it's a simple change to make I expect

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #3146, opened 2023-12-01
Add canonical link to current base_url

hardillb:
fixes https://github.com/FlowFuse/flowfuse/issues/3034

## Description

Add a canonical link to app.flowfuse.com to it updates SOE links to app.flowforge.com

It only gets applied if frontend telemetry is enabled which is fine for FF Cloud.

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/3034

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/ui/index.js
+++ b/forge/routes/ui/index.js
@@ -69,6 +69,10 @@ module.exports = async function (app) {
                 injection += `<script>window.sentryConfig = { dsn: "${telemetry.frontend.sentry.dsn}", production_mode: ${telemetry.frontend.sentry.production_mode ?? true}, version: "flowfuse@${config.version}", environment: "${process.env.SENTRY_ENV ?? (process.env.NODE_ENV ?? 'unknown')}" }</script>`
             }
 
+            if (config.base_url) {
+                injection += `<link rel="canonical" href="${config.base_url}" />`
+            }
+
             // inject into index.html
             cachedIndex = data.replace(/<script>\/\*inject-ff-scripts\*\/<\/script>/g, injection)
         }


knolleary (approved):

forge/routes/ui/index.js
```diff
@@ -69,6 +69,10 @@ module.exports = async function (app) {
                 injection += `<script>window.sentryConfig = { dsn: "${telemetry.frontend.sentry.dsn}", production_mode: ${telemetry.frontend.sentry.production_mode ?? true}, version: "flowfuse@${config.version}", environment: "${process.env.SENTRY_ENV ?? (process.env.NODE_ENV ?? 'unknown')}" }</script>`
             }
 
+            if (config.base_url) {
+                injection += `<link rel="canonical" href="${config.base_url}" />`
```
ZJvandeWeg, line 73: @hardillb @knolleary This doesn't work as you're not including the path element. E.g. `/accounts/create` is now just pointing to the base URL?

Than this seems to be included an _all_ pages, so if any page later will be more open to be more like a github for NR, all cacnonical URLs will be wrong.

hardillb: yes, this applies to all pages.

We can revert it for now and look to see if we can apply the route as well as the host.

ZJvandeWeg: I don't think the revert is worth it, though a follow up should be scheduled. Not sure if VueJS for example can update the link tag dynamically?

hardillb: This might be useful

https://www.digitalocean.com/community/tutorials/vuejs-vue-seo-tips

hardillb: @ZJvandeWeg https://github.com/FlowFuse/flowfuse/issues/3298

Pezmc: @ZJvandeWeg @hardillb A crawler shouldn't really be indexing the pages _after_ login anyway, so setting the canonical URL to the login page should be good enough!

ZJvandeWeg: @Pezmc There's still two different pages, one needed a path set, the other just worked by the base_url as there's no path. So I'm not quite sure what you're saying with your comment.

cstns: The canonical tag is only used by search engines to determine the “main” version of the page. The only pages that the search engines can access are the public ones, mainly login and register (I'm not currently aware of any other pages that are do not require authentication, search engines/crawlers are not able to bypass auth).

Having the canonical tag on pages behind auth is somewhat redundant so we can focus on the tag having the correct values on these two.

As @hardillb mentioned, there are ways to dynamically alter meta tags but we should focus to add them only where they matter. 

I'd go with hardcoded values if the pages mentioned above are the only publicly accessible ones, there's no need for the overhead of an additional library IMO.

PR #3148, opened 2023-12-01
Improve canceled subs handling

knolleary:
## Description

This improves our handling around canceled stripe subscriptions.

We have seen a couple instances where subscription is canceled on stripe, but our local state thinks it is active.

This PR adds a check whenever we come to do billable work to resync our local status with the true stripe status. If we find the status has changed, and that it is now canceled, we suspend all team instances and log it.

It also:
 - adds a proper entry for the `billing.subscription.updated` audit-log event
 - fixes the Team Billing page for when the subscription has expired/canceled - it was previously talking about trials which isn't correct for this state.

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -549,7 +549,7 @@ module.exports.init = async function (app) {
                             await app.billing.updateSubscriptionStatus(subscription, stripeSubscription.status, team)
                             await app.db.controllers.Team.suspendTeam(team)
                         } else {
-                            app.log.warn(`Subscription status for team ${team.hashid} does not match stripe "${subscription.status}" vs "${stribeSubscription?.status}". Subscription ${subscription.subscription}.`)
+                            app.log.warn(`Subscription status for team ${team.hashid} does not match stripe "${subscription.status}" vs "${stripeSubscription?.status}". Subscription ${subscription.subscription}.`)
                         }
                     }
                 } catch (err) {
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -541,13 +541,15 @@ module.exports.init = async function (app) {
             if (subscription.subscription) {
                 try {
                     const stripeSubscription = await stripe.subscriptions.retrieve(subscription.subscription)
-                    if (stripeSubscription.status !== subscription.status) {
+                    if (stripeSubscription?.status !== subscription.status) {
                         // We have got out of sync. We primarily care about being
                         // in the canceled state.
                         // Note: a canceled subscription cannot become uncanceled
                         if (stripeSubscription.status === 'canceled') {
                             await app.billing.updateSubscriptionStatus(subscription, stripeSubscription.status, team)
                             await app.db.controllers.Team.suspendTeam(team)
+                        } else {
+                            app.log.warn(`Subscription status for team ${team.hashid} does not match stripe "${subscription.status}" vs "${stribeSubscription?.status}". Subscription ${subscription.subscription}.`)
                         }
                     }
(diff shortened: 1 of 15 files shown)


forge/ee/lib/billing/index.js
```diff
@@ -531,6 +531,57 @@ module.exports.init = async function (app) {
...
+
+        /**
+         * Resynchronises a subscription with stripe. This ensures are local view
+         * of the subscription status matches stripe's view.
+         */
+        resyncTeamSubscription: async (team, subscription) => {
+            if (subscription.subscription) {
+                try {
+                    const stripeSubscription = await stripe.subscriptions.retrieve(subscription.subscription)
+                    if (stripeSubscription.status !== subscription.status) {
```
Pezmc, line 544: I think we should always log here and perhaps add a guard. e.g. if stripeSubscription is null

Pezmc, line 544:
```suggestion
                    if (stripeSubscription?.status !== subscription.status) {
```

forge/ee/lib/billing/index.js
```diff
@@ -531,6 +531,57 @@ module.exports.init = async function (app) {
...
+         * when we are detect a change from stripe.
+         *
+         * In addition to updating the field, it adds an audit log entry of the change.
+         *
+         * @param {*} subscription
+         * @param {*} newStatus
+         * @param {*} team
+         * @param {*} user
+         */
+        updateSubscriptionStatus: async function (subscription, newStatus, team, user) {
```
Pezmc, line 571: This feel like it should live in the subscription model or be tied to sequelize, but I guess we need access to the user who performed the action, so not quite as straightforward.

knolleary: I agree, however it requires access to the top level `app` object (for the audit logging functions) which models don't by default, so this felt like the next best place to put it for now.

forge/ee/lib/billing/index.js
knolleary: Guard is not strictly necessary as it throws an err if the subscription isn't found - so stripeSubscription should never be null.

forge/ee/lib/billing/index.js
```diff
@@ -531,6 +531,59 @@ module.exports.init = async function (app) {
...
+                    const stripeSubscription = await stripe.subscriptions.retrieve(subscription.subscription)
+                    if (stripeSubscription?.status !== subscription.status) {
+                        // We have got out of sync. We primarily care about being
+                        // in the canceled state.
+                        // Note: a canceled subscription cannot become uncanceled
+                        if (stripeSubscription.status === 'canceled') {
+                            await app.billing.updateSubscriptionStatus(subscription, stripeSubscription.status, team)
+                            await app.db.controllers.Team.suspendTeam(team)
+                        } else {
+                            app.log.warn(`Subscription status for team ${team.hashid} does not match stripe "${subscription.status}" vs "${stribeSubscription?.status}". Subscription ${subscription.subscription}.`)
```
knolleary, line 552:
```suggestion
                            app.log.warn(`Subscription status for team ${team.hashid} does not match stripe "${subscription.status}" vs "${stripeSubscription?.status}". Subscription ${subscription.subscription}.`)
```

Pezmc (approved):

Pezmc (approved):
Still 👍

hardillb:
@knolleary did you want to include this in the billing fix push to prod? It will need the merge conflicts fixing

knolleary:
@hardillb no, need to revisit this before it merges. For another day.

knolleary:
Going to close this PR as our billing system has moved forward since we felt this was needed.

Issue #2793, opened 2023-09-20
Scalability: TimerTasks


knolleary:
### Description

Part of #2782 

The Forge App runs some tasks on a regular schedule as part of its own house keeping. For example
 - Reporting usage telemetry
 - Managing Trial life-cycle
 - Other database housekeeping

Depending on the specific task, we may not want it running in multiple places at once. For example, the Trial Task isn't transactional - it gets the list of all subscriptions to operate on first, then does its thing. It would be possible for two instances to run in parallel, both get the same list and then both try to do the work.

#### Possible Solutions

Specific solutions will depend on the needs of the individual tasks and they will have to be considered on a case-by-case basis.

Some approaches will be:

1. Rewrite tasks to be transactional and concurrency-safe
2. Add jitter to the scheduling of the tasks so they don't run at the same time
3. Add some sort of quorum holder capability so an instance takes a primary role to run the tasks... lots of dragons to be had here.

### Epic/Story

#2782

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

knolleary:
We currently have 4 timerTasks to consider.

 - `expireTokens` - runs `@daily`, removes any tokens that have expired. No real issue this running repeatedly - just needs some jitter as `@daily` translates to `0 0 * * *` and we wouldn't want all instances running at the same time.
 - `licenseCheck` - runs `'0 2 * * *'`. Checks license expiry and emails admins if nearing expiry date and/or expired. Don't want to send multiple emails each day.
 - `licenseOverage` - runs `@weekly` (`0 0 * * 0`). Generates audit log messages if the platform is over license terms. Don't want duplicate audit log entries.
 - `telemetryMetrics` - runs `10s` after startup and at a randomly picked minute/hour each day. Not ideal to have it run multiple times, but at least with the randomly picked reporting time it isn't too bad,

To simplify this task I think we only need to address the scheduling of the `expireTokens` task.

- [ ] update `expireTokens` schedule to a randomly picked minute/hour rather than `@daily`

As for the others:
 - `licenseCheck`/`licenseOverage` - this is an edge case I don't think we need to prioritise fixing today. It would require an platform running with multiple instances *and* an expiring license. Duplicate emails/audit log entries are not ideal, but not show stopping.
 - `telemetryMetrics` - we can live with multiple reports each day

PR #3167, opened 2023-12-05
Move app.monitor to be a housekeeper task

knolleary:
Part of #2793 

## Description

The `monitor` component was added to perform the daily telemetry pings. Later, we added the `housekeeper` component as a common utility for running regularly scheduled tasks - however we didn't update the monitor to use it.

This PR does just that - moves the `monitor` to be a timerTask inside housekeeper.

This is Step Zero to solving the scalability of the TimerTasks - removing the need to solve the scalability of the monitor component as a separate thing.


hardillb (approved):

hardillb:
Review looks good, but postgress test looks to be hanging and UI tests failing for some reason?

hardillb:
@knolleary Postgresql test has failed with an error that points to the telemetry, maybe a shutdown race condition?

https://github.com/FlowFuse/flowfuse/actions/runs/7103096829/job/19336901845?pr=3167

knolleary:
Yup - will be investigating in the morning.

PR #3170, opened 2023-12-05
Extend manual billing options to non-trial teams

knolleary:
## Description

This extends the work done in #3138 so it can be applied to regular teams.

The original PR only considered the user journey of a customer who signs up, is in trial mode and wants to setup an annual billing agreement.

This PR allows us to convert a customer who has already setup monthly billing into annual billing.

In this new scenario, we will cancel their existing subscription as part of setting up the unmanaged mode of billing.

To help with the overall workflow, I've added details of the team's stripe customer/subscription ids to the admin-only tools section. This will make it much easier to correlate without having to do user/email lookups.

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -513,24 +513,44 @@ module.exports.init = async function (app) {
          * Flags the subscription as being unmanaged. This disables all interaction
          * with Stripe for this team.
          *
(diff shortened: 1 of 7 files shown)


knolleary:
UI tests failing as it assumes the particular view only has one dialog, whereas it now has two.

PR #3180, opened 2023-12-07
Flow Blueprints Selection UI

### LINKED ISSUE #2948: Update Blueprint selection UI as per designs
We've shipped the MVP of Blueprints that allows a simple selection to be made as part of the Create Instance form.

This UX will not scale beyond 3 or so blueprints - the more blueprints we add today, the worse the UX will become.

Designs were prepared in #2828 for a more scalable solution with much better UX that we should commit to as soon as we plan to go beyond a small number of Blueprints.

There did remain an open UX question on the design around the interaction between the blueprint selection screen and the CreateInstance form - that will need resolving before this can be implemented.

Addition from @MarianRaphael 
## User Story
As a FlowFuse user,
I want to easily access and view a comprehensive list of all available blueprints,
so that I can efficiently identify and select the one that best suits my specific needs without unnecessary delay.

## Design
270637154-fbd868d8-026e-4247-9f75-8448253bfd9c

### Overview Page

- Image 
- Title
- Description
- Category

### Instance creation page

- "Blank Blueprint" is still pre-selected.
- A button linked to the overview page allows the user to change the blueprint from the overview page.

## Not in Scope (potential follow-ups)

- Filter
- Category icons
- Preview function

Pezmc:
## Description

Adds a blueprint selection UI to the create instance (and application > instance) pages.

Adds admin support for custom icons, ordering and default state of a FlowBlueprint.

Screenshot 2023-12-07 at 16 58 48
Screenshot 2023-12-07 at 16 58 57
Screenshot 2023-12-07 at 17 00 01
Screenshot 2023-12-07 at 16 59 57

## Related Issue(s)

Fixes #2948 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/migrations/20231207-01-add-to-flow-template-table.js
+++ b/forge/db/migrations/20231207-01-add-to-flow-template-table.js
@@ -16,6 +16,10 @@ module.exports = {
             type: DataTypes.NUMBER,
             defaultValue: 0
         })
+        await context.addColumn('FlowTemplates', 'default', {
+            type: DataTypes.BOOLEAN,
+            defaultValue: false
+        })
     },
     down: async (context) => {
     }
--- a/frontend/src/pages/admin/index.vue
+++ b/frontend/src/pages/admin/index.vue
@@ -41,7 +41,6 @@ export default {
     },
     data () {
         return {
-            navigation: null,
             mounted: false,
             icons: {
                 chevronLeft: ChevronLeftIcon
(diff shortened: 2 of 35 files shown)


forge/db/migrations/20231207-01-add-to-flow-template-table.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable no-unused-vars */
+
+const { DataTypes, QueryInterface } = require('sequelize')
+
+module.exports = {
+    /**
+     * Add icon and order to FlowTemplate (renamed FlowBlueprints in the UI)
+     * @param {QueryInterface} context Sequelize.QueryInterface
+     */
+    up: async (context) => {
```
knolleary, line 10: As FlowTemplate is an EE model, this needs to check the table exists before attempting to modify it.

knolleary (changes requested):
I'm not keen on the `default` flag as I don't think it serves a purpose we'll want to make use of. But the work has been done.

Only required change is an EE guard in the migration to check the table exists before trying to modify it

forge/db/migrations/20231207-01-add-to-flow-template-table.js
Pezmc: [787605b](https://github.com/FlowFuse/flowfuse/pull/3180/commits/787605b85a264e50b27fc591135ad1a998ec2370)

knolleary:
> It was that or hardcode that we always look for a blueprint called "Blank" or something similar

Fair enough. It will mean that once deployed, *every* instance will get created with a blueprint applied - it may just be the blank one. Will need to make sure our stats tracking the usage of blueprints takes that into account.

knolleary:
> Any thoughts how I can check

We already have backend tests that validate if a blueprint id is passed to the api that it gets applied:

https://github.com/FlowFuse/flowfuse/blob/fbca84576fb2bd27e07d626a94d993b148b67d95/test/unit/forge/routes/api/project_spec.js#L978

I would suggest for the front end test that it is sufficient to validate it makes the right api call with the expected `blueprintId` set in the payload.

knolleary:
When I run this branch, I get the following console.log at startup. I'm suspecting it's related to the schema changes - this message appears to come from `ajv` which is used to compile the schemas inside fastify. I'll see if I can track it down further.
```
[2023-12-13T11:52:00.716Z] INFO: Loading EE Features
strict mode: missing type "object" for keyword "required" at "#" (strictTypes)
```

The second issue I have is the fact this now relies on admin creating a Blank blueprint - rather than it being implicitly added to the list like we currently do. As soon as we deploy this code to production, user's will loose the ability to create a blank instance until an admin goes in and creates the Blank blueprint. That isn't ideal.

We either accept that is necessary - and be *very* careful about when this is merged and deployed to production to ensure the change happens, or we add some code to create a default Blank blueprint as part of the migration. I'll look at which makes sense to do.

joepavitt:
> The second issue I have is the fact this now relies on admin creating a Blank blueprint - rather than it being implicitly added to the list like we currently do. As soon as we deploy this code to production, user's will loose the ability to create a blank instance until an admin goes in and creates the Blank blueprint. That isn't ideal.

It was my assumption that if no blueprint id was provided in the API, then a blank Node-RED is created anyway? If no blueprints are found at the moment, then no blueprints are presented, and the user can create the Instance as-is.

knolleary:
> It was my assumption that if no blueprint id was provided in the API, then a blank Node-RED is created anyway? If no blueprints are found at the moment, then no blueprints are presented, and the user can create the Instance as-is.

In FlowFuse Cloud production we already have blueprints. When we deploy this change, it will offer up those blueprints and no longer offer the hardcoded 'blank' option until we add the blank blueprint.

joepavitt:
> When we deploy this change, it will offer up those blueprints and no longer offer the hardcoded 'blank' option until we add the blank blueprint.

Apologies, I mis-read, thought you were referring to self-hosters. Yes, when we deploy to production, someone will need to go in and create the Blueprint. Although, we will also need to add icon, etc. to existing ones anyway, so this is a necessary step anyway.

We could also create the "Blank" blueprint _now_, and just put it as not-active. Then it's just a two-click to make it active (and default)

knolleary:
> We could also create the "Blank" blueprint now, and just put it as not-active. Then it's just a two-click to make it active (and default)

Yup - filing that under the heading of 'need for careful co-ordination' to get this feature live :)

knolleary:
I've found the cause of the `strict mode: missing type "object" for keyword "required" at "#" (strictTypes)` warning. It is the schema of the create blueprint endpoint:

```
            summary: 'Create a flow blueprint - admin-only',
            tags: ['Flow Blueprints'],
            body: {
                $ref: 'FlowBlueprint',
                required: ['name']
            },
```

It seemingly doesn't like the `required` property alongside `$ref`. Am looking to see if that's supported syntax or if a different approach is needed.

knolleary:
Have pushed the fix for the schema issue.

knolleary:
What are the rules for category icon?

I'm getting mixed results - `users` works for example, but `tv` and `bug-ant` don't.

joepavitt:
I understood it to be any Heroicon (outline)

joepavitt:
Founded check with version of heroicons we are using, think are a major version behind, and there was a lot of renaming

knolleary:
We are indeed on Heroicons 1.x - I'll update the inline link in the edit template to `https://v1.heroicons.com/`

frontend/src/pages/admin/index.vue
```diff
@@ -33,43 +33,40 @@ import { mapState } from 'vuex'
...
             }
         }
     },
     computed: {
-        ...mapState('account', ['user', 'team'])
+        ...mapState('account', ['features', 'user', 'team'])
     },
     mounted () {
         this.mounted = true
+        const navigation = [
```
Pezmc, line 56: This feels like it should be a computed property

knolleary (approved):

Issue #2559, opened 2023-08-03
Can not delete a trial account


iskerrett:
### Current Behavior

If I open a trial account, then I decide I am finish my trail and want to delete my account. I go to User Settings and press Delete Account. I get the error message 'Can not delete the last owner of a team'.  However, I didn't invite any team members so it is even possible to delete this account?

### Expected Behavior

I would expect we want to allow people to delete their accounts after completing their trial.

### Environment

- FlowForge version: 1.10

## Feature request and Scope (Marian)

Implement a more intuitive account deletion process. Specifically, when a user who is the last member and owner of a team attempts to delete their account, a pop-up should automatically appear. This pop-up will prompt the user with the option to also delete their team(s) where they are the sole remaining owner. This change will replace the current error message ('Cannot delete') with a proactive solution, simplifying the account deletion process for individual users and ensuring a smoother user experience.

### Have you provided an initial effort estimate for this issue?

I am not a FlowForge team member

hardillb:
@MarianRaphael can we schedule something around this. While we want to keep users it shouldn't be because they can't work out how to leave.

MarianRaphael:
FYI - I added a Scope

hardillb:
Yep, doing it steps, but will look at that bit next

hardillb:
Verified deleting teams with Applications, Instances and devices on staging.

Nick to test deleting a team with expired team (but should be no different)

knolleary:
Spotted an issue trying to delete expired team. Investigating.

knolleary:
I can successfully delete a team with an expired trial locally. However there is a team on staging that appears to be in that state, but the delete fails.

Need to poke into the database to understand how/why it is failing.

hardillb:
@knolleary did you get to the bottom of this?

knolleary:
I think this was a result of Stripe cancelling the subscription after 3 months (as it does by default on the test account). That got things in to a weird state. Happy to close this off.

PR #3187, opened 2023-12-08
Allow expired billing teams to be deleted

hardillb:
part of #2559

## Description

Allow a team with expired billing e.g. at end of trial to be deleted even if it has instances/applications.

## Related Issue(s)

#2559 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
Need to work out some tests for this

hardillb:
Change to Billing.vue is just wrapping the whole thing in a `<div>` to clear VSCode lint warning

hardillb:
The failing testcase is not something I've touched.

Wonder if it's a race condition?

forge/routes/api/team.js
```diff
@@ -485,6 +485,40 @@ module.exports = async function (app) {
         // call to destroy the team will throw an error
         try {
             if (app.license.active() && app.billing) {
+                // It is possible the team has Instances but expired billing so
```
knolleary, line 488: We need to delete devices as well.

hardillb: Yep, will add that

forge/routes/api/team.js
```diff
@@ -484,12 +484,54 @@ module.exports = async function (app) {
...
+                for (const instance of instances) {
+                    try {
+                        await app.containers.remove(instance)
+                    } catch (err) {
+                        if (err?.statusCode !== 404) {
+                            throw err
+                        }
+                    }
+
+                    // disconnect any devices from the Instance
```
knolleary, line 500: Given we're about to delete the devices, I don't think we should send them all a command that will cause them to call home and change their state.

knolleary (changes requested):
One more tweak needed - no need to send devices a command if we're about to delete them anyway.

knolleary (approved):

Issue #2997, opened 2023-10-24
Implementing Device Groups Management


MarianRaphael:
### User Story
As a FlowForge user, who manages a large number of devices,
I want to have the ability to create "Device Groups",
So that I can manage and organize my devices more efficiently by grouping and identifying them through common characteristics.

Follow-up story:
https://github.com/FlowFuse/flowfuse/issues/2989

### Scope

- Users can create a new "Device Groups" by providing a unique name 
- Users can view a list of all created "Device Groupss" with their names
- Devices can only be assigned to one Device Group at any time.
- Groups (Tag) are managed at the application level, but viewable/configurable in the pipeline
- Users can view or remove the "Device Group" for a device

### Design
https://github.com/FlowFuse/flowfuse/issues/3103

## License 
~~Teams Tier~~
Enterprise Tier

```[tasklist]
### Tasks
- [ ] https://github.com/FlowFuse/flowfuse/issues/3103
```

```[tasklist]
### Follow-ups
- [ ] https://github.com/FlowFuse/flowfuse/issues/3204
```

MarianRaphael:
Follow-up: https://github.com/FlowFuse/flowfuse/issues/2989

MarianRaphael:
Designs in: https://github.com/FlowFuse/flowfuse/issues/3103

Steve-Mcl:
@MarianRaphael et al, FYI,

After resurrecting the device tagging branch and working through it with the finalised direction of Device Groups (as discussed in the 2 previous meetings and the update to this OP) I have proposed a new, simplified DB schema for introducing Device Groups.  IMO, device groups were moving a sufficient distance away from the mechanics of tagging (of being able to set a devices properties/env vars etc) into its own entity. 

I proposed this to Nick and we are in agreement: "simplifying Device Groups to a standalone, one-to many table joined to the device" (see attached schema)

I feel I can deliver this a lot quicker and believe this will not interfere with the (later) concept of tagging for settings / env vars / logical grouping etc.

![image](https://github.com/FlowFuse/flowfuse/assets/44235289/4d4f51e4-1be6-434b-8482-d5a60e995c80)

MarianRaphael:
Device Groups should be a "Teams Tier"(ee) feature

Steve-Mcl:
@MarianRaphael 

Audit log entries for Device Groups was not in the scope of this work but I _assume_ we/customers will want this?

If yes, can we add / schedule this work?

Thanks.

[...]

Issue #3103, opened 2023-11-22
Design rework for Device Groups


MarianRaphael:
## Original Design

https://www.figma.com/file/rMB2w5ZZUtTC0nxbwLYKJY/FlowForge-Platform?type=design&node-id=2528-18633&mode=design&t=UtJNcyM8PnopNmc5-0

## Task

Based on our discussion, there is a need for a new design scope around the Device Group concept in #2997

joepavitt:
Full set of screens here: https://www.figma.com/file/rMB2w5ZZUtTC0nxbwLYKJY/FlowForge-Platform?type=design&node-id=3636%3A12502&mode=design&t=ti6Xpx7QyoRLfh3O-1

joepavitt:
Golden Path Prototype for device groups: https://www.figma.com/proto/rMB2w5ZZUtTC0nxbwLYKJY/FlowForge-Platform?page-id=3636%3A12502&type=design&node-id=3641-24919&viewport=2746%2C968%2C0.43&t=aiNRiHeoodkgCFir-1&scaling=min-zoom&starting-point-node-id=3641%3A24919&mode=design

joepavitt:
## Highlights

### Team > Devices > Device Groups

![Image](https://github.com/FlowFuse/flowfuse/assets/99246719/d00181ee-afb3-4586-a428-971acf6174d9)

### Device Group > Settings

![Image](https://github.com/FlowFuse/flowfuse/assets/99246719/03b0afce-5d2b-4d9a-8f3b-ec26563ba9fb)

MarianRaphael:
Thanks for the designs. Looks good!

joepavitt:
To scope this back, we can for now, not bother with the Team-level screens, Device Groups 9as first iteration) can be Application-level configured only.

In a meeting between @MarianRaphael, @Steve-Mcl and myself we've agreed with this scope as first iteration. There was discussion around moving everything to live within the scope of DevOps pipelines only, but this would be just as much development effort, without a single view of Device Groups which would hinder UX.

Steve-Mcl:
@joepavitt could do with a bit of a leg up here. 

The screenshots all _seem_ to show management at team level (the `Devices` menu level is highlighted in all)

I feel there is a screenshot design missing on how the Application shows its device groups?

joepavitt:
this was the missing screenshot:

[...]

PR #3192, opened 2023-12-08
2997 device group frontend

Steve-Mcl:
## Description

Adds frontend management for new feature Device Groups.
image

Device Groups are accessible under the an Application page.  To edit a Device Group, click the entry in the list:
chrome_gXFqh1KJy2

NOTES:
* Only Application Assigned Devices are available to Device Groups
* A Device can only belong to one device group
   * This means you will see already grouped devices in the "Device Groups"->"Devices" page under the "Available Devices" list
* Deleting a Device Group will remove devices from a group
* Removing / Adding devices in/out of a group are deliberatly front-end changes only until the "Save Changes" button is operated
   * This is to avoid potential inconsistent state (i.e. User1 is setting up group, User2 causes pipeline deploy before User1 is complete)

## Related Issue(s)

#2997 

### TODO:

[ ] docs

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


knolleary:
Some feedback on the UX side of this.

- I think the default view for a group should be the list of devices in the group rather than its settings tab. That feels like something a user will want to interact with more frequently
- I find the Device Group members view confusing. The first thing I see is list of devices *not* in the group. It doesn't give you the information you want at a glance. Could we add a 'add devices' button in the header and only show the 'available devices' list when clicked?

joepavitt:
> I think the default view for a group should be the list of devices in the group rather than its settings tab. That feels like something a user will want to interact with more frequently

This is the way in the designs.

> I find the Device Group members view confusing. The first thing I see is list of devices not in the group. It doesn't give you the information you want at a glance. Could we add a 'add devices' button in the header and only show the 'available devices' list when clicked?

Steve and I had discussed exactly the same. An "Edit Group Devices" button rather than "Add Devices", but yes, agreed.

Steve-Mcl:
> > I think the default view for a group should be the list of devices in the group rather than its settings tab. That feels like something a user will want to interact with more frequently
> 
> This is the way in the designs.

@joepavitt can you clarify please. Are you saying that is how it is supposed to be (as per Nicks suggestion) OR that it is as per the design (and _dont_ want it to change)?

For reference: see 2nd image [in designs](https://github.com/FlowFuse/flowfuse/issues/3103#issuecomment-1824334250)

The implementation is the same (in layout) but perhaps I missed a subtle detail (i.e. the Devices Tab is the default?)

<br>

From my perspective, I am happy to set "Devices" tab as the default but need clarification on this:
* Default to 2nd tab ("Devices")
* OR
* Move "Devices" to be the 1st tab (and be the default)

joepavitt:
> can you clarify please. Are you saying that is how it is supposed to be (as per Nicks suggestion) OR that it is as per the design (and dont want it to change)?

Agree with Nick, in the prototype, when you click a Device Group in Application/Device Groups you go the the "All Devices" screen.

> but need clarification on this:

Very good point, "Devices" being the first tab (and default) makes way more sense.

Steve-Mcl:
> > I find the Device Group members view confusing. The first thing I see is list of devices not in the group. It doesn't give you the information you want at a glance. Could we add a 'add devices' button in the header and only show the 'available devices' list when clicked?
> 
> Steve and I had discussed exactly the same. An "Edit Group Devices" button rather than "Add Devices", but yes, agreed.

As Joe says, we did and I even wrote it up in an issue comment. But upon development of the original design (with help from Joe) i got the UI (almost) as prescribed in the design so did not end up doing this.

As I am now 3 deep in nested branches (device groups PAI -> Device groups Front end -> Pipeline deploy Device Group) working this is (for me) a pickle since I need to rebase etc. 

I will finish works on Deploy to Device Group, get it in a branch, then back track to this and the comments in the API (then rebase, rebase, rebase) - unless someone has a better idea :)

Steve-Mcl:
> I find the Device Group members view confusing. The first thing I see is list of devices not in the group. It doesn't give you the information you want at a glance. Could we add a 'add devices' button in the header and only show the 'available devices' list when clicked?

Implemented in 
[Only show available devices in "edit mode"](https://github.com/FlowFuse/flowfuse/pull/3192/commits/eb5c684ea705245080fbba91b3aea091010b3c0c) [eb5c684](https://github.com/FlowFuse/flowfuse/pull/3192/commits/eb5c684ea705245080fbba91b3aea091010b3c0c)

chrome_vhKaLIiURR


_just noticed the checkbox still visible in read mode_ 🤦

knolleary (approved):
Approved pending the tests passing following PR merges to this branch

PR #3195, opened 2023-12-11
Enforce password quality rules

hardillb:
## Description

Enforce minimum password quality rules for all users

## Related Issue(s)

FlowFuse/security#36

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -136,7 +136,7 @@ module.exports = function (env, argv) {
                         priority: -10,
                         chunks: 'initial'
                     },
-                    zxcvbn: {
+                    async: {
                         test: /[\\/]node_modules[\\/]/,
                         name: 'async-vendor',
                         priority: -10,
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -138,7 +138,7 @@ module.exports = function (env, argv) {
                     },
                     async: {
                         test: /[\\/]node_modules[\\/]/,
-                        name: 'async-vendor',
+                        name: 'async-vendors',
                         priority: -10,
                         chunks: 'async'
                     }
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -23,7 +23,8 @@ module.exports = function (env, argv) {
         output: {
             path: getPath('frontend/dist/app'),
             publicPath: '/app/',
-            assetModuleFilename: './assets/[hash][ext][query]'
+            assetModuleFilename: './assets/[hash][ext][query]',
+            filename: '[name].js'
         },
         module: {
             rules: [
@@ -133,7 +134,13 @@ module.exports = function (env, argv) {
                         test: /[\\/]node_modules[\\/]/,
                         name: 'vendors',
                         priority: -10,
-                        chunks: 'all'
+                        chunks: 'initial'
+                    },
+                    zxcvbn: {
+                        test: /[\\/]node_modules[\\/]/,
+                        name: 'async-vendor',
+                        priority: -10,
+                        chunks: 'async'
                     }
                 }
             }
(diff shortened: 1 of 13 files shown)


hardillb:
Note, this adds a dependency to the page load javascript that 400kb gzip compressed

hardillb:
@joepavitt if you get 2 mins can you review the webpack changes and see if you can some up with a better async load in the vue.

Thanks

frontend/src/pages/account/PasswordReset.vue
```diff
@@ -45,6 +47,17 @@ export default {
...
     computed: mapState('account', ['settings', 'pending']),
+    watch: {
+        'input.password': function (v) {
+            if (this.errors.password && v.length >= 8 && zxcvbn(v).score >= 2) {
+                this.errors.password = ''
+            }
+        }
+    },
+    async mounted () {
+        const { default: zxcvbnImp } = await import('zxcvbn')
```
joepavitt, line 58: why `import` in here? Why not import at the top level with others?

hardillb: Because it has to be loaded async (it's huge so need to only be loaded on demand) so I assumed it needs to be in an async function

config/webpack.config.js
```diff
@@ -133,7 +134,13 @@ module.exports = function (env, argv) {
                         test: /[\\/]node_modules[\\/]/,
                         name: 'vendors',
                         priority: -10,
-                        chunks: 'all'
+                        chunks: 'initial'
+                    },
+                    async: {
+                        test: /[\\/]node_modules[\\/]/,
+                        name: 'async-vendor',
```
joepavitt, line 141:
```suggestion
                        name: 'async-vendors',
```

frontend/src/pages/account/Security/ChangePassword.vue
```diff
@@ -41,6 +43,17 @@ export default {
...
         }
     },
+    watch: {
+        'input.password': function (v) {
+            if (this.errors.password && v.length >= 8 && this.zxcvbn(v).score >= 2) {
+                this.errors.password = ''
+            }
+        }
+    },
+    async mounted () {
```
joepavitt, line 53: could do this inside the `beforeCreate` - but may slow down initial load?

joepavitt (approved):
Spoke with Ben on a Slack Huddle - functional, and can't think of a better way to do it - happy to proceed.

knolleary (changes requested):
This needs to update the sign-up page as well - `frontend/src/pages/account/Create.vue`

knolleary:
Not seeing it get applied on the Account Create page. I can see you've added logic to `checkPassword` - but that function is unused on the page.

Also `formValid` needs to be updated to ensure the button is disabled if the password isn't good enough.

hardillb:
Sorry, missed that earlier context switching.

Should be right now

hardillb:
Just got to fix the UI tests now

frontend/src/pages/account/Create.vue
```diff
@@ -143,6 +146,14 @@ export default {
...
             if (this.errors.password && v.length >= 8) {
                 this.errors.password = ''
+            } else {
+                this.errors.password = 'Password needs to be longer than 8 chars'
+                return
+            }
+            if (zxcvbn(v).score >= 2) {
+                this.errors.password = ''
+            } else {
+                this.errors.password = 'Password needs to be more complex'
```
knolleary, line 156: Don't think the logic is right here.

As I type I'm seeing the 'longer than 8 chars' message toggle on and off with each character.

hardillb: Should be fixed now

knolleary (approved):

PR #3214, opened 2023-12-14
Ensure editor sessions are revoked when logging out of the platform

knolleary:
## Description

This fixes the handling of logging out of Node-RED sessions when a user logs out of the platform.

It reduces code duplication by moving it into a new controller for the StorageSession object.

Adds unit test for the logic.

Adds audit-log details in the frontend for the event.

### DIFF
--- a/test/unit/forge/db/controllers/StorageSession_spec.js
+++ b/test/unit/forge/db/controllers/StorageSession_spec.js
@@ -1,0 +1,73 @@
+const should = require('should') // eslint-disable-line
+const sinon = require('sinon')
+const setup = require('../setup')
+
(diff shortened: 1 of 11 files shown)


hardillb (approved):

test/unit/forge/db/controllers/StorageSession_spec.js
```diff
@@ -0,0 +1,73 @@
+const should = require('should') // eslint-disable-line
+const sinon = require('sinon')
```
hardillb, line 2:
```suggestion
const sinon = require('sinon')

```

PR #3217, opened 2023-12-15
Send alerts on Instance Crash

### LINKED ISSUE #2518: Send email alerts to users when their instances crash or restart
### Epic

_No response_

### Description

**As a:**

User of FlowForge

**I want to:**

Get alerts (probably by email) when there are issues with my instance which will stop my flows from running.

**So that:**

If an instance crashes, or ends up in a crash loop I can take steps to resolve the problem.

Licensed Edition (EE)
Teams Tier

### Acceptance Criteria

_No response_

I can not provide an initial effort estimate

## Scope (Marian edit)

- [ ] Email Notification Setup: Introduce the capability to add an email address under the Instance Settings General tab. This will serve as the destination for alerts.

- [ ] Notification Preferences: Implement a drop-down menu in the Instance Settings to allow users to subscribe to notifications for specific Audit log events. Users should be able to select multiple events for which they wish to receive notifications. The options in this drop-down menu should include:

- Instance Crash
- Instance Restart
- Node-RED Stopped
- Instance suspended 

- [ ] Alert Behavior: Upon the occurrence of any selected event, the system will automatically send a standard email notification to the specified email address, informing the user of the event.
- [ ] Owner privileges required

hardillb:
fixes #2518

## Description

Allows Team owners to enable emails to be sent when an instance crashes or is put into safe mode due to crashing multiple times.

They can also decide if Team Owners, Members or both receive these emails.

## Related Issue(s)

#2518

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
Will try and work out how to test this

knolleary (changes requested):
`s/Project/Instance/g` in the emails

My only slight pause here is whether we've checked what patterns of events we can get and what that'll mean for the emails. For example, if an application is in a crash loop - are we going to fill their inbox.

frontend/src/pages/admin/TeamTypes/dialogs/TeamTypeEditDialog.vue
```diff
@@ -163,11 +164,14 @@ export default {
                     if (this.input.properties.features.customCatalogs === undefined) {
                         this.input.properties.features.customCatalogs = true
                     }
+                    if (this.input.properties.features.emailAlerts === undefined) {
+                        this.input.properties.features.emailAlerts = true
```
knolleary, line 168:
```suggestion
                        this.input.properties.features.emailAlerts = false
```

Default should be disabled until they are added to a TeamType.

forge/ee/lib/alerts/index.js
```diff
@@ -0,0 +1,54 @@
...
+                    } else if (emailAlerts?.safe && event === 'safe-mode') {
+                        template = 'SafeMode'
+                    }
+                    if (!template) {
+                        return
+                    }
+                    const team = await app.db.models.Team.byId(project.Team.id)
+                    const where = {
+                        TeamId: team.id
+                    }
```
knolleary, line 27: At first glance, this appears to be an unneeded lookup of `Team`. I would expect `project.Team.id` to be the same as `team.id` - which is all it's using `team` for.

That said, given this feature is flagged against the teamType as well, I think this needs something like this (please test it first!)
```
if (!team.TeamType.getFeatureProperty('emailAlerts', false)) {
  return
}
```

knolleary (changes requested):

forge/ee/lib/alerts/index.js
hardillb: I've pushed an update, had to swap the team lookup for a TeamType lookup, but should cover all angles (e.g. when team type is changed to one without the option enabled)

Issue #2989, opened 2023-10-23
Possibility to assign Devices Groups to a DevOps Pipeline


MarianRaphael:
## User Story

As a FlowFuse user,
I want the ability to push a Snapshot through a DevOps Pipeline to a groups of devices,
So that I can efficiently manage and update multiple devices simultaneously 

```[tasklist]
### Requirements
- [ ] https://github.com/FlowFuse/flowfuse/issues/2902
- [ ] https://github.com/FlowFuse/flowfuse/issues/2756
- [ ] https://github.com/FlowFuse/flowfuse/issues/3183
```

MarianRaphael:
Next iteration: https://github.com/FlowFuse/flowfuse/issues/3142

Steve-Mcl:
@MarianRaphael, @joepavitt , @Yndira-FlowForge 


Questions on design specifically around "Device Groups" that will soon block me - looking for direction please?

![Image](https://github.com/FlowFuse/flowfuse/assets/44235289/37097823-8e0b-4b73-a3e1-4758e4cb9b72)

1. Device Group colour?
2. What will we display in the stage?
    * Considering there could be a large about of devices, perhaps simple stats like:
       *  `Device Count` pill 
       * `Targeted Count` pill (the amount of devices with target snapshot set to that of the last pipeline deploy)
       * `Active Count` (the number of devices in the group that have the Active snapshot set to that of the last pipeline deploy)
3. Should we have a stage AFTER a Device Group stage?  Does it make sense?
4. If yes to above ^ then what are the deployment actions and can this me a later iteration or is it MVP?

joepavitt:
> 1. Device Group colour?

I had assumed the same at the "Device Icon". The blue we use as a link, purple is for the dev mode. Green/Yellow have traffic light connotations, so I'd say keep the same. Happy to be overruled by @Yndira-FlowForge though.

> 2. What will we display in the stage?

I'd agree with your ideas here. Although a user is then inevitably going to want to see which devices haven't been deployed to yet, although I guess that's accessible by going to the Device Group page anyway?

> 3. Should we have a stage AFTER a Device Group stage? Does it make sense?

Good question as it's not clear what the selected action could do. There are no Device Group scoped Snapshots or anything, so I'd say, for now, nothing can go after a Device Group in a Pipeline. Thoughts @MarianRaphael?

MarianRaphael:
> Should we have a stage AFTER a Device Group stage? Does it make sense?

[...]

PR #3229, opened 2023-12-18
Assign device group to pipeline stage

Steve-Mcl:
### General

Add support for setting a Device Group as the target of a pipeline execution

### Adds through table `PipelineStageDeviceGroups` 
image

### Add unit tests
```
Pipelines API
  Create Pipeline Stage
    With device group
      ✔ Should create a pipeline stage
      Validates that the first stage cannot be a device group
        ✔ Rejects a pipeline stage if the device group is applied to the 1st stage
      Validates that a stage cannot be added after device group
        ✔ Rejects a pipeline stage if the device group is added where a device group already exists
    With both device and device group
      ✔ Rejects the request gracefully
    With both instance and device group
      ✔ Rejects the request gracefully
    With neither device, instance or device group
      ✔ Fails gracefully with a clear error
  Get Pipeline Stage
    ✔ Should return a single pipeline stage with a device group
  Update Pipeline Stage
    With a new name
      ✔ Should update a single pipeline stage with a new name (44ms)
    With a new device group
      ✔ Should unassign the old device group and assign the new one (62ms)
      ✔ Should require the device group to be part of the same application
      ✔ Should not be allowed to replace 1st stage
      ✔ Should not be allowed set a center stage as a device group (86ms)
  Delete Pipeline Stage
    ✔ should destroy the pipeline stage, but not touch the assigned device group
  Deploy Pipeline Stage
    For instance=>device group
      Copies the existing instance snapshot to the next stages device group
  List Application Pipelines
    ✔ should list all the pipelines within an application including stages, instances, devices and device groups
```

### add e2e tests
```
FlowForge - Application - DevOps Pipelines
    ✔ cannot select device group in the first stage of a pipeline
    ✔ cannot add any more stages after a device group
```

## Related Issue(s)

#2989 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [x] Concepts

## Labels


Pezmc:
I've done a first pass here to get an overall idea, but need to go through again tomorrow with fresh eyes

Pezmc (approved):

Pezmc:
I've been through this again and nothing stood out, I'm still not feeling 100%, perhaps just because of the number of changes, but let's not block this any longer.

PR #3235, opened 2023-12-19
Add device group feature flag in TeamType

knolleary:
This adds a `deviceGroup` feature flag to TeamType so the feature can be limited to certain team types.

As part of this, I found that if you open an application in a team that isn't your default team directly (ie paste the url in, or Ctrl-R on the page), we would load the user's default team first, then load the team of the application you were accessing.

This would cause the view to first load data associated with your default team, before refreshing with data for the team you were accessing - in some cases, it wouldn't refresh because there was no watch happening.

To fix this, I've modified the account load phase to skip loading the user's team if the url is on either `/application`, `/device` or `/instance` - as all of these views do a lookup of the team and call `setTeam` anyway. This avoids the double load.

However, I then hit the problem with the order the `application` page does things. It loads the application information, devices, deviceGroups etc, *then* it ensures the loaded team is the one for the application. This leads to a chicken-vs-egg situation as we now need to check the team's feature flags to decide whether to load the device groups or not.... but we may not have loaded the team yet.

~As a workaround, I've put a try/catch around the loading of deviceGroups, which will 404 if the team isn't allowed to use device groups (we at least have the platform-level guard in there). I then pass a flag down through all of the Pipeline components to say if deviceGroups is enabled or not. I dislike this model - I find it much easier to reason over if a component uses `mapState` to access the team locally, rather than all of this passing down of properties. I'm sure there is a Vue reason it's being done the way it is...~

Updated via https://github.com/FlowFuse/flowfuse/pull/3235/commits/2ced020281e69ad692057c68a9e7128d65c5b00d - I've modified the order the Application loads things, having realised we can await the loading of the application team (if needed). This ensure we have the Team object so can check the device group feature flag. As a consequence, I've also removed the `deviceGroupEnabled` property I had added - cleaner (IMHO) to check the team directly from the store (ie via `mapState...`).

### DIFF
--- a/frontend/src/pages/application/DeviceGroups.vue
+++ b/frontend/src/pages/application/DeviceGroups.vue
@@ -23,7 +23,7 @@
             </template>
         </ff-data-table>
     </div>
-    <EmptyState v-else :featureUnavailable="!featureEnabled" :featureUnavailableMessage="'Device Groups are an enterprise feature'">
+    <EmptyState v-else :featureUnavailable="!featureEnabledForPlatform" :featureUnavailableToTeam="!featureEnabledForTeam" :featureUnavailableMessage="'Device Groups are an enterprise feature'">
         <template #header>Add your Application's First Device Group</template>
         <template #img>
             <img src="../../images/empty-states/application-device-groups.png">
@@ -92,6 +92,10 @@ export default {
         application: {
             type: Object,
             required: true
+        },
+        team: {
+            type: Object,
+            required: true
         }
     },
     data () {
@@ -126,8 +130,14 @@ export default {
     },
     computed: {
         ...mapState('account', ['features']),
+        featureEnabledForTeam () {
+            return !!this.team.type.properties.features?.deviceGroups
+        },
+        featureEnabledForPlatform () {
+            return this.features.deviceGroups
+        },
         featureEnabled () {
-            return this.features['device-groups']
+            return this.featureEnabledForTeam && this.featureEnabledForPlatform
         }
     },
     mounted () {
--- a/frontend/src/pages/application/DeviceGroups.vue
+++ b/frontend/src/pages/application/DeviceGroups.vue
@@ -12,8 +12,11 @@
             <p>The device groups can then be set as the target in a DevOps Pipeline to update multiple devices in a single operation</p>
         </template>
     </SectionTopMenu>
-
-    <div v-if="deviceGroups?.length > 0" class="pt-4 space-y-6" data-el="pipelines-list">
+    <ff-loading
+        v-if="loading"
+        message="Loading Device Groups..."
+    />
+    <div v-else-if="deviceGroups?.length > 0" class="pt-4 space-y-6" data-el="pipelines-list">
         <ff-data-table v-model:search="tableSearch" :columns="tableColumns" :rows="deviceGroups" :show-search="true" search-placeholder="Filter..." :rows-selectable="true" @row-selected="editDeviceGroup">
             <template #actions>
                 <ff-button data-action="create-device-group" :disabled="!featureEnabled" @click="showCreateDeviceGroupDialog">
@@ -100,6 +103,7 @@ export default {
     },
     data () {
         return {
+            loading: false,
             deviceGroups: [],
             input: {
                 name: '',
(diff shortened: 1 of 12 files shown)


frontend/src/pages/application/index.vue
```diff
@@ -174,13 +176,22 @@ export default {
...
                 this.applicationDevices = deviceData?.devices
                 const applicationInstances = await instancesPromise
-
-                if (this.features?.['device-groups']) {
-                    const deviceGroupsPromise = ApplicationApi.getDeviceGroups(applicationId)
-                    const deviceGroupsData = await deviceGroupsPromise
-                    this.deviceGroups = deviceGroupsData?.groups
+                if (this.features?.deviceGroups) {
+                    // DeviceGroups could be disabled for this team. However at this point
+                    // in time we may not have looked up the team object yet, so we cannot be
```
Pezmc, line 181: Where is the first place the team object gets looked up? I'm also not particularly fond of this pattern...

frontend/src/store/account.js
```diff
@@ -189,13 +189,14 @@ const actions = {
...
             }
-
             try {
-                const team = await teamApi.getTeam(teamId || { slug: teamSlug })
-                const teamMembership = await teamApi.getTeamUserMembership(team.id)
-                state.commit('setTeam', team)
-                state.commit('setTeamMembership', teamMembership)
-
+                // Some routes will call `setTeam` once they figure out who they belong to.
+                if (!/^\/(application|device|instance)\//.test(router.currentRoute.value.path)) {
```
Pezmc, line 194: I think this needs to be more explicit about what this doing and why. Right now (without the context of this PR) it reads not applications device instance in current path, which is kind of unclear as to what the objective is...

Steve-Mcl (approved):
Other than Pez's comments for clarification, happy to merge.

frontend/src/store/account.js
knolleary: I've added a longer comment to make it clearer. 
https://github.com/FlowFuse/flowfuse/pull/3235/commits/d524e15f983f600f70c1e27d8c62eed5a6344dcf

frontend/src/pages/application/index.vue
knolleary: Ultimately it comes down to needing to check `team.type.properties.features.deviceGroups` before deciding whether to do *any* device group related work.

I'm not clear why we're loading deviceGroups at such a high-level of the application stack. The DeviceGroups view reloads them regardless. It would seem cleaner to have the Pipeline views decide for itself what it needs to load - and only if necessary.

Steve-Mcl: > I'm not clear why we're loading deviceGroups at such a high-level of the application stack. The DeviceGroups view reloads them regardless. It would seem cleaner to have the Pipeline views decide for itself what it needs to load - and only if necessary.

I dont disagree however this was done to follow suit and not differ from the existing placement of code for all of the instances and devices that are loaded in the same way, same place (with regards to pipelines)

knolleary: Via https://github.com/FlowFuse/flowfuse/pull/3235/commits/2ced020281e69ad692057c68a9e7128d65c5b00d I've reworked the order the application loads things, having discovered we can `await` the setting of the team object before anything else is done. This is much cleaner.

knolleary:
@Steve-Mcl can you have a look at the last set of commits?

 - Added a Loading placeholder on the Device Group view (given the async nature of loading the groups, it was flashing up the empty state before showing the actual list).
 - Reworked the order the application loads things so we will have the `team` object available to check if the feature flag is set.

frontend/src/pages/application/DeviceGroups.vue
```diff
@@ -12,8 +12,11 @@
             <p>The device groups can then be set as the target in a DevOps Pipeline to update multiple devices in a single operation</p>
         </template>
     </SectionTopMenu>
-
-    <div v-if="deviceGroups?.length > 0" class="pt-4 space-y-6" data-el="pipelines-list">
+    <ff-loading
+        v-if="loading"
+        message="Loading Device Groups..."
+    />
+    <div v-else-if="deviceGroups?.length > 0" class="pt-4 space-y-6" data-el="pipelines-list">
```
Steve-Mcl, line 19: Instead of show/hide whole portions of the page, perhaps we should simply utelise the features of the data table like we do for "Loading Teams" and "Loading Users"?


https://flowfuse.github.io/forge-ui-components/#ff-data-table

> Property | Default | Description
> -- | -- | --
> `loading` | false | Shows a placeholder Loading message in place of the data.
> `loading-message` | 'Loading Data...' | If loading=true, then this message will be shown.
>

Pezmc (approved):
Much prefer the new approach!

Issue #3246, opened 2023-12-20
Device Groups: First pipeline stage hides button


Pezmc:
### Description

A device group can't be the first stage in a pipeline, however instead of showing this to the user, we hide the device group button entirely when creating your first stage.

It'd be much clearer if this button was visible but disabled with a tooltip explaining why, similar to the deploy to devices option. /cc @joepavitt 

### First stage

![Screenshot 2023-12-20 at 15 27 05](https://github.com/FlowFuse/flowfuse/assets/507155/ba4fae04-42d1-455a-8f9f-1a9431e88e86)

### Second stage

![Screenshot 2023-12-20 at 15 27 40](https://github.com/FlowFuse/flowfuse/assets/507155/01f30007-c334-42ef-8be0-4ff7916eee59)

### Epic/Story

#3157
https://github.com/FlowFuse/flowfuse/issues/2997

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Agreed

PR #3249, opened 2023-12-20
Device Groups: Disable rather than hide Device Group option

Pezmc:
Screenshot 2023-12-20 at 16 17 15

Easiest to review with [?w=1](https://github.com/FlowFuse/flowfuse/pull/3249/files?w=1)

Fixes https://github.com/FlowFuse/flowfuse/issues/3246

### DIFF
--- a/frontend/src/pages/application/PipelineStage/form.vue
+++ b/frontend/src/pages/application/PipelineStage/form.vue
@@ -45,11 +45,13 @@
                     <template #icon><IconDeviceSolid /></template>
                 </ff-tile-selection-option>
                 <ff-tile-selection-option
-                    v-if="!isFirstStage && deviceGroupsEnabled"
+                    v-if="deviceGroupsEnabled"
                     label="Device Group"
                     :value="StageType.DEVICEGROUP"
                     description=""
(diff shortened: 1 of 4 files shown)


Steve-Mcl (approved):
Not blocking but this change to a component will require the docs to be updated here: https://flowfuse.github.io/forge-ui-components/#ff-tile-selection

knolleary:
@Steve-Mcl those docs are no longer maintained since the ui-components were merged into the core repo.

Steve-Mcl:
@knolleary I use them all the time. We discussed this recently in eng meeting. They are still up & still (mostly) relevant) so should be updated.

Steve-Mcl:
@joepavitt please assess

joepavitt:
> They are still up & still (mostly) relevant) so should be updated.

We don't currently have a pipeline by which to deploy those docs from within the core repo, so if we can get that working, then it's "easy", otherwise, it's going to be a painful process of needing to manually transfer over changes to the retired repo, and update the docs

Steve-Mcl:
I will not block.  Will raise issue on the out of sync situation this PR causes.

PR #3252, opened 2023-12-20
Fix missing instance info on pipeline stage with suspended instance

### LINKED ISSUE #2997: Implementing Device Groups Management

Steve-Mcl:
fixes #2997 

## Description

Fixes issue with missing detail and bad router-link

### Before 
chrome_iWjYkdUfnl
image

### After 

image

### DIFF
--- a/frontend/src/pages/application/Pipelines.vue
+++ b/frontend/src/pages/application/Pipelines.vue
@@ -275,7 +275,7 @@ export default {
                     this.instanceStatusMap = new Map(
                         instances.map((instance) => {
                             const previousStatus = this.instanceStatusMap.get(instance.id)
-                            return [instance.id, { ...previousStatus, ...instance.meta, flowLastUpdatedSince: instance.flowLastUpdatedSince }]
+                            return [instance.id, { ...previousStatus, ...instance.meta, flowLastUpdatedSince: instance.flowLastUpdatedSince, id: instance.id, name: instance.name }]
                         })
                     )
                 })


frontend/src/pages/application/Pipelines.vue
```diff
@@ -275,7 +275,7 @@ export default {
                     this.instanceStatusMap = new Map(
                         instances.map((instance) => {
                             const previousStatus = this.instanceStatusMap.get(instance.id)
-                            return [instance.id, { ...previousStatus, ...instance.meta, flowLastUpdatedSince: instance.flowLastUpdatedSince }]
+                            return [instance.id, { ...previousStatus, ...instance.meta, flowLastUpdatedSince: instance.flowLastUpdatedSince, id: instance.id, name: instance.name }]
```
Pezmc, line 278: I'd expect the instance status map to solely contain statuses, rather than having the ID and name added. Do we know _why_ this change was needed?

Steve-Mcl: No idea. Was very late and this fixes it. 

And as per other maps, where the id and name are present, this change aligns instances. 

Feel free to investigate but this does resolve the issue.

Steve-Mcl: Ps. "no idea"  was a quick and tired way of saying "it's tricky" and depends on which activity happens first. Typically, the map _does_ contain the id and name but that depends on a timing thing. This change ENSURES they are present.

Pezmc: Bug was introduced here (in https://github.com/FlowFuse/flowfuse/pull/3229) by the looks of things: https://github.com/FlowFuse/flowfuse/blame/ef99669cde9ea51390c521ee284a2df7e196af05/frontend/src/components/pipelines/PipelineRow.vue#L126-L144

Previously stage.instance was not mutated. Now it's being modified to contain only the instance status, which is why the fix in this PR was needed. Am looking into how to fix the source.

Steve-Mcl: > Previously stage.instance was not mutated

True, but there were existing (timing) bugs (can provide link for context if needed but it was the ones I brought up with you in a huddle - around misaligned vue event hook ups for the status polling) preventing `stage` in the `map` being populated with the `instance` property. While this change may have been naive, it was there to ensure `instance` was always something. 

That said, that line may actually no longer be required and can possibly be deleted?

Steve-Mcl: I can confirm removing the line `stage.instance = stageInstance` does fix the issue.

possibly related to the event hook up bug or other recent change.

Steve-Mcl: also, the extra code to `find` the instance name is likely now also redundant

Pezmc: I've put a replacement together in https://github.com/FlowFuse/flowfuse/pull/3255

Pezmc (changes requested):

PR #3286, opened 2024-01-03
Fix subscription deletion when putting into unmanaged mode

knolleary:
## Description

A regression was introduced in #3181 that meant when putting teams into unmanaged mode, we'd try to delete a subscription with an id of `''` - not gonna happen.

This fixes that and improves the existing unit test to ensure the delete function is called with the expected subscription id.

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -546,7 +546,7 @@ module.exports.init = async function (app) {
                             // needed by closeSubscription. This is to ensure when the
                             // stripe callback arrives we don't trigger a suspension of
                             // the team resources.
-                            await stripe.subscriptions.del(subscription.subscription, {
+                            await stripe.subscriptions.del(existingSubscription, {
                                 invoice_now: true,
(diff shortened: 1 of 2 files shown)


forge/ee/lib/billing/index.js
```diff
@@ -546,7 +546,7 @@ module.exports.init = async function (app) {
                             // needed by closeSubscription. This is to ensure when the
                             // stripe callback arrives we don't trigger a suspension of
                             // the team resources.
-                            await stripe.subscriptions.del(subscription.subscription, {
+                            await stripe.subscriptions.del(existingSubscription, {
```
Pezmc, line 549: Easy to miss, I wouldn't expect `subscription.subscription` to be `''` but indeed I see it on line 528 above: https://github.com/FlowFuse/flowfuse/blob/9df9a6f4d37452f5322acda65a3a8082e83452a3/forge/ee/lib/billing/index.js#L528

Pezmc (approved):

PR #3287, opened 2024-01-03
Devices: Fleet Mode Label

### LINKED ISSUE #3005: Renaming "Standard" Device Mode
### Description

This issue has been opened to exclusively discuss the proposal of renaming the "Standard" Device Mode to "Fleet Mode" within our system. The suggestion emerged from discussions in another thread see #2884 

The proposition is to adopt the term "Fleet" or "Fleet Mode," aligning with practices seen in platforms like AWS SageMaker, where "Fleets" refer to collections of logically grouped devices.

One significant contention against the use of "Fleet" stems from its traditional application within the context of "Fleet Management," a term entrenched in the logistics and transportation sectors for managing groups of vehicles, like trucks and cars. 

### Epic/Story

_No response_

Pezmc:
## Description

Replaces references to device default mode with "fleet mode".

This is implemented by changing the DeveloperModeBadge to a generic DeviceModeBadge, which is then used to display the icon in the table, text in the overview page and as a badge in the status bar.

We should avoid using the BeakerIcon and IconDeviceSolid directly as this spreads business logic throughout the app. Instead the new component that knows how to display a devices mode should be used.

Screenshot 2024-01-03 at 16 15 12
Screenshot 2024-01-03 at 16 15 02
Screenshot 2024-01-03 at 16 14 17

## Related Issue(s)

Fixes #3005 

## Checklist

 - [x] Suitable unit/system level tests have been added and they pass - Effectively text only change
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
Can we (if we don't already) get a hover tooltip on the "Mode" pill please to explain what this means, as I don't think "Fleet Mode" on it's own is enough.

Assume we'll have some supporting docs to go with this too?

Pezmc:
@joepavitt Ready for re-review. When in icon mode there is also a tooltip for fleet mode (but not badge or text since it would be redundant!

joepavitt (changes requested):
Tooltips still need to eb added as per my comment above

Pezmc:
What are you expecting the tooltip to say @joepavitt? If we copy the pattern for developer mode it'll just say "fleet mode"?

joepavitt:
> What are you expecting the tooltip to say

What does Fleet mode mean? A quick one-liner on the basic principle would be nice.

Pezmc:
It's just the default mode a device is in. It's either in fleet mode, or developer mode. Fleet mode is the opposite of developer mode. I'm really unclear on how to label that other than "not developer mode".

joepavitt:
Yeah, I'm not a fan of the "Fleet Mode" terminology, as I don't think it's at all clear to a user what it actually means. We should be able to come up with _something_ though - @Steve-Mcl or @MarianRaphael - thoughts?

Pezmc:
In the interest of moving forward, I propose we merge this PR as is, and open an issue for @MarianRaphael (with input from @Steve-Mcl, @joepavitt and myself) on how best to label this to reduce and possible user confusion. 

Does that sound alright?

Steve-Mcl:
Just read the OP:  "fleet mode" is IMO, almost as off-putting as "DevOps Pipelines"

Under the hood, the equivalent mode is "autonomous" - that was chosen to signify a device is "auto" doing stuff.  That too might also be off-putting to some but i've seen this term in the likes of cisco APs running Autonomous mode.

Other suggestions we might consider:
* "Linked mode"?
* "Production Mode"
* "Standard Mode"
* "Operational Mode"
* "Automated Mode"
* "Managed Mode"

joepavitt (approved):

PR #3297, opened 2024-01-05
Add teamTrial, teamType, deviceLastSeen and deviceMode to stats end point

knolleary:
The `/api/v1/admin/stats` endpoint is missing some useful information about the platform. This PR adds in some additional stats we want to easily track:

**How many of each team type:**
```
"teamsByType": {
   "starter": 16,
   "Starter": 7,
   "Team": 6,
   "Enterprise": 1
},
```

**Trial states**
 - created: active trial
 - ended: trial expired, has not converted

```
"trialsByState": {
   "created": 1,
   "ended": 8
}
```

**Device mode counts**
```
"devicesByMode": {
   "autonomous": 18
   "developer": 4
},
```
**Device last seen counts**
A simple count of how many devices were last seen within these intervals. Note that a device seen in the last 'day' is not counted as being seen in the last 'week' etc.
```
"devicesByLastSeen": {
   "never": 18,
   "day": 1,
   "week": 0,
   "month": 0,
   "older": 3
},
```

### DIFF
--- a/forge/routes/api/admin.js
+++ b/forge/routes/api/admin.js
@@ -1,20 +1,30 @@
+const { Op } = require('sequelize')
+
 module.exports = async function (app) {
     async function getStats () {
         const userCount = await app.db.models.User.count({ attributes: ['admin'], group: 'admin' })
         const projectStateCounts = await app.db.models.Project.count({ attributes: ['state'], group: 'state' })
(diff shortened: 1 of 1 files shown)


hardillb (approved):

hardillb:
admin token generation failed on postgres, re-running to see if it fails again

knolleary:
Postgres fixed - got the case of a column name wrong which pg doesn't like

PR #3322, opened 2024-01-11
Ensure swagger-ui baseDir is set to non-existent directory

knolleary:
~This sets the `baseDir` option of swagger-ui to *disable* its default behaviour that leads to it exposing its own source code on the routes it serves up. By setting it to a non-existent directory it won't serve up unexpected content.~

Fixes https://github.com/FlowFuse/security/issues/52

This upgrades to the latest release of swagger-ui that addresses the above issue.

### DIFF
--- a/forge/routes/api-docs.js
+++ b/forge/routes/api-docs.js
@@ -66,6 +66,13 @@ module.exports = fp(async function (app, opts, done) {
         },
         logLevel: 'silent',
         hideUntagged: true,
+        staticCSP: true,
+        transformStaticCSP: header => {
+            header.replace(
+                /script-src 'self'/,
+                "script-src 'self' 'unsafe-inline'"
+            )
+        },
         uiConfig: {
             defaultModelsExpandDepth: -1,
             operationsSorter: 'alpha',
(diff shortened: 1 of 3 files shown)


knolleary:
@Pezmc follow up to our discussion earlier, I discovered this is only used to serve up additional static content the swagger doc references. We don't have any such content to serve so I don't need to do any copying around of content.

knolleary:
Put back to draft - looking at `main` of the fastify swagger code they have changed some logic in this area that won't work with this workaround.

Pezmc:
Of note (you likely already saw this) swagger UI recently cut a [2.0](https://github.com/fastify/fastify-swagger-ui/releases/tag/v2.0.1) release, though the breaking change doesn't seem to impact this PR: https://github.com/fastify/fastify-swagger-ui/pull/105

knolleary:
There is a pending swagger-ui release that we will wait for before resolving this PR with a permanent fix.

hardillb:
@knolleary did you get answers from swagger-ui about this (and the pending release on their end)?

hardillb (approved):

PR #3327, opened 2024-01-11
Change expireTokens task schedule

knolleary:
## Description

Part of #2793 

This updates the schedule of the `expireTokens` task to run at a randomly picked time each day. This is in aid of improving our horizontal scalability as describe in my comments in #2793

### DIFF
--- a/forge/housekeeper/tasks/expireTokens.js
+++ b/forge/housekeeper/tasks/expireTokens.js
@@ -1,9 +1,13 @@
 const { Op } = require('sequelize')
 
+const randomInt = (min, max) => { return min + Math.floor(Math.random() * (max - min)) }
+
 module.exports = {
     name: 'expireTokens',
     startup: true,
-    schedule: '@daily',
+    // Pick a random hour/minute for this task to run at. If the application is
+    // horizontal scaled, this will avoid two instances running at the same time
+    schedule: `${randomInt(0, 60)} ${randomInt(0, 24)} * * *`,
     run: async function (app) {
         await app.db.models.Session.destroy({ where: { expiresAt: { [Op.lt]: Date.now() } } })
         await app.db.models.AccessToken.destroy({ where: { expiresAt: { [Op.lt]: Date.now() } } })
--- a/forge/housekeeper/tasks/expireTokens.js
+++ b/forge/housekeeper/tasks/expireTokens.js
@@ -1,13 +1,13 @@
 const { Op } = require('sequelize')
 
-const randomInt = (min, max) => { return min + Math.floor(Math.random() * (max - min)) }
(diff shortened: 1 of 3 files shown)


forge/housekeeper/tasks/expireTokens.js
```diff
@@ -1,9 +1,13 @@
 const { Op } = require('sequelize')
 
+const randomInt = (min, max) => { return min + Math.floor(Math.random() * (max - min)) }
```
Steve-Mcl, line 3: Slightly pedantic but my OCD is twitching here Nick.  In the original version, `min` is inclusive but `max` is exclusive.  i.e. in its current form, `randomInt(/*min*/0, /*max*/1)` will ALWAYS return `0`. 

Can I propose both min and max are inclusive?

Suggestion: 

```suggestion
const randomInt = (min, max) => Math.floor(Math.random() * (max - min + 1) + min)
```

forge/housekeeper/tasks/expireTokens.js
```diff
@@ -1,9 +1,13 @@
...
+
 module.exports = {
     name: 'expireTokens',
     startup: true,
-    schedule: '@daily',
+    // Pick a random hour/minute for this task to run at. If the application is
+    // horizontal scaled, this will avoid two instances running at the same time
+    schedule: `${randomInt(0, 60)} ${randomInt(0, 24)} * * *`,
```
Steve-Mcl, line 10: Only apply if first suggestion is accepted.

```suggestion
    schedule: `${randomInt(0, 59)} ${randomInt(0, 23)} * * *`,
```

Steve-Mcl (commented):
Functionally, works but code messed with my OCD, suggestions applied - please consider.

forge/housekeeper/tasks/expireTokens.js
knolleary: We use this pattern in a couple of the timertasks with this function copied between them. If we change it in one, we'll need to change them all so as not to have inconsistent implementations.

forge/housekeeper/tasks/expireTokens.js
knolleary: Resolved by https://github.com/FlowFuse/flowfuse/pull/3327/commits/9bb33fa445c23daa24226ecce5df43bb1054d309

forge/housekeeper/tasks/expireTokens.js
knolleary: https://github.com/FlowFuse/flowfuse/pull/3327/commits/9bb33fa445c23daa24226ecce5df43bb1054d309

Steve-Mcl (approved):
My likey :)

PR #3330, opened 2024-01-12
Update device on group membership change

### LINKED ISSUE #3260: Automatic snapshot assignment for newly added Devices in a Device Group
## Current Behaviour
In the current DevOps pipeline setup, when a Device Group is targeted for deployment, only the devices that are part of the group at the time of pipeline execution receive the new target snapshot. Devices added to the group after the execution do not automatically receive this snapshot.

## User Story
As a: User managing a fleet of devices

I want: To have the ability to assign a device to a device group after a pipeline execution, and for the device to automatically receive the snapshot that was deployed to the group during that execution.

So that: I can ensure all devices in a group are synchronized with the latest snapshot without the need to re-execute the pipeline for each new device added to the group.

## Scope
- Any device added to the group after the pipeline execution should automatically receive the snapshot that was most recently deployed to that group.
- Display the current target snapshot in the Device Group overview

## License
Enterprise Tier (as Device Groups)

Steve-Mcl:
closes #3260

## Description

Premise: The customer wishes for devices added to a group to receive the snapshot last pushed via a pipeline. Additionally, the customer wishes for the pipeline deployed snapshot to be removed from the device when it is removed from the group.

* ~~Adds field `activePipelineStageId` to `DeviceGroups` so that added/removed devices can query the active pipeline stage and be updated accordingly~~
* Adds field `targetSnapshotId` to `DeviceGroups` so that added/removed devices can query the DeviceGroup and be updated accordingly
* Updates UI to warn user of the consequences by way of a confirmation dialog (shown when user attempts to save changes to a device group)
* Immediately transmits (in MQTT coms mode) an update to the entire group and any devices removed
* Updated test coverage
   * Adds: Add and Remove Devices get an update request
   * ~~Adds: All Devices removed get an update request and clears activePipelineStageId~~
   * Adds: All Devices removed get an update request and clears `DeviceGroup.targetSnapshotId`
* Updated docs 

## Related Issue(s)

#3260

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
@ZJvandeWeg not specifically adding you as a reviewer, but  I thought to tag you as docs around device-groups will be updated as per comment here: https://github.com/FlowFuse/flowfuse/pull/3318#discussion_r1448968509

knolleary:
@Steve-Mcl think I'd like to do a huddle review of this PR with you on Monday. This may be building on choices already made, but I'd like to understand better why the device group needs the `activePipelineStageId` adding - we already have the `PipelineStageDeviceGroup` model that holds the relationships between DeviceGroup and PipelineStage which, I think, could be used instead of adding this column. I appreciate I may be missing something here - easier to step through together on Monday.

Steve-Mcl:
> @Steve-Mcl think I'd like to do a huddle review of this PR with you on Monday. This may be building on choices already made, but I'd like to understand better why the device group needs the `activePipelineStageId` adding - we already have the `PipelineStageDeviceGroup` model that holds the relationships between DeviceGroup and PipelineStage which, I think, could be used instead of adding this column. I appreciate I may be missing something here - easier to step through together on Monday.

Sure, NP.

For the memory banks and later discussion:

1. A device group can be assigned to multiple pipelines. Without `activePipelineStageId` there is no way of knowing which of these pipeline stage was deployed.
2. To permit us to indicate which pipeline stage was/is executing, (and as a nice side-effect) prevent all pipelines in the page animating as if deploying when one is operated, each `PipelineStageDeviceGroup` row holds the `targetSnapshotId` that was deployed. This is the value we need to get to understand _WHICH_ of the multiple pipelines was executed.

forge/db/migrations/20240110-01-add-active-pipeline-id-to-devicegroup.js
knolleary, line 1: We have since merged a migration with a newer filename to main. This migration will need renaming.

Steve-Mcl: done [9e88a2c](https://github.com/FlowFuse/flowfuse/pull/3330/commits/9e88a2cb795ac4bde78f091ee936b84f9655883e)

docs/user/device-groups.md
```diff
@@ -12,6 +12,12 @@ navTitle: Device Groups
 When managing many devices that are intended to run the same [snapshot](./snapshots.md), Device Groups allow you
 to organise your devices into logical groups.
 These groups can then be set as the target of a [DevOps Pipelines](./devops-pipelines.md).
+
+Furthermore:
+* Devices added to an active Device Group will automatically be updated to the active pipeline snapshot
+* Devices removed from an active Device Group will have their active pipeline snapshot cleared
+More details are provided below.
```
ZJvandeWeg, line 19: Can we link these to the anchors? Or do h3 elements not get an anchor by default? "More details provided below" works in this diff as the examples are marked in a nice green shade, but I'm not sure the end-user knows where to scroll to.

Steve-Mcl: @ZJvandeWeg 

Updated in [769e49c](https://github.com/FlowFuse/flowfuse/pull/3330/commits/769e49c44463ff5926cd1946cd5338d7f0cce68b)

Please see how that reads now.

docs/user/device-groups.md
```diff
@@ -12,6 +12,12 @@ navTitle: Device Groups
 When managing many devices that are intended to run the same [snapshot](./snapshots.md), Device Groups allow you
 to organise your devices into logical groups.
 These groups can then be set as the target of a [DevOps Pipelines](./devops-pipelines.md).
+
+Furthermore:
+* Devices added to an active Device Group will automatically be updated to the active pipeline snapshot
+* Devices removed from an active Device Group will have their active pipeline snapshot cleared
+More details are provided below in [Adding a Device to a group](#adding-a-device-to-a-group-which-has-an-active-pipeline-snapshot) and [Removing a Device from a group](#removing-a-device-from-a-group-which-has-an-active-pipeline-snapshot)
```
ZJvandeWeg, line 19: ❤️

Steve-Mcl:
@knolleary review ready following changes discussed. OP updated, tests updated etc.

forge/ee/db/controllers/DeviceGroup.js
```diff
@@ -120,52 +124,89 @@ module.exports = {
             // otherwise, throw a friendly error message along with the original error
             throw new Error(`Failed to update device group membership: ${err.message}`)
         }
+        if (changeCount > 0) {
+            // clean up where necessary
+            // check to see if the group is now empty
+            const remainingDevices = await deviceGroup.deviceCount()
+            if (remainingDevices === 0) {
+                deviceGroup.targetSnapshotId = null
```
knolleary, line 132:
```suggestion
                deviceGroup.targetSnapshotId = null
```

Wondering if setting the snapshot to null when the group is emptied is the right behaviour. A user could be in the middle of changing the devices in the group (albeit doing it in multiple steps) - and clearing the snapshot mid-way through the task could be unexpected.

forge/ee/db/controllers/Pipeline.js
```diff
@@ -432,6 +432,9 @@ module.exports = {
                 // Update the targetSnapshot of the device group
                 await targetDeviceGroup.PipelineStageDeviceGroup.update({ targetSnapshotId: sourceSnapshot.id }, { transaction })
 
+                // Update the targetSnapshotId on the device group
+                await targetDeviceGroup.update({ targetSnapshotId: sourceSnapshot.id }, { transaction })
+
```
knolleary, line 437: A possibly cleaner approach would be to have a controller for DeviceGroup which provides a function to update the targetSnapshotId which takes care of updating the devices *and* sending the update command. As it is we've got that logic spread through this Pipeline controller. Not blocking, just a thought.

forge/ee/db/controllers/DeviceGroup.js
Steve-Mcl: I don't _think_ it is of consequence. 

it just means the pipeline needs to be executed to make it re-sync.

The issue this was added for is adding devices to a _spare_ empty group (months/years later) and suddently they fire up running old (dangerous?) snapshots.

I feel quite strongly that we DONT actually provide a checkbox `[x] remove current snapshot` / `[x] apply current snapshot` - I think it is dangerous and suspect most (other) users will not like the behaviour we are adding at all


From the ISSUE https://github.com/FlowFuse/flowfuse/issues/3260#issuecomment-1866211727:

> 
> @MarianRaphael I believe it would be prudent to consider this:
> 
> Personally, I believe this should be done via an option when adding a device to a group since you may NOT actually want this to occur immediately and without warning that "things" could start moving/happening.
> 
> Consider the scenario:
> 
> You have x devices that you need to add to a device group BUT FIRST you need to update the source snapshot
> The current snapshot is NOT what needs to be deployed (because changes have not yet been made to accommodate the 10 new devices)
> You CANNOT/DO NOT want to deploy the new changes to the existing devices without the other devices being online/ready to receive the new snapshot.
> By simply adding a checkbox "[x] Auto Deploy current device group snapshot 'version 2.1'?" at the time of adding the device to a group can make this problem go away!

forge/ee/db/controllers/Pipeline.js
Steve-Mcl: I _could_ add a method to DeviceGroup controller but I dont want to be updating devices until the transaction is committed.

knolleary (approved):

PR #3331, opened 2024-01-12
Add CommandResponse pattern for handling scalable platform

### LINKED ISSUE #2792: Scalability: redesign MQTT Command/response pattern
### Description

Part of #2782 

The flow to create a Device Editor session involves the following initial steps ([ref](https://flowfuse.com/docs/contribute/workflows/device-editor/))

image

- An HTTP request is made to trigger the session
- An MQTT message is sent to the device to get it to connect
- The device connects its websocket
- The device published an MQTT message to confirm
- The HTTP request from the first step is completed with the session identifier.

The key point is that, in the above flow, the initial HTTP request is only responded to when a reply to the initial MQTT publish has been received over MQTT. In a scalable system, the Forge App will use a Shared Subscription to the broker - so that work can be distributed and will mean the reply could end up at a different instance.

The other place the command/response pattern is used is when the request to create a new Device Snapshot. The command is published and the device sends back the snapshot over MQTT. The original HTTP request to create the snapshot is blocked and only responds once the snapshot has been created.

#### Possible Solutions

1. Do not use a shared subscription for command/response MQTT pattern - and include a reply-to topic in the command message to ensure the response comes back to the topic subscribed to the correct instance.
    - **Cons**
        - Doesn't really solve the stateful nature of the command/response pattern
        - The current Device Agent has a hardcoded response topic to use - so we'd have a legacy issue with existing devices.

2. Redesign the whole command/response pattern to support stateless behaviour. The initial HTTP response will respond once the initial MQTT message is sent with some sort of 'pending' response including a request identifier. An API can then be used to track the completion of the pending request. The exact details of which will necessarily differ between the two current use-cases of the Command/Response pattern.
    - **Pros**
        - Cleanest solution that doesn't require any Device Agent changes
        - Properly stateless
     - **Cons**
        - Not trivial to implement properly - needs some thought on database table structure.
        - Changes the external API for settings up Device Editor and capturing Device Snapshot

### Epic/Story

#2782

knolleary:
Closes #2792 

## Description

This modifies our command/response messaging pattern to support horizontal scaling of the forge app.

In summary - once scaled, we cannot guarantee the response message comes back to the specific app instance that published the request and that is holding an unresolved promise. In that scenario, we store the message in the database and republish a notification message that is received by all forge app instances - whichever one recognises the correlationData value can retrieve the message from the db and continue processing as normal.

### Summary of changes
 - Change core platform subscriptions (and ACL checks) to use `$share/platform/` shared-subscription prefix (all except the logs end point.. that needs more work in a separate item)
 - Add `CommandResponse` table and migration
 - Add logic to store the response in the db if not recognised as a 'local' item
 - Add handling for new `.../response/broadcast` topic used to prompt the platform to process a stored response if it recognises it

### Remaining tasks

 - Add a TimerTask to periodically clean up the `CommandResponse` table of anything stale - just in case they are orphaned
 - Add unit tests - I cannot see any tests for the existing command/response flows so need to add something in that area first

### DIFF
--- a/forge/comms/aclManager.js
+++ b/forge/comms/aclManager.js
@@ -104,15 +104,18 @@ module.exports = function (app) {
         forge_platform: {
             sub: [
                 // Receive status events from project launchers
-                // - ff/v1/+/l/+/status
-                { topic: /^ff\/v1\/[^/]+\/l\/[^/]+\/status$/ },
+                // - ff/v1/<team>/l/<instance>/status
+                { topic: /^ff\/v1\/[^/]+\/l\/[^/]+\/status$/, shared: true },
                 // Receive status events, logs and command responses from devices
-                // - ff/v1/+/d/+/status
-                { topic: /^ff\/v1\/[^/]+\/d\/[^/]+\/status$/ },
-                // - ff/v1/+/d/+/logs
+                // - ff/v1/<team>/d/<device>/status
+                { topic: /^ff\/v1\/[^/]+\/d\/[^/]+\/status$/, shared: true },
+                // - ff/v1/<team>/d/<device>/logs
                 { topic: /^ff\/v1\/[^/]+\/d\/[^/]+\/logs$/ },
-                // - ff/v1/+/d/+/logs
-                { topic: /^ff\/v1\/[^/]+\/d\/[^/]+\/response$/ }
+                // - ff/v1/<team>/d/<device>/response
+                { topic: /^ff\/v1\/[^/]+\/d\/[^/]+\/response$/, shared: true },
+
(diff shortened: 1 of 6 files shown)


hardillb (approved):

knolleary:
This still needs unit tests adding. Sorry, should have been more explicit to review without merge.

knolleary:
Whilst writing the tests I've found a different, much simpler, approach that I'll be PRing. It will revert much of this PR - but as the migration has already hit staging, it'll need to do something to tidy that up as the CommandResponse table is no longer required.

PR #3334, opened 2024-01-12
Allow device to be add to platform regardless of agent version recorded in DB

### LINKED ISSUE #3248: Cannot add device to group if it has not yet connected to the platform
### Description

1. When creating a device group, we only allow you to pick from the devices that have been assigned to the current application.
2. To assign a device to an application, we require the device to have connected so we know what version of the agent it is running (older versions do not support application-assignment).

If I want to setup a bunch of devices and get them assigned to a group, I know need to fully setup each device individually - including getting it connected. This adds a bunch of friction to getting started with device groups.

We should relax the restriction on adding a device to a group to allow it be done if we don't know the agent version - only block if if we *know* the agent is too old.

We then also need to handle the case of that device eventually coming online and reporting its version as too told. But given a device could be downgraded at any time anyway, this should be handled.

### Epic/Story

_No response_

Steve-Mcl:
closes #3248 

## Description

* Permits devices to be assigned to an application regardless of `agentVersion` reported by the device
* Updates the device status to show the agent version number in a pill (with status and tooltip) so the user can understand if/why an application device is not receiving its snapshot
   * See screenshots below
* Tests (`test/unit/forge/routes/api/device_spec.js`)
   * Removes
      * "cannot assign to an application if device agent version is not present"
      * "cannot assign to an application if device agent version is too old"
   * Adds
      * "can assign to an application even if device agent version is not present"
      * "can assign to an application if device agent version even is too old to support application assignment"
* Tests (`test/unit/forge/comms/devices_spec.js`)
   * Adds
      * "sends update to clear application device configuration if device agent is older than 1.11.0"
* Docs have been scanned and appear to remain relevant (no doc updates included in this PR)

### Device with agent version that supports Application assignment

image

### Device with agent version too old
_Application assignment was introduced in V1.11.0_

image

### Device with no version info
_Either too old (i.e. pre v0.7.0, before `agentVersion` was added) OR the DA has not yet checked in_

image

## Related Issue(s)

 #3248

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
Looks great - great work Steve. Will leave Nick to comment/review server-side

knolleary (approved):

PR #3338, opened 2024-01-15
Do not allow password to be same as email/username

knolleary:
## Description

This adds an additional requirement for passwords to not match a user's email/username.

I've tried to standardise the various forms which allows a user to enter/change their password to be consistent in feedback.

Fixes https://github.com/FlowFuse/security/issues/47

### DIFF
--- a/forge/db/controllers/User.js
+++ b/forge/db/controllers/User.js
@@ -30,6 +30,11 @@ module.exports = {
             if (zxcvbn(newPassword).score < 3) {
                 throw new Error('Password Too Weak')
             }
+            if (newPassword === user.username) {
+                throw new Error('Password must not match username')
+            } else if (newPassword === user.email) {
+                throw new Error('Password must not match email')
+            }
             user.password = newPassword
             user.password_expired = false
             return user.save()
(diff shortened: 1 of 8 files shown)


knolleary:
Looking at test failures...

knolleary:
I enabled password strength checks on the reset-password route - but forgot to update the tests to use a strong enough password to pass

forge/db/controllers/User.js
```diff
@@ -39,9 +44,14 @@ module.exports = {
     },
 
     resetPassword: async function (app, user, newPassword) {
-        // if (zxcvbn(newPassword.score < 3)) {
-        //     throw new Error('Password Too Weak')
-        // }
+        if (zxcvbn(newPassword).score < 3) {
```
hardillb, line 47:
```suggestion
        if (zxcvbn(newPassword).score < 2) {
```

Everywhere else in the code the strength is tested to `< 2`

hardillb (changes requested):
Looks ok, but the reset rule is stronger than the rest ( `score < 3` vs  `score < 2` everywhere else.

forge/db/controllers/User.js
knolleary: I just uncommented your code ;)

PR #3342, opened 2024-01-15
Improved handling of command/response pattern in scaled environment

knolleary:
This is a reworking of #3331 to improve how we make the command/response MQTT pattern scalable.

The work done in 3331 introduced a CommandResponse db table and a mechanism for a response to be stored and retrieved by the correct platform instance.

Whilst writing tests for that, I discovered a critical piece I had missed in the original design work. The command sent to the device already includes a `responseTopic` and the existing Device Agent will use that if set.

This PR reverts the DB changes in 3331 (leaving the migration file in place, albeit empty, as it had already been merged to main).

Instead, the commsClient generates a `platformId` that will unique to this running instance. That is appended to the response topic the platform subscriptions to. This means each instance of the platform will subscription to a unique topic.

We can then include the `responseTopic` in the command sent to the device and the response will come straight back to the required instance.

Much cleaner than the db table approach.

This also adds a simple test for some of the existing command/response logic.


Steve-Mcl (commented):
Functional checks of code along side branch prior to revered PR #3331 all good.

Will finalise review after test coverage additions.

knolleary:
Tests added

Steve-Mcl (approved):

Issue #2569, opened 2023-08-08
Possibility of orphaned devices and storage items connected to teams


Pezmc:
### Current Behavior

When instances, teams, devices, applications or users are deleted, several related models are left behind orphaned in the database.

On a quick review; this includes: 
 - Devices for Instances
 - ProjectSnapshots for Instances
 - DeviceSettings for Devices
 - AccessTokens for Users
 - Invitations for Users

### Expected Behavior

These tables should also be cleaned up to prevent orphaned data being left behind; there are potential privacy and security implications if they sit in the database forever.

Proposal:
 - Research into what data is left orphaned
 - For each decide whether to:
   - Keep the data around for audit purposes and have the default SET NULL
   - Set ON DELETE for immediate removal
   - Manually clean-up as part of related object deletion
   - Block deletion of the parent object if there are still children (how teams/application currently behave for instances)
   - Allow orphaned record, but have a cron that cleans them up after X days (to allow recovery) 
   - Set a "deleted" flag on those tables and schedule a cron that cleans up after X days (to allow recovery)

### Steps To Reproduce

Delete a team/instance and check the database for any orphaned r

### Environment

- FlowForge version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Pezmc:
Related to https://github.com/FlowFuse/CloudProject/issues/335

joepavitt:
@knolleary any idea on the status of this? Was marked for `2.1`, and has been sat in "todo" for a while?

knolleary:
@joepavitt Need to re-review the full list of potentially orphaned items. Some bits (like Devices) may have been addressed by other work in the meantime.

I'll put this on my pile - its a pretty small shelf-contained chore that'll be good to get done.

knolleary:
Have reviewed the list. Only one immediate action to take - #4063 - then we can close this off.

PR #3356, opened 2024-01-18
Quieten down `Device.Team` is `null` exceptions

### LINKED ISSUE #3221: Exception caused by Device.Team is null
Sentry Issue: [FLOWFUSE-BACKEND-2A](https://flowfuse.sentry.io/issues/4720465020/?referrer=github_integration)

```
TypeError: Cannot read properties of null (reading 'hashid')
  File "/usr/src/forge/app/node_modules/@flowforge/flowforge/forge/db/controllers/Device.js", line 77, in Object.sendDeviceUpdateCommand
    app.comms.devices.sendCommand(device.Team.hashid, device.hashid, 'update', payload)
  File "/usr/src/forge/app/node_modules/@flowforge/flowforge/forge/comms/devices.js", line 108, in DeviceCommsHandler.handleStatus
    await this.app.db.controllers.Device.sendDeviceUpdateCommand(device)
  File "node:internal/process/task_queues", line 95, in process.processTicksAndRejections
```

Steve-Mcl:
closes #3221

## Description

The PR aims to simply remove the exceptions (currently flooding sentry) by _handling_ the situation and logging out the problem device(s) hashID in `app.warn`. 

### Background

I have checked through the calls to `app.comms.devices.sendCommand` via `sendDeviceUpdateCommand` in forge/app/node_modules/@flowforge/flowforge/forge/db/controllers/Device.js"

As far as I can see `device` is always a fully loaded model (e.g. loaded via await `app.db.models.Device.byId(targetDevice.id)`) with the required associations

So if these ARE orphaned devices (no `Team` associated) we should at least prevent the exceptions and extensive sentry logging (the orphaned devices part of this was raised in https://github.com/FlowFuse/flowfuse/issues/2569)

## Related Issue(s)

#3221

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/controllers/Device.js
+++ b/forge/db/controllers/Device.js
@@ -80,7 +80,20 @@ module.exports = {
             } else {
                 delete payload.application // exclude application property to avoid triggering the wrong kind of update on the device
             }
-            app.comms.devices.sendCommand(device.Team.hashid, device.hashid, 'update', payload)
+
+            // ensure the device has a team association
+            let team = device.Team
+            if (!team) {
+                // reload the device with the team association
+                const _device = await app.db.models.Device.byId(device.id)
+                team = _device?.Team
+                if (!team) {
+                    app.log.warn(`Failed to send update command to device ${device.hashid} as it has no team association`)
+                    return
+                }
+            }
+            // send out update command
+            app.comms.devices.sendCommand(team.hashid, device.hashid, 'update', payload)
         }
     },
     /**


knolleary:
How has an orphaned device got this far for us to be sending a command to it? One theory is we have already detected this is an orphaned device so we're telling it to stop runnning - but to do that, we need the team id for the publish... which we don't have.

Should we be doing something earlier in the process for this scenario such as revoking the devices credentials etc? We don't have the ability to kick the device off the broker, but we should be able to stop it reconnecting.

hardillb:
If we revoke it's ACL entry then (once the broker cache expires) it should just dump the check in/ping messages.

Steve-Mcl:
> As for this PR, I still think this is addressing the issue too late

I dont disagree, but in a future where a mistake or manual operation lands us here again, it would guard AND would log a warn - helping to a) avoid the noisy sentry and b) diagnose/trace the issue.

Happy either way.

knolleary:
I'm going to merge this as-is so we can capture the device id that is leading to this error being logged. The theory has been this is an orphaned device - however, of the orphaned devices, none have a recent `lastSeenAt` value - something that would have been updated *before* go down the path that leads to this error. So we need to identify the device to figure out what it is doing.

knolleary (approved):

PR #3386, opened 2024-01-25
docs: Kube stacks %s/flowforge/flowfuse/

ZJvandeWeg:


hardillb:
We published the containers on both names as part of v2.0.0, we didn't update the helm/docker deployments to use the new names (as the only way to test the push to the new names was to do the release).

We will update the names in the 2.1.0 release

https://github.com/FlowFuse/helm/pull/274
https://github.com/FlowFuse/docker-compose/pull/117

Will merge this with the 2 above PRs

PR #3399, opened 2024-01-29
Limit referers to just origin for cross origin

hardillb:
## Description

Set the Referer-Policy flag in helmet to `origin-when-cross-orgin` so no query string parameters are shared

## Related Issue(s)

part of FlowFuse/security#67

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/forge.js
+++ b/forge/forge.js
@@ -315,7 +315,9 @@ module.exports = async (options = {}) => {
             frameguard: {
                 action: 'deny'
             },
-            referrerPolicy: 'origin-when-cross-origin'
+            referrerPolicy: {
+                policy: 'origin-when-cross-origin'
+            }
         })
 
         // Routes : the HTTP routes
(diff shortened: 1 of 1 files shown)


forge/forge.js
```diff
@@ -314,7 +314,8 @@ module.exports = async (options = {}) => {
             strictTransportSecurity,
             frameguard: {
                 action: 'deny'
-            }
+            },
+            referrerPolicy: 'origin-when-cross-origin'
```
knolleary, line 318: The docs for helmet suggest this needs to be an object, not just a plain string. Have you verified this sets the header as you expect?

Also need to verify this will still send useful information to posthog/hubspot as they depend on the referer information.

hardillb: And yes it was setting the header, but I've updated to match the doc.

hardillb: Tested posthog, no email reset tokens in the referrer header, but the full page paths are still available in the posthog logs

knolleary (approved):

PR #3409, opened 2024-01-30
Handle null status events from devices

knolleary:
## Description

Devices can publish a `null` status update after then reconnect their MQTT connection (https://github.com/FlowFuse/device-agent/blob/93e7b488271c7241df4a6e42db80ef074c84458d/lib/mqtt.js#L66) - as this call is unguarded unlike other status checkins.

There is a question on whether the device should be doing that, but given its the current behaviour, we need to handle it.

This PR updates the code to deal with a null payload - inferring its an `updating` device - and skips any further checks. The device will finish updating and its regular checkin will resync its state.

### DIFF
--- a/forge/db/controllers/Device.js
+++ b/forge/db/controllers/Device.js
@@ -12,29 +12,35 @@ module.exports = {
         return device.activeSnapshotId !== device.targetSnapshotId
     },
     updateState: async function (app, device, state) {
-        if (state.state) {
-            device.set('state', state.state)
-        }
-        if (state.agentVersion) {
-            device.set('agentVersion', state.agentVersion)
-        }
         device.set('lastSeenAt', literal('CURRENT_TIMESTAMP'))
-        if (!state.snapshot || state.snapshot === '0') {
-            if (device.activeSnapshotId !== null) {
-                device.set('activeSnapshotId', null)
-            }
+        if (!state) {
+            // We have received a `null` state from the device. That means it
+            // is busy updating itself. Update out local state to infer as much
+            // as we can from that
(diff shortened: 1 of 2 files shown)


forge/db/controllers/Device.js
```diff
@@ -12,29 +12,35 @@ module.exports = {
...
         device.set('lastSeenAt', literal('CURRENT_TIMESTAMP'))
-        if (!state.snapshot || state.snapshot === '0') {
-            if (device.activeSnapshotId !== null) {
-                device.set('activeSnapshotId', null)
-            }
+        if (!state) {
+            // We have received a `null` state from the device. That means it
+            // is busy updating itself. Update out local state to infer as much
+            // as we can from that
+            device.set('state', 'updating')
```
Steve-Mcl, line 20:
```suggestion
        if (!state) {
            // We have received a `null` state from the device. That means it
            // is busy updating itself. Update our local state to infer as much
            // as we can from that
            device.set('state', 'updating')
```

Super minor typo (had to read comment several times)

Steve-Mcl (approved):
All good bar minor comment that tripped my brain!

knolleary:
@Steve-Mcl for future reference, if you propose a change for a minor typo, you are encouraged to self-merge the proposal to get it fixed and then merge the PR as a whole - rather than wait for me to get back to this.

Steve-Mcl:
> @Steve-Mcl for future reference, if you propose a change for a minor typo, you are encouraged to self-merge the proposal to get it fixed and then merge the PR as a whole - rather than wait for me to get back to this.

Noted

PR #3407, opened 2024-01-30
User Identification via HubSpot Integration

Yndira-FlowForge:
## Description

Enabled HubSpot
Added the user identification script according to https://developers.hubspot.com/docs/api/events/tracking-code#identify-a-visitor

## Related Issue(s)

closes [#3392](https://github.com/FlowFuse/flowfuse/issues/3392)

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [x] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Backport needed? -> add the `backport` label
 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #3411, opened 2024-01-30
Identify user to hubspot api if tracking enabled

knolleary:
Alternative to #3407 

Moves the logic to identify to the HS api into the Vue code that holds the user identity already.

I have had mixed results in testing locally - although doing so on http (not https) and localhost could well impact the results.

On one occassion HS properly identified my logged in user and created a new contact for the chat. On any subsequent attempt (incognito window etc), it didn't identify me.

Given the hubspot api is a blackbox, it's hard to know if this is having the intended result. I suggest we get this into production and see what we get.

### DIFF
--- a/frontend/src/services/product.js
+++ b/frontend/src/services/product.js
@@ -14,6 +14,11 @@ function identify (userId, set, setonce) {
             ...setonce
         })
     }
+    const _hsq = window._hsq = window._hsq || []
+    _hsq.push(['identify', {
+        id: userId,
+        email: set.email
+    }])
 }
 
 /**
--- a/frontend/src/store/account.js
+++ b/frontend/src/store/account.js
@@ -148,6 +148,13 @@ const actions = {
             const user = await userApi.getUser()
             state.commit('login', user)
 
+            if (window._ffhstc) {
+                const _hsq = window._hsq = window._hsq || []
+                _hsq.push(['identify', {
+                    email: user.email,
+                    id: user.id
+                }])
+            }
             // User is logged in
             if (router.currentRoute.value.name === 'VerifyEmail' && user.email_verified === false) {
                 // This page has `meta.requiresLogin = false` as it needs to be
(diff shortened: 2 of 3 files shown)


knolleary:
Note this also attempts to *un*identify the user when they logout - otherwise any existing HS chats the user had will be visible to the next user of the computer.

frontend/src/store/account.js
```diff
@@ -148,6 +148,13 @@ const actions = {
             const user = await userApi.getUser()
             state.commit('login', user)
 
+            if (window._ffhstc) {
```
Pezmc, line 151: We have a service that handles this for posthog and sentry, this identify command can live there too:

https://github.com/FlowFuse/flowfuse/blob/aadd6091dd905fb0ee1626a39e75874c5c14c69c/frontend/src/services/product.js#L9-L17

Pezmc (changes requested):

frontend/src/services/product.js
```diff
@@ -14,6 +14,11 @@ function identify (userId, set, setonce) {
             ...setonce
         })
     }
+    const _hsq = window._hsq = window._hsq || []
+    _hsq.push(['identify', {
+        id: userId,
+        email: set.email
+    }])
```
Pezmc, line 21: I might be missing something, but why not:

```suggestion
    if (window._hsq) {
        window._hsq.push(['identify', {
            id: userId,
            email: set.email
        }])
    }
```

Pezmc: Edit, oh perhaps hubspot loads deferred, and still checks window._hsq?

Pezmc (approved):

frontend/src/services/product.js
knolleary: This is how all of their examples are coded. I *think* its because they recommend using deferred loading of their module, so it is possible it hasn't been loaded when you want to use `window._hsq` - I assume their code checks to see if you've created _hsq yourself and then processes what's there before swapping it for their own object.

PR #3436, opened 2024-02-02
Generate database structure via migration

### LINKED ISSUE #3426: Scalability: Database Migration strategy
Part of #2782 

Our existing database initialisation and migration strategy is based on the assumption that a single process is going to run the logic. That won't work when we have two instances of the platform that start at the same time and both decide they need to apply the migrations.

In other applications, applying DB migrations is handled as an admin step prior to starting the new code. Or, at the other extreme, every code change is written to cope with old/new schemas with feature flags to control the behaviour.

I'm not looking to completely overhaul how we do this - want to keep it simple.

The plan is:

1. keep the existing strategy of the platform applying migrations on startup
2. modify the migration logic to obtain a table lock on `MetaVersions` before checking and applying migrations.

This locking logic will **only apply when running against postgres**. In other words, we do not plan to support horizontal scaling with sqlite as the backend.

If two instances start at the same time, one will get the lock. It can then check for and apply migrations, then release the lock. In my testing I see the second instance blocks and waits to get the lock, rather than throwing an error. It then finds there are no migrations left to apply and continues starting up.

This has revealed a secondary issue that we've known we would need to address at some point. Currently we use `sequelize.sync()` after registering our DB models. This allows sequelize to check all the tables are present and correct. For an upgrade, it *should* make no changes as the migrations will have done the work. For a clean install, that will create all the tables to match the model definitions. We know that causes a delay on startup - made worse by the fact we do it twice (once for core models and once for EE models). In a scaled-out environment, I am seeing that cause errors when the two instances both run the sync at the same time.

To fix it, we will create a new base-line migration that explicitly creates all tables (if they don't exist) to the current required state. For a clean install, it will skip previous migrations, but start at the new base-line and apply that plus any subsequent ones. For an upgrade, it will get applied as a normal migration, so it will need to detect the tables exist already and do no further work.

The hard part here is creating the base-line migration. We cannot dump a database schema as we need this to work for both sqlite and postgres. We're going to have to generate it via script and hand tweaking. The testing will be to verify the resulting db schema matches what we expect in both sqlite and postgres.

knolleary:
Closes #3426 

## Description

As detailed in #3426, to enable scalability of the app we need to change how we handle migrations and the initialisation of the database.

When running on Postgres, the database migration logic that is applied when the app starts is now wrapped in a transaction that holds a lock on the `MetaVersions` table. This ensures two concurrently started apps do not both try to apply migrations.

With that fixed, the next issue is that we use `sequelize.sync()` to initialise the database. We have seen this taking 10+s in staging. In a scaled-out environment, having two instances running sync at the same time causes errors.

To fix that, this PR introduces a new 'baseline' migration that, if no tables exist, will create the entire database structure as expected *today*.

For a new install, the migration logic will skip 'old' migrations (as it does today), but then apply the baseline migration and any ones that follow.

For an existing install, the baseline migration will no-op as the tables already exist.

Creating this baseline migration has been quite the task. I'm describing the process here for future reference, but also to document what testing I've done.

I wrote some code to generate the bulk of it based on our models. I then used the DDL of each table from an sqlite instance to check all of the generated fields and fill in anything missing - including indexes.

I then ran the migration to generate a new database (sqlite), and manually compared the constraints, indexes and DDL of each table against an existing database.

I then ran against postgres.

Finally I ensured our unit tests passed cleanly against both sqlite and postgres.

Whoever reviews this, please do a test of running this up with a new sqlite and/org pg database and verify some basic ff functionality.

### DIFF
--- a/test/unit/forge/db/models/TeamType_spec.js
+++ b/test/unit/forge/db/models/TeamType_spec.js
@@ -1,6 +1,9 @@
 const should = require('should') // eslint-disable-line
 const setup = require('../setup')
 
+let objectCount = 0
+const generateName = (root = 'object') => `${root}-${objectCount++}`
+
 describe('TeamType model', function () {
     let app
 
@@ -14,7 +17,7 @@ describe('TeamType model', function () {
     describe('getInstanceTypeProperty', async function () {
         it('Should read the (nested) property from an instances config when passed a hashid', async function () {
             const teamType = await app.db.models.TeamType.create({
-                name: 'Test Team Type',
+                name: generateName('Test Team Type'),
                 properties: {
                     instances: {
                         one: {
@@ -38,18 +41,16 @@ describe('TeamType model', function () {
             properties.instances[instanceType.hashid] = {
                 active: true
             }
-
             const teamType = await app.db.models.TeamType.create({
-                name: 'Test Team Type',
+                name: generateName('Test Team Type'),
                 properties
             })
-
             teamType.getInstanceTypeProperty(instanceType, 'active').should.equal(true)
         })
 
         it('Should return the default value if property is not found', async function () {
             const teamType = await app.db.models.TeamType.create({
-                name: 'Test Team Type',
+                name: generateName('Test Team Type'),
                 properties: {
                     instances: {
                         one: {
@@ -63,7 +64,7 @@ describe('TeamType model', function () {
 
         it('Should also return the default value if property is null', async function () {
             const teamType = await app.db.models.TeamType.create({
-                name: 'Test Team Type',
+                name: generateName('Test Team Type'),
                 properties: {
                     instances: {
                         one: {
--- a/forge/db/index.js
+++ b/forge/db/index.js
@@ -67,6 +67,7 @@ module.exports = fp(async function (app, _opts) {
         controllers,
         utils
     }
+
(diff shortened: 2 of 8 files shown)


test/unit/forge/db/models/TeamType_spec.js
```diff
@@ -1,6 +1,9 @@
 const should = require('should') // eslint-disable-line
 const setup = require('../setup')
 
+let objectCount = 0
+const generateName = (root = 'object') => `${root}-${objectCount++}`
```
knolleary, line 5: The `TeamType` sequelize model doesn't have 'unique: true' set on the name field. This means, for a clean install, the constraint doesn't get applied and the tests pass.

However, the migration that was added to introduce TeamType *does* set the unique constraint - so that is what we have in production.

I chose to match production with out baseline - so the constraint is applied. So I had to update the tests to use unique names below.

hardillb:
I've given everything but the 1000 line migration the once over and looks ok so far.

Pezmc (approved):
I've done a full pass of the migration file and `forge/db/migrations/index.js`, nothing stood out of the ordinary including a comparison with my local PostgreSQL tables.

If this works on temporary staging (create new), and real staging (no-op upgrade existing), I'd expect it to behave in production too!

PR #3446, opened 2024-02-05
Remove blueprints from instance duplication

hardillb:
## Description

When duplicating an instance don't show the Blueprint selector and don't pass a Blueprint id to the API

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/application/createInstance.vue
+++ b/frontend/src/pages/application/createInstance.vue
@@ -130,6 +130,7 @@ export default {
         createInstance (instanceDetails, copyParts) {
             const createPayload = { ...instanceDetails, applicationId: this.application.id }
             if (this.sourceInstance?.id) {
+                delete createPayload.flowBlueprintId
                 createPayload.sourceProject = {
                     id: this.sourceInstanceId,
                     options: { ...copyParts }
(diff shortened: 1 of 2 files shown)


frontend/src/pages/application/createInstance.vue
```diff
@@ -130,6 +130,7 @@ export default {
         createInstance (instanceDetails, copyParts) {
             const createPayload = { ...instanceDetails, applicationId: this.application.id }
             if (this.sourceInstance?.id) {
+                delete createPayload.flowBlueprintId
```
Pezmc, line 133: It feels like this shouldn't be there in the first place, but if this is the first spot to remove it! 👍

Pezmc (approved):

frontend/src/pages/application/createInstance.vue
hardillb: It looks like it's initialised to an empty string which then gets mapped to the default in the backend.

So it's either strip it here or add a test in the backend.

Pezmc:
You should also add something here that skips blueprint loading entirely if `isCopyProject`

https://github.com/FlowFuse/flowfuse/blob/fc628b20cbca84be5dd6e1aec58ac4ae3c130e75/frontend/src/pages/instance/components/InstanceForm.vue#L676

PR #3481, opened 2024-02-14
Implement device auto snapshot

Steve-Mcl:
## Description

Hooks into new device audit events and performs an auto snapshot

### Finer details

* Upon an audit event `flows.set`, spawn an auto snapshot action
* fifo clean up, max 10 auto snapshots
   * does NOT auto clean any auto snapshot that was found to be the target of a device, device group or pipeline stage so there can be more than 10
* Adds a setting option to disable (enabled by default)
  * This utilises the `DeviceSettings` table
* Adds teamtype flag to enable/disable the feature
* EE only

### Tests

`test/unit/forge/auditLog/device_spec.js`
```log
  Audit Log > Device
    ✔ Provides a logger for changing a settings of a device
```

`test/unit/forge/db/controllers/ProjectSnapshot_spec.js`
```log
  ProjectSnapshot controller
    device auto snapshot
      ✔ throws an error when deviceAutoSnapshot feature is not enabled
      ✔ throws an error when team type feature flag deviceAutoSnapshot is not enabled
      ✔ throws an error when device setting autoSnapshot is not enabled
      when deviceAutoSnapshot feature is enabled
        ✔ creates an autoSnapshot for a device following a 'full' deploy
        ✔ only keeps 10 autoSnapshots for a device (87ms)
        ✔ keeps 11 autoSnapshots for a device if one of them is assigned as target snapshot to another device (101ms)
```

`test/unit/forge/db/models/Device_spec.js`
```log
  Device model
    Settings hash
      ✔ is not updated when the device option autoSnapshot is changed
```

`frontend/src/pages/device/Snapshots/index.vue`
```log
  Logging API
    device instance audit logging
      When an audit event for flows.set is received
        ✔ Generates a snapshot for device when deploy type === full (58ms)
        ✔ Does not generates a snapshot for device if the deploy type is not one of full, flows or nodes (45ms)
```

### Screenshots / demo

image

chrome_UmKnugx4BD

image

## Related Issue(s)

3358

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [x] Concepts

## Labels

### DIFF
--- a/forge/db/controllers/ProjectSnapshot.js
+++ b/forge/db/controllers/ProjectSnapshot.js
@@ -82,22 +82,26 @@ const deviceAutoSnapshotUtils = {
         candidateIds = removeFromArray(candidateIds, inUseAsActive)
 
         // Check `DeviceGroups` table
-        const snapshotsInUseInDeviceGroups = await app.db.models.DeviceGroup.findAll({
-            where: {
-                targetSnapshotId: { [Op.in]: candidateIds }
-            }
-        })
-        const inGroupAsTarget = snapshotsInUseInDeviceGroups.map((group) => group.targetSnapshotId)
-        candidateIds = removeFromArray(candidateIds, inGroupAsTarget)
+        if (app.db.models.DeviceGroup) {
+            const snapshotsInUseInDeviceGroups = await app.db.models.DeviceGroup.findAll({
+                where: {
+                    targetSnapshotId: { [Op.in]: candidateIds }
+                }
+            })
+            const inGroupAsTarget = snapshotsInUseInDeviceGroups.map((group) => group.targetSnapshotId)
+            candidateIds = removeFromArray(candidateIds, inGroupAsTarget)
+        }
 
         // Check `PipelineStageDeviceGroups` table
-        const snapshotsInUseInPipelineStage = await app.db.models.PipelineStageDeviceGroup.findAll({
-            where: {
-                targetSnapshotId: { [Op.in]: candidateIds }
-            }
-        })
-        const inPipelineStageAsTarget = snapshotsInUseInPipelineStage.map((stage) => stage.targetSnapshotId)
-        candidateIds = removeFromArray(candidateIds, inPipelineStageAsTarget)
+        if (app.db.models.PipelineStageDeviceGroup) {
+            const snapshotsInUseInPipelineStage = await app.db.models.PipelineStageDeviceGroup.findAll({
+                where: {
+                    targetSnapshotId: { [Op.in]: candidateIds }
+                }
+            })
+            const inPipelineStageAsTarget = snapshotsInUseInPipelineStage.map((stage) => stage.targetSnapshotId)
+            candidateIds = removeFromArray(candidateIds, inPipelineStageAsTarget)
+        }
 
         return autoSnapshots.filter((snapshot) => candidateIds.includes(snapshot.id))
     },
@@ -110,6 +114,11 @@ const deviceAutoSnapshotUtils = {
         }
     },
     doAutoSnapshot: async function (app, device, deploymentType, options, meta) {
+        // sanitise options
+        options = options || {}
+        options.clean = options.clean ?? true // default to true
+        options.setAsTarget = options.setAsTarget ?? false // default to false
+
         // eslint-disable-next-line no-useless-catch
         try {
             // if not permitted, throw an error
--- a/forge/db/controllers/ProjectSnapshot.js
+++ b/forge/db/controllers/ProjectSnapshot.js
@@ -93,7 +93,8 @@ const deviceAutoSnapshotUtils = {
         }
 
         // Check `PipelineStageDeviceGroups` table
-        if (app.db.models.PipelineStageDeviceGroup) {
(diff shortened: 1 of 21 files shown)


forge/db/controllers/ProjectSnapshot.js
```diff
@@ -1,3 +1,185 @@
...
+     * Get all auto snapshots for a device
+     *
+     * NOTE: If a `limit` of 10 is provided and some of the snapshots are in use, the actual number of snapshots returned may be less than 10
+     * @param {Object} app - the forge application object
+     * @param {Object} device - a device (model) instance
+     * @param {boolean} [excludeInUse=true] - whether to exclude snapshots that are currently in use by a device, device group or pipeline stage device group
+     * @param {number} [limit=0] - the maximum number of snapshots to query in the database (0 means no limit)
+     */
+    getAutoSnapshots: async function (app, device, excludeInUse = true, limit = 0) {
+        // TODO: the snapshots table should really have a an indexed `type` column to distinguish between auto and manual snapshots
```
Pezmc, line 34: 👍

forge/db/controllers/ProjectSnapshot.js
```diff
@@ -1,3 +1,185 @@
...
+        if (snapshots.length > limit) {
+            const toDelete = snapshots.slice(0, snapshots.length - limit).map((snapshot) => snapshot.id)
+            await app.db.models.ProjectSnapshot.destroy({ where: { id: { [Op.in]: toDelete } } })
+        }
+    },
+    doAutoSnapshot: async function (app, device, deploymentType, options, meta) {
+        // sanitise options
+        options = options || {}
+        options.clean = options.clean ?? true // default to true
+        options.setAsTarget = options.setAsTarget ?? false // default to false
```
Pezmc, line 120: This could also be a destructured objected in the function params

```suggestion
    doAutoSnapshot: async function (app, device, deploymentType, options = {}, meta) {
        // sanitise options
        options.clean = options.clean ?? true // default to true
        options.setAsTarget = options.setAsTarget ?? false // default to false
```

forge/db/controllers/ProjectSnapshot.js
```diff
@@ -1,3 +1,185 @@
...
+
+            // 3. clean up older auto snapshots
+            if (options?.clean === true) {
+                await app.db.controllers.ProjectSnapshot.cleanupDeviceAutoSnapshots(device)
+            }
+
+            return snapShot
+        } catch (error) {
+            // TODO: device snapshot:  implement audit log
+            // await deviceAuditLogger.device.snapshot.created(request.session.User, error, request.device, null)
```
Pezmc, line 174: Another todo!

Pezmc (approved):
👍 Assuming the tests are tweaked to go green. There are some caveats to this approach, as Steve has already noted in the comments, but this is more than good enough for an MVP. Ideally it will be followed up with the addition of a database column.

forge/db/controllers/ProjectSnapshot.js
```diff
@@ -1,3 +1,185 @@
...
+                device.Application,
+                device,
+                user,
+                saneSnapshotOptions
+            )
+            snapShot.User = user
+
+            // 2. log the snapshot creation in audit log
+            // TODO: device snapshot: implement audit log
+            // await deviceAuditLogger.device.snapshot.created(request.session.User, null, request.device, snapShot)
```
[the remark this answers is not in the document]
Steve-Mcl: @Pezmc 

I was tempted to add it carte blanch but I resisted for 2 reasons:
1. not in scope (well not explicitly anyhow)
2. would generate a LOT of noise in the audit log

So I left it as a TODO / placeholder for possible future iteration

@MarianRaphael @joepavitt we could either:
1. remove the TODO / never add audit logging of an auto snapshot
2. keep the TODO and raise a follow up task
3. add the extra audit logging now before merging

Please advise.

Pezmc:
For those following along (cc @knolleary @joepavitt), the tests were failing as `PipelineStageDeviceGroup.findAll` was throwing `Error: ConnectionManager.getConnection was called after the connection manager was closed!`. 

Usually this is because a test ends before all the commands have finished running, caused e.g. by a missing await. 

However in this case, for a reason not fully explained yet `app.db.models.PipelineStageDeviceGroup` is defined. Despite the test instance being OS not EE. 

The failures only occur if an EE test, e.g. [Broker Auth API Auth ACL Project EE](https://github.com/FlowFuse/flowfuse/blob/3413ad284e91a76b2686403fb95841246e42cedf/test/unit/forge/comms/authRoutes_spec.js#L691) has run before this test.

Previously `cleanupAutoSnapshots`, checked if app.db.models.PipelineStageDeviceGroup was truthy before calling `.findAll` on it, the fix was to also check if the license was active (35c6af8). 

As to why the tests are allowing `app.db.models.PipelineStageDeviceGroup` to persist between tests. We did not get to the bottom of it. It is likely connected to the fact that all database tables now always exist even in OS.

Going forward, all checks that a EE model exists should also check if the app is also licensed in case:
 
 - License state has changed but model has/hasn't been reloaded (i.e. no restart)
 - The model is loaded when the app isn't licensed

PR #3507, opened 2024-02-21
Implement instance auto snapshots

### LINKED ISSUE #3355: Implementation - Automated Snapshot creation on Flow Deployment for Instances
## Description

This feature introduces an automated snapshot creation system triggered by the generation of an "Flow Deployed" audit log entry. Each snapshot will be automatically named with the prefix [auto], followed by the current date and time. This feature includes configurability for the number of auto snapshots retained; older snapshots beyond this limit will be automatically deleted, provided they are not currently in use as target snapshots.

## User Story

https://github.com/FlowFuse/flowfuse/issues/2423

## Scope

- Trigger Condition: Snapshot creation occurs upon the generation of an "Flow Deployed" audit log entry.

- Naming Convention: Snapshots will be named following the format [auto]Date+Time, incorporating the date and time of creation.

- The maximum number of auto snapshots retained is 10.

- Snapshot Retention: Snapshots currently designated as target snapshots will be exempt from automatic deletion.

- For Instances only

## Which customers would this be available to

Licensed Edition (EE)
Teams Tier

Steve-Mcl:
closes #3355

## Description

Hooks into instance audit events and performs an auto snapshot

### Details

* Upon an audit event `flows.set`, spawn an auto snapshot action
* fifo clean up, max 10 auto snapshots
   * does NOT auto clean any auto snapshot that was found to be the target of a device, device group or pipeline stage so there can be more than 10
* Adds teamtype flag to enable/disable the feature
* EE only

### Tests

`test/unit/forge/db/controllers/ProjectSnapshot_spec.js`
```log
  ProjectSnapshot controller
    createSnapshot (instance)
      auto snapshots
        ✔ throws an error when instanceAutoSnapshot feature is not enabled
        ✔ throws an error when team type feature flag instanceAutoSnapshot is not enabled
        with instanceAutoSnapshot feature enabled
          ✔ creates an autoSnapshot for a device following a 'full' deploy
          ✔ only keeps 10 autoSnapshots for an instance (192ms)
          ✔ keeps 11 autoSnapshots for an instance when one of them is assigned as target snapshot to a device (255ms)
          ✔ keeps 11 autoSnapshots for an instance when one of them is assigned as the instances target snapshot (238ms)
```

`test/unit/forge/db/models/Device_spec.js`
```log
  Device model
    Settings hash
      ✔ is not updated when the device option autoSnapshot is changed
```

`test/unit/forge/routes/logging/index_spec.js`
```log
  Logging API
    instance audit logging
      When an audit event for flows.set is received
        ✔ Generates a snapshot for instance when deploy type === full (95ms)
        ✔ Does not generates a snapshot for instance if the deploy type is not one of full, flows or nodes (74ms)
```

TODO:
- [x] documentation
- [ ] (as per feedback on question raised in issue) Raise a separate follow up issue to address the name matching situation (i.e. a DB field enum that specifies the snapshot type for both device and instance auto snapshots should be introduced.  this was deliberately NOT introduced in this PR to avoid having differences between the two features when they share the same table. )

## Related Issue(s)

#3355
 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/logging/index.js
+++ b/forge/routes/logging/index.js
@@ -78,6 +78,35 @@ module.exports = async function (app) {
         }
 
         response.status(200).send()
+
+        // perform an auto snapshot
+        if (event === 'flows.set' && ['full', 'flows', 'nodes'].includes(auditEvent.type)) {
+            if (!app.config.features.enabled('instanceAutoSnapshot')) {
+                return // device auto snapshot feature is not available
+            }
+
+            const teamType = await request.project.Team.getTeamType()
+            const instanceAutoSnapshotEnabledForTeam = teamType.getFeatureProperty('instanceAutoSnapshot', false)
+            if (!instanceAutoSnapshotEnabledForTeam) {
+                return // not enabled for team
+            }
+            const instanceAutoSnapshotEnabledForProject = true // FUTURE: await request.project.getSetting('autoSnapshot')
+            if (instanceAutoSnapshotEnabledForProject === true) {
+                setImmediate(async () => {
+                    // when after the response is sent & IO is done, perform the snapshot
+                    try {
+                        const meta = { user: request.session.User }
+                        const options = { clean: true, setAsTarget: false }
+                        const snapshot = await snapshotController.doInstanceAutoSnapshot(request.project, auditEvent.type, options, meta)
+                        if (!snapshot) {
+                            throw new Error('Auto snapshot was not successful')
+                        }
+                    } catch (error) {
+                        console.warn('Error occurred during auto snapshot', error)
+                    }
+                })
+            }
+        }
     })
 
     /**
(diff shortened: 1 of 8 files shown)


forge/routes/logging/index.js
```diff
@@ -78,6 +78,35 @@ module.exports = async function (app) {
...
+                    // when after the response is sent & IO is done, perform the snapshot
+                    try {
+                        const meta = { user: request.session.User }
+                        const options = { clean: true, setAsTarget: false }
+                        const snapshot = await snapshotController.doInstanceAutoSnapshot(request.project, auditEvent.type, options, meta)
+                        if (!snapshot) {
+                            throw new Error('Auto snapshot was not successful')
+                        }
+                    } catch (error) {
+                        console.warn('Error occurred during auto snapshot', error)
```
knolleary, line 105: Shouldn't be console.warn - needs to use `app.log.error`.

Noticed the device auto-snapshot path also does a console.warn - so I can't propose a change in this PR that fixes both up. Will manually push a fix.

knolleary (approved):
Approved pending a couple tidy ups I'll merge

Issue #3367, opened 2024-01-22
Scalability: Persistent Device Editor token


knolleary:
Part of #2782 and https://github.com/FlowFuse/device-agent/issues/222

Currently the device editor access token is stored in memory within the forge app that handles the request to enable the device editor.

This token needs to be stored in the db for two reasons:

1. allow a device editor to reconnect after a platform restart without having to toggle device mode on/off
2. to allow any instance of the platform to validate the ws connection, not just the one holding the token in memory

Issue #2514, opened 2023-07-21
Device Editor: change websocket handling for scalability


knolleary:
### Description

When the Device Editor is enabled the Device Agent creates a WebSocket connection back to the Forge App.

When a user access the Node-RED editor, the requests are sent to the Forge App (including its own WebSocket) and then tunnelled over the Device Agent's WebSocket.

If there are two instances of the Forge App, only one of them will hold the Device Agent WebSocket connection. If the requests to access the Node-RED editor are load balanced between the instances, only those that arrive at the 'connected' instance will be handled - the others will be dropped.

Even if we enable some form of 'sticky' session to ensure requests for a single browser session are all sent to one instance, we would need a way to direct that to the 'right' instance.

#### Possible solutions

1. Move the proxy to a separate service.
    - **Pros**
         - removes the load of the device editor from the main forge app.
    - **Cons**
         - Device Agent needs to be told where to connect to. All existing agents assume its the forge app - so we have a legacy support issue. Might be possible to tell the websocket to redirect... not sure.
         - Another service to maintain that will have all the same scalability challenges over session/websocket stickiness if the service were to be scaled horizontally.

2. Sticky sessions in nginx
    - **Cons**
        - Still need a way to get the requests to the 'right' instance where the Device Agent tunnel is connected.

knolleary:
With the current architecture, restarting the forge app interrupts ongoing editor sessions. As we move to a CI/CD approach, this could potentially happen more often and start impacting users.

@MarianRaphael We should prioritise this task, or investigate other potential mitigations to ensure any impact is minimal.

knolleary:
I've pulled this item under the epic #2782 which is a broader topic around scalability. Moving the tunnel handling to a separate service might *not* be the right solution, but this issue should be used to track identifying and implementing whatever is needed to address this particular scalability issue.

MarianRaphael:
Investigation issue: https://github.com/flowforge/flowforge/issues/2725

knolleary:
@hardillb @ppawlowski would like your input here on what we can do with sticky sessions in the ingress.

To summarise the problem - for the device editor, we use persistent websocket connections. One is from the Device to the platform, the other is from the user's browser to the platform. For the Device Editor to work, those two websockets need to land on the *same running copy* of the platform. Do we have any means to use some sort of hashing of some attribute of a request to direct requests to a consistent instance?

hardillb:
[...]

PR #3511, opened 2024-02-22
Improve scalability of device editor

knolleary:
This updates the Device Editor tunnel handling to make it scalable.

Closes:
 - #2514 
 - #3367 

# Summary of changes

 - Adds news fields to the Device model to store editor tunnel state:
    - `editorToken` - the access token to be used. Unlike other tokens, this has to be stored in the clear as it needs to be retrievable to give to both the device and frontend when someone tries to access the editor
    - `editorAffinity` - if set, this is the value of the `FFSESSION` affinity cookie the device has been given by the `ingress-nginx` load balancer. This allows us to set a matching cookie on any api request for this device's details.
    - `editorConnected` - whether the editor tunnel is currently connected or not

 - Adds a new `local` property to the device editor status object. This is a boolean flag that indicates if the device agent websocket is connected to the instance that is servicing the api request.

The key logic is all around the `editorAffinity` property and the `FFSESSION` cookie.

When a user enables the Device Editor, the device agent will connect to the platform and the ingress will give it a cookie to ensure affinity with a particular upstream platform instance. The Device Agent communicates the value of that cookie back to the platform (both in response to the MQTT command to enable the editor, *and* as part of its regular status checkin).

The load balancer is configured to set the path of the `FFSESSION` cookie to `/api/v1/device` (because we cannot do wildcard locations in the nginx config).

Any api request to `/api/v1/device/<hashid>` will get its own `FFSESSION` cookie set to match that of the device (if the device is connected). This means subsequent rquests from the client should hit the *same* instance the device is connected to.

### Older Device Agent support

Older device agents do not know about the affinity cookie - they will connect without the cookie and end up wherever the load balancer puts them.

To try to cater for that, the `local` property on the device editor status object is used. If the frontend wants to connect to the device it can check the `local` property. If its false, it knows it is not talking to the instance that holds the connection. But in making that request, the server will also notice the same thing and attempt to set/clear the FFSESSION cookie. The frontend can then retry - hopefully to get load balanced to a different instance. We optimistically hope it will eventually get load-balanced to the right instance (it tries 10 times before stopping.. the user can retry themselves).

### Testing

This has been tested in our scaling environment using `ingress-nginx` with 2 instances of the forge app, with both older device agents and the ones with the improved cookie handling.

My main goal is two-fold; keep things working in the non-scaled case (so these changes to ship asap), and have confidence they will work when we scale the platform.

### DIFF
--- a/forge/routes/api/device.js
+++ b/forge/routes/api/device.js
@@ -101,11 +101,9 @@ module.exports = async function (app) {
         }
     }, async (request, reply) => {
         const result = app.db.views.Device.device(request.device)
-        app.log.info(`NOL: Clearing top-level FFSESSION cookie for device ${request.device.hashid}`)
         // ingress-nginx will set a cookie on /api/v1/devices - which will cannot allow to override
         // that of the device-specific cookie. So clear it out.
         if (request.cookies.FFSESSION) {
-            app.log.info('NOL: Clearing FFSESSION cookie on generic path')
             reply.clearCookie('FFSESSION', { path: '/api/v1/devices/' })
         }
         if (result.editor && result.editor.enabled) {
@@ -118,8 +116,6 @@ module.exports = async function (app) {
                         // In this case, we *know* the cookie used by the device, so
                         // use that.
 
-                        // TODO: remove before shipping
-                        app.log.info(`NOL: Setting FFSESSION cookies to known affinity for device ${request.device.hashid} : ${request.device.editorAffinity}`)
                         reply.setCookie('FFSESSION', request.device.editorAffinity, {
                             httpOnly: true,
                             path: request.url,
@@ -129,8 +125,6 @@ module.exports = async function (app) {
                             encode: s => s
                         })
                     } else {
-                        // TODO: remove before shipping
-                        app.log.info(`NOL: Clearing FFSESSION cookies as we don't have affinity for device ${request.device.hashid}`)
                         // An older device agent doesn't tell us about its affinity cookie
                         // So the best we can do is clear the session cookies and have
                         // the load balancer pick another route to try
--- a/forge/routes/api/device.js
+++ b/forge/routes/api/device.js
@@ -115,7 +115,11 @@ module.exports = async function (app) {
 
                     reply.setCookie('FFSESSION', request.device.editorAffinity, {
                         httpOnly: true,
-                        path: request.url
+                        path: request.url,
+                        // By default, it will uriEncode the value, which changes | to %7C
+                        // We don't want that change to happen, so provide a cleaner
+                        // encode function
+                        encode: s => s
                     })
                 } else {
                     // TODO: remove before shipping
--- a/forge/routes/api/device.js
+++ b/forge/routes/api/device.js
@@ -101,6 +101,13 @@ module.exports = async function (app) {
         }
     }, async (request, reply) => {
         const result = app.db.views.Device.device(request.device)
+        app.log.info(`NOL: Clearing top-level FFSESSION cookie for device ${request.device.hashid}`)
+        // ingress-nginx will set a cookie on /api/v1/devices - which will cannot allow to override
+        // that of the device-specific cookie. So clear it out.
+        if (request.cookies.FFSESSION) {
+            app.log.info('NOL: Clearing FFSESSION cookie on generic path')
+            reply.clearCookie('FFSESSION', { path: '/api/v1/devices' })
+        }
         if (result.editor && result.editor.enabled) {
             if (result.editor.connected && !result.editor.local) {
                 // This device has the editor enabled, is connected, but not to
@@ -112,7 +119,6 @@ module.exports = async function (app) {
 
                     // TODO: remove before shipping
                     app.log.info(`NOL: Setting FFSESSION cookies to known affinity for device ${request.device.hashid} : ${request.device.editorAffinity}`)
-
                     reply.setCookie('FFSESSION', request.device.editorAffinity, {
                         httpOnly: true,
                         path: request.url,
@@ -130,7 +136,6 @@ module.exports = async function (app) {
                     // We have to clear both the top level cookie and the one for
                     // this specific device.
                     reply.clearCookie('FFSESSION', { path: request.url })
-                    reply.clearCookie('FFSESSION', { path: '/api/v1/devices' })
                 }
             }
         }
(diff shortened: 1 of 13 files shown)


hardillb:
I have built versions of the containers and stored them in my container repo, we should be able to pull them in the morning.

Will review code first thing

forge/routes/api/device.js
```diff
@@ -100,7 +100,61 @@ module.exports = async function (app) {
             }
         }
     }, async (request, reply) => {
-        reply.send(app.db.views.Device.device(request.device))
+        const result = app.db.views.Device.device(request.device)
+        app.log.info(`NOL: Clearing top-level FFSESSION cookie for device ${request.device.hashid}`)
```
hardillb, line 104:
```suggestion
```

forge/routes/api/device.js
```diff
@@ -100,7 +100,61 @@ module.exports = async function (app) {
...
+        // ingress-nginx will set a cookie on /api/v1/devices - which will cannot allow to override
+        // that of the device-specific cookie. So clear it out.
+        if (request.cookies.FFSESSION) {
+            app.log.info('NOL: Clearing FFSESSION cookie on generic path')
```
hardillb, line 108:
```suggestion
```

forge/routes/api/device.js
```diff
@@ -100,7 +100,61 @@ module.exports = async function (app) {
...
+                if (!result.editor.local) {
+                    // This device has the editor enabled, is connected, but not to
+                    // this local platform instance. We need to clear the session
+                    // cookies so the client gets routed to another platform instance
+                    if (request.device.editorAffinity) {
+                        // In this case, we *know* the cookie used by the device, so
+                        // use that.
+
+                        // TODO: remove before shipping
+                        app.log.info(`NOL: Setting FFSESSION cookies to known affinity for device ${request.device.hashid} : ${request.device.editorAffinity}`)
```
hardillb, line 122:
```suggestion
```

forge/routes/api/device.js
```diff
@@ -100,7 +100,61 @@ module.exports = async function (app) {
...
+                            httpOnly: true,
+                            path: request.url,
+                            // By default, it will uriEncode the value, which changes | to %7C
+                            // We don't want that change to happen, so provide a cleaner
+                            // encode function
+                            encode: s => s
+                        })
+                    } else {
+                        // TODO: remove before shipping
+                        app.log.info(`NOL: Clearing FFSESSION cookies as we don't have affinity for device ${request.device.hashid}`)
```
hardillb, line 133:
```suggestion
```

frontend/src/pages/device/index.vue
```diff
@@ -314,37 +310,63 @@ export default {
...
-                } else {
-                    this.openingTunnel = true
-                    this.$refs.dialog.show()
-                    try {
-                        // * Enable Device Editor (Step 1) - (browser->frontendApi) User clicks button to "Enable Editor"
-                        const result = await deviceApi.enableEditorTunnel(this.device.id)
-                        if (result.affinity) {
-                            document.cookie = `FFSESSION=${result.affinity}`
+            try {
+                console.warn(`openTunnel launchEditor:${launchEditor} ${JSON.stringify(this.device.editor)}`)
```
hardillb, line 314:
```suggestion
```

frontend/src/pages/device/index.vue
```diff
@@ -314,37 +310,63 @@ export default {
...
+                if (this.device.status === 'running') {
+                    if (this.device.editor?.enabled && this.device.editor?.connected && this.device.editor?.local) {
+                        console.warn(' - open editor')
```
hardillb, line 317:
```suggestion
```

frontend/src/pages/device/index.vue
```diff
@@ -314,37 +310,63 @@ export default {
...
+                        this.openEditor()
+                    } else {
+                        console.warn(' - show dialog')
```
hardillb, line 320:
```suggestion
```

frontend/src/pages/device/index.vue
```diff
@@ -314,37 +310,63 @@ export default {
...
+                        this.openingTunnel = true
+                        this.$refs.dialog.show()
+
+                        // Polls the tunnel status until we see it connected to the
+                        // 'local' platform instance - will give up after 10 attempts
+                        const pollTunnelStatus = (attempt = 0, timeout = 500) => {
+                            console.warn(' - pollTunnelStatus', attempt, timeout)
```
hardillb, line 327:
```suggestion
```

frontend/src/pages/device/index.vue
```diff
@@ -314,37 +310,63 @@ export default {
...
+                            if (attempt < 10) {
+                                this.openTunnelTimeout = setTimeout(async () => {
+                                    await this.loadDevice()
+                                    console.warn(` - poll result ${JSON.stringify(this.device.editor)}`)
```
hardillb, line 331:
```suggestion
```

frontend/src/pages/device/index.vue
```diff
@@ -314,37 +310,63 @@ export default {
...
+                                    if (this.device.editor?.enabled && this.device.editor?.connected) {
+                                        if (this.device.editor?.local) {
+                                            if (launchEditor) {
+                                                console.warn(' - open editor')
```
hardillb, line 335:
```suggestion
```

frontend/src/pages/device/index.vue
```diff
@@ -314,37 +310,63 @@ export default {
...
                         }
-                        this.updateTunnelStatus(result)
-                        setTimeout(() => {
-                            this.loadDevice()
-                            if (launchEditor && this.device.editor?.enabled && this.device.editor?.connected) {
-                                this.openEditor()
+
+                        try {
+                            if (!this.device.editor?.enabled || !this.device.editor?.connected) {
+                                console.warn(' - enabling editor tunnel')
```
hardillb, line 348:
```suggestion
```

hardillb (changes requested):
First pass looks good

I've flagged all the console.log/console.warns

knolleary:
Unit tests passing. Final task is to tidy up the Device Editor buttons on the new home screen. Choice is to either get the buttons working, or hide them for now.

Will time-box an hour tomorrow on getting the buttons working, otherwise will hide them and move forward.

knolleary:
This is now ready for full review.

hardillb (approved):
I've given it another read through and it looks good.

knolleary:
Okay - I think this is good to merge. Will do further validation on staging as well.

Issue #3488, opened 2024-02-15
Instance Protection Mode


MarianRaphael:
## User Story

As a FlowFuse Team Owner,
I want the ability to restrict modifications to production instances exclusively to Owners,
so that changes to any protected/productive instance require authorization from an Owner, ensuring controlled and secure updates.

## Scope

Introduce a Protected Mode setting within Instance Settings to safeguard production instances.

- Implement a button/switch within Instance Settings to activate or deactivate Protected Mode.

**Functionality in Protected Mode:**

- Editor Restrictions: The instance editor can only be accessed in viewer mode as an owner to prevent unauthorized modifications. Member and viewer team members still get viewer access.
- DevOps Pipeline Execution Control: Only Owners can initiate and execute pipelines targeting a protected instance, ensuring that critical environments are modified under strict supervision.

### Which customers would this be available to

Team + Enterprise Tiers (EE)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate
```[tasklist]
### Follow-up
- [ ] https://github.com/FlowFuse/flowfuse/issues/3139
```

MarianRaphael:
Customer request: https://github.com/FlowFuse/flowfuse/issues/3093#issuecomment-1830293431

hardillb:
I think this needs to be broken into 2 separate items

- viewer only editor access for owner, all others blocked
- Dev Pipeline work

hardillb:
@MarianRaphael I need an answer to the ^ question or I'll be blocked today.

Thanks.

MarianRaphael:
member and viewer team members still get viewer access

robmarcer:
This is for https://app-eu1.hubspot.com/contacts/26586079/record/0-1/3228251

Steve-Mcl:
Checked NR editor access is restricted. 
Checked session token is revoked when protection is toggled.
Checked owner can still operate pipeline

Last thing to check is non owner cannot operate pipeline 
Member Invite sent to Ben for "Steve on Staging" team, application `funny-crested-lark-6873`

Steve-Mcl:
> Last thing to check is non owner cannot operate pipeline
> Member Invite sent to Ben for "Steve on Staging" team, application funny-crested-lark-6873

@hardillb verified

PR #3520, opened 2024-02-27
Instance Protection 

hardillb:
fixes https://github.com/FlowFuse/flowfuse/issues/3488
## Description

Adds the concept of Protected Instances.

- These can only be deployed to via a pipeline and then only by the Team Owner.
- Access to the Node-RED Editor is read only for all Team memebers including the Owner.

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/3488

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
Instruction to enable this:

 - Team Types > Select Team > Features > Toggle ON "Protected Instances"

joepavitt:
We would benefit from Cypress tests here too, sorry to be the pain...

joepavitt (changes requested):
Functionally, seems to do the job fine, I have a few pain points around disabling of elements without feedback to the user on _why_ that's happening, but will open a separate issue for that, as that's an app-wide problem, not just here.

We are missing Cypress tests though, which should be added to the `pipelines.spec.js` test suite at least. These things csan be fragile, and given it's permissions related, want to make sure it's stable.

Steve-Mcl (changes requested):
As discussed on Slack: If the editor has been opened previously, then you switch on the protection & relaunch the editor, you still have write access to the flows.  

Additionally, the Pr has no docs update - are the docs update in a separate PR or is there a follow up issue/task?

hardillb:
@joepavitt I've started on some cypress tests, but this is going to be a pain, it needs to run things as 2 separate users

1. Team Owner to set the protected mode
2. Team Member to not be allowed to run the pipeline

joepavitt:
We have the cy.login command wrapper, possibly even a cy.logout wrapper too? 

The login wrapper atm I think runs in the "beforeAll" script for each test suite.

Perfectly plausible to do what you need, just can't remember what's in place already. Will confirm at desk tomorrow

joepavitt:
Checkout: https://github.com/FlowFuse/flowfuse/blob/main/test%2Fe2e%2Ffrontend%2Fcypress%2Fsupport%2Fcommands.js

That details the commands we've added to help

hardillb:
OK, got a mostly working test, just don't know how to check for the disabled play button on the stage.

Steve-Mcl:
> just don't know how to check for the disabled play button on the stage

It may depend on the button type but other parts of the 3e3 tests use `cy.get('button.ff-btn.ff-btn--danger').should('be.disabled')`

hardillb:
@Steve-Mcl It should now log out any active NR instance (only to be discovered on a deploy) and there is some "docs"...

@joepavitt Might need 5mins with you tomorrow to add a little bit more to the Cypress tests

joepavitt:
`should('be.disabled')` is the line you're looking for to check a button cannot be clicked.

Steve-Mcl:
> > just don't know how to check for the disabled play button on the stage
> 
> It may depend on the button type but other parts of the e2e tests use `cy.get('some-selector').should('be.disabled')`

Did this help Ben?

Is this ready for review now or is there more tests you wish to get in first?

joepavitt (approved):

Steve-Mcl (commented):
@hardillb in testing, I disabled the "Protected Instances" option in team types and the protection remains on for those instances that were set protected. The badge remains visible and the protection is still in place (flows cannot be edited).  While I would not call it not a bug (more a side-effect of me leaving instances in protected mode when disabling the capability), should we not also actually disable the feature (i.e. permit editor access / permit pipeline deploys / hide the "protected" badge)?

Happy to approve as is but perhaps a follow up task should be added?

This may be more for @MarianRaphael to answer?

Steve-Mcl:
> as we are not likely to enable then immediate disabled this feature

we "no", people, "likely"

Will approve and merge having raised the point for consideration.

@MarianRaphael please consider this point about disabling the option

Steve-Mcl (approved):

PR #3527, opened 2024-02-29
Add billing status check when creating device

knolleary:
Fixes https://github.com/FlowFuse/security/issues/75

## Description

This adds a check that a team's subscription is in a valid state when trying to create a device.

This mirrors the approach already taken for creating instances; where a `checkDeviceCreateAllowed` function is added to the `Team` model - and overloaded when billing is enabled.


hardillb (approved):

knolleary:
Test fix pushed

Issue #2964, opened 2023-10-17
FlowFuse PAT for Node-RED Http Endpoints


MarianRaphael:
### Description

As a FlowFuse User
I want to use the FlowFuse access tokens (PAT) to access my Node-RED HTTP endpoints, which are secured with "FlowFuse User Authentication"
So that I can securely access my HTTP endpoints even with SSO enabled.

### Which customers would this be availble to

Licensed Edition (EE)
Team

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate
```[tasklist]
### Follow-up
- [ ] https://github.com/FlowFuse/flowfuse/issues/2703
```

knolleary:
Increasing the sizing as this is potentially quick tricky to get right.

robmarcer:
An alternative  would be to allow a team to CRUD a list of tokens (seems like a pretty standard feature for various other platforms I use) which can then be used to protect their APIs. It would need to be on a per-instance basis. I think if all things are equal, putting it in HTTP Node Security would be a slightly better solution than using the FF PATs.

https://d.pr/i/rkp9Lw

robmarcer:
We have a customer who has specifically requested the ability to issue tokens to protect APIs.

knolleary:
@robmarcer if there's a requirement for per-instance tokens, then doing this via our existing access tokens wouldn't be appropriate (because they are tied to the user). So you are right to say having some instance-specific token capability would be needed.

There is also the issue that we currently don't have the ability to reduce the scope of a token - so you effectively have to provide full access to your FF account. We've already had feedback on that - see #2703 for enabling reduced-scope tokens.

To summarise:

 - Option 1: Allow existing User PATs to secure HTTP endpoints
    - Pros: we already have PATs
    - Cons: they are user-scoped, so would apply to *all* instances by default
    - Cons: leaking of the token gives *full* access to the whole FF user account, not just the individual instance.
    - Mitigation: provide a way to reduce scope of a token - potentially only specific instances. I would be super uncomfortable shipping this without the ability to reduce the token scope first.

 - Option 2: Allow user to create 'instance' tokens to secure HTTP endpoints
     - Pros: specific to an instance
     - Cons: another set of tokens to manage (although we have all the bits in place to do it)

MarianRaphael:
@robmarcer I want to better understand the customer requirements. Is it about a general way to secure the HTTP endpoints, or is a more granular approach (per instance) requested?

[...]

PR #3535, opened 2024-03-01
Add Bearer token support to FF Auth

hardillb:
part of https://github.com/FlowFuse/flowfuse/issues/2964
## Description

Allow HTTP Bearer tokens for Node-RED HTTP endpoints when using FF auth

image

requires https://github.com/FlowFuse/nr-launcher/pull/212

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/2964

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/pages/instance/Settings/dialogs/TokenDialog.vue
```diff
@@ -0,0 +1,142 @@
...
+    <ff-dialog ref="dialog" data-el="add-token-dialog" :header="(token ? 'Edit' : 'Add') + ' Token'" :confirm-label="token ? 'Save' : 'Create'" :disablePrimary="disableConfirm" @confirm="confirm()">
+        <template #default>
+            <form class="space-y-4" @submit.prevent>
+                <FormRow v-model="input.name" data-form="token-name" :disabled="edit">
+                    Token name
+                </FormRow>
+                <ff-checkbox v-model="input.expires" data-form="expiry-toggle" label="Add Expiry Date" />
+                <FormRow v-model="input.expiresAt" data-form="token-expiry" type="date" :disabled="!input.expires">
+                    Expires
+                    <!-- <template v-slot:description>Expires</template> -->
```
Steve-Mcl, line 11:
```suggestion
```

Steve-Mcl (commented):
Tested locally in conjunction with current and associated nr-launcher branch. Working as expected however there are a few minor points that might want to be addressed:

1. The token creation dialog permits a user to chose a date in the past. I have verified the token is not useable but ideally we would add a `min` value on the input element to inhibit this. Alternatively, client side logic that toasts a message?.
2. A token created for todays date does not work. Is that expected? A user may think they are creating a token for the remainder of the day (I did) but looking at the database, I can see the expire time is set as `yyyy/mm/dd 00:00:00` (i.e. this morning 00:00am)
3. There is no update to the docs for this. A short nod to this feature in `docs/user/instance-settings.md#Security` would be enough IMHO. 
   * Nice to have: Include a `cURL` example. Ideally this would by a copyable/usable code div on the token creation dialog, but at minimum a code example in the docs so the users know not to base64 encode the token for example?

hardillb:
>  1. The token creation dialog permits a user to chose a date in the past. I have verified the token is not useable but ideally we would add a min value on the input element to inhibit this. Alternatively, client side logic that toasts a message?

The token will be removed as soon as it is tested after the expiry date.

If you know how to add a min value to a the FF-Form-Row to make it add `min` attribute please share:

```
<FormRow v-model="input.expiresAt" data-form="token-expiry" type="date" min="{ new Date().toJSON().split('T')[0] }" :disabled="!input.expires">
```

doesn't work
.
  > 2. A token created for todays date does not work. Is that expected? A user may think they are creating a token for the remainder of the day (I did) but looking at the database, I can see the expire time is set as yyyy/mm/dd 00:00:00 (i.e. this morning 00:00am)

Tokens are valid until the expiry date, this is working as expected

   > 3. There is no update to the docs for this. A short nod to this feature in docs/user/instance-settings.md#Security would be enough IMHO.
        Nice to have: Include a cURL example. Ideally this would by a copyable/usable code div on the token creation dialog, but at minimum a code example in the docs so the users know not to base64 encode the token for example?

Docs I can add.

Steve-Mcl:
@hardlib, I added date validation in commit [5d23583](https://github.com/FlowFuse/flowfuse/pull/3535/commits/5d235830b182fed6823cd912ac8c3cbc663d8704)


chrome_xsnDtj1Y2a

PR #3537, opened 2024-03-03
fix: allow mixed-case usernames in backend validation

hardillb:
## Description

Backend username validation enforces lowercase only, front end is case insensitive, but the error message if tripped is for lowercase only.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/auth/index.js
+++ b/forge/routes/auth/index.js
@@ -389,7 +389,7 @@ async function init (app, opts) {
             return
         }
 
-        if (/^(admin|root)$/.test(request.body.username) || !/^[a-z0-9-_]+$/.test(request.body.username)) {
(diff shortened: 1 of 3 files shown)


knolleary:
We currently have 1831 users with mixed-case usernames - it is something we have always allowed. Rather than stop allowing it we should update the backend validation added in #3514 to allow for it.

knolleary (changes requested):
As per previous comment - rather than block this in the UI, we should update the backend validation to allow mixed case usernames.

knolleary (approved):

knolleary:
I've just thrown in a couple tests for mixed-case usernames.

hardillb:
Tests have passed, merging

PR #3562, opened 2024-03-12
Update licensing to combine device and instance limits

### LINKED ISSUE #3533: Update license handling to combine instance/device counts as one
Part of #3477 

**Objective**: Simplify the license generation process by removing the distinction between Devices and Instances.

**New Process**: Generate licenses based on the total number of Node-RED Runtimes. This total will include all Hosted Instances and Devices on the platform. Customers will have the flexibility to allocate these runtimes across their devices and hosted instances as they see fit, without any predefined restrictions.

knolleary:
Closes #3533
## Description

With this change, newly generated licenses will have a single instance/device count - rather than two separate limits.

It also reduces the limits of the unlicensed code to 5 users, 5 teams and 5 devices.

Existing licenses will continue to be honoured with their separate instance and device counts.

### DIFF
--- a/forge/db/models/Device.js
+++ b/forge/db/models/Device.js
@@ -66,9 +66,8 @@ module.exports = {
                 // if the product is licensed, we permit overage
                 const isLicensed = app.license.active()
                 if (isLicensed !== true) {
-                    const deviceLimit = app.license.get('devices')
-                    const deviceCount = await M.Device.count()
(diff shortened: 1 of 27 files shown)


knolleary:
A lot of the changes in the tests were because they need to create > 5 of a particular resource - which is now above the default limit.

hardillb (approved):

PR #3589, opened 2024-03-18
Validate instance property on provisioning token api

knolleary:
Fixes https://github.com/FlowFuse/security/issues/79


Steve-Mcl:
approved pending tests

Issue #3581, opened 2024-03-15
Allow "debugMaxLength" to be configured


freefos:
### Description

As a FlowFuse admin/developer
I would like to be able to set the value on the "debugMaxLength"
This is really easy when working with larger datasets.

### Scope 

- Add debugMaxLength to Instance and Device settings

### Which customers would this be available to

Team/Enterprise Tier (EE)

### Have you provided an initial effort estimate for this issue?

I am not a FlowFuse team member

MarianRaphael:
Updated the Tier

Issue #3577, opened 2024-03-14
Allow HTTP max payload size to be configured


hardillb:
### Description

**As a** FlowFuse Admin 
**I would like to** be able to change the default 5mb HTTP payload size
**So that** I can support larger data uploads

Currently Instances have the default 5mb max payload size.

### Which customers would this be available to

Team/Enterprise Tier (EE)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Requested by: https://app-eu1.hubspot.com/contacts/26586079/record/0-3/10928370628/

zackwasli:
Requested by: https://app-eu1.hubspot.com/contacts/26586079/record/0-3/10928370628/

MarianRaphael:
Updated the Tier

Steve-Mcl:
verify criteria:

### Using a known flow size (e.g. 5000 bytes)

- [ ] Set limit `apiMaxLength` to 6kb, save, restart instance
- [ ] deploy a simple change - should succeed

<br>

- [ ] Set limit `apiMaxLength` to 4kb, save, restart instance
- [ ] deploy a simple change - should fail

### Using an inject with a string of 500 characters connected to a debug

```
[{"id":"0cdd246f428efea3","type":"inject","z":"3642c7ee286f9c17","name":"512 chars","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"FlowFuse stands out as a remarkable orchestration platform tailored specifically for Node-RED. Its intuitive interface coupled with robust features empowers users to seamlessly automate workflows, streamline processes, and enhance productivity. With FlowFuse, managing Node-RED flows becomes a breeze, offering unparalleled flexibility and scalability. Whether orchestrating complex sequences or simplifying routine tasks, FlowFuse proves indispensable for harnessing the full potential of Node-RED. (512 chars)","payloadType":"str","x":250,"y":1060,"wires":[["e633a8ebd6c3b5ab"]]},{"id":"e633a8ebd6c3b5ab","type":"debug","z":"3642c7ee286f9c17","name":"debug 5","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","statusVal":"","statusType":"auto","x":430,"y":1060,"wires":[]}]
```

- [ ] Set limit `debugMaxLenth` to 400, save, restart instance
- [ ] inject should limit debug output

<br>

[...]

PR #3595, opened 2024-03-18
Add apiMaxLength & debugMaxLength settings

hardillb:
part of #3577 and #3581

## Description

This adds the option to configure the apiMaxLength and debugMaxLength Node-RED settings from the template.

This only available to Team/Enterprise licensed instances

Requires "API/Debug Length Limits" to be enabled for team types.

Has a paired nr-launcher PR https://github.com/FlowFuse/nr-launcher/pull/217

image

image

Staging/Prod config change Issue: https://github.com/FlowFuse/CloudProject/issues/377

## Related Issue(s)

#3577 and #3581

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/controllers/ProjectTemplate.js
+++ b/forge/db/controllers/ProjectTemplate.js
@@ -204,7 +204,7 @@ module.exports = {
             }
         }
         if (result.debugMaxLenth) {
-            if (!Number.isInteger(result.debugMaxLength)) {
+            if (!Number.isInteger(result.debugMaxLength) || result.debugMaxLength < 0) {
                 throw new Error('Invalid settings.debugMaxLength')
             }
         }
--- a/forge/db/controllers/ProjectTemplate.js
+++ b/forge/db/controllers/ProjectTemplate.js
@@ -198,6 +198,16 @@ module.exports = {
         if (typeof result.httpNodeAuth?.pass === 'string' && result.httpNodeAuth.pass.length > 0) {
             result.httpNodeAuth.pass = hash(result.httpNodeAuth.pass)
         }
+        if (result.apiMaxLength) {
+            if (!/^[0-9]+[km]?$/.test(result.apiMaxLength)) {
+                throw new Error('Invalid settings.apiMaxLength')
+            }
+        }
+        if (result.debugMaxLenth) {
+            if (!Number.isInteger(result.debugMaxLength)) {
+                throw new Error('Invalid settings.debugMaxLength')
+            }
+        }
         return result
     },
 
--- a/frontend/src/pages/instance/Settings/Editor.vue
+++ b/frontend/src/pages/instance/Settings/Editor.vue
@@ -1,6 +1,6 @@
 <template>
     <form class="space-y-6">
-        <TemplateSettingsEditor v-model="editable" :editTemplate="false" :team="team" />
+        <TemplateSettingsEditor v-model="editable" :editTemplate="false" :team="team" :instance="project" />
         <div class="space-x-4 whitespace-nowrap">
             <ff-button size="small" :disabled="!unsavedChanges" @click="saveSettings()">Save settings</ff-button>
         </div>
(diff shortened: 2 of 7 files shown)


frontend/src/pages/instance/Settings/Editor.vue
```diff
@@ -1,6 +1,6 @@
 <template>
     <form class="space-y-6">
-        <TemplateSettingsEditor v-model="editable" :editTemplate="false" />
+        <TemplateSettingsEditor v-model="editable" :editTemplate="false" :team="team" />
```
Steve-Mcl, line 3: Ben, is there a commit missing from this PR? 

`TemplateSettingsEditor` does not have a `team` prop.

Steve-Mcl: never mind, i am blind

hardillb:
PostgreSQL Test failure is GH/npm network thing, will re-run when able to.

hardillb:
Hold off on this. It needs a nr-launcher version test

forge/db/controllers/ProjectTemplate.js
```diff
@@ -198,6 +198,16 @@ module.exports = {
...
         }
+        if (result.apiMaxLength) {
+            if (!/^[0-9]+[km]?$/.test(result.apiMaxLength)) {
+                throw new Error('Invalid settings.apiMaxLength')
+            }
+        }
+        if (result.debugMaxLenth) {
+            if (!Number.isInteger(result.debugMaxLength)) {
+                throw new Error('Invalid settings.debugMaxLength')
+            }
```
Steve-Mcl, line 209:
```suggestion
            if (!Number.isInteger(result.debugMaxLength) || result.debugMaxLength < 0) {
                throw new Error('Invalid settings.debugMaxLength')
            }
```

Should prevent invalid negative numbers at API level

forge/db/controllers/ProjectTemplate.js
```diff
@@ -198,6 +198,16 @@ module.exports = {
         if (typeof result.httpNodeAuth?.pass === 'string' && result.httpNodeAuth.pass.length > 0) {
             result.httpNodeAuth.pass = hash(result.httpNodeAuth.pass)
         }
+        if (result.apiMaxLength) {
+            if (!/^[0-9]+[km]?$/.test(result.apiMaxLength)) {
```
Steve-Mcl, line 202: Should we consider an upper limit?

forge/db/controllers/ProjectTemplate.js
Steve-Mcl: Should we not consider an upper limit? greater than `10000` is pushing it but allowing unlimited is possibly a good thing

Steve-Mcl (commented):
1 suggestion and a couple of queries. 

Also, I note that if the admin switches the feature flag "API/Debug Length Limits" off, existing limits will continue to persist (but hidden) - should we condition inclusion of these 2 settings based on the feature flag setting?

forge/db/controllers/ProjectTemplate.js
hardillb: We can't set a sensible upper limit, we don't know how much memory the NR instance has or what might get dropped by any reverse proxy

forge/db/controllers/ProjectTemplate.js
hardillb: This is a performance thing, if a user cripples themselves it's on them.

hardillb:
>Also, I note that if the admin switches the feature flag "API/Debug Length Limits" off, existing limits will continue to persist (but hidden) - should we condition inclusion of these 2 settings based on the feature flag setting?

Won't that add a bunch of DB queries as part of getting the settings? e.g. need to look up the team -> teamType to check the properties each time the project settings are requested.

Steve-Mcl:
> > Also, I note that if the admin switches the feature flag "API/Debug Length Limits" off, existing limits will continue to persist (but hidden) - should we condition inclusion of these 2 settings based on the feature flag setting?
> 
> Won't that add a bunch of DB queries as part of getting the settings? e.g. need to look up the team -> teamType to check the properties each time the project settings are requested.

Fair point. Not certain of the _cost_ but we can always come back to this if deemed necessary.

Approving as is.

Steve-Mcl (approved):

Issue #3612, opened 2024-03-21
Navigating straight to a device-groups page results in missing content


knolleary:
### Current Behavior

If you open a device-groups url directly, it does not list any groups. If you navigate to another view in the UI and click back to device groups, the groups appear.

The same is true if you open an individual device-group directly - except then you get an even more blank page:

<img width="1024" alt="image" src="https://github.com/FlowFuse/flowfuse/assets/51083/2016eaef-0e29-4c81-88ce-6a5f04ff938c">

No obvious errors in the browser console.

### Expected Behavior

You should see the application device-groups as expected.

### Steps To Reproduce

Open a device-group url in a new browser tab.

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

knolleary:
I can recreate the missing device-group data when navigating to the application/device-group page - and have got the fix for that.

I cannot recreate the blank screen (as per screenshot above) when navigating directly to a specific device-group. For it not even to have the side navigation entries, *something* is going wrong high-up in the page structure - but there are no console errors and as far as I can tell, the same set of api calls are made (so no clues there as to where it is getting stuck).

Steve-Mcl:
I can get it to occur on FlowFuse cloud staging but there is no logs indicating any issue.

I have tested this locally, using latest code base but cannot recreate.  i have tried various throttling to see if it was a resource+timing issue but so far have not managed to gleen anything useful

I will see if the flame charts and a dump of the network requests can reveal any clues.

PR #3614, opened 2024-03-22
Ensure device group data loaded on direct navigation

knolleary:
Fixes https://github.com/FlowFuse/flowfuse/issues/3612

Guards access to computed props and ensures device-group data is loaded on direct-navigation to the dg page.


joepavitt (approved):

joepavitt:
How did you find the culprit `?` in the other file in the end? Did the "caught exceptions" flag anything, or more manual?

joepavitt (approved):

knolleary:
@joepavitt I wish I could say it was more scientific than "I just added lots of `?` to the computed props on the vue"

PR #3616, opened 2024-03-22
Clear Device Group when device owner is changed

### LINKED ISSUE #3613: Moving a device between applications does not remove it from a device group
### Current Behavior

Given a device that is assigned to an application and added to a device group in that application, if that device is then reassigned to another application, it is not removed from the old application's device group.

This results in the device not showing up as a candidate device for adding to a device group in the new application.

This was hit by a FF Cloud customer. I verified the root cause by examining the database and could see the devices that had been moved had `ApplicationId` set to the new application, but `DeviceGroupId` still set to the group of the old application.

### Expected Behavior

Unassigning a device from an application should also remove it from any device group in that application.

### Steps To Reproduce

 - Add device to application, and then to a device group in that application
 - Re-assign the device to another application
 - Create a new device group in the new application - edit the devices in the group. Note the just-moved device does not appear as a candidate.

### Environment

- FlowFuse version: 2.2

Steve-Mcl:
fixes #3613

## Description

Clears devices DeviceGroup when the device owner is changed

Includes tests:
```
  Application Device Groups API
    Update Device Group Membership
      Changing a device owner removes the device from the group
        ✔ Changing a device owner application removes the device from the group (168ms)
        ✔ Clearing device owner removes the device from the group (122ms)
```

## Related Issue(s)

#3613

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/api/device.js
+++ b/forge/routes/api/device.js
@@ -442,6 +442,8 @@ module.exports = async function (app) {
                     // Clear its target snapshot, so the next time it calls home
                     // it will stop the current snapshot
                     await device.setTargetSnapshot(null)
+                    // Since the project/application is being cleared, clear the deviceGroup
+                    device.DeviceGroupId = null
                     sendDeviceUpdate = true
                     // disable developer mode
                     device.mode = 'autonomous'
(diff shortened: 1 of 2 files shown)


knolleary (approved):

knolleary:
Had one more thought seconds after hitting merge - what happens with the `Device.targetSnapshotId` in this scenario? Does it get cleared properly?

Steve-Mcl:
gonna do some manual checks on DB to establish. Will raise another PR if that is the case - but, there are other checks in place where the target snapshot is verified to be same team/project/application when a device update is sent (so its kinda covered after the event) - still worth a follow up IMO

Steve-Mcl:
Verified, target snapshot is cleared on removing from application

PR #3645, opened 2024-03-27
Implement blueprint availability to team type option 

Steve-Mcl:
### High level
* Adds a default [x] Available to all team types UI dialog
* When selected, a list of check boxes and team types are presented, permitting availability to be turned on/off by team type/tier

chrome_bCv41o6ftd

### Details
* Adds DB field `FlowTemplates.teamTypeScope` which is a JSON field (type TEXT, parsed)
   * Default value `null` means "available to all, current and future team types"
   * when configured, the array will hold teamtype IDs that the blueprint can used
* Adds query param `team` to the existing GET `/api/v1/flow-blueprints/` route
   * Specifying a team returns only blueprints that are applicable to that teams tier
* Adds front end API fn `getFlowBlueprintsForTeam`
* Adds elements in `frontend/src/pages/admin/FlowBlueprints/dialogs/FlowBlueprintFormDialog.vue` for configuring the blueprint availability
* Adds tests (listed below)

### Tests
```
  Flow Blueprints API
    Create Flow Blueprints
      ✔ Availability is null by default (available to all team types)
      ✔ Availability is set to empty array when no team types are selected
      ✔ Availability is set to an array of team types
    List Flow Blueprints
      Blueprint availability
        ✔ Lists blueprints available to a starter tier
        ✔ Lists blueprints available to an enterprise tier
    Update Flow Template
      ✔ Updates availability
      ✔ Discards invalid team types
```

### TODO
- [ ] docs

## Related Issue(s)

#3599

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Backport needed? -> add the `backport` label
 - [x] Includes a DB migration? -> add the `area:migration` label

PR #3649, opened 2024-03-27
Add credentialSecret to snapshot model and update logic to use it

### LINKED ISSUE #3644: Migrate credentialSecret into Snapshot model
Part of #3622 

Task is to add credentialSecret to the ProjectSnapshots table so we are not reliant on the originating instance/device existing in order to access the snapshot.

The open question is whether the db migration should populate this column for all existing snapshots. We *could* choose to only populate for new snapshots; the logic to retrieve the credentialSecret could fall back to existing behaviour if not set on the snapshot. This would mean existing snapshots are still vulnerable to being orphaned, but the risk would be diminishing as applications tend to move forwards.

Will need to evaluate the cost/risk of doing full migration.

```[tasklist]
### Tasks
- [x] Add column to ProjectSnapshots table
- [x] Update snapshot creation logic to populate credentialSecret column
- [x] Update logic that retrieves the credentialSecret to look in the new place
```

knolleary:
Fixes #3644 

## Description

This adds `credentialSecret` to the snapshot model so we no longer depend on the parent of the snapshot existing. This makes it easier to reason about the snapshots as they get moved around.

The db migration will populate the new column with the appropriate value from with the Device or ProjectSettings tables. I've tested this migration in both sqlite and postgres.

I have looked at many of the pipeline tests that relate to snapshots. They checked the result had a credentials property, but didn't try to validate the value. I've updated the tests in a few places to validate the credentials are properly encrypted with the expected key.

I've also done various bits of manual testing with rollbacks, instance-assigned devices, app-assigned devices, pipeline deploys to instances and devices/device groups.

### DIFF
--- a/forge/db/controllers/ProjectSnapshot.js
+++ b/forge/db/controllers/ProjectSnapshot.js
@@ -364,9 +364,12 @@ module.exports = {
     createSnapshot: async function (app, project, user, options) {
         const projectExport = await app.db.controllers.Project.exportProject(project)
 
+        const credentialSecret = await project.getCredentialSecret()
+
         const snapshotOptions = {
             name: options.name || '',
             description: options.description || '',
+            credentialSecret,
             settings: {
                 settings: projectExport.settings || {},
                 env: projectExport.env || {},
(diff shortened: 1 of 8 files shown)


knolleary:
Am going to spin up a test env to do some more pointing and clicking before we merge, but please continue the review of the code.

knolleary:
Hitting a postgres-only test failure. Can recreate locally on pg, but on sqlite. Investigating...

knolleary:
Test fixed.

Steve-Mcl:
code and tests read fine but i am pulling this branch for a quick run through.

Steve-Mcl (approved):
code review and local testing with app device and other instance snapshot deploys 👍

Steve-Mcl:
@knolleary Approved.  Had to update branch and so tests are re-running. Merge when tests pass.

Steve-Mcl:
@knolleary now that #3645 is merged, we can merge this. 

_#3645 had migration "20240325-01-add-teamTypeScope-to-FlowTemplates.js" which has a date 2 days prior to this PR "20240327-01-add-credentialSecret-to-snapshots.js"_

I've merged main to this and will merge as soon as tests pass unless you should up

knolleary:
@Steve-Mcl all good - thanks

PR #3676, opened 2024-04-03
Ensure InstanceForm applies InstanceType availability for manual billing teams

knolleary:
If a team was set to manual billing, the code to filter the list of available instance types was being skipped. This hadn't been an issue because only Starter teams had a restricted selection of types and they don't ever get put into manual billing mode.

But as we now want to apply some restrictions to Team/Enterprise teams, this needs to work.


Steve-Mcl (approved):
It looks like the `unmanaged` flag is an iterative/later addition and therefore default/null value means "managed". Therefore the logic adjustment in this PR seems sane. You permit the body of code to execute solely on  `this.billingEnabled` but inhibit/guard the parts of code not intended to run using `!this.team.billing?.unmanaged`.

 Happy to approve.

PR #3679, opened 2024-04-04
Make limits tougher on common auth routes

knolleary:
## Description

This makes the limits tougher on the forgot-email and verify-email routes - 2 requests/minute. That should be more than enough for any human interaction with the routes. Also added a 'try again later' message to the UI in case of a human being too quick.

This should reduce some of the noise we get on the security mailing list.


hardillb:
Looks good, but obviously needs test update

hardillb (approved):

Issue #3712, opened 2024-04-11
Device overview page continues to poll after 404


hardillb:
### Current Behavior

If you have the device overview page open it will continue to poll for updates even after the device is deleted

### Expected Behavior

If device api returns 404 it should stop polling

May also navigate to device list (optional)

### Steps To Reproduce

- Open the device overview page
- In second window, delete device

### Environment

- FlowFuse version: 2.3.0
- Node.js version: 18.x
- npm version: 9.x
- Platform/OS: all
- Browser: all

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #3714, opened 2024-04-12
Stop polling a device if it's deleted

hardillb:
fixes https://github.com/FlowFuse/flowfuse/issues/3712

## Description
Stop polling for device status if it gets a 404 HTTP status code which implies that the device has been deleted.

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/3712

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
@joepavitt can you have a look at this please, I'd like to get it merged

joepavitt:
@Steve-Mcl in reviewing this PR I noticed that the "last seen" status refreshed every 5 seconds when on the Overview page, but not the others. Can see [this](https://github.com/FlowFuse/flowfuse/blob/10c670e8093a71803b2ad91e48c7873d6ce0aefa/frontend/src/pages/device/index.vue#L300) was added in by you a month ago to protect the Settings page, but is there a reason why it's not also running on Snapshots, Audit Log, Device Logs pages?

Steve-Mcl:
> @Steve-Mcl in reviewing this PR I noticed that the "last seen" status refreshed every 5 seconds when on the Overview page, but not the others. Can see [this](https://github.com/FlowFuse/flowfuse/blob/10c670e8093a71803b2ad91e48c7873d6ce0aefa/frontend/src/pages/device/index.vue#L300) was added in by you a month ago to protect the Settings page, but is there a reason why it's not also running on Snapshots, Audit Log, Device Logs pages?

It is a bit more than that Joe.

It refreshes "always" on pages without user input or long lists (i.e. overview AND developer pages)
It refreshes on ALL pages if the state unknown or is intermediate.

```
            // Only refresh device via the timer if we are on the overview page, developer mode page
            // the device status is empty or the device is in a transition state
            // This is to prevent settings pages from refreshing the device state while modifying settings
            // See `watch: { device: { handler () ...  in pages/device/Settings/General.vue for why that happens
            const settingsPages = ['DeviceOverview', 'DeviceDeveloperMode']
            try {
                if (settingsPages.includes(this.$route.name)) {
                    await this.loadDevice()
                } else if (typeof this.device?.status === 'undefined') {
                    await this.loadDevice()
                } else if (deviceTransitionStates.includes(this.device?.status)) {
                    await this.loadDevice()
                }
```


This was necessary as prior to this, status was not being updated on certain pages and this was a quick win to (better) show actual state.

the better/harder/longer/slower solution was to re-engineer the pages with inputs and their own polling to use a store/vuex

joepavitt:
> It refreshes "always" on pages without user input or long lists (i.e. overview AND developer pages)

It doesn't refresh on Snapshots, or Audit Log or Device Logs though, and that's annoying... I'll raise this as a separate issue because it's not linked (as I thought it might have been) to Ben's PR.

joepavitt (approved):

Steve-Mcl:
> It doesn't refresh on Snapshots, or Audit Log or Device Logs though

And that was true prior to the code you linked. What that code did was actually allow status to be updated when on those other pages when state was unknown or undetermined. 

Prior to the linked code, polling only occurred on overview and only until it reached a non indeterminate state. 

As I said, it was a quick win improvement. The better solution is ...

PR #3718, opened 2024-04-15
Remove NR Session when user role changes

hardillb:
fixes https://github.com/FlowFuse/security/issues/82

## Description

Ensure that any open NR sessions for the user are removed so they no longer have access

Need to work out test cases for this.

## Related Issue(s)

https://github.com/FlowFuse/security/issues/82

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


knolleary (changes requested):
Needs some tests, but otherwise looks okay

hardillb:
Need to work out how to test this and get it merged

forge/db/controllers/Team.js
```diff
@@ -90,8 +90,9 @@ module.exports = {
             }
             await userRole.destroy()
 
+            await app.db.controllers.StorageSession.removeUserFromTeamSessions(user, team)
```
ZJvandeWeg, line 93: Wouldn't the API: `await team.removeUserFromSessions(user)` be nicer, and make the model fat and keep the controller skinny?

https://medium.com/marmolabs/skinny-models-skinny-controllers-fat-services-e04cfe2d6ae

hardillb: We already have similar code in the controller, this keeps it all in the same place

knolleary (approved):
Noting ZJ's comment about fat controllers - we have some general work to do to tidy up what belongs in the controllers and what in the models. We will have to be opportunistic to do that work. For this PR, I'm okay with adding this functionality to the existing controller alongside the other session invalidation logic for now in order to keep things moving forward.

PR #3737, opened 2024-04-16
Revert "Bump @node-saml/passport-saml from 4.0.4 to 5.0.0"

hardillb:
Reverts FlowFuse/flowfuse#3723

Breaks SSO on staging


knolleary:
The underlying `@node-saml/node-saml` is updated which includes a breaking change:  https://github.com/node-saml/node-saml/blob/master/CHANGELOG.md#-major-changes

Specifically https://github.com/node-saml/node-saml/pull/343

knolleary (approved):

PR #3748, opened 2024-04-18
billing: Start billing cycle on the 1st each month

ZJvandeWeg:
For billing insights and understand early in the month about where our revenue is going to end I've updated the billing cycle to start the 1st of each month. The default behaviour is to start the month the date + time they start their subscription.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


knolleary (changes requested):
`billing_cycle_anchor_config` isn't a valid parameter for the `stripe.checkout.sessions.create(..)` function we use.

According to [the docs](https://docs.stripe.com/payments/checkout/billing-cycle#create-session), we have to use `subscription_data.billing_cycle_anchor` which lacks the convenice of `billing_cucle_anchor_config` as we'll have to calculate the timestamp of the next 1st of the month ourselves.

Instead it has to be added under the existing `subscription_data` parameter (line 64) -

ZJvandeWeg:
@knolleary I'm sorry; I don't quite now how to validate my changes. The only test I could think of was to `spy` on the map/hash constructed, which felt silly. But I don't have a great Stripe test setup. Do you know how to achieve this and validate my hacking?

knolleary:
@ZJvandeWeg I'll take a look. I have stripe setup locally; what is harder to test is time-based events in the future where you want to check what happens at invoice time etc. That's something we have to do fairly manually.

knolleary:
Initial test looks good. Here is what the checkout page shows:

image

Are we happy with that?

ZJvandeWeg:
I like this a lot, thanks @knolleary !

knolleary (approved):

PR #3766, opened 2024-04-23
Allow user setting node red version setting for application assigned devices

### LINKED ISSUE #2802: Define Node-RED version for Devices not bound to an Instance
### Description

Originally found with https://github.com/flowforge/flowforge-device-agent/issues/173

Devices that are not bound to Instance do not have a way of pulling a defined Stack, as such, they default to the `latest` instance of Node-RED.

We should be able to configure (not sure where/how?) which version of Node-RED a Device runs.

## Background
In the current FlowForge Device setup, devices that are not bound to an Instance automatically use the 'latest' version of Node-RED. This default behavior limits users who may need to run specific versions of Node-RED for development, compatibility, or testing purposes.

## User Story
As a FlowFuse User who works with Devices,
I want the ability to specify and change the Node-RED version (stack) for my devices,
So that I can deploy different versions of Node-RED to my development devices, catering to the specific needs of each project.

## Acceptance Criteria
- Default Stack Selection: During the initial creation phase of a Device, the UX remains unchanged, automatically selecting the 'latest' Node-RED stack as the default.
- Stack Modification Option: In the device settings, users must have the option to change the stack. This includes a list of all available Node-RED versions (stacks) that the user can select from.
- Developer Mode Requirement: Changing the Node-RED version on a device is only permissible if the device is in 'Developer Mode'. 

## Which customers would this be available to
All Starter + Team + Enterprise Tiers (CE)

Steve-Mcl:
**NOTE: Part 1 of 2. Only merge when both this and the [device agent counterpart](https://github.com/FlowFuse/device-agent/pull/254) are both approved.**

## Description

As per [Approach](https://github.com/FlowFuse/flowfuse/issues/2802#issuecomment-2071738509) in issue discussion:

> 1. Add `editor` side tab to device->settings
> 2. Get/Set an object `editor` in `DeviceSettings` Table
>    1. this provides a place for future editor based settings
> 3. Override the starter & actual snapshots `.modules['node-red']` version if `editor.nodeRedVersion` is set
>    1. This acts like an override of any snapshot applied to the device
>    2. Clearing the value reverts to using the version in the snapshot (latest for starter/whatever was stack was used for the snapshot)

further details and considerations are written in the linked issue

## Related Issue(s)

Closes #2802

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
Arrrrgggghhhh

unrelated e2e test still failing

```
  1) FlowForge - Team Overview (Home) - With License
       shows a list of cloud hosted instances:
     AssertionError: Timed out retrying after 4000ms: Expected to find content: 'application-device-a' within the element: <li> but never did.
      at Context.eval (webpack:///../test/e2e/frontend/cypress/tests-ee/team/overview.spec.js:123:27)
```

Tests run cleanly in local env with cypress open. will try full run without cypress open locally - perhaps something related to other tests is causing order of elements to be different?

knolleary (changes requested):
As this has some dependency on the agent version, I think we should inform the user what behaviour they will get based on the known agent version.

 - older agents: NR version will only get applied when a new snapshot is deployed
 - new agents: NR version will get updated after settings saved (and/or new snapshot is deployed).

Exactly what version number to test for depends on us releasing the Device Agent - which we can do first

Steve-Mcl:
> Exactly what version number to test for depends on us releasing the Device Agent - which we can do first
> * older agents: NR version will only get applied when a new snapshot is deployed
> * new agents: NR version will get updated after settings saved (and/or new snapshot is deployed).



@Steve-Mcl 
_For the personal memory banks - reminder to follow up once new release of device agent is cut_

Add feedback when user applies a NR version to the device - so that they know when/how the set version will be applied (based on the agent version)

Steve-Mcl:
Since Agent is currently v2.4.0 and xxx has been merged, i can use a semvar <= 2.4.0 to indicate the alternative meeans of applying the NR version update.

Here is a demo - lay out was discussed with @joepavitt 


chrome_WiS2CuKSuS


Footnotes:

| agent version | footnote |
| --------: | :-------- |
| <= 2.4.0 | The device will be updated to the specified version when a different snapshot is deployed |
| > 2.4.0 | The device will be updated to the specified version and restarted upon saving the settings |

frontend/src/pages/device/Settings/Editor.vue
```diff
@@ -0,0 +1,130 @@
...
+        changed () {
+            const changed = this.initial.nodeRedVersion !== this.input.nodeRedVersion
+            return changed
+        },
+        deviceOwnerType: function () {
+            return this.device?.ownerType || ''
+        },
+        versionChangeFootnote: function () {
+            if (!this.device?.agentVersion || semver.lte(this.device.agentVersion, '2.4.0')) {
+                return 'The device will be updated to the specified version when a different snapshot is deployed'
```
knolleary, line 83:
```suggestion
                return 'The device will be updated to the specified version the next time a snapshot is deployed'
```

Slight rewording suggestion

knolleary (approved):
Approved - with just one small rewording suggestion

knolleary:
Only pause is now that this setting *doesn't* get applied if the device is in developer mode. Lets move forward with this for now, but I think we need a step back look at developer mode and how we communicate what does/doesn't happen whilst in that state and how we can improve things. For example, we don't let a pipeline stage get triggered from a developer-mode device.... which feels odd if its your development device you want to snapshot and deploy... but I digress and will follow up separately.

Issue #3744, opened 2024-04-17
Exporting Team Library flows breaks the UI interface when users don't manually fill in the .json extension


cstns:
### Current Behavior

Users are able to export a team library flow without appending the .json extension in the NR editor which then breaks the FF Team Library UI when displaying said export.

### Expected Behavior

The FF UI should be able to load and render said export even if the .json extension is missing

### Steps To Reproduce

- select a couple of nodes from the NR editor
- export nodes to team library and set a name that does not include the .json extension
- navigate to the FF Library page and click on the newly created export
- notice that the page is not displayed correctly

### Environment

- FlowFuse version: "2.3.0"
- Node.js version: n/a
- npm version: n/a
- Platform/OS: n/a
- Browser: n/a

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

sherbastian:
When exporting the flow from the NR editor, the placeholder name contains the .json placeholder. If we're going to solve this under the hood and make it transparent/seamless to the end user, this should also be addressed

PR #3777, opened 2024-04-25
Fix lib viewer when user does not specify an extension

Steve-Mcl:
## Description

fixes the issue where FF cannot display a user saved flow or function if they did not include the extension

### Detail

Existing logic inferred the type from the name. (flows were *.json, functions *.js)
This doesnt handle names without an extension (node-red itself has no issue with this and existing users libraries likely contain entries like this)

Since the backend API is used for both FF & Node-RED, I could not simply add the `type` info into the data (data is not wrapped or and provides no meta info - it is the essentially flows array or a function string!). In this instance, the MVP was to append the `type` info (that is available server side) as a header in the response and parse it back out in the FF FrontEnd API wrapper. 

### Before: 
chrome_RCz4FqbYU1

_^ JS and flow files without extensions behave incorrectly (and differently)_

## Related Issue(s)

#3744

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/pages/team/Library.vue
```diff
@@ -30,7 +30,7 @@
                     <ff-data-table-row v-for="row in rows" :key="row" :selectable="true" @click="entrySelected(row)">
                         <ff-data-table-cell><TypeIcon :type="row.type" /></ff-data-table-cell>
                         <ff-data-table-cell>{{ row.name }}</ff-data-table-cell>
-                        <ff-data-table-cell>{{ formatDateTime(row.updatedAt) }}</ff-data-table-cell>
+                        <ff-data-table-cell>{{ row.updatedAt ? formatDateTime(row.updatedAt) : ''}}</ff-data-table-cell>
```
Steve-Mcl, line 33:
```suggestion
                        <ff-data-table-cell>{{ row.updatedAt ? formatDateTime(row.updatedAt) : '' }}</ff-data-table-cell>
```

Fix lint warn

knolleary (approved):

PR #3814, opened 2024-05-02
Generate snapshot-specific filename on download

### LINKED ISSUE #3813: Downloaded snapshot filename should include id/date rather than just `snapshot.json`
### Description

The 'download snapshot' feature always downloads the snapshot as `snapshot.json`. It would be better to have a a more unique name so you don't end up with `snapshot.json`, `snapshot (1).json` etc.

One option would be to include the snapshot name in the filename - but with our auto-generated names, that could get unwieldy.

Propose something like: `snapshot-<id>-YYYYMMDD.json`

### Epic/Story

#3627

knolleary:
Closes #3813 

Generates a snapshot-specific filename rather than hardcoding to `snapshot.json`, in the format: `snapshot-<id>-<date>-<time>.json`. For example: `snapshot-WYNXyNORQK-20240122-175224.json`

Also reduced the generated cred secret length to be a bit more manageable.

### DIFF
--- a/config/cypress-shared.config.js
+++ b/config/cypress-shared.config.js
@@ -1,3 +1,5 @@
+const { readdir } = require('fs/promises')
+
 module.exports = {
     viewportWidth: 1024,
     viewportHeight: 768,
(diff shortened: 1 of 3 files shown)


Steve-Mcl:
Pretty sure there is an e2e test that will need updating Nick.

Steve-Mcl (approved):
love it.

Steve-Mcl:
approved and tested on pre-staging env.

chrome_7DyKXMuH57

Issue #324, opened 2022-02-18
Custom Instance Domains


hardillb:
### Description

As a Team Owner,
I want to be able to supply a custom domain to be used with my projects to differentiate them from other projects hosted on the system

### Customer requests
1. https://app-eu1.hubspot.com/contacts/26586079/record/0-2/8209301981
2. https://app-eu1.hubspot.com/contacts/26586079/record/0-1/7563151
3. https://app-eu1.hubspot.com/contacts/26586079/record/0-2/8763467964
4. https://app-eu1.hubspot.com/contacts/26586079/contact/8464752608

```[tasklist]
### 1st Iteration Development Work
- [ ] https://github.com/FlowFuse/driver-k8s/pull/151
- [ ] https://github.com/FlowFuse/flowfuse/pull/3830
- [x] nr-launcher authentication plugin will need updating (and possible forge app) to handle correct redirect back to the editor/dashboard if request comes from custom hostname.
- [ ] https://github.com/FlowFuse/helm/pull/396
- [x] Testing on dedicated scaling cluster
```

```[tasklist]
### Production Updates
- [ ] https://github.com/FlowFuse/website/issues/2152
- [ ] https://github.com/FlowFuse/CloudProject/issues/418
- [ ] https://github.com/FlowFuse/CloudProject/issues/422
```

```[tasklist]
### Future work
- [ ] https://github.com/FlowFuse/terraform-aws-flowfuse/issues/8
- [ ] https://github.com/FlowFuse/flowfuse/issues/3982
```

sammachin:
Just to confirm this should be a FQDN per project rather than a domain per team

cgsmith:
Would this mean the FQDN would be `domain.com` for flowforge then each project would be a a subdomain or project folder?

`project.domain.com` would point to `localhost:7880` or w/e?

hardillb:
@cgsmith This is specifically talking about running with the docker or kubernetes container drivers not the localfs.

This is about allow teams to have their own root domain. e.g. team one can have `<project name>.one.com` and team two would have `<project name>.two.com`

You can not host projects (Node-RED instances) on different paths (e.g. `example.com/<project name>`)  due to the way browser local storage works as Node-RED use it as part of the editor authentication.

protocolus:
I have customers that would love to use the FF service but, depend on custom domains to serve their clients. Being able to tie a wildcard SSL to an account would be even better. That way new projects could all be under that domain automatically.

[...]

PR #3830, opened 2024-05-07
324 - Allow users to set custom hostnames for instances

hardillb:
## Description

UI work for #324

Will also include the ee work

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/instance/Settings/General.vue
+++ b/frontend/src/pages/instance/Settings/General.vue
@@ -35,6 +35,21 @@
         <FormRow v-model="input.templateName" type="uneditable">
             Template
         </FormRow>
+        <div>
+        <FormHeading class="mb-6">Hosting</FormHeading>
+        <FormRow v-model="instance.url" type="uneditable">
+            Default URL
+        </FormRow>
+        <FormRow v-model="input.hostname">
+            Custom Hostname
+            <template #description>
+                This needs to be a fully qualified hostname
+            </template>
+            <template #append>
+                <ChangeIndicator :value="changed.hostname" />
+            </template>
+        </FormRow>
+        </div>
         <DangerSettings
             :instance="instance"
             @instance-updated="$emit('instance-updated')"
@@ -49,6 +64,7 @@ import { mapState } from 'vuex'
 
 import FormHeading from '../../../components/FormHeading.vue'
 import FormRow from '../../../components/FormRow.vue'
+import ChangeIndicator from '../../admin/Template/components/ChangeIndicator.vue'
 
 import DangerSettings from './Danger.vue'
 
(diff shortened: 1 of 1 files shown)


knolleary (changes requested):
Some initial comments based on discussion:

 - modify the customHostname setting to be an object ('hostname:string', 'dnsVerified':boolean) - which allows us to track if the dns 'looks' okay for the provided hostname
 - add a verification step when the user sets/changes the custom hostname - see if we can check the dns for the provided hostname is properly configured
 - add UI feedback on what to set the dns entry to
     - `myCustomDomain.net CNAME dns-entry.flowfuse.com` (or whatever makes sense)
 - add a top level configuration to specify the subdomain assigned to the ingress entry point

hardillb:
If we turn the Custom hostname setting into an object, how can we do a squelize query on a field in the object to check if the name is in use?

knolleary:
Good point. Scrap that idea then - but we still want a way to know if the name is verified or not. So either another ProjectSetting value, or we introduce a new db model for InstanceHostname. Suspect the former is more straight forward

forge/ee/lib/customHostnames/index.js
```diff
@@ -0,0 +1,34 @@
+const { KEY_CUSTOM_HOSTNAME } = require('../../../db/models/ProjectSettings')
+
+module.exports.init = function (app) {
+    app.log.debug(`EE CustomHostname ${app.config.driver.type}\n ${JSON.stringify(app.config.driver.options)}`)
+    // TODO localfs should be removed before check in
+    if (app.config.driver.type === 'kubernetes' || app.config.driver.type === 'stub' || app.config.driver.type === 'localfs') {
```
hardillb, line 6:
```suggestion
    if (app.config.driver.type === 'kubernetes' || app.config.driver.type === 'stub') {
```

hardillb: Just making sure we don't forget to remove localfs from supported options

frontend/src/pages/instance/Settings/General.vue
```diff
@@ -78,21 +129,75 @@ export default {
...
+        customHostnameLauncherVersion () {
+            const launcherVersion = this.instance?.meta?.versions?.launcher
+            if (!launcherVersion) {
+                // Not sure the launcher version - could be suspended/started
+                // Be optimistic
+                return true
+            }
+
+            // needs to be  v2.5.0 or better
+            return SemVer.satisfies(SemVer.coerce(launcherVersion), '>=2.4.0')
```
hardillb, line 175:
```suggestion
            return SemVer.satisfies(SemVer.coerce(launcherVersion), '>=2.5.0')
```

Again bumping this to match what it will ship in.

knolleary (approved):

knolleary:
Going to hit merge once tests pass

PR #3861, opened 2024-05-15
Fix tile-select option sizing

knolleary:
Whilst working on some improvements around changing team type, I notice the layout of the tile-selection-options was not aligning the vertical sizes.

image

On investigation, I found #3249 had wrapped the `tile-selection-option` in a div in order to get `ff-tooltip` working on the option. The wrapping div is then getting the right vertical size, but the option doesn't grow to fill it.

On further digging into *why* the div was added, it was done to get the tooltip working. The reason it wasn't working was a disabled tile-selection-option sets `pointer-events:none`, so the hover state is never triggered to show the tooltip.

This PR removes the wrapping div and modifies the css to remove the need to use `pointer-events: none` in the disabled state.

End result:

image

### DIFF
--- a/frontend/src/pages/team/changeType.vue
+++ b/frontend/src/pages/team/changeType.vue
@@ -15,7 +15,7 @@
         <ff-loading v-if="loading" message="Updating Team..." />
         <div v-else class="m-auto">
             <form class="space-y-6">
-                <div class="max-w-2xl m-auto">
+                <div>
                     <FormHeading>
                         <template v-if="isTypeChange">
                             Change your team type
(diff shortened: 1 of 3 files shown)


knolleary:
Added a small change to allow the Team Type select to use full select, whilst keeping the form its reduced width.

Given we have 3 types in production,optimising to have them in a single row rather than split makes for a nicer layout.

image

joepavitt:
Agree with the change of widht for the tiles, but it looks very odd center-align, moved to left aligned instead:

Screenshot 2024-05-16 at 14 04 43

joepavitt (approved):
Approved subject to final tests passing

knolleary:
> Agree with the change of widht for the tiles, but it looks very odd center-align, moved to left aligned instead:

I specifically didn't do that as all of our other forms (create app/instance etc) are center-aligned and constrained width. So was trying to keep some consistency with those - whilst allowing the type selection to 'expand' outwards.

But if you're happy with these changes, then no worries.

PR #3863, opened 2024-05-16
Add Blueprints preview

### LINKED ISSUE #3838: Blueprint preview
Leverage the flow viewer to show the flow before it's selected

Designs: https://github.com/FlowFuse/flowfuse/issues/3768#issuecomment-2076731668

cstns:
## Description

Adding support to preview blueprints using the flow-renderer across blueprint tiles

## Related Issue(s)

closes #3838

## Checklist

 - [ ] ~~Documentation has been updated~~
    - [ ] ~~Upgrade instructions~~
    - [ ] ~~Configuration details~~
    - [ ] ~~Concepts~~
 - [ ] ~~Changes `flowforge.yml`?~~
    - [ ] ~~Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template~~
    - [ ] ~~Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production~~

## Labels

 - [ ] ~~Includes a DB migration? -> add the `area:migration` label~~

### DIFF
--- a/forge/ee/db/views/FlowTemplate.js
+++ b/forge/ee/db/views/FlowTemplate.js
@@ -11,12 +11,13 @@ module.exports = function (app) {
             icon: { type: 'string' },
             order: { type: 'number' },
             default: { type: 'boolean' },
+            flows: { type: 'array' },
             createdAt: { type: 'string' },
             updatedAt: { type: 'string' }
         }
     })
     function flowBlueprintSummary (blueprint) {
-        return {
+        const newVar = {
             id: blueprint.hashid,
             active: blueprint.active,
             name: blueprint.name,
@@ -25,9 +26,11 @@ module.exports = function (app) {
             icon: blueprint.icon,
             order: blueprint.order,
             default: blueprint.default,
+            flows: blueprint.flows.flows,
             createdAt: blueprint.createdAt,
             updatedAt: blueprint.updatedAt
         }
+        return newVar
     }
 
     app.addSchema({
--- a/frontend/src/ui-components/components/DialogBox.vue
+++ b/frontend/src/ui-components/components/DialogBox.vue
@@ -1,13 +1,13 @@
 <template>
     <div ref="container" class="ff-dialog-container" :class="'ff-dialog-container--' + (open ? 'open' : 'closed')">
-        <div class="ff-dialog-box" :class="boxClass">
+        <div v-click-outside="close" class="ff-dialog-box" :class="boxClass">
             <div class="ff-dialog-header" data-sentry-unmask>{{ header }}</div>
             <div ref="content" class="ff-dialog-content" :class="contentClass">
                 <slot></slot>
             </div>
             <div class="ff-dialog-actions">
                 <slot name="actions">
-                    <ff-button kind="secondary" data-action="dialog-cancel" @click="cancel()">Cancel</ff-button>
+                    <ff-button v-if="canBeCanceled" kind="secondary" data-action="dialog-cancel" @click="cancel()">Cancel</ff-button>
                     <ff-button :kind="kind" data-action="dialog-confirm" :disabled="disablePrimary" @click="confirm()">{{ confirmLabel }}</ff-button>
                 </slot>
             </div>
@@ -46,6 +46,10 @@ export default {
         contentClass: {
             type: String,
             default: ''
+        },
+        canBeCanceled: {
+            type: Boolean,
+            default: true
         }
     },
     emits: ['cancel', 'confirm'],
--- a/frontend/src/components/blueprints/BlueprintTile.vue
+++ b/frontend/src/components/blueprints/BlueprintTile.vue
@@ -12,6 +12,16 @@
                 <CheckCircleIcon class="ff-icon-lg" />
                 <label class="text-green-800">Default</label>
             </div>
+            <ff-button
+                v-if="!editable"
+                data-action="show-blueprint"
+                class="ff-btn--secondary"
+                @click="$refs['dialog'].show()"
+            >
+                <template #icon>
+                    <ProjectIcon />
+                </template>
+            </ff-button>
             <ff-button v-if="!editable" data-action="select-blueprint" @click="choose(blueprint)">
                 Select
             </ff-button>
@@ -19,6 +29,18 @@
                 Edit
             </ff-button>
         </div>
+        <ff-dialog
+            ref="dialog"
+            data-el="preview-blueprint-dialog"
+            :header="`Flow Preview: ${blueprint.name}`"
+            :closeOnConfirm="true"
+            confirmLabel="Close"
+            :canBeCanceled="false"
+        >
+            <template #default>
+                <ff-flow-viewer :flow="blueprint.flows" />
+            </template>
+        </ff-dialog>
     </div>
 </template>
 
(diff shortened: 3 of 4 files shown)


forge/ee/db/views/FlowTemplate.js
```diff
@@ -11,12 +11,13 @@ module.exports = function (app) {
...
             order: { type: 'number' },
             default: { type: 'boolean' },
+            flows: { type: 'array' },
             createdAt: { type: 'string' },
             updatedAt: { type: 'string' }
         }
     })
     function flowBlueprintSummary (blueprint) {
-        return {
+        const newVar = {
```
joepavitt, line 20: why make it a const and not just return as before? Also, naming of this is a bit odd. Guessing this was a debug thing at some point?

frontend/src/ui-components/components/DialogBox.vue
```diff
@@ -1,13 +1,13 @@
...
-        <div class="ff-dialog-box" :class="boxClass">
+        <div v-click-outside="close" class="ff-dialog-box" :class="boxClass">
             <div class="ff-dialog-header" data-sentry-unmask>{{ header }}</div>
             <div ref="content" class="ff-dialog-content" :class="contentClass">
                 <slot></slot>
             </div>
             <div class="ff-dialog-actions">
                 <slot name="actions">
-                    <ff-button kind="secondary" data-action="dialog-cancel" @click="cancel()">Cancel</ff-button>
+                    <ff-button v-if="canBeCanceled" kind="secondary" data-action="dialog-cancel" @click="cancel()">Cancel</ff-button>
```
joepavitt, line 10: So now the default behaviour for _all_ dialogs across the platform is that there is no "cancel" button, and you must click outside?

forge/ee/db/views/FlowTemplate.js
cstns: true, that was for debugging purposes, reverted!

joepavitt:
Clicking outside dialog to close I find really annoying, as in my first attempt to create a blueprint just now, I accidentally clicked out, which closed the dialog, and I lost all progress.

https://github.com/FlowFuse/flowfuse/assets/99246719/c77d2e38-820e-4d1c-90f3-6fe694d40a95

I'm okay with this being functionality for the viewer, as there is no risk of progress-loss, but if there is a "cancel" button, I don't think click-outside should be enabled

cstns:
It can be easily removed, I added it because I have a habit of clicking outside of dialogs to close them.

joepavitt:
I agree it's a useful feature, but only if the dialog doesn't have a form in place. For the `v-click-outside`, wrap that in a check to see if the "Cancel" button is already present, and then we can close respectively.

joepavitt:
Also noticed this odd behaviour if we click on the zoom controls of the flow-viewer?

cstns:
It happened only on the instance creation screen because that was the only place where the flow viewer was inside a form. Somehow the click on any of the controls that are buttons caused the form submission even if they weren't of type submit.

I fixed it by preventing the click event from propagating outside of the FlowViewer component

joepavitt:
Screenshot 2024-05-16 at 16 31 38

The default zoom and rendering of a blueprint is broken when first viewing - if I recall correctly, @Steve-Mcl hit this issue with the snapshot preview? Are you able to advise on the workaround Steve?

Steve-Mcl:
> The default zoom and rendering of a blueprint is broken when first viewing - if I recall correctly, @Steve-Mcl hit this issue with the snapshot preview? Are you able to advise on the workaround Steve?

Will take a look @joepavitt 

I raised an issue about scroll bar sizing (quick small task) - that _might_ be related but this looks like a rendering error.  

Do you have the flow for that blueprint?

joepavitt:
> Do you have the flow for that blueprint?

It happens for all flows for me, but this particular flow was just a basic hello world:

```{"flows":[{"id":"0370b6b0630603a1","type":"inject","z":"9c18a4bf6ab4e5c1","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"","payloadType":"date","x":100,"y":100,"wires":[["00554bbc221b63b5"]]},{"id":"00554bbc221b63b5","type":"debug","z":"9c18a4bf6ab4e5c1","name":"debug 381","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","statusVal":"","statusType":"auto","x":290,"y":100,"wires":[]}]}```

If I recall, it's linked tot he dialog not quite being ready before the flow renderer tries to occupy the space? I know you made a specific renderer dialog to handle this, and this PR doesn't utilise that - which it may need to do.

joepavitt:
@Steve-Mcl fyi @cstns will update flow-renderer version as part of this PR too

Steve-Mcl:
@cstns I see you have renamed the snapshot viewer dialog to "flow viewer" and changed the `show` function to accept only a flow. 

It was intended this component would be able to show other parts of a snapshot in future iterations (e.g. the module list, the env vars etc). 

I think a better option would be to instead add an additional `show` function like `showBlueprint(blueprint)` instead of making `show(flows)` less future proof. 

This would mean in a future iteration, we could also support viewing the modules for blueprints modules and env vars too.

joepavitt:
> It was intended this component would be able to show other parts of a snapshot in future iterations (e.g. the module list, the env vars etc).

We haven't implemented that functionality _now_, so I think the approach @cstns has taken is the correct one. It doesn't prevent any future iteration to then adapt this and include the extra points you've mentioned.

Steve-Mcl:
> > It was intended this component would be able to show other parts of a snapshot in future iterations (e.g. the module list, the env vars etc).
> 
> We haven't implemented that functionality _now_, so I think the approach @cstns has taken is the correct one. It doesn't prevent any future iteration to then adapt this and include the extra points you've mentioned.

No it doesnt, but renaming it "flow viewer" and reducing scope from what it currently is means it will need another refactoring. With up front knowledge (as provided) a 3rd rename can be avoided.  All I am asking is to bear these point in mind.

joepavitt:
I think flow viewer is still appropriate here - it's more generalised than Blueprints, Library Entry or Snapshot which are the uses cases you're referring to?

Steve-Mcl:
> I think flow viewer is still appropriate here 

Not really, it will be (once we can show modules/env vars/etc) even more generic than that (more like _thingViewer_), but i'll stop quibbling over the name, it'll just be an under the hood implementation oddity ;) 

As for reducing the parameter scope for the `show` function to accept only `flow`, I am requesting instead, rename the existing function to `showSnapshot` and add a new `showBlueprint` so that when it comes time to implement the features we don't need to undo all the changes made in this PR back to how they were & it will work for blueprints too out of the gate.

joepavitt:
> As for reducing the parameter scope for the show function to accept only flow, I am requesting instead, rename the existing function to showSnapshot and add a new showBlueprint so that when it comes time to implement the features we don't need to undo all the changes made in this PR back to how they were & it will work for blueprints too out of the gate.

Sounds reasonable - @cstns please take action accordingly 😄

cstns:
I'm not sure I follow. Both snapshots and blueprints have similar structures and the component's show could cover them, in which case the component name becomes problematic because it's not only a flow viewer but a more detailed viewer

Steve-Mcl:
> I'm not sure I follow. Both snapshots and blueprints have similar structures and the component's show could cover them, in which case the component name becomes problematic because it's not only a flow viewer but a more detailed viewer

For those following along, we had a chat about this.  The function parameter was named `flows` but really it was a `thing` object that contained `flows` so was really a none issue.

@cstns & I then agreed we can now keep a singular `show` function that shows the `thing` (be it a blueprint, snapshot, whatever) and it will be able to have `modules` and other future stuff in the function parameter object.  All is well in the world. :)

cstns:
We just have to find a name for that `thing` :crossed_fingers:

joepavitt:
Please proceed as you had already named it @cstns - I don't want us holding up a PR over bikeshedding a name for a component

joepavitt:
Code all looks good, just updating to `main` as there are 14 file changes here, so want to reduce risk of a clash when merged

PR #3873, opened 2024-05-16
Adds creatable flag on TeamType/Instance settings

### LINKED ISSUE #3859: Fix team upgrade when it includes instance types unavailable at target type
### Description

The logic that handles changing team types verifies that the current team's usage is within the restrictions of the target team type and rejects the change if not.

This logic prevents a 'Team' tier team with lots of instances from 'downgrading' to the 'Starter' tier where they would only pay the flat rate.

However, this same logic also prevents a 'Starter' or 'Team' tier with a Small instance from upgrading to 'Enterprise' as we have disabled small instances in Enterprise. We need to define the expected behaviour when upgrading in this scenario.

There are two possible solutions:

 1. For an 'upgrade', allow a team to bring in 'inactive' instance types, whilst preventing creation of new ones.
     - Challenges:
         - We do the same check of 'isAllowed' in both the create and unsuspend paths. We'd need to differentiate the two - prevent creation, but allow unsuspend
         - We don't currently have a concept of 'upgrade' versus 'downgrade' - so would need a way to do so
 2. When trying to change team type, if there are disallowed instances, ask the user what to do (delete/upgrade type).
    - Challenges:
         -  That's a lot of UX work and API work

I think the quicker solution is option 1 - allow a team to bring in 'inactive' instance types when upgrading - but disallow in downgrading.

### Workaround

There are current two possible workarounds to this, although neither are ideal - hence the need to improve.

1. Delete all Small instances before upgrading
2. First upgrade to 'Team' tier and upgrade Small instances to Medium. Then upgrade to 'Enterprise'. This will generate a lot of Stripe churn.

knolleary:
Fixes #3859 

The existing model for how we enable/disable instance types for a particular team type means its all or nothing; a team can either have instances of a particular type or not.

This poses a problem in FFC when a Starter Team with a Small instance wants to upgrade to Enterprise Team that does not have Small instance type enabled.

The linked issue lists a few ways to tackle the problem - this PR is the quickest way forward:

 - In the TeamType configuration for what instances are available (`active`), it now also has a `creatable` flag.
 - A Team Type can have an Instance Type marked as `active` but if `creatable` is not enabled, then the team will not be allowed to create *new* instances of that type.
 - This means a Starter Team with a Small Instance can upgrade to Enterprise, but Small Type will not be an option when creating new instances.

In discussion on #3859 there were ideas around auto-upgrading instances to an 'allowed' type. This gets quite complicated to configure as you have to tell the platform how to map between instance types when changing teams. In the interests of unblocking the core issue, I've gone for this more straight-forward approach.

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '16'
+          node-version: '18'
       # Install NPM dependencies
       - name: Install Dependencies
         run: npm ci
--- a/forge/db/models/Team.js
+++ b/forge/db/models/Team.js
@@ -376,6 +376,11 @@ module.exports = {
                     await this.ensureTeamTypeExists()
                     return this.TeamType.getInstanceTypeProperty(instanceType, 'active', false)
                 },
+                isInstanceTypeCreatable: async function (instanceType) {
+                    await this.ensureTeamTypeExists()
+                    // Defaults to true unless explicit disabled
+                    return this.TeamType.getInstanceTypeProperty(instanceType, 'creatable', true)
+                },
                 getInstanceTypeLimit: async function (instanceType) {
                     await this.ensureTeamTypeExists()
                     if (!await this.isInstanceTypeAvailable(instanceType)) {
@@ -398,6 +403,15 @@ module.exports = {
                  */
                 checkInstanceTypeCreateAllowed: async function (instanceType) {
                     await this.ensureTeamTypeExists()
+
+                    const typeAvailable = await this.isInstanceTypeAvailable(instanceType)
+                    const typeCreatable = await this.isInstanceTypeCreatable(instanceType)
+                    if (!typeAvailable || !typeCreatable) {
+                        const err = new Error()
+                        err.code = 'instance_not_available'
+                        err.error = `Instance type '${instanceType.name}' not available for this team`
+                        throw err
+                    }
 
                     const instanceTypeLimit = await this.getInstanceTypeLimit(instanceType)
                     // Note that if the instanceType is unavailable for this team type,
--- a/frontend/src/pages/admin/TeamTypes/dialogs/TeamTypeEditDialog.vue
+++ b/frontend/src/pages/admin/TeamTypes/dialogs/TeamTypeEditDialog.vue
@@ -53,7 +53,10 @@
                 </div>
                 <div v-for="(instanceType, index) in instanceTypes" :key="index">
                     <FormHeading>Instance Type: {{ instanceType.name }}</FormHeading>
-                    <FormRow v-model="input.properties.instances[instanceType.id].active" type="checkbox" class="mb-4">Available</FormRow>
+                    <div class="grid gap-3 grid-cols-4">
+                        <FormRow v-model="input.properties.instances[instanceType.id].active" type="checkbox" class="mb-4">Available</FormRow>
+                        <FormRow v-if="input.properties.instances[instanceType.id].active" v-model="input.properties.instances[instanceType.id].creatable" type="checkbox" class="mb-4">Creatable</FormRow>
+                    </div>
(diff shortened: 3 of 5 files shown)


knolleary:
@Steve-Mcl ah - prestaging doesn't have billing setup and a crucial bit of logic is inside a billing-guarded block. I'll sort that out.

knolleary:
@Steve-Mcl the prestaging env is now properly configured.

 - There are now two instance types - Default and Large
 - Starter Teams can create Default instances but not Large
 - Enterprise Teams can create Large instances but not Default
 - `Plant Starter` team is a Starter type. It has a couple Default instances.

Prior to this PR, in this situation it would not be possible to upgrade the `Plant Starter` team to Enterprise - because Enterprise had 'Default' instance disabled.

With this PR, The Enterprise Team is configured to now allow Default instances, but not to allow new ones to be created. As such, it is possible to upgrade the Starter team to Enterprise. (I haven't done so - will let you click the button to verify).

Steve-Mcl:
> * Starter Teams can create Default instances but not Large

✅ check


> * Enterprise Teams can create Large instances but not Default

✅  check

> * `Plant Starter` team is a Starter type. It has a couple Default instances.

> With this PR, The Enterprise Team is configured to now allow Default instances, but not to allow new ones to be created. As such, it is possible to upgrade the Starter team to Enterprise. (I haven't done so - will let you click the button to verify).

✅  check

--- 

What I did note is, i can upgrade the Starter team to Enterprise but I can ALSO downgrade the team back to starter - is this expected Nick?

chrome_DZ2pIFqywQ

knolleary:
> What I did note is, i can upgrade the Starter team to Enterprise but I can ALSO downgrade the team back to starter - is this expected Nick? Your comment " (I haven't done so - will let you click the button to verify)." infers to me that that should not be possible?

An enterprise team can downgrade to starter if it doesn't contain anything disallowed in the Starter tier. If you had created a large instance in the enterprise team before attempting the downgrade, I would expect it to then fail.

Steve-Mcl:
> > What I did note is, i can upgrade the Starter team to Enterprise but I can ALSO downgrade the team back to starter - is this expected Nick? Your comment " (I haven't done so - will let you click the button to verify)." infers to me that that should not be possible?
> 
> An enterprise team can downgrade to starter if it doesn't contain anything disallowed in the Starter tier. If you had created a large instance in the enterprise team before attempting the downgrade, I would expect it to then fail.

I checked and the enterprise team cannot be switched to a starter team - this is what we see below. Not super friendly but it is prevented:
image

On a similar note, adding an default instance to starter team is not permitted but again, it is not great UX (lets user chose the default type then errors upon creation. A future iteration might want to disable the un-creatable types.
image

Happy to approve as is but will not merge in case you wish to address these.

Steve-Mcl (approved):
Approved but will leave for you to merge Nick as per previous comment.

knolleary:
Hmm, your last video isn't quite my expectation. Will have a look at the prestaging branch to understand exactly what state its in.

knolleary:
@Steve-Mcl when I go to create an instance in the starter team, I'm presented with the expected banner about being at the team limit, and the options are disabled.

I cannot recreate what you show in the video.

I wondered if it was an artefact of downgrading the team from Enterprise and then attempting to create without doing a page reload (just in case the Team object still had the Enterprise type attached)... but that all worked as well.


image

Steve-Mcl:
> I cannot recreate what you show in the video.

As you suspected, it is related to refresh.


In the starter team, with `creatable` enabled and browser refreshed:
1. switch off `creatable` for starter team
2. go to the application
3. create instance - you see default type is selectable and pressing the button generates the toast
4. refresh browser - you get now get no types and no banner
5. navigate away, then back to create instance - now it looks like your screenshot.

chrome_GvMIIbbnSJ


In essence it works, but a bit quirky.

knolleary:
The only quirky scenario is when an admin toggles the setting then tries to update a team without a browser reload. End users won't be able to hit that scenario :)

So all in all, good to merge.

PR #3877, opened 2024-05-16
Set secure flag on session cookie when running on https

knolleary:
Part of https://github.com/FlowFuse/security/issues/88

## Description

If we know we're running with https configured, we should set the secure flag on the session cookie. We cannot set it always on as that will break localfs installs without https configured.

I believe the test on `base_url` is the right one to do here. I've checked our helm chart sets that with a `https` prefix if the `https` flag is set.

### DIFF
--- a/forge/routes/auth/index.js
+++ b/forge/routes/auth/index.js
@@ -172,6 +172,11 @@ async function init (app, opts) {
         if (session) {
             const cookieOptions = { ...SESSION_COOKIE_OPTIONS }
             cookieOptions.maxAge = SESSION_MAX_AGE
+            if (/^https:/.test(app.config.base_url)) {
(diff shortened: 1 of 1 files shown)


knolleary:
Going to verify this works as I expect it to on the prestaging setup as I don't have a full https enabled localfs setup.

hardillb:
I'll have another look at how hard it would be to do HTTPs on localfs, in theory we should be able to share a single certificate between the forge app and the all the instances, since the certs don't care about port numbers.

hardillb (approved):
Approved as logically it looks good, but not merged while waiting on pre-staging testing results.

knolleary:
Have verified the cookie has the secure flag set in the prestaging env. Will merge in the morning.

Issue #3807, opened 2024-05-01
Improve feedback on Blueprint creation


joepavitt:
### Description

<img width="646" alt="Screenshot 2024-05-01 at 13 27 36" src="https://github.com/FlowFuse/flowfuse/assets/99246719/d6790913-6949-494b-8042-6eab68947a7f">

Trying to create a blueprint, and I get no feedback from the API/UI as to what the problem is, and we have no docs (that I can find) which help me understand what the problem may be

### Which customers would this be available to

Everyone - CE/Starter/Team/Enterprise

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Issue #2945, opened 2023-10-13
Flow Blueprints: Admin UI improvemenets


Pezmc:
Follow up to #2929 as part of https://github.com/FlowFuse/flowfuse/issues/2872

Need to look at the validation of the Flows object. In the database, we store it as:
```
{
    flows: [ usual array of nodes ],
    credentials: { object of credentials }
}
```

If some exports a flow from the editor, they just get the `[ usual array of nodes]` bit - and pasting that in here would be invalid.

I think the UI could provide a bit more help here to avoid an unexpected error if they get that format wrong.

I suggest splitting the Flows box into two - Flows and Credentials. And doing the work to split them apart/recombine in the expected format.

_Originally posted by @knolleary in https://github.com/FlowFuse/flowfuse/issues/2923#issuecomment-1761077094_

joepavitt:
Closed by #3887

PR #3887, opened 2024-05-20
Double down on flow validation

cstns:
## Description

Doubling FE & BE validation for flow creation while making error messages more explicit

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/3807

## Checklist

 - [ ] ~~Suitable unit/system level tests have been added and they pass~~ 
 - [ ] ~~Documentation has been updated~~
    - [ ] ~~Upgrade instructions~~
    - [ ] ~~Configuration details~~
    - [ ] ~~Concepts~~
 - [ ] ~~Changes `flowforge.yml`?~~
    - [ ] ~~Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template~~
    - [ ] ~~Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production~~

## Labels

 - [ ] ~~Includes a DB migration? -> add the `area:migration` label~~


joepavitt:
The biggest improvement I think required here is the lack of information on what data structure is required in each field. A placeholder would be a good addition too, or more detailed error mesage re: _why_ the form submission broke.

For example, it's not obvious that the "flow" form entry point actually requires an object of:

```json
{
    "flows": [...]
}
```

Similarly, no indication on the data format required for "modules" either

forge/ee/db/models/FlowTemplate.js
```diff
@@ -62,6 +62,15 @@ module.exports = {
...
                 return JSON.parse(rawValue)
+            },
+            validate: {
+                isValidModules (value) {
+                    try {
+                        JSON.parse(value)
+                    } catch (err) {
+                        throw new Error('Invalid \'modules\' json')
+                    }
+                }
```
Steve-Mcl, line 73: Could I suggest we do a little more validation - I have been caught out with the lack of validation on blueprints myself in the past!

for example `null`, `""`, `"123"`, 123, `[]` and `[ {"name: "node-red-dashboard", "version": "1.9.1"}]` are ALL valid JSON but they will break the blueprint.

```suggestion
            isValidModules (value) {
                // modules should be a single level key:string-value:string object
                try {
                    const parsed = JSON.parse(value)
                    if (!parsed || typeof parsed !== 'object' || Array.isArray(parsed)) {
                        throw new Error()
                    }
                    for (const key in parsed) {
                        if (typeof key !== 'string' || typeof parsed[key] !== 'string') {
                            throw new Error()
                        }
                    }
                } catch (err) {
                    throw new Error("Invalid 'modules' json")
                }
            }
```

frontend/src/pages/admin/FlowBlueprints/dialogs/FlowBlueprintFormDialog.vue
```diff
@@ -179,17 +179,33 @@ export default {
...
+            } catch (err) {
+                this.error = err
                 this.errors.flows = err
                 return
             }
+
+            flowBlueprintProps.flows = JSON.parse(flowBlueprintProps.flows)
+
             try {
                 flowBlueprintProps.modules = JSON.parse(flowBlueprintProps.modules)
```
Steve-Mcl, line 205: Same issue as before, this validation will permit null, "", "123", 123, [] and [ {"name: "node-red-dashboard", "version": "1.9.1"}] since they are ALL valid JSON, but they will break the blueprint!

frontend/src/pages/admin/FlowBlueprints/dialogs/FlowBlueprintFormDialog.vue
```diff
@@ -179,17 +179,33 @@ export default {
...
             } catch (err) {
                 this.error = 'Invalid JSON for modules'
-                this.errors.modules = err
+                this.errors.modules = 'Invalid JSON'
```
Steve-Mcl, line 208: Suggestion - tell the users what we expect:

```suggestion
                this.errors.modules = 'Modules should be an object of "module" : "version" pairs'
```

or something less technical (but equally informative)?

Steve-Mcl (commented):
Some suggestions to slightly improve the feedback and interlocking

forge/ee/db/models/FlowTemplate.js
Steve-Mcl: _If you are to accept my suggestion, please double check indentation - looks like I am 1 block off_ (sorry)

frontend/src/pages/admin/FlowBlueprints/dialogs/FlowBlueprintFormDialog.vue
cstns: I was lazy enough to copy over the validation from BE. Will update now

cstns: On the long run we should transition to more robust validation libraries like https://vuelidate-next.netlify.app/ or https://vee-validate.logaretm.com/v3/guide/basics.html

frontend/src/pages/admin/FlowBlueprints/dialogs/FlowBlueprintFormDialog.vue
```diff
@@ -179,17 +179,43 @@ export default {
...
+                this.errors.flows = 'Invalid JSON'
+                return
+            }
+
+            try {
+                if (!JSON.parse(flowBlueprintProps.flows).flows) {
+                    throw new Error('Flow json missing \'flows\' property')
+                }
+                if (!Array.isArray(JSON.parse(flowBlueprintProps.flows).flows)) {
+                    throw new Error('Flow json \'flows\' property not an Array')
```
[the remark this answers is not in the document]
cstns: i'll take it!

cstns:
Something strange happened though, a postgress pipeline test failed oh gh (didn't do it locally).

Passed on a [re-run](https://github.com/FlowFuse/flowfuse/actions/runs/9187985103?pr=3887). 

Can it be a cause of concern?

Steve-Mcl (approved):
Approved ~~with 1 v minor suggestion - so will let you decide and merge~~ you already merged suggestion - good to go

Steve-Mcl:
> Can it be a cause of concern?

The single unit test fail on PG was a seemingly random issue that could not be recreated on GH or locally (tried several times). 

You re-ran and tested locally. 
I scanned the test for missing awaits and a quick look through the core code but nothing stands out.
So as due diligence was applied and this is unrelated to original works, we should just log this move on for now. If further occurrences are encountered, we should investigate further.

<br>

For future readers (or anyone searching), the fail was `expected Object to have property isDeploying of true (got false)`

```
  1) Pipelines API
       Deploy Pipeline Stage
         With action=use_latest_snapshot
           For instance=>device group
             Copies the existing instance snapshot to the next stages device group:
     AssertionError: expected Object {
  id: '4VanVZnKEX',
  name: 'device-group-b',
  description: 'an unnamed device group',
  deviceCount: 1,
  targetMatchCount: 0,
  activeMatchCount: 0,
  developerModeCount: 0,
  runningCount: 0,
  isDeploying: false,
  hasTargetSnapshot: false
} to have property isDeploying of true (got false)
      at Assertion.value [as property] (node_modules/should/cjs/should.js:356:19)
      at Context.<anonymous> (test/unit/forge/ee/routes/api/pipeline_spec.js:2437:49)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

```

joepavitt:
Closes https://github.com/FlowFuse/flowfuse/issues/2945

PR #3891, opened 2024-05-20
Ensure proration_behaviour set when adding first device

### LINKED ISSUE #3890: Proration billing option not applied when adding first device
### Current Behavior

We added proration options to billing via #2758 - which means billing changes are applied to their next scheduled invoice, rather than generating an instant invoice.

However, one scenario was missed - when a team adds its very first device. This is still triggering an instant invoice.

### Expected Behavior

The appropriate proration behaviour should be applied.

### Steps To Reproduce

 - Create a **Team** plan team and setup billing
 - Create an Instance and delete it a couple minutes later - observe no invoice generated
 - Register a Device - note an invoice is generated immediately

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Fixes #3890 

Adds the missing `proration_behaviour` option to the code path when adding a teams first device.

This is outside the scope of our unit tests, and I don't believe the pre-staging setup has billing configured. I have verified this locally with the test stripe account - and the change is bringing this one call inline with all of the other calls to modify a users subscription.

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -378,14 +378,15 @@ module.exports.init = async function (app) {
                     }
                 } else if (billableCount > 0) {
                     // Need to add the device item to the subscription
+                    app.log.info(`Updating team ${team.hashid} subscription device count to ${billableCount}`)
                     const update = {
+                        proration_behavior: prorationBehavior,
                         items: [{
                             price: deviceBillingIds.price,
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):
> This is outside the scope of our unit tests, and I don't believe the pre-staging setup has billing configured

Understand.

Approved.

Issue #3851, opened 2024-05-14
Date ordering on Instance Last Updated uses the string rather than the date


robmarcer:
### Current Behavior

When you order the instances by Last Updated, it uses the shown string (such as 33 minutes ago) to do the ordering.

https://d.pr/i/0KrJEl

### Expected Behavior

It should use the underlying timestamp so when ordered the Instances are in the actual chronological order rather.

### Steps To Reproduce

Go to the Instance page, order by Last Updated

### Environment

- FlowFuse version: v2.4.1-951664c-202405100858.0
- Node.js version: v18.20.2
- npm version: unknown
- Platform/OS: FF Cloud
- Browser: Chrome Version 124.0.6367.158 (Official Build) (arm64)

### Have you provided an initial effort estimate for this issue?

I can not provide an initial effort estimate

PR #3893, opened 2024-05-20
Use original timestamp for sorting instances on Last Updated column

Steve-Mcl:
## Description

* Keeps the original datetime field in data
* Adds a SimpleTextCell (generic, reusable text cell) for displaying in markRaw
* Use the actual original field datetime value for the column key

No tests were added (its just a minor cosmetic fix TBH). 
If tests are deemed necessary, they can be added (but it is not a super simple task as we would need to add several instances each with differing updated times, operate client side sorting, read values from table fields to ensure order - tests can be done but it feels like a case of maximum effort, minimum gain!)

## Related Issue(s)

#3851

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


cstns (approved):

cstns:
Not really. @Steve-Mcl, is it good to go?

Steve-Mcl:
@cstns yes, not sure why it wasn't merged TBH. possibly slipped through the net while waiting for branch update I imagine!?

PR #3898, opened 2024-05-21
Allow Create Team view to go full width

knolleary:
This copies the changes made for the Upgrade Team Type view in [3861](https://github.com/FlowFuse/flowfuse/pull/3861) to the Create Team view for consistency.

image

### DIFF
--- a/frontend/src/pages/team/create.vue
+++ b/frontend/src/pages/team/create.vue
@@ -11,12 +11,12 @@
     <ff-page>
         <ff-loading v-if="redirecting" message="Redirecting to Stripe..." />
         <ff-loading v-else-if="loading" message="Creating Team..." />
-        <div v-else class="max-w-2xl m-auto">
+        <div v-else class="m-auto">
             <form class="space-y-6">
                 <FormHeading>Create a new team</FormHeading>
                 <div class="mb-8 text-sm text-gray-500">Teams are how you organize who collaborates on your projects.</div>
                 <!-- TeamType Type -->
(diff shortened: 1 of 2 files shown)


frontend/src/pages/team/create.vue
```diff
@@ -11,12 +11,12 @@
...
         <ff-loading v-if="redirecting" message="Redirecting to Stripe..." />
         <ff-loading v-else-if="loading" message="Creating Team..." />
-        <div v-else class="max-w-2xl m-auto">
+        <div v-else class="m-auto">
             <form class="space-y-6">
                 <FormHeading>Create a new team</FormHeading>
                 <div class="mb-8 text-sm text-gray-500">Teams are how you organize who collaborates on your projects.</div>
                 <!-- TeamType Type -->
-                <div class="flex flex-wrap gap-1 items-stretch">
+                <div class="grid grid-1">
```
joepavitt, line 19:
```suggestion
                <div class="grid">
```

Not sure `grid-1` does anything? It's not a tailwind class that I can find, and it renders the same in my browser if I remove it

joepavitt (changes requested):

frontend/src/pages/team/create.vue
knolleary: Was part of the other pr. Will apply and tidy up the other instance to match

knolleary: Have pushed a commit that fixes this in both this form and the upgrade form to match.

PR #3902, opened 2024-05-21
Add charges table to device create dialog

### LINKED ISSUE #3892: Device create UX does not inform about billing cost
### Current Behavior

Unlike when a user creates an instance, the device creation UX does not provide any information about the additional cost. This can lead to a user to unexpectedly rack up charges whilst exploring the platform.

We should include cost information on the Device creation form in some way that is informative but doesn't get in the way.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Closes #3892 

This reuses the existing `InstanceChargesTable` component to provide a summary of charges that will incurred by creating a new device:

As part of this, I've updated the message in the charges table based on the team proration behaviour. For Team/Enterprise we have it set to add items to the next invoice. For that case the message has been updated to say so:

image

The other behaviour we have is to trigger an invoice for each change made. For that case, the message remains as it was -indicating an immediate charge will be made:

image

### DIFF
--- a/frontend/src/pages/team/Devices/dialogs/TeamDeviceCreateDialog.vue
+++ b/frontend/src/pages/team/Devices/dialogs/TeamDeviceCreateDialog.vue
@@ -27,8 +27,13 @@
                     <template #description>Assign the device to an application (optional).</template>
                     Application
                 </FormRow>
-                <div v-if="billingDescription">
-                    <b>Price: {{ billingDescription }}</b>
+                <div v-if="deviceIsBillable">
+                    <InstanceChargesTable
+                        :project-type="deviceBillingInformation"
+                        :subscription="subscription"
+                        :trialMode="false"
+                        prorationMode="team?.type?.properties?.billing?.proration"
+                    />
                 </div>
             </form>
         </template>
@@ -38,6 +43,7 @@
 <script>
 import { mapState } from 'vuex'
 
+import billingApi from '../../../../api/billing.js'
 import devicesApi from '../../../../api/devices.js'
 import teamApi from '../../../../api/team.js'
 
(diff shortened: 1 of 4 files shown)


frontend/src/pages/team/Devices/dialogs/TeamDeviceCreateDialog.vue
```diff
@@ -126,15 +148,13 @@ export default {
...
-            // By passing 0 in as the limit, it will just return the `count`
-            // and not retrieve any actual devices
-            const teamData = await teamApi.getTeamDevices(this.team.id, null, 0)
-            this.totalDeviceCount = teamData.count
-        }
         this.loadApplications()
+        if (this.features.billing && !this.team.billing?.unmanaged) {
+            try {
+                this.subscription = await billingApi.getSubscriptionInfo(this.team.id)
+            } catch (err) {
```
joepavitt, line 155: why nothing in the catch here? Will suppress an error that we should be made aware of, no?

joepavitt:
Not being prompted with the charges table when looking at https://3902.flowfuse.dev/team/plant-team/devices

knolleary:
The auto-deploy setup does not have billing configured - so this cannot be verified there.

frontend/src/pages/team/Devices/dialogs/TeamDeviceCreateDialog.vue
knolleary: Because if the team does not have billing setup, this end point returns a 404. This is consistent with all other places we retrieve the subscriotuin

joepavitt (approved):

PR #3903, opened 2024-05-21
Docs - Update some of the "Stacks" docs and include instructions for local node development

joepavitt:
## Description

- Adds details on how to add a stack to FlowFuse such that you can have a local development environment for testing your own nodes, e.g. for Node-RED Dashboard 2.0
- Adds in some screenshots and attempts to centralise the "Manage Stacks" instructions which were a bit scattered in our documentation.


docs/admin/introduction.md
```diff
@@ -120,23 +120,50 @@ for teams to use.
...
+##### Create New Stack
+
+Alternatively, click 'Create stack' to create an entirely new stack.
+
+![Screenshot of FlowFuse showing the "Create Stack" dialog](./images/admin-stack-create.png){data-zoomable}
+<em>Screenshot of FlowFuse showing the "Create Stack" dialog</em>
+
+When prompted for the Node-RED version, provide the exact stack name (Node-RED version)
+that was installed. For example, if you ran the script with `latest` and it
+resulted in `3.1.9` being installed, you should enter `3.1.9`.
```
hardillb, line 158: This is specific to localfs stacks, it does not apply to docker/k8s where the user must supply a container image name.

The exact version is only used to locate the correct directory when using localfs stacks.

hardillb (commented):
Looks good, but the new section in the admin docs is specific to localfs stacks only.

I will try and find some suitable words for the other 2

docs/admin/introduction.md
joepavitt: We can always reference the relative instructions from here - but need to know where to point to, and those instructions need to be accurate.

hardillb: The docker stacks info is here https://flowfuse.com/docs/install/docker/stacks/ and the same for k8s here https://flowfuse.com/docs/install/kubernetes/stacks/

That is accurate, just no images

hardillb (approved):

PR #3925, opened 2024-05-24
Add LDAP support for SSO

knolleary:
## Description

This adds the ability to add an LDAP-based SSO configuration.

This has been done in a consistent way with our existing SAML SSO configurations. When an add creates an SSO configuration they chose if its SAML or LDAP as part of the first step. They are then presented the corresponding set of options - it is *not* possible to switch type between SAML and LDAP after it has been created.

Under the covers, this adds a `type` field to the `SAMLProviders` table to indicate if its a `saml` or `ldap` configuration. I've avoided renaming the table for now.

I believe this represents the right minimal set of options that should work for common LDAP servers. It *should* work for MS Entra - but I haven't got that fully setup to test against yet. This is based on research of other applications that provide similar ldap integration.

I have tested this against a local OpenLDAP docker instance, with both TLS and non-TLS enabled.

image

Remaining tasks:

 - [x] docs
 - [ ] improved test coverage (don't really want to have to run up an ldap server to test against every time...)
 - [ ] manual testing with MS Entra


knolleary:
@hardillb whilst I haven't fully testing with Entra yet, I believe there's enough here to review in the mean time. I've tried to structure the commits to help organise the changes.

docs/admin/sso/ldap.md
```diff
@@ -0,0 +1,57 @@
...
+You can save the configuration at any time by clicking the `Update configuration`
+button. The configuration will only be enabled when you tick the `active` checkbox
+and save the changes.
+
+
+#### User Search Filter
+
+The search filter is used when checking if a user exists within the directory, using
+the standard LDAP query notation. The default search filter is `(uid=${username})` - where
+`${username}` will be replaced with the user's username when they attempt to login.
```
hardillb, line 56: As disscussed

I think `username` should be the username set for the user in FlowFuse not just what was entered in the username box on the login page.

And we should support both `username` and `email` substitution in the filter

knolleary: Changes made

hardillb (approved):
Tested locally against my ldap (once I remembered that email was actually in the `mail` field)

Issue #3934, opened 2024-05-27
Refreshing the page while creating an application instance does not load the team correctly


cstns:
### Current Behavior

The team is not loaded after accessing the application > app > create instance page and reloading which prevents users from selecting instance types and instance stacks.

### Expected Behavior

Team data should be present after a page refresh.

### Steps To Reproduce

see current behaviour

### Environment

- FlowFuse version: 2.4.0
- Node.js version: N/A
- npm version: N/A
- Platform/OS: N/A
- Browser: N/A

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Closed by #3945

Issue #3884, opened 2024-05-17
Clean up UI/UX when creating an instance


ZJvandeWeg:
### Description

To increase level of UX when creating a new instance we should fix a few things:

## Better name selection

GitHub for example has better UX when typing what this thing does and is, it keeps the repository name empty at first and lets the user type:

![Screenshot 2024-05-17 at 15 42 49](https://github.com/FlowFuse/flowfuse/assets/2529595/2761a2a6-f655-4842-856e-776d1cc9f17f)

Suggestions it does make, but like so:

![Screenshot 2024-05-17 at 15 46 29](https://github.com/FlowFuse/flowfuse/assets/2529595/b5c0adb9-2df8-49bb-9abb-9436b0e5665d)

## Blueprint selection

I think the Blueprint selection should be about the last thing to select, not the first thing. The name selection should be first. Especially as the blueprint selected doesn't prefill the form fields after them.

## Hide unavailable sizes

Come to think of it, maybe an instance size should just be bound to the tier. Either way; for the time being we should hide "Small" here:

![Screenshot 2024-05-17 at 15 47 36](https://github.com/FlowFuse/flowfuse/assets/2529595/8369f7c7-f04c-4b37-8a22-678009a869d4)

## Stack

The latest stable Node-RED version is selected, which is great. I think if a user wants to deviate, it should look and feel more like changing the blueprint. 

![Screenshot 2024-05-17 at 15 49 58](https://github.com/FlowFuse/flowfuse/assets/2529595/8ec665fd-aa90-4873-b3c8-fd112fe4d02a)
![Screenshot 2024-05-17 at 15 51 35](https://github.com/FlowFuse/flowfuse/assets/2529595/cec4a374-bac2-42b3-b94f-e7e146f7306b)

### Epic/Story

https://github.com/FlowFuse/flowfuse/issues/3839

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate
```[tasklist]
### Tasks
- [ ] https://github.com/FlowFuse/flowfuse/issues/3894
- [ ] https://github.com/FlowFuse/flowfuse/issues/3895
- [ ] https://github.com/FlowFuse/flowfuse/issues/3896
- [ ] https://github.com/FlowFuse/flowfuse/issues/3914
- [ ] https://github.com/FlowFuse/flowfuse/issues/3918
- [ ] https://github.com/FlowFuse/flowfuse/issues/3929
[...]

PR #3945, opened 2024-05-28
Fix instance creation form getting stuck without a team after refresh

cstns:
## Description

Extended the loading screen state to include the time it takes to load the teams.

The issue was caused by a combination of factors:
- we are passing the team from the createInstance.vue page down to the InstanceForm 
- there is no persistent state for teams in local storage and each full page refresh needs to call all API endpoints again
- creation form also loads teams

There is a short moment after the page re-loads and the team call is not finalized but the InstanceForm gets mounted with a null team prop and fails it's created method's api calls due to the team being null.

## Related Issue(s)
#3884 
#3934

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/application/createInstance.vue
+++ b/frontend/src/pages/application/createInstance.vue
@@ -138,6 +138,9 @@ export default {
             delete createPayload.isHA
 
             return instanceApi.create(createPayload)
+        },
+        qwe () {
+
         }
     }
 }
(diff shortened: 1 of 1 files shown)


joepavitt:
Does this close #3934?

joepavitt (approved):

joepavitt:
Just need confirmation of whether it's closing #3934, or just related to, can then close.

PR #3952, opened 2024-05-29
Revert billing_cycle_anchor change to fix coupons

### LINKED ISSUE #3950: Cannot apply coupons on checkout when billing_anchor_date set
### Current Behavior

We recently updated our stripe setup to anchor a customer's billing date to the 1st of the month.

This has broken our ability to apply coupons on checkout.

Documented [here](https://docs.stripe.com/payments/checkout/billing-cycle?locale=en-GB#limitations), it is not currently possible to apply an `amount_off` type coupon on checkout if the `proration_behaviour` is set to `create_prorations`.

We have three options:

1. revert the billing date change - billing date will be aligned to their sign-up date
2. abandon coupons - not a real option in terms of business case
3. modify the checkout to set `proration_behaviour` to `none`. This will mean no partial invoice is generated for the period between sign-up and the billing date. Essentially, they get that period for free.

On the assumption we need coupons to still work, the choice is between option 1 and 3.

I need to play a bit more to understand the full behaviour of option 3. For example, if a user has 2xSmall instances and they choose to upgrade to Team tier, does that mean they get those 2 small instances for 'free' for the remainder of the month - and what if they then add a 3rd instance... what does their bill look like then.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Fixes #3950 - reverting the billing_cycle_anchor change as I don't see any other practical way to support both that and coupons.

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -35,24 +35,6 @@ module.exports.init = async function (app) {
         return null
     }
 
-    /**
-     * @returns {Date} The next month start date in UTC
-     */
-    function getNextMonthStartUTC () {
-        const now = new Date()
-        let year = now.getUTCFullYear()
-        let month = now.getUTCMonth() + 1
-
-        // If this is December, increment year and set month to January (0)
-        if (month === 12) {
-            year++
-            month = 0
-        }
-
-        // Create a new Date object for the first day of the next month in UTC
-        return new Date(Date.UTC(year, month, 1))
-    }
-
     return {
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):
Not pulled/tested locally but I have confirmed this reverts back to state of play prior to https://github.com/FlowFuse/flowfuse/pull/3748

PR #3970, opened 2024-05-31
Add port details to Device Agent docs

knolleary:
Add more explicit details on what domains/ports the Device Agent will require access to when working with FF Cloud


ZJvandeWeg (approved):

ZJvandeWeg:
Thanks and sorry

PR #3981, opened 2024-06-05
Docs update for custom hostnames

knolleary:
Updates custom hostname docs - adds dedicated page and clarifies use of subdomain.

Frontend updated to point at new docs page

(also a drive-by update to the SSO docs as I noticed it needed updating)

### DIFF
--- a/forge/routes/api/settings.js
+++ b/forge/routes/api/settings.js
@@ -44,6 +44,9 @@ module.exports = async function (app) {
                 base_url: app.config.base_url,
                 license: app.license.status()
             }
+            if (app.config.features.enabled('customHostnames')) {
+                response.cnameTarget = app.config.driver.options?.customHostname?.cnameTarget
+            }
 
             if (request.session.User.admin) {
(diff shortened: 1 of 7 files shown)


forge/routes/api/settings.js
```diff
@@ -44,6 +44,9 @@ module.exports = async function (app) {
                 base_url: app.config.base_url,
                 license: app.license.status()
             }
+            if (app.config.features.enabled('customHostnames')) {
+                response.cnameTarget = app.config.driver.options?.customHostname?.cnameTarget
```
hardillb, line 48: Does this want to be a little more explicit that it's to do with custom hostnames?

hardillb (approved):
Only question if cnameTarget on it's own in the UI settings object is a little vauge

forge/routes/api/settings.js
knolleary: It's fine for now. We are free to move it later if we want.

PR #3991, opened 2024-06-06
Snapshot compare dx

cstns:
## Description

Implements review changes from #3978

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/pull/3978

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/components/dialogs/AssetCompareDialog.vue
```diff
@@ -0,0 +1,158 @@
...
+<script>
+
+import FlowRenderer from '@flowfuse/flow-renderer'
+
+import SnapshotsApi from '../../api/snapshots.js'
+
+import Alerts from '../../services/alerts.js'
+
+export default {
+    name: 'FlowViewerDialog',
```
Steve-Mcl, line 62: this is same name as the other dialog (causes confusion when debugging)

frontend/src/components/dialogs/AssetCompareDialog.vue
```diff
@@ -0,0 +1,158 @@
...
+            this.changeIndex = (this.changeIndex - 1 + this.changes.length) % this.changes.length
+            this.changes[this.changeIndex].highlight()
+        },
+        cleanup () {
+            while (this.$refs.compareViewer?.firstChild) {
+                this.$refs.compareViewer.removeChild(this.$refs.compareViewer.firstChild)
+            }
+            while (this.$refs.viewer?.firstChild) {
+                this.$refs.viewer.removeChild(this.$refs.viewer.firstChild)
+            }
```
Steve-Mcl, line 148: Where is the `viewer` element?  I only see [compareViewer](https://github.com/FlowFuse/flowfuse/pull/3991/files#diff-7b95391d1d30ff5519730ac8213cbdfb75523aeb454ae3826526e037b57da997R42) element

frontend/src/components/dialogs/AssetCompareDialog.vue
```diff
@@ -0,0 +1,158 @@
...
+        }
+    },
+    setup () {
+        return {
+        /**
+         * Shows the compare flows dialog and presents the user with a list of snapshots to compare against
+         * @param {{flows: { flows :[]}}} v1Snapshot - A snapshot object as the base for comparison
+         * @param {[{label: String, value: String}]} snapshotList - A list of snapshots to compare against where label is the snapshot name and value is the snapshot id
+         */
+            showCompareSnapshots (v1Snapshot, snapshotList) {
```
Steve-Mcl, line 76: Since the whole comparison functionality is broken out into its own dialog, is it necessary to use the long descriptive function name?

When this was part of asset dialog it made sense to use a long descriptive name.  It feels superfluous here!

I would probably rename this to just `show` as per asset viewer dialog.

Will leave it up to you.

Steve-Mcl:
Overall, i understand now the use of props to avoid unnecessary recomputing however i am less sure about making this its own dialog.  I fear there will be more duplication when the asset viewer gets the ability to display the modules list and instance settings (we will need to do that also on this new asset compare dialog)

The above is only an observation and I will not block this PR based on that - but rather it may sway the direction of this?

frontend/src/components/dialogs/AssetCompareDialog.vue
cstns: It must have slipped by me when I duplicated the component. I'll fix it

frontend/src/components/dialogs/AssetCompareDialog.vue
cstns: It must have slipped by me when I duplicated the component. I'll fix it

cstns:
We can avoid duplication by extracting common functionality into separate mixins and only duplicating the template where needed. I didn't pay much attention to it this time round, I admit!

cstns:
On a related note, I found a way to get rid of the setTimout function on the render function. It was causing async issues when creating instances on the createInstance form in the BlueprintSelectorDialog on the deployBlueprint landing page.

I realized that  we're using the HTMLDialogElement show method which causes concurrency issues when rendering. It is a best practice when using vue to use only vue when manipulating DOM elements.

A side effect of this is that all dialogs are present at all times in pages where they are used (but hidden with display:none) and: 
- unnecessarily loading the DOM
- creating difficulties when testing with cypress because we have to pick the exact dialog we're referring to
- adding state management problems in which certain methods/watchers are being called unnecessarily on these modals even though they are not in use

Calling child component methods from the parent is similar to calling parent component methods from the child component in that it tightly couples the components together and is considered an anti-pattern. The recommended way is for two way parent-child is props passed down to child components by the parent and events emitted by child components and caught by parent components.

cstns:
One example I can give, which was my own doing (I now realize), is for the blueprint tiles. Each blueprint tile has the AssetDetailDialog inside.

This implies that, with our current configuration, we have a hidden AssetDetailDialog present for each blueprint tile present on the page. 

At the moment there are no real implications on this other than the fact that we have multiple hidden modals. But if we were passing props, or the dialog was using computed store props or state keys and was reacting via watchers/computed props, it would unnecessarily add computational overhead to page renders.

What I've done in https://github.com/FlowFuse/flowfuse/pull/3994/files#diff-0d94818a781a7387946662af02b3ed79ce7d1999841488af4d4985121cb3565b addresses this issue by conditionally rendering the modal in the parent. It's a small change, with the implication that the modal's state is determined externally from the dialog which somewhat breaks encapsulation but it resolves the aforementioned side effects because the dialog is not mounted until it is decided by the parent.

I feel that this is out of scope for this task, but I'll have to address it soon.

cstns:
Created https://github.com/FlowFuse/flowfuse/issues/4028 to address the above

cstns:
I addressed all the inline comments and tests are passing locally. Please give the UI another go as a precaution, I didn't spot anything wrong

frontend/src/components/dialogs/AssetCompareDialog.vue
```diff
@@ -0,0 +1,156 @@
...
+            this.changeIndex = (this.changeIndex - 1 + this.changes.length) % this.changes.length
+            this.changes[this.changeIndex].highlight()
+        },
+        cleanup () {
+            while (this.$refs.compareViewer?.firstChild) {
+                this.$refs.compareViewer.removeChild(this.$refs.compareViewer.firstChild)
+            }
+            while (this.$refs.compareViewer?.firstChild) {
+                this.$refs.compareViewer.removeChild(this.$refs.compareViewer.firstChild)
+            }
```
Steve-Mcl, line 146:
```suggestion
            while (this.$refs.compareViewer?.firstChild) {
                this.$refs.compareViewer.removeChild(this.$refs.compareViewer.firstChild)
            }
```

Am I missing something? Why twice?

Steve-Mcl (approved):

PR #3994, opened 2024-06-07
Add a new instance creation landing page

### LINKED ISSUE #3973: Users with accounts should be able to try out existing blueprints by using a configured URL
Authenticated users should be able to access a url that contains the blueprint id and be taken to an instance creation form.

The url format:
`<FlowFuse.domain>/deploy/blueprint?blueprintId=<xxx>`

At this moment in time the user's default team will be used and the user will have the option to select the application where the new instance will be created

The landing page is implemented as a distinct page which allows for greater flexibility in the sense that we could integrate gtag tags when reaching this stage.

cstns:
## Description

Add a new landing page where users can create instances with predefined blueprints.
Adds the ability for existing users to create an instance with a pre-defined blueprint when following an external URL.

## Related Issue(s)

closes #3973

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


cstns:
something weird is happening. Some e2e tests are failing in cypress cli but not in the browser session

joepavitt:
> something weird is happening. Some e2e tests are failing in cypress cli but not in the browser session

Often an ordering problem, i.e. the cli runs _all_ tests, so some leftovers from previous tests influence future tests. In the browser, you just run the tests you're interested in.

cstns:
yes, that is the case. I narrowed it down to the fact that we're somehow ending up with two default blueprints in the test DB.

cstns:
I was close. We don't end up with two default blueprints, but in a suite of blueprint tests that's running prior to the landing page ones, we're changing the default blueprint. This is a side effect of the fact that we're not refreshing the DB between test specs.

I'll need to intercept the response on the failing ones

frontend/src/components/dialogs/BlueprintSelectorDialog.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <ff-dialog ref="dialog" data-el="blueprint-selector-dialog" class="blueprints-selector-dialog" header="Choose a Blueprint">
+    <ff-dialog ref="dialog" data-el="blueprint-selector-dialog" class="blueprints-selector-dialog" header="Choose a Blueprint" style="display: block;">
```
joepavitt, line 2: ??

cstns: It's best explained here: https://github.com/FlowFuse/flowfuse/issues/4028

joepavitt: Right, but that's not in scope of this task, and doesn't explain why you've got hardcoded in-line `style` tag here?

cstns: The hardcoded inline `style` tag was just laziness on my behalf, I'll address it shortly.

#4028 contains the a more generic explanation of why I propose this approach when it comes to dialogs, but for this case in particular the setTimeout on the flowRenderer was causing problems on the instance form because although the dialog was hidden, it was still present in the DOM and was reacting to state changes.

In this particular case, It was trying to render flows even when it didn't have a DOM element to attach itself to right after creating an instance before the page would have been redirected to the instance details page.

Steve-Mcl:
@cstns I have read the linked issue but it not clear to me what this PR is doing. There are no screenshots or details about the included works.

To help me sign off on this, could you point me to screenshots and/or stepwise "how to" or something that explains what this is all about?

My apologies if I am missing something obvious.

cstns:
> My apologies if I am missing something obvious.

No need, you did not miss anything, I didn't create a feature demo just yet! 

I'll try to address it today

frontend/src/pages/landing/DeployBlueprint.vue
joepavitt, line 1: Not sure I see the need for this file? We have a lot of duplicate content in here from `team/createInstance.vue` - why not just re-use this?

cstns: I felt that altering the `team/createInstance.vue` component would possibly add too much complexity, but indeed, I ended up with a lot of duplication. I'll see if I can switch to it easily

cstns: successfully switched to the team/createInstance.vue

frontend/src/components/dialogs/BlueprintSelectorDialog.vue
cstns: I removed the inline style attribute

Steve-Mcl:
> * [ ]  Suitable unit/system level tests have been added and they pass
> * [ ]  Documentation has been updated

@cstns this Pr does have tests but the check box in the OP is not checked. Are there more tests to come?

Also, are there any docs and/or screenshots that need updating in the docs to accompany this?

---

Side note: I have clicked update branch on GH UI to ensure this is most up to date before review.

joepavitt (approved):

PR #3996, opened 2024-06-07
Export/import Blueprints

### LINKED ISSUE #3969: Blueprint import/export APIs
Need API endpoints:

 - Download/Export single or multiple blueprints
 - Import single or multiple blueprints
     - Ben raised points around merging/overriding
     - I vote first iteration: just create copy with a `(1)` on the end of it - iterate, iterate, iterate

hardillb:
fixes #3969
## Description

Adds API to export or import Blueprints as an admin

## Related Issue(s)

#3969 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/ee/routes/flowBlueprints/index.js
```diff
@@ -243,6 +243,94 @@ module.exports = async function (app) {
...
+        const newBlueprints = request.body
+        const existingBlueprints = await app.db.models.FlowTemplate.getAll()
+        const response = {
+            count: 0,
+            blueprints: []
+        }
+        for (const nbp of newBlueprints.blueprints) {
+            // check for clash
+            let clash = false
+            for (const existing of existingBlueprints.templates) {
```
cstns, line 313: My only concern is this nested for loop which is an O(n²) notation. If you don't think it will cause problems it can be ignored

cstns (commented):
See inline comment

forge/ee/routes/flowBlueprints/index.js
hardillb: I can't think of a better way to do it so as to check all the names for duplicates.

We can't remove them from the "existing" list as there may be multiple matches (but that would still end up with name clashes in the newly generated names).

cstns (approved):

forge/ee/routes/flowBlueprints/index.js
cstns: We'll change it if needed

PR #4037, opened 2024-06-17
Revert incorrect flowforge -> flowfuse

hardillb:
## Description

This should not have been changed.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/docs/install/docker/stacks.md
+++ b/docs/install/docker/stacks.md
@@ -44,7 +44,7 @@ FROM flowfuse/node-red
 WORKDIR /usr/src/node-red
 RUN npm install node-red-dashboard
 
-WORKDIR /usr/src/flowfuse-nr-launcher
+WORKDIR /usr/src/flowforge-nr-launcher
 ```
 
 To build the container run the following:


ZJvandeWeg:
@hardillb When is that being changed though? We shouldn't be using `flowforge` anymore.

ZJvandeWeg (approved):
@joepavitt Can you make sure the follow up task is created?

joepavitt (approved):

hardillb:
@ZJvandeWeg No this will not change as it will break any users custom stack containers.

joepavitt:
so, even new users are still having to use `flowforge`?

hardillb:
Yes

joepavitt:
We aren't called `flowforge` anymore, so this shouldn't be the case. Why will it break for users with custom stack containers? It's a breaking chage we will need to introduce at some point, otherwise our re-brand isn't fulfilled

hardillb:
That ship has already sailed, the time to do it was when we changed the container names from `flowforge/...` to `flowfuse/...`

The only time we could do it now is when we ship v3.0

joepavitt:
okay, which repo is appropriate to ensure we track this?

hardillb:
FlowFuse/helm & FlowFuse/docker-compose

PR #4044, opened 2024-06-19
Fix posthog csp header

knolleary:
The [posthog docs ](https://posthog.com/docs/advanced/content-security-policy)recommend setting `*.posthog.com` in the CSP header rather than particular subdomains - so that is what we should do.

### DIFF
--- a/forge/forge.js
+++ b/forge/forge.js
@@ -277,23 +277,17 @@ module.exports = async (options = {}) => {
                 }
             }
             if (runtimeConfig.telemetry?.frontend?.posthog?.apikey) {
-                let posthogHost = 'app.posthog.com'
-                if (runtimeConfig.telemetry.frontend.posthog.apiurl) {
-                    posthogHost = new URL(runtimeConfig.telemetry.frontend.posthog.apiurl).host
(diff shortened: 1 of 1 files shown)


forge/forge.js
```diff
@@ -277,23 +277,17 @@ module.exports = async (options = {}) => {
...
-                const posthogConnect = [
-                    posthogHost,
-                    'eu.i.posthog.com'
-                ]
                 if (contentSecurityPolicy.directives['connect-src'] && Array.isArray(contentSecurityPolicy.directives['connect-src'])) {
-                    contentSecurityPolicy.directives['connect-src'].push(...posthogConnect)
+                    contentSecurityPolicy.directives['connect-src'].push(posthogHost)
                 } else {
-                    contentSecurityPolicy.directives['connect-src'] = posthogConnect
+                    contentSecurityPolicy.directives['connect-src'] = posthogHost
```
cstns, line 290:
```suggestion
                    contentSecurityPolicy.directives['connect-src'] = [posthogHost]
```

cstns (approved):

Issue #3999, opened 2024-06-10
Users without accounts should be redirected to the prefilled try blueprint page after registration


cstns:
Follow up on #3973.

Redirects users that access the blueprint landing page and follow the registration process to continue their instance creation with the pre-defined blueprint.

Issue #3946, opened 2024-05-28
Allow direct linking to Create Instance view with blueprint pre-selected


knolleary:
Pre-requisite for https://github.com/FlowFuse/flowfuse/issues/3917

It should be possible to have a url that opens the create instance view with a particular blueprint pre-selected.

This would enable the marketing website to provide 'try this blueprint' button without requiring the user to reselect the blueprint.

The normal create instance form url includes the team slug. We would need a URL that doesn't include a team slug, which can then redirect to the user's default team create instance page.

There are two scenarios:

1. user is already logged in. Can go straight to create form for their *default* team.
2. user is not logged in. Should redirect once they log in
3. user has no account. Not quite sure if we would be able to redirect after account creation *and* email verification steps... that's a lot of steps to get through
```[tasklist]
### MVP
- [ ] https://github.com/FlowFuse/flowfuse/issues/3973
- [x] Unauthenticated users should be redirected to the try blueprint page after authenticating successfully
- [ ] https://github.com/FlowFuse/flowfuse/issues/3999
```

```[tasklist]
### Follow On Tasks
- [ ] https://github.com/FlowFuse/flowfuse/issues/3974
```

cstns:
We could leverage the browser's localStorage to store any required data and redirect the user after completing all signup steps.

cstns:
I'll add a series of tasks so we can split this feature into atomic deliverable chunks.

cstns:
Would also like to offer the user the possibility to select the team where to use the blueprint. This could be done after the gross of the implementation is done.

cstns:
Pre-selecting blueprints will always be pointing to an existing blueprint so it can safely be a GET request as it eases implementation as opposed to https://github.com/FlowFuse/flowfuse/issues/3917 deploy flows functionality where we are somewhat forced to use POST in order to avoid url length restrictions.

Given that the GET request will be somewhat visible/customer facing should we pay attention to how the route will look like? Should consider SEO even though the FF app is behind an auth wall and also an SPA?

I'm going with `<FlowFuse.domain>/instance/create?blueprintId=<xxx>` for development purposes but `/blueprint/use` / `blueprint/try/` variations would make more sense if links pointing back th the FF app are relevant.

@knolleary, any thoughts?

joepavitt:
Will close this as the primary tasks are complete, #3974 is a follow on task that can be scheduled in it's own time

Issue #3917, opened 2024-05-23
"Deploy to FlowFuse" Button


ZJvandeWeg:
### Description

For a snapshot or Flow.json it would be great if there was a "Deploy to FlowFuse" button to allow faster deployment to FlowFuse.

Things to consider
- Integration with FlowViewer
- Putting this button on flows.nodered.org

### Which customers would this be available to

None

joepavitt:
Core driver for this feature is improving virality of product, and reach the button would have

PR #4020, opened 2024-06-13
users without accounts should be redirected to the blueprint landing page after registration

cstns:
## Description
Adds support for unregistered users to successfully follow the instance creation process when accessing a blueprint landing page

Adds support for 

## Related Issue(s)

closes #3999
depends on #4043 
depends on #4045 #4050 

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #4045, opened 2024-06-19
Add smtp server for e2e testing

cstns:
## Description

Adds a simple smtp server for e2e tests that cypress can interact with to grab email confirmation links/password change, etc

## Related Issue(s)

Pre-requisite for https://github.com/FlowFuse/flowfuse/issues/3999
part of https://github.com/FlowFuse/flowfuse/issues/3946
part of https://github.com/FlowFuse/flowfuse/issues/3917

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/test/e2e/frontend/cypress/support/commands.js
+++ b/test/e2e/frontend/cypress/support/commands.js
@@ -1,3 +1,5 @@
+import 'cypress-mailpit'
+
 // ***********************************************
 // This example commands.js shows you how to
 // create various custom commands and overwrite
--- a/test/e2e/frontend/environments/smtp.js
+++ b/test/e2e/frontend/environments/smtp.js
@@ -1,0 +1,49 @@
+const childProcess = require('child_process')
+
+const mailpitLogger = (data) => {
+    console.info(`Mailpit: ${data}`)
+}
+
+module.exports = async function (config = {}) {
+    let { webPort, smtpPort } = config
+
+    webPort = webPort || 8025
+    smtpPort = smtpPort || 1025
+
+    mailpitLogger('Starting e-mail server...')
+
+    const dockerCommand = 'docker'
+    const args = [
+        'run',
+        '--rm',
+        `-p=${webPort}:8025`,
+        `-p=${smtpPort}:1025`,
+        '-e=MP_MAX_MESSAGES=5000',
+        '-e=MP_SMTP_AUTH_ACCEPT_ANY=1',
+        '-e=MP_SMTP_AUTH_ALLOW_INSECURE=1',
+        'axllent/mailpit'
+    ]
+
+    const dockerProcess = childProcess.spawn(dockerCommand, args)
+
+    dockerProcess.on('spawn', () => {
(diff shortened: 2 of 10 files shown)


test/e2e/frontend/cypress/support/commands.js
```diff
@@ -1,3 +1,5 @@
+import 'cypress-mailpit'
```
hardillb, line 1: Does this actually do anything?

test/e2e/frontend/environments/smtp.js
```diff
@@ -0,0 +1,49 @@
...
+    const args = [
+        'run',
+        '--rm',
+        `-p=${webPort}:8025`,
+        `-p=${smtpPort}:1025`,
+        '-e=MP_MAX_MESSAGES=5000',
+        '-e=MP_SMTP_AUTH_ACCEPT_ANY=1',
+        '-e=MP_SMTP_AUTH_ALLOW_INSECURE=1',
+        'axllent/mailpit'
+    ]
```
hardillb, line 25: Should we be starting this as a child process or using a GHA docker service (and have people start it manually for local testing?) 

See how we use a Postgres container here https://github.com/FlowFuse/flowfuse/blob/0d5e6ec5e778ffb7aace9e34294787df1421177c/.github/workflows/tests.yml#L98

hardillb (commented):
Just a couple of questions

test/e2e/frontend/environments/smtp.js
cstns: That was my initial thought as well, but we need the service to be available when we're testing locally as well without adding additional complexity for developers.

test/e2e/frontend/cypress/support/commands.js
cstns: not yet, it just adds convenience methods for cypress to interact with the smtp api. The commands will be used when creating tests in #4020

hardillb: OK, but I'm surprised this passes the lint test if it's not being used yet.

test/e2e/frontend/environments/smtp.js
hardillb: OK, so only other question is does this clean up properly when the tests exit? I see the `--rm` so the container will get removed when shutdown, and I assume the child process dies with the test process which triggers an exit?

test/e2e/frontend/cypress/support/commands.js
cstns: It must be because I'm not storing anything in unused const/let just extending

test/e2e/frontend/environments/smtp.js
cstns: Yes, the process is killed when the parent process is because it's not running in a detached state. The `--rm` is just a precaution

hardillb (approved):

PR #4054, opened 2024-06-20
Security Settings - Improve language & hide user/pw fields when not used

joepavitt:
## Description

- Was taking a screenshot for an article of the security settings, and the "username"/"password" fields are distracting when they're not in use. This changes it so that they only show when you have "Basic Auth" option selected.
- Also changes the language for the FF User Auth option to be a little clearer

## Related Issue(s)

No linked issue, small PR


joepavitt:
### Before

Screenshot 2024-06-20 at 11 51 27

### After

Screenshot 2024-06-20 at 11 51 03

Steve-Mcl (approved):

PR #4016, opened 2024-06-11
Ensure target editor title doesn't get overwritten

hardillb:
fixes #3920

## Related Issue(s)

#3920

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #4059, opened 2024-06-21
Fix Pipeline overwriting name

### LINKED ISSUE #4058: Instance name is copied over when deploying DevOps Pipelines
### Current Behavior

In DevOps Pipelines the first instance (name ending in 2741) is the source instance and the one ending in 4231 is the destination one. When deploying, the flows are deployed correctly but the name of the destination instance is now the same as the other one, which is confusing

image

### Expected Behavior

The name of the destination instance should be the same as before (the one ending in 4231)

### Steps To Reproduce

Create both instances, link them in DevOps Pipelines, and deploy from 2741 to 4231

### Environment

- FlowFuse version: Launcher version: 2.5.2-5843274-202406170826.0
- Node.js version: 20.14.0
- npm version:
- Platform/OS: cloud
- Browser: Chrome

hardillb:
fixes #4058 

## Description

Looks like the logic got flipped during the changes to https://github.com/FlowFuse/flowfuse/pull/4016

## Related Issue(s)

#4058

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/controllers/ProjectTemplate.js
+++ b/forge/db/controllers/ProjectTemplate.js
@@ -231,7 +231,7 @@ module.exports = {
         const result = JSON.parse(JSON.stringify(existingSettings))
         const skipList = ['disableEditor', 'page_title', 'header_title']
         templateFields.forEach((name) => {
-            if (mergeEditorSettings && skipList.includes(name)) {
+            if (!mergeEditorSettings && skipList.includes(name)) {
                 return
             }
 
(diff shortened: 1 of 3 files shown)


knolleary:
I disagree. The logic should be: if !merge then overwrite everything. If merge, we skip if on the skip list.

Is the problem a layer up where this function is being called by the pipeline with a `false` rather than a true` for the merge flag?

We need a unit test to verify the behaviour.

hardillb:
@knolleary ready for another look

knolleary (approved):
Looks good. Added just the test case to my local and it failed (as expected). Then applied the fix and it passes (as expected).

PR #4076, opened 2024-06-29
Remove trickle down router props

cstns:
Remove trickle down router props and cohesive way of using team, teamMembership and isVisitingAdmin

## Description

Focusing mainly on teamMembership & team props which can be used directly from the permissionsMixin, or using accessing the account store directly where the permission mixin is not used. 

The isVisitingAdmin computed prop has also been moved in the permissionsMixin

## Related Issue(s)
none

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/application/createInstance.vue
+++ b/frontend/src/pages/application/createInstance.vue
@@ -58,7 +58,6 @@ import instanceApi from '../../api/instances.js'
 import NavItem from '../../components/NavItem.vue'
 import SideNavigation from '../../components/SideNavigation.vue'
 import applicationMixin from '../../mixins/Application.js'
-import permissionMixin from '../../mixins/Permissions.js'
 import Alerts from '../../services/alerts.js'
 import InstanceForm from '../instance/components/InstanceForm.vue'
 
@@ -69,7 +68,7 @@ export default {
         NavItem,
         SideNavigation
     },
-    mixins: [applicationMixin, permissionMixin],
+    mixins: [applicationMixin],
     inheritAttrs: false,
     props: {
         sourceInstanceId: {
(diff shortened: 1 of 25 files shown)


joepavitt:
I won't be able to review this today, so if someone else can step in, would appreciate it. My main point of concern is that if a user is logged in and viewing Team 1, and they directly navigate in the browser to a Team 2 page (e.g. viewing instances) that the new view reflects the _correct_ team's data

cstns:
I added two new tests to cover these specific scenarios

frontend/src/pages/application/createInstance.vue
```diff
@@ -68,7 +69,7 @@ export default {
         NavItem,
         SideNavigation
     },
-    mixins: [applicationMixin],
+    mixins: [applicationMixin, permissionMixin],
```
joepavitt, line 72: I can't see where this is used in this component? We use the `team`, but that's already imported in the `mapState` directly, so need for the mixin here.

joepavitt (changes requested):
You're missing an opportunity for the `permissionsMixin` in `pages/application/Overview.vue`

Steve-Mcl (approved):
@joepavitt I went through this line by line with Serban . We found 1 left over prop and fixed that.  

did a full review on pre-staging, tested all pages touched (with dev tools console open) and direct url access to alt teams etc. 

I am happy to approve but will not merge as you may see something i dont

joepavitt:
@Steve-Mcl thanks Steve, just left my own comments on this too

Steve-Mcl (approved):
all good from my pov.

frontend/src/pages/application/createInstance.vue
cstns: Good catch! It doesn't seem to be used anywhere

joepavitt (approved):

PR #4092, opened 2024-07-03
Update frameguard to sameorigin

knolleary:
Current helmet/frameguard policy is set to `deny` - which prevents the app being embedded in an iframe.

This *might* be causing issues with the immersive editor in production - yet it works on staging. Raising PR to get this change (from `deny` to `sameorigin`) and test in a dev env.


knolleary:
Need to get this onto production - it's the only way to ultimate test it properly. Can always revert it necessary is it won't impact anything other than the immersive editor.

PR #4097, opened 2024-07-03
Handle httpAdminRoot in template when calculating dashboard link

### LINKED ISSUE #4093: Setting the Editor Path in the Template breaks the Dashboard link
### Current Behavior

This is an extension of #3926 it appears that setting the editor path at the template level rather than in the instance settings still causes the wrong path to be generated.

last fix: https://github.com/FlowFuse/flowfuse/pull/3971

### Expected Behavior

Be able to set the editor path both on the instance and the template and have the correct dashboard path be calculated.

### Steps To Reproduce

- create template with editor path set
- create instance
- install Dashboard v2 into instance
- follow link to dashboard from instance overview

### Environment

- FlowFuse version: 2.5.0
- Node.js version: 20.x
- npm version: na
- Platform/OS: all
- Browser: all

knolleary:
Fixes #4093 

When calculating the dashboard link, this now takes into account if the template sets `httpAdminRoot`, but the individual instance hasn't.

### DIFF
--- a/frontend/src/pages/instance/components/DashboardLink.vue
+++ b/frontend/src/pages/instance/components/DashboardLink.vue
@@ -56,9 +56,10 @@ export default {
                 return
             }
             let baseURL = removeSlashes(this.instance.url, false, true)
+            const adminRootSetting = this.instance.settings?.httpAdminRoot || this.instance.template?.settings?.httpAdminRoot
             // Check to see if the editor path has been set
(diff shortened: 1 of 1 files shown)


knolleary:
I checked the FF Team Auth for the dashboard with `httpAdminRoot` set in the template and not the instance - all appeared fine. This is because the launcher only sees the consolidated settings - it doesn't need to care where they came from.

hardillb (approved):

Issue #4078, opened 2024-06-30
Logged in users that manipulate the url and remove everything after the application domain land on a blank page.


cstns:
### Current Behavior

Logged in users that manipulate the url and remove everything after the application domain land on a blank page.

### Expected Behavior

Users should be taken to their default team applications page

### Steps To Reproduce

Log in the platform, remove everything after the application domain and press enter.

### Environment

- FlowFuse version: 2.5.0
- Node.js version: N/A
- npm version: N/A
- Platform/OS: N/A
- Browser: Firefox

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

knolleary:
I cannot reproduce this myself. When logged in, if I open a new tab to `app.flowfuse.com`, or edit the url in an existing tab to that, then I end up on my default team's page as expected.

cstns:
Apparently It's only happening on Firefox. Did not notice it until now

cstns:
I updated the description

knolleary:
Hmm, working for me with Firefox as well.

cstns:
That's weird.. Here's how I reproduce it consistently

https://github.com/FlowFuse/flowfuse/assets/20094749/5adaa480-26f5-4aa9-ab64-7e7965fc7065

I cannot replicate it on chrome though.

ZJvandeWeg:
@cstns I ran into it with Chrome today

knolleary:
Can anyone who reproduces this check the browser console for any errors?

cstns:
https://github.com/FlowFuse/flowfuse/assets/20094749/5f628d35-4cbe-4c97-bef0-46122f8530a0

This ignoreSurvey.js seems to be interacting with localStorage. Might be clashing with the recent vuex persistence in localStorage ?

cstns:
The redirectUrlAfterLogin url is not cleared after the redirect occurred and that causes side effects in other locations in the app. This is now a problem because we're storing the redirectUrlAfterLogin in localStorage and it doesn't get cleared on page refresh.

The fix contains a 3 part fix:
- clear the `redirectUrlAfterLogin` after a route was accessed
- attempt to redirect to the `redirectUrlAfterLogin` if it's present when accessing the HomePage
- fix logo not doing anything when clicked after reaching this state

PR #4106, opened 2024-07-04
4078 blank page after redirect

cstns:
## Description

The `redirectUrlAfterLogin` url is not cleared after the redirect occurred and that causes side effects in other locations in the app. This is now a problem because we're storing the `redirectUrlAfterLogin` in localStorage and it doesn't get cleared on page refresh

Three part fix:
- making sure that the `redirectUrlAfterLogin` is cleared after the route is hit for logged in users
- making sure that users that have the `redirectUrlAfterLogin` get redirected when they reach the home page
- fix the FF home logo not following any path when clicked in this state

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4078

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/PageHeader.vue
+++ b/frontend/src/components/PageHeader.vue
@@ -76,7 +76,7 @@ export default {
     mixins: [navigationMixin],
     computed: {
         ...mapState('account', ['user', 'team', 'teams']),
-        ...mapGetters('account', ['notifications', 'defaultUserTeam']),
+        ...mapGetters('account', ['notifications']),
         navigationOptions () {
             return [
                 {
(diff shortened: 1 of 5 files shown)


knolleary:
Looks better testing locally. One odd bit:

1. non-logged in User navigates to team url they do not have access to
2. they log in
3. they are shown the 404 page
4. the 404 page is missing the team select dropdown. They only thing they can click on is the FF logo
5. Clicking the FF logo takes to 'home' - but is blank.
6. Reload the page, all looks okay again

Missing team select:
image

I think we can ship this PR as-is (unless I'm missing something) so we can get today's release moving. The items I've identified above can raised separately and fixed after today.

knolleary (approved):

PR #4115, opened 2024-07-05
PostHog - Add .group call to automatically map all events to the relevant team

joepavitt:
## Description

Following chat with Greg this week, want to be more team-centric in our analysis on PostHog. These changes enable that.

- Adds in tier data for each team
- centralises the team assignment
- Calls `.group` within a wrapped `setTeam` function. This runs every time user logs in, or switches team, all events in that session _after_ this are then automatically mapped to the relevant team.
- This will mean we can do team-centric analysis on all insights in PostHog

Team is attached to _every_ event after as the property "Group 1" (it's just how PostHog tracks the concept of "groups")

Context: [PostHog Docs](https://posthog.com/docs/product-analytics/group-analytics)

### DIFF
--- a/frontend/src/store/account.js
+++ b/frontend/src/store/account.js
@@ -6,6 +6,7 @@ import settingsApi from '../api/settings.js'
 import teamApi from '../api/team.js'
 import userApi from '../api/user.js'
 import router from '../routes.js'
+import product from '../services/product.js'
 
 // initial state
 const state = () => ({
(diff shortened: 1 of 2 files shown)


joepavitt:
Must be related to my changes, as it's failed two times in a row, but my other PR passed first time

frontend/src/store/account.js
```diff
@@ -115,6 +116,8 @@ const mutations = {
         state.user = user
     },
     setTeam (state, team) {
```
knolleary, line 118: Is it possible that setTeam is being called with `null`? Suggest wrapping in try/catch to rule out.

joepavitt:
You called it Nick - was the tests for when a user is invited to FF Team, but currently doesn't have a team

knolleary (approved):

PR #4147, opened 2024-07-09
Run e2e tests locally without spinning smtp servers

cstns:
## Description

add the ability to run e2e tests locally without spinning the docker smtp server processes

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


cstns:
the only downside to this approach is that e2e tests that require the smtp server will always fail locally when running the `cy:web-server-light` script

Steve-Mcl:
> the only downside to this approach is that e2e tests that require the smtp server will always fail locally when running the `cy:web-server-light` script

Could the same env var not be used to `.skip` the failing tests?

package.json
```diff
@@ -37,6 +37,7 @@
         "test:docs": "node test/e2e/docs/valid-links.js ./docs",
         "test:system": "mocha 'test/system/**/*_spec.js' --timeout 10000 --node-option=unhandled-rejections=strict",
         "cy:web-server": "npm-run-all --parallel cy:web-server:os cy:web-server:ee",
+        "cy:web-server-light": "cross-env NO_SMTP_SERVER=true npm-run-all --parallel cy:web-server:os cy:web-server:ee",
```
Steve-Mcl, line 40: Instead of yet another script, can we not simply do the inverse?

i.e.
* local tests always runas they always used to
* tests on CI detect the flag and do run
* If local wants to run smtp tests then local sets an ENV VAR e.g. `SMTP_TESTS=true`

cstns: Can't think of a solution where we'd get away with not duplicating an existing script, but I'm open to suggestions.

We'd need to be able to run the e2e tests locally with and without the smtp server (this applies to all three scenarios running os, running ee and running them together) while CI needs relies on a smtp service (so no smtp server required via docker)

> > the only downside to this approach is that e2e tests that require the smtp server will always fail locally when running the `cy:web-server-light` script
> 
> Could the same env var not be used to `.skip` the failing tests?

This would be a good approach but would come with some caveats.

Running the e2e tests locally without the smtp server and associated tests would result in an incomplete run and would render incomplete result. That's why I think there should be another script or method that would run the entire suite with the smtp server enabled locally.

cstns:
closing, handled differently in https://github.com/FlowFuse/flowfuse/pull/5457

PR #4155, opened 2024-07-10
Add session ownerId/type to log output if set

knolleary:
Adds info on who is making http requests when using an access token rather than user session. For device/team tokens, ensures the hashid is provided. Project tokens are uuids already.

### DIFF
--- a/forge/routes/api/assistant.js
+++ b/forge/routes/api/assistant.js
@@ -63,6 +63,22 @@ module.exports = async function (app) {
 
         // post to the assistant service
         const headers = {}
+        if (request.session?.ownerType) {
+            switch (request.session.ownerType) {
+            case 'team':
+                headers['ff-owner-type'] = 'team'
+                headers['ff-owner-id'] = app.db.models.Team.encodeHashid(request.session.ownerId)
+                break
+            case 'device':
+                headers['ff-owner-type'] = 'device'
+                headers['ff-owner-id'] = app.db.models.Device.encodeHashid(request.session.ownerId)
+                break
+            case 'project':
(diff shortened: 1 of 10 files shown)


Steve-Mcl (approved):
Approved in the interest of time with the knowledge `ownerType` and `ownerId` are (currently) only set to specific values in the src, however we should have tests that ensure IDs are alphanumeric.

forge/routes/api/assistant.js
```diff
@@ -63,6 +63,22 @@ module.exports = async function (app) {
...
+                headers['ff-owner-id'] = app.db.models.Team.encodeHashid(request.session.ownerId)
+                break
+            case 'device':
+                headers['ff-owner-type'] = 'device'
+                headers['ff-owner-id'] = app.db.models.Device.encodeHashid(request.session.ownerId)
+                break
+            case 'project':
+            case 'instance':
+                headers['ff-owner-type'] = request.session.ownerType
+                headers['ff-owner-id'] = request.session.ownerId
```
Steve-Mcl, line 79:
```suggestion
            case 'instance':
                headers['ff-owner-type'] = request.session.ownerType
                headers['ff-owner-id'] = request.session.ownerId
                break
            default:
                // Don't log the id as we don't know how to hash it
                // Log the type so we can spot cases we aren't handling and address it
                response.request.ownerType = reply.request?.session?.ownerType
```

Apply same default as logging

Steve-Mcl: discussed. not required.

Steve-Mcl (approved):

Issue #4157, opened 2024-07-11
Search/Filter applications based on instance/device name


cstns:
Follow up on https://github.com/FlowFuse/flowfuse/issues/4142

Add the ability to search/filter applications based on device/instance names.

Use cases:
- search query matches application name whilst no device/instance match will render the application with a full list of instances/devices
- search query doesn't match application name but matches instance name but no device match will render the application with empty devices and matching instance
  - the inverse applies for devices
- search query doesn't match application name but matches both instance and device name will render the application with filtered instance and device
- the `x more...` tile should react accordingly based on filtered results

Part of https://github.com/FlowFuse/flowfuse/issues/4014

cstns:
The only thing that worries me is that I had to remove the ASSOCIATIONS_LIMIT that restricted app instances/devices count in order for the search function to work on entities that are hidden by the 'x more...' tile.

cstns:
implemented via https://github.com/FlowFuse/flowfuse/pull/4159

PR #4159, opened 2024-07-11
Search applications by instances and devices

cstns:
## Description

Add the ability to search through an application's devices and instances

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4157

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/api/team.js
+++ b/forge/routes/api/team.js
@@ -292,7 +292,7 @@ module.exports = async function (app) {
         const includeInstances = true
         const includeApplicationDevices = true
         const associationsLimit = request.query.associationsLimit
-        const includeApplicationSummary = !!associationsLimit
+        const includeApplicationSummary = !!associationsLimit || request.query.includeApplicationSummary
 
         const applications = await app.db.models.Application.byTeam(request.params.teamId, { includeInstances, includeApplicationDevices, associationsLimit, includeApplicationSummary })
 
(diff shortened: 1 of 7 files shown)


cstns:
The only thing that worries me is that I had to remove the ASSOCIATIONS_LIMIT that restricted app instances/devices count in order for the search function to work on entities that are hidden by the 'x more...' tile.

Steve-Mcl (approved):
Very nice. 

I do have one super simple suggestion regarding UX:

When I search for something and realise there are more matching my search criteria, when I click `More` I expected my search term to be carried over.  Instead, i had to re-type my search term.

chrome_owWRavH3Tt


Carrying my search term to the next page would feel much more consistent.

Otherwise, excellent.

Steve-Mcl:
Sorry, one more thing.

Do you have access to the ID and Type when performing the search?

I tried searching `NUC` (a type value) and `nkDqdOYgxM` (a device ID) - but the search did NOT reveal any results whereas on the next page it does.

chrome_1Fxicw5DL7

cstns:
Thank you for the positive feedback!

> Carrying my search term to the next page would feel much more consistent.

I have to admit I did not think of that use case, there's added value in it!

> Do you have access to the ID and Type when performing the search?

Yes, the search function can be extended easily to include id search for app/instance/device.

I'll create a follow up task to include them

PR #4163, opened 2024-07-11
Only permit instance/device token to call upon assistant endpoint

Steve-Mcl:
## Description

* Ensures assistant endpoint requests are made by a device or instance token only
* Removes unnecessary permission `"assistant:method"` now that we limit to devices/instances via `preHandler` 
* Adds test to ensure user tokens are rejected (401)

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/lib/permissions.js
+++ b/forge/lib/permissions.js
@@ -109,10 +109,7 @@ const Permissions = {
     'platform:debug': { description: 'View platform debug information', role: Roles.Admin },
     'platform:stats': { description: 'View platform stats information', role: Roles.Admin },
     'platform:stats:token': { description: 'Create/Delete platform stats token', role: Roles.Admin },
-    'platform:audit-log': { description: 'View platform audit log', role: Roles.Admin },
-
-    // assistant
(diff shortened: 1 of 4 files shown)


Steve-Mcl:
> I think this means you can remove the `needsPermission` check (and the associated permission from `permissions.js` as all the useful checks are already done.

Thanks for the tip. I'll check that out and update accordingly.  2 mins.

Steve-Mcl:
> > I think this means you can remove the `needsPermission` check (and the associated permission from `permissions.js` as all the useful checks are already done.
> 
> Thanks for the tip. I'll check that out and update accordingly. 2 mins.

Quite right you were.

I have removed all traces in commit [remove unnecessary permission assistant:method now that we limit to d…](https://github.com/FlowFuse/flowfuse/pull/4163/commits/e43da5e3cbeda832f5f052acf1e7685489efbf6f)

knolleary (approved):

Issue #4113, opened 2024-07-04
Eng. Design: Notifications Schema


joepavitt:
We should establish an agreed upon schema for the `Notifications` objects to be stored in the database, and used around the application.

Some considerations:

- Short Title
- Long form description/detail
- Actions: probably a link to somewhere in the application (or external to docs/change logs?)
- Source:
    - Device - events linked to unexpected device state changes
    - Instance - events linked to unexpected Instance state changes
    - User? - Team invites may be linked to the sending user?
    - Team - Team invites may be linked to a team?
    - Platform/FlowFuse Team - Notifications manually triggered by us, detailing scheduled downtime, or new features that users may want to see?
- Timestamp
- State
    - Unread
    - Read
    - ???

knolleary:
Having doing some research on how this type of feature is implemented elsewhere, I've come to the following high-level design points. tl;dr: a similar model to our audit log can be used

1. We will have a well-defined (in code) list of notification types.
2. The notification table (exact schema to follow) will store, `user`, `type`, `content` (json blob), `created_at` (sequelize default column), `unread` (boolean)
3. Each notification type will have a Vue component that renders the notification according to a template using the `content` property.

The `content` property, as a json blob, will contain any notification-specific meta data - for example, in the case an a team invite, it can contain team id/name, inviter name, role. The Vue component will have the template for how what title/description should be shown to the user for the notification - no need to store that in the database.

knolleary:
## External API

 - `GET /api/v1/user/notifications` - returns notifications, most recent first. Uses pagination as per standard design
 - `PUT /api/v1/user/notifications` - bulk mark notifications read. Will mark *all* read by default
     - body params:
         - `id` - csv list of notifications to mark as read (optional)
         - `unread` - `true`/`false` (optional)
 - `PUT /api/v1/user/notifications/:id` - mark notifications read - defaults to marking all read for the user
     - body params:
         - `unread` - `true`/`false`
 - `DELETE /api/v1/user/notifications/:id` - delete a notification
 - `DELETE /api/v1/user/notifications` - bulk delete
     - body params:
         - `id` - csv list of notifications to delete (required)

knolleary:
As the primary task is complete, closing this off - and moving on to #4158 to do implementation

PR #4164, opened 2024-07-11
Initial user notification backend

### LINKED ISSUE #4158: Add server-side support for notifications (API/Database)
See https://github.com/FlowFuse/flowfuse/issues/4113 for design details

knolleary:
Closes #4158 

## Description

Implements the backend design for user notifications - as detailed in https://github.com/FlowFuse/flowfuse/issues/4113

## Database

New `Notification` model added, including db migration to create the table:

```
CREATE TABLE Notifications (
    id        INTEGER       PRIMARY KEY AUTOINCREMENT,
    type      VARCHAR (255) NOT NULL,
    reference VARCHAR(255),
    read      TINYINT (1)   DEFAULT 0,
    data      TEXT,
    createdAt DATETIME      NOT NULL,
    UserId    INTEGER       REFERENCES Users (id) ON DELETE CASCADE
                                                  ON UPDATE CASCADE
);

```

## External API

 - `GET /api/v1/user/notifications` - returns notifications, most recent first. Uses pagination as per standard design
 - `PUT /api/v1/user/notifications/:id` - mark notification read
     - body params:
         - `read` - `true`/`false`
 - `DELETE /api/v1/user/notifications/:id` - delete a notification

### Not included

The spec in #4113 included bulk end points - I haven't implemented them in this PR as unnecessary for first iteration where we can make individual api requests to update individual notifications
 - `PUT /api/v1/user/notifications` - bulk mark notifications read. Will mark *all* read by default
 - `DELETE /api/v1/user/notifications` - bulk delete

## Runtime API

A bare bones API is provided to send a notification to a user. Not going to over specify this until work is done to start adding real notifications.

```
app.notifications.send(user, type, data, reference)
```
 - `user` - the User to send the notification to (a full `app.db.models.User` object)
 - `type` - `string` - the type of notification. This will be one from a well-defined list. How they get defined (and whether we add convenience functions like we have for the auditLog) can be determined as we start adopting it.
 - `data` - `object` - contextual data needed for the notification. This will get exposed on the API, so must not include raw database ids - only hashids.
 - `reference` - (optional) `string` - a reference that can be used to clear this notification later

```
app.notifications.remove(user, reference)
```
 - `user` - the User to remove the notification for (a full `app.db.models.User` object)
 - `reference` - `string` - the notification reference to remove

The idea behind `reference` is that we can associate the notification with *something* - such as the hashid of an invitation. When that invitation is processed, we can remove the notification.


Steve-Mcl (approved):
pulled and tested with curls locally. works as expected.

One suggestion would be to consider the `read` field - Is the a possibility of a future where notifications could be "paused", or "forwarded" or something other than `read: true/false`? Implementing this now would save a future migration. The original proposal did request a `status` field which would cater for such instances?

However, that said, not gonna block this.


One other suggestion, since this is very early days, it might be wise to mark the API as private to avoid any lock in (however unlikely that is at this early stage)

knolleary:
>  Is the a possibility of a future where notifications could be "paused", or "forwarded" or something other than read: true/false

None of the apis I've researched around notifications have multiple states like this. A notification is either read or unread. If we find a need for other states, they can be a new field that is independent of whether the user has 'seen' a notification or not.

knolleary:
Not going to merge this until I make a bit more progress with #4165 to help validate the assumptions in this PR

knolleary:
Updated to include the new `reference` field to allow clearing of notifications.

PR #4195, opened 2024-07-17
Move to email verification codes rather than links on signup

### LINKED ISSUE #4174: Require users to insert confirmation code on signup
Draft workflow in Figma: https://www.figma.com/board/mkOzyvy2b5Iu86xBpFaBZA/User-SignUp-Flows?node-id=526-263&t=dGhMaHUwpM9tfZvL-4

This keeps users in the same browser session, and enables better tracking for us to ensure users are onboarding correctly.

It also prevents the need for user to click (what may appear to be) dodgy URLs in their onboarding e-mails to verify.

knolleary:
Closes #4174 

This reworks our how we verify a user's email address in the account registration flow.

On registration, we would generate a signed JWT token that we email to the user as a url for them to follow. They then open the url, and are presented with a page to click a button to verify. They then had to login.

With this change, we now email a code to them and prompt them to enter the code on the screen shown after initial registration. This means they stay in the same browser session through-out - and are logged in automatically.

Demo video of it in action (internal link) https://flowforgeworkspace.slack.com/archives/C04GW82DJFK/p1721207608385039

 - The route `POST /account/verify/:token` was previously used by the button the user clicks on. That is now replaced with:
     - `POST /account/verify/token` (note `token` is now part of the url, not a param), which accepts the body of. `{ token }` - this receives the submitted verification token from the new verify page.

 - We now log the user in in response to the initial registration - this means we don't need the additional log-in step
 - If the email is sso-enabled, we still prompt them to login via SSO provider as that does the email verification step. This may change with the auto-provisioning users for SSO task that we'll look at soon.
 - The Change Email flow has not been modified - that still uses the email link approach. Changing that was out of scope for this task.


frontend/src/pages/UnverifiedEmail.vue
```diff
@@ -31,16 +38,41 @@ export default {
...
+            try {
+                await userApi.verifyEmailToken(this.token)
+                clearTimeout(this.resendTimeout)
+                window.location = '/'
+            } catch (err) {
+                // Verification failed.
+                this.token = ''
+                this.error = 'Verification failed. Click resend to receive a new code to try again'
+                clearTimeout(this.resendTimeout)
+                this.resentTimeout = 0
```
knolleary, line 59:
```suggestion
                this.resendTimeout = 0
```

forge/postoffice/templates/VerifyEmail.js
```diff
@@ -3,13 +3,16 @@ module.exports = {
     text:
 `Hello, {{{safeName.text}}},
 
-Use the link below to verify your email address.
+Please use this code to verify your email address:
 
-{{{ confirmEmailLink }}}
+{{{ token.token }}}
+
+Do not share this code with anyone else.
```
Steve-Mcl, line 10:
```suggestion
Do not share this code with anyone else.

This token will expire in 30 minutes.
```


Suggestion only: Inform user of expiry duration?

forge/postoffice/templates/VerifyEmail.js
```diff
@@ -3,13 +3,16 @@ module.exports = {
...
 `,
     html:
 `<p>Hello, <b>{{{safeName.html}}}</b>,</p>
-<p>Use the link below to verify your email address.</p>
-<p><a href="{{{ confirmEmailLink }}}">Confirm Email address</a></p>
+<p>Please use this code to verify your email address:</p>
+<p><b>{{{ token.token }}}</b></p>
+<p>Do not share this code with anyone else.
```
Steve-Mcl, line 16:
```suggestion
<p>Do not share this code with anyone else.</p>
```

Missing end tag

Steve-Mcl, line 16:
```suggestion
<p>Do not share this code with anyone else.</p>
<p>This token will expire in 30 minutes.</p>
```

Suggestion only: Inform user of expiry duration?

frontend/src/pages/UnverifiedEmail.vue
```diff
@@ -1,18 +1,25 @@
...
             <p>
                 Before you can access the platform, we need to verify your email
                 address.
             </p>
             <p>
-                We sent you an email with a link to click when you signed up.
+                We sent you an email with a code in it. Enter the code below to continue.
+            </p>
+            <div>
+                <ff-text-input v-model="token" data-form="verify-token" maxlength="6" label="token" @enter="submitVerificationToken" />
```
Steve-Mcl, line 12: Ideally this would have `autocomplete="one-time-code"` ([reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#values)) to prevent suggesting passwords and such

without correct autofill we get this: 
image

Steve-Mcl (commented):
On top of the comments, I dont see any docs update, in particular: [the signup workflow](https://flowfuse.com/docs/contribute/workflows/signup/)

frontend/src/pages/UnverifiedEmail.vue
knolleary: Noted - I'd like to improve the style of the OTC input as we have it in multiple places now. Will include this suggestion in that, rather than diverge.

knolleary:
Good spot on signup flow - will update that now. I don't *think* we have any user-facing docs that describe the sign-up process - but will double check.

knolleary:
@Steve-Mcl sign-up workflow doc updated. Cannot find any other docs that need updating.

docs/contribute/workflows/signup.md
```diff
@@ -21,17 +20,16 @@ sequenceDiagram
...
     end
 
     UI->>UI: Show 'Check your email' page
     UE-->>US: Email received
-    US->>+UI: Opens /account/verify/{token}
-    UI->>US: Displays page asking user to "Please verify your email"
-    US->>UI: Click "Verify my email" button
-    UI->>+RT: POST /account/verify/{token}
+    US->>+UI: Enters verification code
+    US->>UI: Click "Continue button
```
Steve-Mcl, line 31:
```suggestion
    US->>UI: Click "Continue" button
```

Steve-Mcl:
@knolleary 
Any specific timing on merge once tests complete?

knolleary:
@Steve-Mcl I'm going to merge this in the morning when I can also verify all is working as expected - I'm out of time to do that verification tonight.

PR #4202, opened 2024-07-18
Ensure team slug is passed for nested views

### LINKED ISSUE #4200: Inconsistent behaviour of "Invite Members" toolbar button
### Current Behavior

chrome_UCJacseJyx

* button only works on overview pages
* button does not work on the invites page (not too problematic as there is a duplicate in the table search bar that does work)

### Expected Behavior

Not sure, but as it is it only works on certain pages.

### Steps To Reproduce

visit any instance, application or device - button does not work

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

joepavitt:
## Description

When on a nested view, the `to.params.team_slug` isn't populated because the `team` slug doesn't exist in the nested URL. To be safe, we should pass this explicitly into the `$route.push()`

## Related Issue(s)

Closes #4200

### DIFF
--- a/frontend/src/components/PageHeader.vue
+++ b/frontend/src/components/PageHeader.vue
@@ -37,7 +37,7 @@
         </div>
         <div class="hidden lg:flex">
             <ff-team-selection data-action="team-selection" />
-            <div class="px-4 flex flex-col justify-center ff-border-left">
+            <div class="px-4 flex flex-col justify-center ff-border-left" v-if="showInviteButton">
                 <ff-button kind="secondary" @click="inviteTeamMembers">
                     <template #icon-left><UserAddIcon /></template>
                     Invite Members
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
Tested on pre-staging.

Button now works on most screens but still does not work on the screen where 2 of them are visible:

Additionally, I noticed the breakpoint button is not configured correctly (can raise as separate issue if I must)

YsEsLjeSe5

joepavitt:
> Button now works on most screens but still does not work on the screen where 2 of them are visible

Raise this as a separate issue as this is a _lot_ more hands on to try and fix.

We'd likely need to go down the Vue route to fix that, currently we just rely on the query parameter in the URL, and the `mounted` check

Steve-Mcl:
> > Button now works on most screens but still does not work on the screen where 2 of them are visible
> 
> Raise this as a separate issue as this is a _lot_ more hands on to try and fix.
> 
> We'd likely need to go down the Vue route to fix that, currently we just rely on the query parameter in the URL, and the `mounted` check

Would simply disabling the button when on this view not make more sense?

joepavitt:
> Would simply disabling the button when on this view not make more sense?

Done

Steve-Mcl (approved):

PR #4209, opened 2024-07-18
Bulk device delete api

### LINKED ISSUE #4208: Bulk Device Operations API (MVP: bulk delete)

Steve-Mcl:
closes #4208 

## Description

Adds bulk operation for deleting multiple devices in a team.

Basic premise:
1. Operates on a team level API 
   * I initially wrote this as a top level API `api/v1/devices/bulk` but there were difficulties with `needsPermission` and team membership validation (would have to be handled manually).  It made sense at this time to have this as a team level API as it is to be initially implemented in the UI at the device browser component (happy to move to top level if deemed better/necessary)
1. All devices specified in the `devices` body property **must** exist otherwise operation throws
3. All devices specified in the `devices` body property **must** must be in the same team otherwise operation throws
4. Calls device `destroy` on where `id` in `ids` as a singular operation
5. Calls `billing.updateTeamDeviceCount` is licenced
6. logs bulk delete count
   * only count of deletion at this time 

### Endpoint
* Adds endpoint `DELETE` `/bulk` to `forge/routes/api/teamDevices.js`
   * This equates to `/api/v1/team/:teamId/devices/bulk`
      * This is open to debate. I figured since we might one day have bulk update, bulk create etc - it feels clear and separate. Though in writing this up, it might be enough to simply leave the sub path at `/`  e.g. DELETE `/api/v1/team/:teamId/devices`
   * Device IDs are expected in the body as an array of hash-ids in the property `devices`
* Adds permission `"team:device:bulk-delete"`
* Adds team level audit log event handler

### Tests
* Adds audit logger test
```
  Audit Log > Team
    ✔ Provides a logger for bulk deleting team devices
```
* Adds route tests:
```
 Team Devices API
    Team Devices
      Supports bulk operations
        delete
          ✔ Non owner cannot delete devices
          ✔ Delete a single device
          ✔ Delete multiple devices
          ✔ Cannot delete device belonging to a different team
          ✔ Prevents deleting any devices mixed in with other team devices
```

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


knolleary (approved):
All looks good. My only pause has been the choice of api endpoint - whether having `/bulk` in the url is the right choice. I've spent some time researching other APIs to see how (and if) they support bulk operations.... and it's pretty inconclusive. Having a dedicated bulk endpoint (rather than overloading an existing api) does appear to be one common approach - so on the basis of that, lets roll with what's proposed here.

Issue #4184, opened 2024-07-16
Front-End: Add Notifications Inbox view


joepavitt:
As discussed, first iteration will be:

- Slide out tray (from right)
- Notifications icon - `inbox` from https://heroicons.com/
- Could be added as a square section between the team dropdown and the user icon, then that triggers the slide-out tray
- Update the team invite "notifications" to use the new schema/approach

PR #4210, opened 2024-07-18
Add notifications drawer

cstns:
## Description

Adds a notifications drawer boilerplate. Currently only supports team invitations but is easily extended to future notifications

## Related Issue(s)
closes https://github.com/FlowFuse/flowfuse/issues/4184

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/components/NotificationsButton.vue
```diff
@@ -0,0 +1,71 @@
...
+  .notifications-button {
+    border-left: 1px solid $ff-grey-500;
+    color: white;
+    display: flex;
+    align-items: center;
+    flex: 1;
+    justify-content: center;
+    width: 100%;
+    height: 100%;
+    padding: 10px;
```
joepavitt, line 48:
```suggestion
    padding: 18px;
```

frontend/src/components/NotificationsButton.vue
```diff
@@ -0,0 +1,71 @@
...
+    position: relative;
+
+    &:hover {
+      background-color: $ff-grey-700;
+    }
+
+    svg {
+      flex: 1;
+      width: 35px;
+      height: 35px;
```
joepavitt, line 58:
```suggestion
      width: 24px;
      height: 24px;
```

joepavitt (changes requested):
- Odd glitch on close where it seems to span the full width for a brief millisecond or two, then pull back?

https://github.com/user-attachments/assets/a6082067-0bcd-4ce3-bd85-6237dd8007d9

- I'd also expect to be able to click the notifications icon _again_ to close too, not just have to click into empty space.
- We should also change the red of the `ff-notification-pill` to `#d12c2b` as it's too dark at the moment
- We also need to improve the padding and vertical alignment of the notifications themselves:

Screenshot 2024-07-18 at 18 20 49

I'm happy to do that myself though if you don't mind handling the other recommendations

joepavitt:
As an FYI too, we have `ff-icon`, `ff-icon-lg` and `ff-icon-sm` which you can use here too. The icon in the notification itself for example can just use `ff-icon ff-icon-lg` and it'll then be 24px squared.

cstns:
I'm looking into the collapse bug! Feel free to make any adjustments you feel necessary!

joepavitt:
Made some structural improvmenets by having all of the structure live inside a `Notification.vue` component with relevant slots that the bespoke notifications can extend.

I've also then modified the CSS slightly to implement the following changes:

### Before

Screenshot 2024-07-19 at 12 36 54

### After

Screenshot 2024-07-19 at 12 34 07

joepavitt:
Waiting for tests to run - but will approve once they're successful

joepavitt (approved):

PR #4219, opened 2024-07-19
Add bulk delete for devices browser

### LINKED ISSUE #4218: Add multiple selection and bulk deletion to Device Browser
```[tasklist]
### Tasks
- [x] check boxes for each row
- [x] select all in header
- [x] trash button in device browser toolbar adjacent to search input (enable/disable based on selection)
- [x] e2e tests
- [x] update rbacs doc
```

Steve-Mcl:
closes #4218

## Description

* Adds support for checkbox column to table component
* Enables checkboxes in devices browser
* Adds Delete Bin icon in device browser toolbar

### Demo
chrome_8Wx6xgUHun

### TODO:

- [ ] Fix Checkbox styling in table (@joepavitt )
- [x] Tests
- [x] Docs

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
> * Fix Checkbox styling in table (@joepavitt )

@joepavitt this is now ready to switch out of draft - please pull and update your local branch and switch out of draft when you have had a moment to sort out the funky column sizing.

Steve-Mcl:
@joepavitt if you havent already pulled this local ignore.

I had this targeting main - I just re-targeted it as it was based off the API - you might want to pull local once more (sorry)

knolleary:
How scalable is listing all of the devices in the dialog? What happens if they have 50 devices to delete?

joepavitt:
Yeah, I'd say just say "Are you sure you want to delete N devices" - no need to list them explicitly.

Steve-Mcl:
> Yeah, I'd say just say "Are you sure you want to delete N devices" - no need to list them explicitly.

I disagree. I explicitly want to spell out what is to be deleted to avoid ambiguity (e.g. when a filter is applied AFTER selection).  I've been stung in other applications like this.

However, Nick was right to point out scalability of large number of devices - currently this is what a user would see:

image


However, with a simple parent div with max-height and auto scroll overflow, we can still show devices but in a contained manor.

image

Steve-Mcl:
> How scalable is listing all of the devices in the dialog? What happens if they have 50 devices to delete?

contained with a max-height auto scroll parent div in [contain device delete list to sensible size then scroll](https://github.com/FlowFuse/flowfuse/pull/4219/commits/42979d90aca872e10d282136e9556566e49c6cf5)

Steve-Mcl:
@joepavitt In an effort to move this along, do you want me to investigate the funky sizing on the new checkbox column and the existing kebab menu column?

joepavitt:
Sorry Steve, this completely slipped off my radar. Will ping you on Slack

joepavitt:
Functionally great, I've pushed fixes for the CSS alignment and table width problems, although seeing:

Screenshot 2024-07-24 at 11 59 31

in the console. The problem is caused by `<ff-checkbox v-model="checks[row[checkKeyProp]]" />` equating to `undefined`, as we need to ensure we've mapped the relevant rows into `checks[]` to at least show as `false`

Steve-Mcl:
> Screenshot 2024-07-24 at 11 59 31
> in the console. The problem is caused by `<ff-checkbox v-model="checks[row[checkKeyProp]]" />` equating to `undefined`, as we need to ensure we've mapped the relevant rows into `checks[]` to at least show as `false`.

I seen these too.  I didnt get around to reading up on this - my approach was going to be to look for something  like an "ignore" or "infer" directive for vue to ignore these warnings. Effectively, we don't _really_ care.  So long as "not true" can implicitly mean false (we dont care that it is undefined/null/whatever)?

joepavitt (approved):

PR #4221, opened 2024-07-19
Better sanitization of team names in invite email

knolleary:
## Description

Email clients like to render things that look like URLs as clickable links. We had taken some precautions against this, but some gaps still remained.

This PR sanitises the team name in the HTML email by inserting invisible `<br>` elements after any `.`.


cstns (approved):
Looking good:
image

PR #4207, opened 2024-07-18
Sso create new user

hardillb:
closes #4051
## Description

Allows SSO configurations to be configured to create new users when they sign in for the first time.

Currently only supports SAML SSO providers.

Adds new flag to SSO config.

## Related Issue(s)

#4051 

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #4237, opened 2024-07-22
Allow admin to be set from SSO group

### LINKED ISSUE #4085: Create the possibility to define FlowFuse Admins by AD group.
### Description

When SSO and Group Assertions is enabled, please  create the possibility to define which accounts are FlowFuse Admins by AD group.

Enterprise Tier Only (EE)

hardillb:
fixes #4085

## Description

Members of SSO group `ff-admin-admin` will be marked as FF admins

does require `admin` to be added to the list of teams. May want to do this with a checkbox rather than explicitly add it.

## Related Issue(s)

#4085 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/ee/lib/sso/index.js
```diff
@@ -323,8 +324,27 @@ module.exports.init = async function (app) {
                         }
                     }
                 }
+                if (providerOpts.groupAdmin && providerOpts.groupAdminName === ga) {
+                    adminGroup = true
+                }
             })
 
+            if (user.admin && !adminGroup) {
+                if (!Object.hasOwn(desiredTeamMemberships, 'admin')) {
```
knolleary, line 333: Not sure that `desiredTeamMemberships` is relevant here; `admin` isn't a team. What's the case this is handling?

forge/ee/lib/sso/index.js
```diff
@@ -323,8 +324,27 @@ module.exports.init = async function (app) {
                         }
                     }
                 }
+                if (providerOpts.groupAdmin && providerOpts.groupAdminName === ga) {
+                    adminGroup = true
+                }
             })
 
+            if (user.admin && !adminGroup) {
```
knolleary, line 332: We should only be touching admin if `providerOpts.groupAdmin` is set - otherwise we have to leave the flag alone.

As far as I can tell, this logic, if `groupAdmin` is false, will have left `adminGroup` set to `false` and the next piece will then strip them of the admin flag... unless that is what the desiredTeamMemberships check is handling? Would be good to make it a bit clearer (via more explicit logic, or comments).

knolleary:
Updating the branch to have all the latest goodness in after #4207 was merged to main. Will then test out fully

knolleary:
All looks good. Have pushed a small docs update to cover this feature in the SAML docs.

knolleary (approved):

knolleary:
Going to add docs for the create new user feature as well - then will approve and merge

knolleary:
Docs updated.

docs/admin/sso/saml.md
```diff
@@ -72,6 +72,18 @@ The general points are:
 Once you have setup both sides of the configuration you can enable it for use
 by ticking the `active` checkbox and clicking `Update configuration`.
 
+## Creating new users
+
+With FlowFuse 2.6, the SSO Configuration now includes an option to automatically
```
hardillb, line 77:
```suggestion
With FlowFuse 2.7, the SSO Configuration now includes an option to automatically
```

We've shipped 2.6

hardillb:
UI test failure is the bulk delete stuff.

knolleary:
I'm going to merge this. I cannot get the UI tests to fail locally on this branch. We can revert if the ui test failure continues on `main` after merge; but I can't see any reason that it would.

PR #4247, opened 2024-07-24
show full buttons when instance count is 1

### LINKED ISSUE #4199: Show full "Open Editor" text when only one instance
### Description

We're seeing low conversino of new users opening the Editor, and with the recent change in the compressed view for Applications page, the "Open Editor" button is more hidden than it was before:

Now:

Screenshot 2024-07-17 at 18 13 54

I propose that, if we only have a single instance/device in an application, then we should be showing the full descriptive action for the "Open Editor" button

Screenshot 2024-07-17 at 18 14 32

Steve-Mcl:
closes #4199

## Description

Show full buttons when only 1 instance in application condensed view.

<details><summary>Outdated</summary>
<p>

**superseded - see threaded comments below**

image

</p>
</details> 

### TODO - follow up issue to raise once merged: 

- [ ] Buttons no longer provide a visual cue as to whether they are internal (immersive) or external (pop-out)

## Related Issue(s)

Closes #4199

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/instance/components/DashboardLink.vue
+++ b/frontend/src/pages/instance/components/DashboardLink.vue
@@ -6,17 +6,21 @@
         :disabled="buttonDisabled"
         @click.stop="openDashboard()"
     >
-        <template v-if="showExternalLink" #icon-right>
-            <ExternalLinkIcon />
+        <template v-if="showText" #icon-right>
+            <ChartPieIcon />
         </template>
-        <slot name="default">
+        <template v-else #icon>
+            <ChartPieIcon />
+        </template>
+        <template v-if="showText">
             Dashboard
-        </slot>
+        </template>
     </ff-button>
 </template>
 
 <script>
-import { ExternalLinkIcon } from '@heroicons/vue/solid'
+
+import { ChartPieIcon } from '@heroicons/vue/outline'
 
 // utility function to remove leading and trailing slashes
 const removeSlashes = (str, leading = true, trailing = true) => {
@@ -31,7 +35,7 @@ const removeSlashes = (str, leading = true, trailing = true) => {
 
 export default {
     name: 'DashboardLink',
-    components: { ExternalLinkIcon },
+    components: { ChartPieIcon },
     inheritAttrs: false,
     props: {
         disabled: {
@@ -46,7 +50,7 @@ export default {
             default: null,
             type: Object
         },
-        showExternalLink: {
+        showText: {
             type: Boolean,
             default: true
         }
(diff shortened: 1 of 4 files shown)


joepavitt:
To be consistent on iconography, I think we should be including the nodes/pie icons in the "full" buttons too and removing the "eternal link" icon

Steve-Mcl:
> To be consistent on iconography, I think we should be including the nodes/pie icons in the "full" buttons too and removing the "eternal link" icon

hmmm. but then wouldnt that make them inconsistent with the instance view (these buttons are the originals, only being allowed to surface)

Also, as a user, i prefer to know if the button opens internally or externally (the PIE/flow icon dont tell me that) - perhaps some other visual cue could be employed alongside the iconography Joe?

joepavitt:
We can introduce a new visual clue for new tab after this, I'd rather have consistency in the iconography. The Pie/Flow icon should be used there these buttons are present

Steve-Mcl:
> The Pie/Flow icon should be used there these buttons are present

Can I assume you meant to say **"The Pie/Flow icon should be used where these buttons are present"** ?

In which case:

* This Pr update should apply the iconography at the base level (meaning all instance/device overviews will get same button + icons)?
* Additionally, we would also need to change the existing show/hide behaviour of the button icon (Currently Edit Flows button shows/hides its icon based on internal/external linking)
   * i.e. Always show the icons and raise an Issue around visual cues for internal/external?

joepavitt:
> (Currently Edit Flows button shows/hides its icon based on internal/external linking)

Not sure I understand this, but fundamentally, yes, they should be consistent.

Steve-Mcl:
> > (Currently Edit Flows button shows/hides its icon based on internal/external linking)
> 
> Not sure I understand this,

V4.0.2 Node-RED shows the "Editor" button without link-out icon (to signify it is an internal link / immersive)
Older NR versions that do not support immersive editor have the link-out icon.

image

image






> yes, they should be consistent.

So then this PR needs updating to:
* apply the iconography at the base level of the component (meaning all instance/device overviews align with the Applications Overview - all will get same button + icons)
* change the existing show/hide behaviour of the "Editor" button to always show icon regardless of NR version / immersive 
* Raise a follow-up Issue around visual cues for internal/immersive vs external/non-immersive

Steve-Mcl:
Updated so that the underlying buttons use the same icon throughout all application/instance views

chrome_HOrQMgGPzM

Steve-Mcl:
@joepavitt ready for review.

joepavitt (approved):

Issue #4074, opened 2024-06-27
Improve handling of expired licenses


knolleary:
### Description

We don't do *much* if a platform license expires. We should improve that as we are increasingly providing shorter-term trial licenses to customers.

WE check the license state in two places:

1. On startup
2. A daily timer task that runs at 2am. If the license has expired, this task will email the admins of the platform on the day of expiry and every subsequent Sunday.

If the license has expired, then we show this banner to all users on all views:

<img width="392" alt="image" src="https://github.com/FlowFuse/flowfuse/assets/51083/1c294427-f5d7-421a-aee5-14bdf26e8484">

However we do not limit any functionality at this point.

We need to decide how far we want to go on license expiry. Some options are:

 - Remove access from EE features entirely
 - Add intrusive dialog prompts on certain EE-only actions - still allow them to use the feature
 - Add more noisy logging

Open to suggestions on what direction to take in terms of balancing business requirements around trial licenses.

```[tasklist]
### Tasks
- [x] Add admin-only warning banner if license expires within 7 days
- [x] Add daily audit log entry if the license has expired (https://github.com/FlowFuse/flowfuse/pull/4268)
- [x] Suspend all instances if license has expired (https://github.com/FlowFuse/flowfuse/pull/4268)
- [x] Prevent instance resume if license has expired (https://github.com/FlowFuse/flowfuse/pull/4268)
- [x] Prevent resource creation if license has expired (https://github.com/FlowFuse/flowfuse/pull/4268)
- [x] Block broker ACL checks if license has expired (https://github.com/FlowFuse/flowfuse/pull/4268)
- [x] Ensure applying a valid license then unblocks everything
- [ ] Release blog post content - warn of shutdown if upgrading with expired license
```

hardillb:
If we add the ability to remove features when license expires can we also look at not needing a restart after applying a license?

joepavitt:
This needs to kill resource - otherwise, what incentive does a user have to come back to us? They just get free FF indefinitely?

knolleary:
I believe the consensus here is, if the license expires, we:

 - suspend all instances
 - do not allow suspended instances to be resumed
 - update ACL checks on the broker to cut off devices

We need to make sure all alerting/logging/messaging we provide in the days leading up to expiry make it very clear what will happen.

Will update the task list above

joepavitt:
[...]

PR #4268, opened 2024-07-29
Shutdown Instances when license expires

hardillb:
part of #4074

## Description

- suspend all running projects when license expires
- prevent projects from being started
- audit log every day the license is expired

## Related Issue(s)

#4074

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
Need to work out how to write tests for this lot.

Can we mock `app.license`?

knolleary:
A key part of #4074 is adding good notifications in the lead up to license expiry. That is the first item on the todo list, and following the discussion in that issue, something we need in place before we start shutting things down.

We cannot merge this part with out that piece in place as well.

knolleary:
@hardillb this is still in draft - not sure the current state. Please see my comment above - we need to cover the notifications prior to expiry.

knolleary (dismissed):
All looks good so far.

Just need the additional warnings prior to expiry before we can merge.

hardillb:
@knolleary still draft because I wasn't sure we'd decided on just when to show the warnings. Will have another look when I've done with the nr-launcher file api and Steve's comments on the CSV logs

test/unit/forge/licensing/index_spec.js
```diff
@@ -403,4 +427,117 @@ describe('License API', async function () {
...
+                )
+            } catch (err) {
+                return
+            }
+            should.fail()
+        })
+        it('should not allow new device to be created', async function () {
+            await app.license.apply(TEST_LICENSE_EXPIRED)
+            try {
+                const device = await factory.createDevice(
```
hardillb, line 531:
```suggestion
                await factory.createDevice(
```

joepavitt (approved):

Issue #3818, opened 2024-05-03
Rename Snapshots


joepavitt:
### Description

Split from https://github.com/FlowFuse/flowfuse/issues/3509

Want the ability to rename a Snapshot, after it has been created.

### Which customers would this be available to

None

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate
```[tasklist]
### Tasks
- [ ] https://github.com/FlowFuse/flowfuse/issues/4281
- [ ] https://github.com/FlowFuse/flowfuse/issues/4283
- [ ] https://github.com/FlowFuse/flowfuse/issues/4286
- [ ] https://github.com/FlowFuse/website/issues/2408
```

knolleary:
All tasks complete - closing off

Issue #3358, opened 2024-01-18
Implementation - Automated Snapshot creation on Flow Deployment for Devices


MarianRaphael:
### Description

This feature introduces an automated snapshot creation system triggered by the generation of an "Flow Deployed" audit log entry. Each snapshot will be automatically named with the prefix [auto], followed by the current date and time. This feature includes a number of auto snapshots retained; older snapshots beyond this limit will be automatically deleted, provided they are not currently in use as target snapshots.

## User Story
https://github.com/FlowFuse/flowfuse/issues/2423

## Scope

- Trigger Condition: Snapshot creation occurs upon the generation of an "Flow Deployed" audit log entry.
- Naming Convention: Snapshots will be named following the format [auto]Date+Time, incorporating the date and time of creation.
- The maximum number of auto snapshots retained is 10.
- Snapshot Retention: Snapshots currently designated as target snapshots will be exempt from automatic deletion.
- For Devices only

### Which customers would this be available to

Team + Enterprise Tiers (EE)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate
```[tasklist]
### Requirement
- [ ] https://github.com/FlowFuse/flowfuse/issues/3357
```

Steve-Mcl:
@MarianRaphael please clarify

Is this for Application assigned devices only (or am I to do the work for instance assigned devices too)?

MarianRaphael:
For Application assigned devices only

Steve-Mcl:
@MarianRaphael 

> Naming Convention: Snapshots will be named following the format [auto]Date+Time, incorporating the date and time of creation

Do we want to consider a unified format?

![Image](https://github.com/FlowFuse/flowfuse/assets/44235289/3fa651fb-004d-418a-b1cd-7b85eb3a2be1)


The pipelines snapshot look like this

![Image](https://github.com/FlowFuse/flowfuse/assets/44235289/3ffde963-c1d0-4dbd-b507-63b1d2e72909)

Steve-Mcl:
@knolleary @MarianRaphael 

The `ProjectSnapshots` table expects a valid `UserId` (though it does accept `null` )

Since the snapshot is _kinda_ generated by the system (in response to a flows deploy audit log entry), we dont have a user to set against the snapshot. 

Is this a blocker? 
Should we look to first update auth routines on device agent to report the logged in user back to FF in the API calls?
Perhaps we should we fake it and put a System user in the missing icon? Other?

[...]

PR #4282, opened 2024-07-30
Edit snapshot

### LINKED ISSUE #4281: Add Edit snapshot capability to all snapshot views

Steve-Mcl:
closes #4281

## Description

**TLDR;**
Introduces a common dialog and supporting endpoint for editing snapshots.

**Implementation**
* All snapshot table views (instance, application, device) now have the kebab menu item "Edit Device".
* When clicked, a form with `Name` and `Description` are pre-populated with the current values from the snapshot.
* Validation
   * Name has non empty validation only
   * "Confirm" button should be disabled when name is invalid
   * Description has no validation (can be blank)
* Upon clicking Confirm, the API is called with the new Name/Description
   * upon success the form is closed, a "success" toast is shown, and the table is updated (client side only)
   * upon failure, an "error" (warning) toast is shown and the dialog remains open to permit a retry without having to re-type

### Finer detail
* Adds `put` to the main route of the top-level snapshot API
  * expects `data` to contain `name` and/or `description` only at this time
* Implements `updateSnapshot` in controller `forge/db/controllers/Snapshot.js`
* Adds permission `snapshot:edit` that is used for both FE and BE routes
* Adds FE API route `updateSnapshot` to `frontend/src/api/snapshots.js`
* Adds shared dialog `frontend/src/components/dialogs/SnapshotEditDialog.vue`
* Updates kebab in the instances, devices and applications snapshots views (see demo GIF)

### Unit Tests added in `test/unit/forge/routes/api/snapshots_spec.js`
```
    Update snapshot
      instance
        ✔ Returns 404 for non-existent snapshot
        ✔ Non-member cannot update snapshot (72ms)
        ✔ Owner can update snapshot (52ms)
        ✔ Can update name only (51ms)
        ✔ Can update description only (53ms)
        ✔ Member cannot update snapshot (39ms)
      device
        ✔ Returns 404 for non-existent snapshot
        ✔ Non-member cannot update snapshot (44ms)
        ✔ Owner can update snapshot
        ✔ Can update name only
        ✔ Can update description only
        ✔ Member cannot update snapshot
```

### e2e Test added in `test/e2e/frontend/cypress/tests/applications/snapshots.spec.js`
```
FlowForge - Application - Snapshots
  ✔ provides functionality to edit a snapshot
```

### e2e Test added in `test/e2e/frontend/cypress/tests/instances/snapshots.spec.js`
```
FlowForge - Instance Snapshots
  ✔ provides functionality to edit a snapshot
```

### e2e Test added in `test/e2e/frontend/cypress/tests-ee/devices/snapshots.spec.js`
```
FlowForge - Devices - With Billing
  ✔ provides functionality to edit a snapshot
```

### Demo
chrome_qOr84XRdzp

### TODO
- [x] Tests (unit and e2e)

## Related Issue(s)

Owner #4281
Parent #3818

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
How does this impact any snapshots deployed to a device? Is there a need to push updates at that point? I'm assuming we're just running from Snapshot ID, so no need?

Steve-Mcl:
> How does this impact any snapshots deployed to a device? Is there a need to push updates at that point? I'm assuming we're just running from Snapshot ID, so no need?

Hmmm, I was 100% certain there was no need to update the device until you mentioned this 🤔 

_scurries off to double check_

Steve-Mcl:
> > How does this impact any snapshots deployed to a device? Is there a need to push updates at that point? I'm assuming we're just running from Snapshot ID, so no need?
> 
> Hmmm, I was 100% certain there was no need to update the device until you mentioned this 🤔
> 
> _scurries off to double check_

@joepavitt thanks for flagging this.

So, the device does get `FF_SNAPSHOT_NAME` and as this PR stands, the device is not informed/not restarted meaning the env var is out of step.

This is however purely informational, and does not impact operation of the device.

The question becomes, do we _need_ to perform a device update just because of name change?  If we do, then it would be advisable to consider the UX (e.g. do we check ALL devices/instances associated with the snapshot, do we issue restart device automatically?)

It might be as well to toast info to the user (and update docs) to state something like:
_"Renaming a snapshot that is in use will not cause device to automatically restart however they will need an update at some point to synchronise the `FF_SNAPSHOT_NAME` environment variable."_

@knolleary any thoughts please?

knolleary:
Agreed this *should not* trigger a device restart - if I'm doing housekeeping to mark a snapshot as 'good', I don't want it to redeploy and restart everything.

I would not use a toast notification either. It can be a simple one-line in the dialog (so the user knows this information *before* they make the change) to say something like "Changes to the snapshot will not be reflected on devices already running the snapshot".

Steve-Mcl:
Note to reviewers: 

the TODO for audit logging [in lines 195-201](https://github.com/FlowFuse/flowfuse/pull/4282/files#diff-b0a947e68f7754337961ad87c95a14ba4565d8ffbdbbda9e5bb3a7713e8a0608R195-R201) of this PR are handled in the incoming branch PR #4287 [here](https://github.com/FlowFuse/flowfuse/pull/4287/files#diff-b0a947e68f7754337961ad87c95a14ba4565d8ffbdbbda9e5bb3a7713e8a0608R196-R209)

Steve-Mcl:
> would not use a toast notification either. It can be a simple one-line in the dialog (so the user knows this information _before_ they make the change) to say something like "Changes to the snapshot will not be reflected on devices already running the snapshot".

Will do that. Thanks.

Steve-Mcl:
Just hit a speed bump.

image

### Issue:

In short, because I renamed an auto snapshot, its name no longer matches the prescribed format and we can no longer infer this is a system auto snapshot.

We currently infer a snapshot was an auto snapshot due to lack of flags in db  (this has been mentioned in previous GH issue as being an MVP and will require a DB flag in the future).  It may be that future is now OR we can "handle it"

### Options (from simplest to hardest)
1. Leave it blank - size 0
2. Editor inherits the snapshot (current user info gets applied) - size 1
3. We update the database to have a field that records the originator or source of snapshot - size 2~3


Switching PR to draft for now until solution is determined/implemented.

hardillb:
Gut feel is that if you edited, you own it...

knolleary:
What other scenarios mean a snapshot has a null user property? Auto-generated snapshots is the only one I can think of - so why not use the lack of a user on the object to be the trigger to show the 'auto snapshot' label rather than parsing the name?

My only pause for thought is I cannot remember if our MVP for auto-snapshots included deleting old ones - and what logic that uses to determine what is an old auto-snapshot versus a snapshot a user wants to keep.

Steve-Mcl:
> What other scenarios mean a snapshot has a null user property?

Stretching my memory here but 99.9% certain there should NOT be any operations other than auto snapshot that are permitted to leave `UserId` `null`.  My only pause is that I may have a distant memory of some snapshots already having no UserId (though that is very possibly due to local dev work)



> Auto-generated snapshots is the only one I can think of - so why not use the lack of a user on the object to be the trigger to show the 'auto snapshot' label rather than parsing the name?

That is an option (though it does move from "infer" to "assume")


> My only pause for thought is I cannot remember if our MVP for auto-snapshots included deleting old ones - and what logic that uses to determine what is an old auto-snapshot versus a snapshot a user wants to keep.

The premise was: if the name matches the specific format, there are 10 of them, delete the one with lowest `id` 

What will happen here, after an auto snapshot is renamed, there will now effectively be only 9 Auto snapshots. The next auto snapshot will not cause a deletion (but rather make up the 10th) - so this is a non issue.




> Gut feel is that if you edited, you own it...

I am slightly leaning to this also but there is merit in what Nick said about assuming `null` `UserId` as an Auto Snapshot - as keeps the status quo!


Since modifying the DB to identify snapshot source (discussed [here ](https://github.com/FlowFuse/flowfuse/issues/3358#issuecomment-1935790720) and [here](https://github.com/FlowFuse/flowfuse/issues/3355#issuecomment-1948246802)) are not absolutely necessary at this point (only auto snapshots apply `null` UserId) we can move move forwards with the smaller iteration at this time.  Therefore, could I get consensus on the 2 offerings please: 

1. Assume any snapshot NOT matching the name pattern and has `null` `UserId` is an Auto Snapshot
2. User who changes, now owns it

knolleary:
Please go with my suggestion. The user field is who created it, not who owns it. If the field is blank, assume its an auto-snapshot - ie the system generated it. That is a safe assumption to make - *unless* the user who created has since been deleted. But that is an edge case for another day.

Steve-Mcl:
> I would not use a toast notification either. It can be a simple one-line in the dialog (so the user knows this information _before_ they make the change) to say something like "Changes to the snapshot will not be reflected on devices already running the snapshot".

Addressed in [937807d](https://github.com/FlowFuse/flowfuse/pull/4282/commits/937807d60fbe083e05cfbbf9dec29343c7238d59)

image

Steve-Mcl:
> If the field is blank, assume its an auto-snapshot - ie the system generated it.

Addressed in [d46a226](https://github.com/FlowFuse/flowfuse/pull/4282/commits/d46a2267e10878b9a8b70307fb2988760de40978)

image

joepavitt (approved):
FE all good. Server-side looks good too, but will wait for Nick's eyes before merging.

knolleary:
I'm not going to be able to review this properly until much later today.

knolleary (approved):

knolleary:
Merging once the tests go green

PR #4293, opened 2024-07-31
Remove limits on img-src

hardillb:
## Description

The last update tried to add a wildcard for all the different google world wide domains but it was not valid.

Rather than playing wack-a-mole decided to just allow images from everywhere.

This feels better than an incomplete list and not that large a threat.

## Related Issue(s)

FlowFuse/security#90

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


knolleary:
As I understand it, if `img-src` is not set, it will fall back to `default-src` instead - which is not going to be what we want.

To remove the limits, we need to still set `img-src`, but do so in the most permissive way.

knolleary (approved):
Looks good.

PR #4315, opened 2024-08-02
Site wide audit log better formatting when updates object is to be displayed

### LINKED ISSUE #4071: Large entries in audit log
### Current Behavior

Following addition of support for multiline env vars [#4069](https://github.com/FlowFuse/flowfuse/pull/4069) it was noted that imports of large fields cause rather large audit log entries.

Example (zoomed out on a 4k monitor):
image

### Expected Behavior

We should either drop logging of before and after values, log keys changed only or log a summary of changes

### Steps To Reproduce

import large env file or add several large env vars to an instance/device

### Environment

- FlowFuse version: current/prod

Steve-Mcl:
closes #4071

## Description

As discussed in the issue, there are several places where audit log can be a bit too verbose. Rather than handling idividual cases it was suggested that details should be hidden by default and a "better default formatter" should be employed.

From this, the following has been implemented:

* Removes individual use of `AuditEntryUpdates` component and places it at the end
  * New autit log entries will no longer have to explicitly include this
* By default (unless old/new are BOOLS or explicitly a permitted "named" key (e.g. `name`, `type`, `enabled`, `version`) the OLD and NEW values are NOT shown.
  * This inhibits large values from things like description fields being rendered
  * This prevents env values being show
* There is scope for adjust the logic around what is NOT shown in the `AuditEntryUpdates` component as it now receives the full audit entry (e.g. there are now possibilities of inspecting the audit event name and explicitly hiding update details.)

Additionally, a monospaced font has been set and `whitespace-pre` on the details for updates AND errors has been further improve readability.

I have been through a whole bunch of audit log entries with updates and this PR provides a good balance of detail and readability. Future iterations can expand or reduce further based on explicit criteria.

### Screenshots of before and after 

~~As I cannot expect the reviewer to perform every action and check formatting I will update this PR with screenshots before witching out of draft.~~

image

image

image

image

image

#### Errors before and after
image

#### Overview before & after

image

image

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/audit-log/AuditEntryUpdates.vue
+++ b/frontend/src/components/audit-log/AuditEntryUpdates.vue
@@ -14,7 +14,7 @@
                         Old value: '{{ update.old }}'
                     </template>
                 </template>
-                <template v-if="!update.old && update.new">
+                <template v-else-if="!update.old && update.new">
                     Set: '{{ update.key }}'.
                     <template v-if="includeValues(update)">
                         Value: '{{ update.new }}'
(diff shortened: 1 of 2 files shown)


hardillb (approved):
Quick local test looks good.

joepavitt (approved):

PR #4328, opened 2024-08-06
Only clear PH data if there is user session to clear

### LINKED ISSUE #4326: PostHog cookies reset when going to /account/create
### Current Behavior

- User visits our website, PostHog creates a cookie to track their interaction
- User clicks "sign up"
- Platform gets a `401` to the `/user` API call, which in turn calls `posthog.reset()`
- This clears the PH cookies, and then, when the user signs up, we get a newPostHog ID and that doesn't get bound to the PostHog "Person" doing the website interaction - so we're losing all conversion and referral data

### Expected Behavior

PostHog should track the user as a single "Person" from initial Website landing through to sign up and in-Platform activity

### Steps To Reproduce

Problem was introduced here: https://github.com/FlowFuse/flowfuse/pull/3733

It's intentions were correct, when a user times out, we should be clearing the posthog cookies, but because we (for some reason) do a `/user` API call on the `/account/create` page, then the 40 causes the cookie reset.

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

joepavitt:
## Description

- We cleared PH cookies if a user times out
- However, this had the effect that anonymous users (signing up for the first time for example) were having ID's reset mis sign up, and we lost conversion data
- Now, we check if we have any known `state.user` before clearing the cookies on an unauthenticated request.

## Related Issue(s)

Closes #4326

### DIFF
--- a/frontend/src/store/account.js
+++ b/frontend/src/store/account.js
@@ -255,7 +255,10 @@ const actions = {
         } catch (err) {
             // Not logged in
             commit('clearPending')
-            window.posthog?.reset()
+            // do we have a user session to clear?
+            if (state.user) {
+                window.posthog?.reset()
+            }
 
             if (router.currentRoute.value.meta.requiresLogin !== false) {
                 if (router.currentRoute.value.path !== '/') {


knolleary (approved):
Not sure that will cover all the cases the original change was intended for.

`state.user` will only be set if the current browser session was logged in and the session expires whilst they have the tab open.

However, if they have closed the tab and reload the page after the session has expired, `state.user` will not yet have been initialised - so they are logged out, but the posthog reset doesn't happen.

Happy to go with this for now in absence of a better test; I'm not sure what is available under `window.posthog` to differentiate the two scenarios.

joepavitt:
> Happy to go with this for now in absence of a better test; I'm not sure what is available under window.posthog to differentiate the two scenarios.

That's just an interface to their API for pushing data, it doesn't provide access to any data directly.

gstout52:
To clarify, with this change PostHog is now accurately collecting data from page visitors. So PostHog charts related to website conversion will be accurate starting today, but charts created today that look at (say) July will not be accurate. Is that correct?

joepavitt:
That's correct. I haven't verified, as needed to give it some time to get a few people through the system, but 100% certain this was the cause

Issue #4290, opened 2024-07-31
Devices UX: Moving devices between applications


joepavitt:
As part of Bulk Device Actions, users should be able to move multiple (even a single) Device easily from an Application to (a) another Application and (b) an Instance.
```[tasklist]
### Tasks
- [ ] https://github.com/FlowFuse/flowfuse/issues/4335
- [ ] https://github.com/FlowFuse/flowfuse/issues/4337
- [ ] https://github.com/FlowFuse/flowfuse/issues/4345
- [ ] https://github.com/FlowFuse/website/issues/2453
```

Steve-Mcl:
Initial thoughts around how to surface this in the UI:

Replace the new Trashcan bulk "delete" with an "Actions" drop-down that includes "Delete" and "Move..." options.
The actions button should behave similarly in that it should be disabled when no devices are selected.

Steve-Mcl:
Now I have spent some time looking though the application and code, more thoughts occur.

Below is a slightly disjointed brain dump that is intended to be thought provoking to get to the most intuitive behaviour of moving device(s). Please feel free to comment/correct/dismiss any or all of the below (feedback is wanted and required please)

Scope
- moving devices to another team is not in scope

Handling movement of mixed application/instance source devices?
- i.e. it is not unreasonable for an automation script to request to move a list of devices
   that may or may not belong to different applications/instances
- The alternative is to require the user to move "like" devices in separate requests
  which makes me lean towards the following:
  - Since this first iteration is a UI focused bulk operation, we should only support operations
    on devices that are "like" (i.e. moving devices from a single application or a single instance)
  - If the user wants to move devices from different applications/instances, they MUST do so in separate UI ops anyway
  - Therefore, any API ops that attempt to do a move of devices from APP-A-DEVICE + APP-B-DEVICE + INSTANCE-C-DEVICE to APP-D will raise an exception and fail (400)
  - This will simplify the implementation in this first iteration, and we can revisit if/when we decide to support a more flexible implementation
- RBACs to consider: 
  - device:edit (or device:delete/device:create?)
    - device edit feels more like "settings" than "move" but it is a possible candidate
  - device:move (new permission, owner only)
  - My current position on this:
    - The permission(s) should have the "Owner" role (device create and delete are already owner roles)
    - To minimise the combination of checks on various RBACS, we should have a single permission
      that allows the user to move devices between applications/instances
      - That choice IMO would be a new permission "device:move" (owner only) as it differs
[...]

Issue #4313, opened 2024-08-02
Prompt warning or ask conformation while removing devices from the application 


sumitshinde-84:
### Description

Removing devices from the application should prompt a warning or ask for user confirmation, as the user will lose all their previously done work if they remove the devices.

### Which customers would this be available to

None

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Steve-Mcl:
NOTE: for who ever picks this up.

There was a lot of detail in the slack thread that help us understand why this occurred. As that is not visible here and may get lost, the short version is Sumit was moving devices from one place to another and the only option available was to unassign then reassign. This is what caused the devices to flip back into fleet mode and consequently delete his work.  It was noted that we may want to revisit that particular behaviour.

Also note, once bulk move operations are introduced, this will be less of an issue when moving devices between applications.

REF: https://github.com/FlowFuse/flowfuse/pull/4336

PR #4336, opened 2024-08-06
Bulk update team devices API (bulk move support only)

### LINKED ISSUE #4335: Support for bulk move devices (API)

Steve-Mcl:
closes #4335

## Description

Adds backend API route for team devices bulk update:
* `test/unit/forge/routes/api/teamDevices_spec.js`
* Method:  `PUT`
* URI: `/api/v1/teams/:teamId/devices/bulk`

The route was debated in design phase in the issue but after much debate, I have stuck with the same verb and definition already established in the existing `devices` API (that are responsible for single device combined update / device move endpoint).  More than happy to change this around if desired.

### Tests Added
```
  Team Devices API
    Team Devices
      Supports bulk operations
        update
          move devices
            ✔ Rejects invalid application (404)
            ✔ Rejects invalid instance (404)
            ✔ Rejects moving a device to another team application (400)
            ✔ Rejects moving a device to another team instance (400)
            ✔ Rejects invalid input
            ✔ Member cannot move devices (403)
            ✔ Non team member cannot move devices (404)
            ✔ Move all to Application 2 (48ms)
            ✔ Move all to Project 2 (50ms)
            ✔ Only updates devices not already assigned to the target application (46ms)
            ✔ Only updates devices not already assigned to the target project (49ms)
            ✔ Only unassigns devices not already unassigned
```

## Related Issue(s)

Owner #4335
Parent #4290

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
NOTE: I mentioned in Eng Meeting that there was a sensible refactoring - that is the controller method `moveDevices` _should_ replace the code in https://github.com/FlowFuse/flowfuse/blob/f58bc20fb1ffa838dbf24b57240bb91f04eb5955/forge/routes/api/device.js#L435-L518

There reason I was in two minds was related to one of our guys [raising a concern](https://github.com/FlowFuse/flowfuse/issues/4313) about devices being reset when moved (even though they were in developer mode) and whether I would include or exclude that "feature".

As it happens, the implementation of the bulk update is a refactoring of this original route (with obvious modifications for multiple device support) INCLUDING the switching of a device to `"autonomous"` (fleet) mode when unassigned.

As it happens, with the introduction of bulk move feature, the user can side step the unassignment when moving devices from application to application and thus the switching out of developer mode so it kinda becomes  _less of an issue_

<br>
In short, we can and should refactor the single version in the API Route to use the new controller method.

Will raise an issue as a follow up.

forge/db/controllers/Device.js
```diff
@@ -271,5 +271,176 @@ module.exports = {
...
+                    })
+                }
+            } else if (assignTo === 'instance') {
+                // ### Add device to instance ###
+                if (device.ownerType === 'instance' && device.Project?.id === assignToProject.id) {
+                    // Device is already assigned to this instance - nothing to do
+                    continue
+                } else {
+                    device.ProjectId = assignToProject.id
+                    device.ApplicationId = assignToProject.ApplicationId
```
knolleary, line 370: When assigning to an instance, `ApplicationId` should be null. My understanding is ProjectId and ApplicationId should never both be set.

knolleary (changes requested):
One issue spotted.

forge/db/controllers/Device.js
Steve-Mcl: Hmmmm. very well spotted.

I believe, since the target is being set to the value of an existing/valid source `project` the value is inconsequential and the tests (indirectly) confirm this: https://github.com/FlowFuse/flowfuse/pull/4336/files/05cc75601c14f978b6d5d068c25768c135063863#diff-23b65c9650241a9666db727191beff98fe97e80a3a0175102bebd291692ca6abR709 but for correctness here and now, this should probably be `null`ed

I will switch branch and double check.

ta.

Steve-Mcl: checked this over. we do need to set this null.

resolved in [f070a56](https://github.com/FlowFuse/flowfuse/pull/4336/commits/f070a56a4f81f17edf4ca6eb1f906be96d411a11)

Thanks.

knolleary: Conceptually, I was always in favour of setting applicationId when assigned to an instance in that application - as it makes retrieving a list of all devices within the application trivial. But that wasn't the path taken, so yes, we now need to ensure only one or the other is set.

knolleary (approved):

knolleary:
@Steve-Mcl approved - I appreciate there are a few other PRs behind this. Will let you merge in the appropriate order.

Steve-Mcl:
* All branches merged into this.
* Tested operations on pre-staging
* Once all green - good to go.

PR #4343, opened 2024-08-07
Add option to require contact via HS form for create/upgrade of team type

### LINKED ISSUE #4317: Require users to contact sales to upgrade to Enterprise team
Part of https://github.com/FlowFuse/customer/issues/233

- Add option to TeamType object to flag it as requiring admin permission to upgrade to
- Type should still be listed as an option on the Upgrade Type page; but, when viewed by a non-admin user, should show a 'contact us' button when selected.
- Clicking the button should open a new tab on the Contact Us page.
- The API should reject upgrade requests from non-admin users.

Need to decide how much of this is configurable; eg the url the contact us button takes the user to, the text 'contact us' itself. We have successfully avoided hardcoding FFC specific choices into the billing code - and made things like this configurable.

The workflow would then be:

1. User creates regular team - contacts us as they want enterprise
2. Discussions had, agreement reached.
3. Invoice is manually created and sent out
4. Admin uses the 'unmanaged billing' option to upgrade the team to enterprise and disconnect from managed billing.

knolleary:
Closes #4317 

Adds option on TeamType configuration to require users contact us, rather than allow self-service creation/upgrade to that type.

### TeamType config

If the Require contact to upgrade option is selected, hubspot portal/form ids can be provided.

image

### UX

When creating a new team, or changing an existing team type, if they select a type that has been configured to require contact, they are shown this:

image

This *only* applies to non-admin users. If the user is an admin, then can select the type and create the team 'normally'.

### Hubspot

When clicking the 'contact sales' button, the configured hubspot form is submitted with the user's email and the details of the page they are on. The HS form can be then configured to trigger whatever business process is required.

I have created a suitable form in HS already called 'Enterprise Upgrade Request'.

### DIFF
--- a/frontend/src/api/billing.js
+++ b/frontend/src/api/billing.js
@@ -39,10 +39,49 @@ const setTrialExpiry = async (teamId, trialEndsAt) => {
     })
 }
 
+const sendTeamTypeContact = async (user, teamType, pageName) => {
+    const hsPortalId = teamType.properties?.billing?.contactHSPortalId
+    const hsFormId = teamType.properties?.billing?.contactHSFormId
+    if (hsPortalId && hsFormId) {
+        return new Promise((resolve, reject) => {
+            const url = `https://api.hsforms.com/submissions/v3/integration/submit/${hsPortalId}/${hsFormId}`
+            const xhr = new XMLHttpRequest()
+            xhr.open('POST', url)
+            xhr.setRequestHeader('Content-Type', 'application/json')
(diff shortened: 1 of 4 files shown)


joepavitt:
Is "Free" an appropriate pricing offer there?

knolleary:
@joepavitt that's just how I happen to have my teams configured locally as part of validating all the different options. It *isn't* a reflection of the changes in this PR

joepavitt (approved):

Issue #4292, opened 2024-07-31
Delete Account with Team Deletion


joepavitt:
### Description

When deleting an account, users need to delete their team first. Whilst prompted, the settings for this are in a different place and not intuitive path to go between these locations.

If I am the only member of a team, that team has no instances, and I choose to delete my account - it should deleted both my account _and_ my team in one hit.

### Which customers would this be available to

None

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

cstns:
@joepavitt, if the user trying to delete the team is the only owner but the team has other members, should we proceed with the team deletion?

joepavitt:
Good question - @knolleary - thoughts? I'd probably go with "no", as they're the kind of teams we want reaching out to us so that we can understand _why_ they're churned

knolleary:
I'd stick with only deleting if they are the last person in the team.

cstns:
OK, sound's good!

Should I also check for device presence?

joepavitt:
@cstns two options:

- We do a sanity check first to see _which_ teams can be deleted, _then_ delete if they all can
- As you've said, delete what we can, don't delete the full account, and then flag the remaining teams need to be dealt with.

The first option is the "proper" way imo

knolleary:
I'm reverting the PR that fixes this as it introduced a bug that left teams orphaned with active stripe subscriptions. I'm out of time to fix today, so reverting to prevent it happening further over the weekend. Will rework the original PR on Monday.

knolleary:
The original fix has been reapplied with the additional fixes - https://github.com/FlowFuse/flowfuse/pull/4531

PR #4354, opened 2024-08-08
Allow users to delete accounts even when they have active teams

cstns:
## Description

Allows users to delete their account if they have teams they are owner of.
Users are still prevented from deleting their accounts if the owner team has members, instances or devices assigned to it.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4292

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/models/User.js
+++ b/forge/db/models/User.js
@@ -101,11 +101,31 @@ module.exports = {
                 const teams = await app.db.models.Team.forUser(user)
                 for (const team of teams) {
                     const owners = await team.Team.getOwners()
+                    // const teamMemberCount = await team.Team.memberCount()
+
                     const isOwner = owners.find((owner) => owner.id === user.id)
 
-                    // if this user is the only owner of this team, throw an error
                     if (isOwner && owners.length <= 1) {
-                        throw new Error('Cannot delete the last owner of a team')
+                        const instanceCount = await team.Team.instanceCount()
+                        const deviceCount = await team.Team.deviceCount()
+                        const members = await team.Team.memberCount()
+
+                        // throw error if the team has other members assigned to it
+                        if (members > 1) {
+                            throw new Error(`Team ${team.Team.name} which is being deleted alongside your account still has users in it.`)
+                        }
+
+                        // throw error if the team has remaining instances assigned to it
+                        if (instanceCount > 0) {
+                            throw new Error(`Team ${team.Team.name} which is being deleted alongside your account still has instances assigned to it.`)
+                        }
+
+                        // throw error if the team has remaining devices assigned to it
+                        if (deviceCount > 0) {
+                            throw new Error(`Team ${team.Team.name} which is being deleted alongside your account still has devices assigned to it.`)
+                        }
+
+                        await team.Team.destroy()
                     }
                 }
 
--- a/forge/db/models/User.js
+++ b/forge/db/models/User.js
@@ -134,7 +134,9 @@ module.exports = {
                 }
 
                 // delete remaining owned teams
-                teamOwnerships.forEach(team => team.Team.destroy())
+                for (const ownedTeam of teamOwnerships) {
+                    await ownedTeam.destroy()
+                }
 
                 // Need to do this in beforeDestroy as the Session.UserId field
                 // is set to NULL when user is deleted.
--- a/forge/db/models/User.js
+++ b/forge/db/models/User.js
@@ -100,7 +100,7 @@ module.exports = {
                 // throw an error if we would orphan any teams
                 const teams = await app.db.models.Team.forUser(user)
                 const teamBlockers = []
-                const teamOwnerships = []
+                const ownedTeams = []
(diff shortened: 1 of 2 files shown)


forge/db/models/User.js
```diff
@@ -99,16 +99,43 @@ module.exports = {
...
+                        const instanceCount = await team.Team.instanceCount()
+                        const deviceCount = await team.Team.deviceCount()
+                        const members = await team.Team.memberCount()
+
+                        teamOwnerships.push(team)
+
+                        // throw error if the team has other members assigned to it
+                        if (members > 1) {
+                            teamBlockers.push(`Team ${team.Team.name} which is being deleted alongside your account still has users in it.`)
+                        }
```
Steve-Mcl, line 118: Is it a problem if other non owners are in the team?

@joepavitt if a user wants to leave and delete their account, should we prevent this just because a team they are the sole owner of has members?

also, the wording is a little verbose though i cannot think of a good alternative at this time!

forge/db/models/User.js
```diff
@@ -99,16 +99,43 @@ module.exports = {
...
                     }
                 }
 
+                if (teamBlockers.length) {
+                    throw new Error(teamBlockers[0])
+                }
+
+                // delete remaining owned teams
+                teamOwnerships.forEach(team => team.Team.destroy())
```
Steve-Mcl, line 137: destroy is an `async` operation (needs await)

```suggestion
                for (const ownedTeam of teamOwnerships) {
                    await ownedTeam.destroy()
                }
```

forge/db/models/User.js
```diff
@@ -99,16 +99,43 @@ module.exports = {
                 // determine if this user owns any teams
                 // throw an error if we would orphan any teams
                 const teams = await app.db.models.Team.forUser(user)
+                const teamBlockers = []
+                const teamOwnerships = []
```
Steve-Mcl, line 103: Nit pic suggestion: `ownedTeams` is more readable (sorry to be pedantic) (tis a suggestion only, take with salt ;) )

forge/db/models/User.js
cstns: I asked the same question here https://github.com/FlowFuse/flowfuse/issues/4292#issuecomment-2275725114

forge/db/models/User.js
cstns: I agree

forge/db/models/User.js
Steve-Mcl: thanks, i missed that.

test/unit/forge/routes/api/user_spec.js
```diff
@@ -773,6 +836,24 @@ describe('User API', async function () {
...
+            const amidala = await app.db.models.User.create({ username: 'amidala', name: 'Padme Amidala', email: 'amidala@example.com', email_verified: true, password: 'paPassword', admin: false, sso_enabled: false })
+            const nabooTeam = await app.db.models.Team.create({ name: 'nabooTeam3', TeamTypeId: app.defaultTeamType.id })
+
+            await nabooTeam.addUser(amidala, { through: { role: Roles.Owner } })
+
+            await login('amidala', 'paPassword')
+
+            await app.factory.createApplication({ name: 'senate-app-2' }, nabooTeam)
+
+            // Padme now attempts to delete own account: should fail as she still has instances attached to her team
```
Steve-Mcl, line 849:
```suggestion
            // Padme now attempts to delete own account: should succeed even though it has instances attached to the team
```

test/unit/forge/routes/api/user_spec.js
```diff
@@ -753,7 +757,66 @@ describe('User API', async function () {
...
+            await app.factory.createDevice(
+                {
+                    name: 'some-device'
+                },
+                nabooTeam,
+                null,
+                nabooTeamApplication
+            )
+
+            // frank now attempts to delete own account: should fail as frank is the last owner of team B
```
Steve-Mcl, line 811:
```suggestion
            // Padme now attempts to delete own account: should fail as owned team has devices attached
```

Steve-Mcl (approved):
out of date comments, otherwise good.

verified all operations on pre-staging as per issue discussion.

Summary of expectation: _can delete account and owned team(s) if 
* "no additional members" 
* "no instances in team" 
* "no devices in team"_


Deleting account when team still has an instances:
image

Deleting account when team still has a device (unassigned device):

image


Deleting account when team still has additional members:

image

Steve-Mcl:
NOTE: before I merge this, i need to check an issue I witnessed with creating users on this branch.

i was getting no feedback when creating a user. Clicking the "Create User" button again caused an "already exists" toast error.

I dont see how this PR would cause that issue but I need to do due diligence before merging,


Screenshot of issue:
image


Demo of issue:

chrome_2xXOaEqIhZ

PR #4360, opened 2024-08-09
Ensure personal instance is created in personal team on signup

knolleary:
We had received a couple reports that after invite a user to the platform, they were getting a 'verification failed' error when entering the verification code. Clicking the 'resend email' didn't do anything. However they found they could click logout and log in again and all was well.

Via the audit logs, I could see we were:
1. creating their personal team
2. creating an application
3. *not* creating an instance

On further examination, I could create this locally based on the following conditions:

1. An Enterprise team is created by existing user. No instances are created.
2. An external user is invited to join the team
3. The external user receives invite and signs up
4. They receive the verification code, which they submit correctly
5. The verification code is confirm and their email is marked verified
6. The signup code then completes their signup by:
    1. Creates a personal team
    2. Accepts the invite - so they are now a member of both personal team and enterprise team
    3. Checks there are no instances in any of their team - so auto-creates a new instance
    4. The code picks the first team in their list to create the new app/instance in. However, *this happens to be the Enterprise team*.
    5. Because the Enterprise team is picked, the attempt to create a Starter instance fails - as starter isn't allowed in enterprise.
    6. The error bubbles up and appears in the front end as a 'verification failure' - even though the verification succeeded

 This fixes that by:

1. Guarantees that *if* it decides to create the starter instance, it will get created in their personal team
2. Ensures the verify end point always reports success once its gets past the verification step. Any subsequent errors creating their team/etc are *not* verification issues and are logged separately.

### DIFF
--- a/forge/lib/userTeam.js
+++ b/forge/lib/userTeam.js
@@ -6,6 +6,7 @@ const crypto = require('crypto')
  *   - accepts any invitations matching the email
  */
 async function completeUserSignup (app, user) {
+    let personalTeam
     if (app.settings.get('user:team:auto-create')) {
         const teamLimit = app.license.get('teams')
         const teamCount = await app.db.models.Team.count()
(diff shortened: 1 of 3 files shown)


Steve-Mcl (approved):
Code review looks good and makes sense however we should probably unit test this scenario but I can see how this would be difficult to achieve and there is a more immediate need to get this fix in without additional delay. Perhaps a follow up issue around tests would be in order?

knolleary:
@Steve-Mcl I have added a couple tests to the existing suite that covers the auto-creation of team/instance on signup

        ✔ Creates instance in personal team when invites to other teams present (317ms)
        ✔ Skips creating instance in personal team when invited to other team with instance (211ms)

The first test covers the main scenario. The second one was a piece of existing logic I saw that was quick to add a test for.

PR #4369, opened 2024-08-12
Add 'none' option as pipeline stage action

### LINKED ISSUE #4311: Add "None" action type for a DevOps Stage
### Description

When creating a stage on a DevOps Pipeline users are presented with the following options for the "Action" that pipeline should perform.

Screenshot 2024-08-01 at 18 22 13

This can be unintuitive though as when adding hte last stage, there is no action to take, but the user must still select _something_.

It is proposed to add a "No Action" or "None" option here and make it the default option. This makes it more obvious as to the purpose of the stage in question.

We already do not show the "Deploy Action" row for the last stage, so there is no need for extra effort here.

Team + Enterprise Tiers (EE)

knolleary:
Closes #4311 

Adds a 'do nothing' option (stored as 'none') when creating a pipeline stage. 

 - The API no-ops calls to deploy a stage with its action set to none
 - The UI disables the deploy button if the action is set to none
 - DB migration added to add 'none' to the enum on postgres


knolleary:
Investigating postgres failure

Steve-Mcl (commented):
Testing on pre-staging does not appear to operate as expected

1. Getting a double toast when adding a stage with

chrome_e2yzP8TmH1

`SequelizeDatabaseError: invalid input value for enum "enum_PipelineStages_action": "none": undefined`
`Request Failed: SequelizeDatabaseError: invalid input value for enum "enum_PipelineStages_action": "none"`


---

2. As per the issue, the `None` option should only pre presented for the last stage. E2E test should also check this is the case. 
image

knolleary:
@Steve-Mcl there was a bug in the migration with the first commit, fixed in the second commit. However, the prestaging env will have applied the original version of the migration. I don't know if it is possible to completely tear down the prestaging env and recreate clean.

knolleary:
> As per the issue, the None option should only pre presented for the last stage. E2E test should also check this is the case.

The issue doesn't explictly specify that, although the primary use is for the last stage. Currently, the edit form for a pipeline stage doesn't know it's the end stage - it has no knowledge of the pipeline its in. I'll see if there's a simple way to plumb in that additional state, but I doubt its worth the effort.

knolleary:
Okay - took another look and spotted the information was available.

Have updated to only show the option when adding a new stage, or editing the last stage.

I have not added UX tests for that due to time. Please review regardless.

Steve-Mcl:
> The issue doesn't explictly specify that

My bad, I think it was implied (thats how I read it)







> Okay - took another look and spotted the information was available.

Yeah, I was busy writing up a reply along the lines of "there are signals in the vue because we disable the "Device Group" option when adding/editing as first stage" - but you have obv spotted it :)

frontend/src/pages/application/PipelineStage/form.vue
```diff
@@ -448,6 +451,9 @@ export default {
             } else if (this.input.stageType === StageType.DEVICE) {
                 options.unshift({ value: StageAction.USE_ACTIVE_SNAPSHOT, label: 'Use active snapshot' })
             }
+            if (this.isLastStage) {
```
Steve-Mcl, line 454:
```suggestion
            if (!this.isFirstStage && this.isLastStage) {
```

Prevent first stage selecting "none"

Steve-Mcl (commented):
Pulled and tested locally as i couldnt quite understand if/how using just `isLastStage` would work. 

Apart from the suggestion that fixes selection of "Do nothing" on the first stage, all good.

knolleary:
Have added a couple checks to the existing UI tests for the presence/absence of 'do nothing' option as appropriate. Not completely exhaustive, but should be sufficient to spot any regression.

Steve-Mcl (approved):

PR #4378, opened 2024-08-14
Provide proactive feedback when trying to change team type to incompatible type

### LINKED ISSUE #4377: Improve feedback when changing team type blocked due to resource usage
Ref https://github.com/FlowFuse/customer/issues/237

If a user attempts to change their team type to a lower level that doesn't permit the level of usage they currently have, the feedback is poor.

For a Trial user, they pick their desired team type, then they are taken to stripe to setup billing. Once that is done, it completes the request - which fails due to the 'can upgrade' check happening too late.

From a UX perspective, the UI should inform the user they cannot change to the desired team type and why - *before* they try to do so.

knolleary:
Closes #4377 

This provides better feedback if a team upgrade is blocked due to the current usage exceeding any limits on the target team.

My first iteration was to add the check on the backend for the path where a team is setting up billing - it was already handled for the case of changing type when billing is already setup. This ensures any issue is reported *before* they are sent to stripe.

However, the feedback isn't great - and would be better to be more proactive in the UI. So the second iteration (3rd commit) copies the limit checks into the front end and lets us show a more helpful message to the user:

image

The specific errors shows above are based on how I happen to have my Starter team type setup locally; so don't get distracted by the specific numbers.

Also need to highlight the checks are based purely on Members/Device/Instance count limits. It doesn't check if they are using features (eg Team Library) that are unavailable at the target tier. We don't maintain flags on feature usage within a team, so that's out of scope (and not proposing it comes into scope any time soon).

---

There is currently limit test coverage for this. We're missing base coverage in a couple different areas here for the existing code.

I have added basic verification the UX blocks changing to a team that is not compatible with current usage - but it isn't exhaustive.


knolleary:
Have added basic coverage to verify the UI blocks changing to a 'lower' type.

Steve-Mcl (approved):
Happy with the code. 

Do we have a team on prod that we can verify this after it is deployed?

hardillb:
@Steve-Mcl do we want to merge this and then I can test it?

Steve-Mcl:
@hardillb - yes, fine by me Ben

hardillb:
Verified on Staging where I had a Teams, team with more users and instances than allowed in starter.

Issue #4357, opened 2024-08-09
Create /api/v1/instance/:id/files/... api for basic CRUD operations


joepavitt:

knolleary:
Have updated the description with an initial API spec to work from. This will be the basic structure used in the launcher as well.

hardillb:
I think there is an error in the sample above, each file object should have `name` not `path`

```
{
    "meta": {
        "next_cursor": "",
    },
    "files": [
        {
            "type": "file",
            "path": "filename",
            "size": "1024",
            "lastModified": "2024-08-08T19:40:05.000Z"
        },
        {
            "type": "directory",
            "path": "filename",
            "lastModified": "2024-08-08T19:40:05.000Z",
            "share": { ... }
        }
    ],
    "count": 8
}
```

should be

```
{
    "meta": {
        "next_cursor": "",
    },
    "files": [
        {
            "type": "file",
            "name": "filename",
            "size": "1024",
            "lastModified": "2024-08-08T19:40:05.000Z"
        },
        {
            "type": "directory",
            "name": "filename",
            "lastModified": "2024-08-08T19:40:05.000Z",
            "share": { ... }
        }
    ],
    "count": 8
}
```

joepavitt:
Task Completed

PR #4384, opened 2024-08-15
Implement files api

knolleary:
Part of #4357 

## Description

Implements the API specified in 4357 to allow file operations on the instance.

 - Adds `staticAssets` platform feature flag - registered when running with license
 - Adds `staticAssets` team feature flag - must be enabled on the TeamType to access the api
 - Implements all of the end points described in #4357 
 - Requires driver update
     - LocalFS PR: https://github.com/FlowFuse/driver-localfs/pull/144
     - Once reviewed/approved, equivalent PRs will be needed for Docker/K8S drivers
 - Includes setting the `share` property to directories.
 - Adds `KEY_SHARED_ASSETS` as a `ProjectSettings` key that will contain the share settings object.
 - File uploads could be made more efficient; couldn't get the stream api to work properly to avoid pulling the whole file into memory. One option would be to stream to disk to a temp location, then stream off disk when upload to the launcher. Short term, I propose we have a fairly tight limit on file upload size for initial iteration.

This is as far as I could get in 1 day. Here's what I think still needs to be handled whilst I'm out:

 - [ ] Tests! To get to this point, I've been manually testing locally. We need test coverage across the board here. I've added placeholder functions in the stub driver for the file functions. They can be extended to provided basic mock responses
 - [ ] File size limits on file upload. This is something that can be configured in `@fastify/multipart` - [search 'limits' on this page for reference](https://github.com/fastify/fastify-multipart). Should limit to a single file and a small file size initially - say 10Mb? Sufficient for first iteration? But will need *early* verifying that requests that large get through the ingress etc okay.
 - [ ] Testing unusual characters in filenames
 - [ ] Handle talking to an old launcher - either fail gracefully, or add a version check in the preHandler.

### Useful CURL commands

These are commands I've been using to test with. You will need to update the instance ID in the url, and you will need to create your own access token.

**Get files at /**
```
curl -s -X GET -H "content-type: application/json" \
     -H "authorization: Bearer <token>"\
     http://localhost:3000/api/v1/projects/ba2bbf57-e491-46c4-a3a5-41adbd3be413/files/_/
```

**Get files at /test/**
```
curl -s -X GET -H "content-type: application/json" \
     -H "authorization: Bearer <token>"\
     http://localhost:3000/api/v1/projects/ba2bbf57-e491-46c4-a3a5-41adbd3be413/files/_/test
```

**Rename file /test/old.txt to /test/new.txt**
```
curl -X PUT -d '{"path":"/test/new.txt"}' \
    -H "content-type: application/json" \
    -H "authorization: Bearer <token>"\
    http://localhost:3000/api/v1/projects/ba2bbf57-e491-46c4-a3a5-41adbd3be413/files/_/test/old.txt
```

**Move file /test/old.txt to /abc/old.txt**
```
curl -X PUT -d '{"path":"/abc/new.txt"}' \
    -H "content-type: application/json" \
    -H "authorization: Bearer <token>"\
    http://localhost:3000/api/v1/projects/ba2bbf57-e491-46c4-a3a5-41adbd3be413/files/_/test/old.txt
```
**Delete file /test/old.txt**
```
curl -X DELETE
    -H "content-type: application/json" \
    -H "authorization: Bearer <token>"\
    http://localhost:3000/api/v1/projects/ba2bbf57-e491-46c4-a3a5-41adbd3be413/files/_/test/old.txt
```

**Upload file**
```
curl -F "file=@/Users/nol/Downloads/test.png" \
  -H "authorization: Bearer <token>" \
  http://localhost:3000/api/v1/projects/ba2bbf57-e491-46c4-a3a5-41adbd3be413/files/_/test.png
```

**Create dir `/test/one/two**
```
curl -X POST -d '{"path": "one/two"}' \
    -H "content-type: application/json" \
    -H "authorization: Bearer <token>" \
    http://localhost:3000/api/v1/projects/ba2bbf57-e491-46c4-a3a5-41adbd3be413/files/_/test
```


forge/ee/routes/staticAssets/index.js
```diff
@@ -0,0 +1,242 @@
...
+                // Not a directory or 404
+                reply.code(404).send({ code: 'not_found', error: 'Not Found' })
+                return
+            }
+
+            reply.code(200).send({})
+        }
+    })
+
+    app.delete('/_/*', {
```
Steve-Mcl, line 145: I am not comfortable with the API paths permitting `*`.  I think, as the original issue stated, this should be a named parameter 

> DELETE /api/v1/projects/:instanceId/files/_/:path

As @hardillb discovered in https://github.com/FlowFuse/flowfuse/pull/4392#issuecomment-2304401935, fastify resolves URLs before routing.  i.e. a call to `api/v1/projects/ba2bbf57-e491-46c4-a3a5-41adbd3be413/files/_/../../` would be resolved to `api/v1/projects/ba2bbf57-e491-46c4-a3a5-41adbd3be413` and in theory, delete the instance.

My recommendation is that file path value is URL encoded and treated as a single parameter (on all routes)

hardillb: Fixed by URI encoding the path

hardillb:
This should be everything for the forge app now in one PR.

hardillb (approved):
All looks good, but i wrote some of this so would like another look.

Only thing in testing the Assets tab is shown to Viewer users, who can't use this feature. Not a stop ship, but should be fixed in next iteration

Steve-Mcl (approved):
Pulled and tested locally (would have preferred staging before pushing to main - something I will raise in retro)

3 v minor issues remain that will be raised soon after merge:

- [ ] If I select the same file to upload a 2nd time, it does not populate the text element nor does it enable confirm button
- [ ] As discussed in slack thread, if i specify a "bad" folder name, no error is surfaced.  Happy to concede this may be a local issue but without a staging env, not entirely clear.
- [ ] the folder create dialog lets me create a folder named / (but nothing happens) (same as creating a folder named // or \\ or ///// or \\\\ - which TBF may be a non issue

PR #4385, opened 2024-08-15
Send notification to inviter when a team invite is accepted

joepavitt:
## Description

- Records a notification for the inviter when the invitee accepts an invitation as a new `team-invite-accepted-invitor` type
- Adds relevant `.vue` components to display the notification, which navigates to the team member's page when clicked.

Screenshot 2024-08-15 at 20 10 34

Not sure what tests, if any, need to be added here?

## Related Issue(s)

Closes #4379 

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass

PR #4386, opened 2024-08-15
Track server-side invitation events with PostHog

### LINKED ISSUE #4381: Improve tracking of accepted invites
### Description

Currently, when a user receives a team join invitation, and has to create a FlowFuse user in the process, the user does not count as "accepting" the invitation in PostHog.

**Desired Behavior**
Any time a user receives a team join invitation, clicks the invitation, and ends up a part of a team as a result, that user should be counted as "accepting" the invitation in PostHog.

### Epic/Story

_No response_

joepavitt:
## Description

- Registers a new "product" service for the `app`
- Adds `posthog-node` as a new dependency
- Wraps the `capture` function for PostHog, which is initiated using the same `config` as the frontend
- Wire the product service into the accepting of team invites server-side
- Remove the `invite-accepted` event from client-side so as to not duplicate recording.

### Todo

- Unit tests?

## Related Issue(s)

Closes #4381 

## Checklist


joepavitt:
Have a working end-to-end:

Screenshot 2024-08-16 at 15 20 36

Steve-Mcl (commented):
looks good to me Joe but in the spirit of "done" there are no tests.

I think at minimum we should have unit tests that checks `app.product.capture` is a function. Better still would be to perform an invite, spy `app.notifications.send` and check the arguments in the call are sane (no raw `id`s)

joepavitt:
Yep, haven't touched the server-side unit tests before, so wasn't sure how deep to go with it.

joepavitt:
Not sure why the `notifications.send` is picked up in the diff, as that should be on `main` already

joepavitt:
@Steve-Mcl I've updated the branch now so that the diff _doesn't_ include all the stuff from https://github.com/FlowFuse/flowfuse/pull/4385/files too

package.json
```diff
@@ -95,6 +95,7 @@
         "pg": "^8.11.3",
         "pino": "^8.15.1",
         "pino-pretty": "^10.0.0",
+        "posthog-node": "^4.1.0",
```
Steve-Mcl, line 98: there was a guard for null values added in 4.1.1.  also V4.2.0 is released.

We should probably switch to 4.1.1 at min.

Steve-Mcl (changes requested):

Steve-Mcl (approved):
good to go

PR #4392, opened 2024-08-20
Add tests for file API

hardillb:
## Description

Adds basic tests for file API

There needs to be more, but it's a start

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/containers/stub/index.js
+++ b/forge/containers/stub/index.js
@@ -11,6 +11,10 @@
  */
 const list = {}
 const forgeUtils = require('../../db/utils')
+const { util }  = require('@node-red/util')
+const { text } = require('node:stream/consumers')
+
+const files = {}
 
 module.exports = {
     START_DELAY: 500,
@@ -275,10 +279,48 @@ module.exports = {
         if (!list[instance.id] || list[instance.id].state === 'suspended') {
             throw new Error('Cannot access instance files')
         }
-        return {
-            meta: {},
-            files: [],
-            count: 0
+        if (files[instance.id]) {
+            const pathDots = filePath.replace('/','.')
+            const response = {
+                meta: {},
+                files: [],
+                count : 0
+            }
+            try {
+                const dir = pathDots ? util.getObjectProperty(files[instance.id],pathDots) : files[instance.id]
+                Object.keys(dir).forEach(entry => {
+                    if (typeof dir[entry] === 'object') {
+                        response.files.push({
+                            name: entry,
+                            type: 'directory',
+                            lastModified: new Date().toISOString()
+                        })
+                    } else {
+                        response.files.push({
+                            name: entry,
+                            type: 'file',
+                            size: dir[entry].length,
+                            lastModified: new Date().toISOString()
+                        })
+                    }
+                    response.count++
+                })
+                return response
+            } catch (err) {
+                if (err.message === 'Cannot convert undefined or null to object') {
+                    const newErr = new Error('not found')
+                    newErr.statusCode = 404
+                    throw newErr
+                } else {
+                    throw err
+                }
+            }
+        } else {
+            return {
+                meta: {},
+                files: [],
+                count: 0
+            }
         }
     },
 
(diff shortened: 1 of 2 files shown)


forge/containers/stub/index.js
```diff
@@ -275,32 +279,125 @@ module.exports = {
...
 
     updateFile: async (instance, filePath, update) => {
         if (!list[instance.id] || list[instance.id].state === 'suspended') {
             throw new Error('Cannot access instance files')
         }
+        if (!files[instance.id]) {
+            files[instance.id] = {}
+        }
+        // const pathDots = filePath.replace('/','.')
+        // const dir = pathDots ? nrUtil.util.getObjectProperty(files[instance.id], pathDots) : files[instance.id]
```
Steve-Mcl, line 335:
```suggestion
```

Dead code

test/unit/forge/ee/routes/staticAssets/index_spec.js
```diff
@@ -0,0 +1,229 @@
+const FormData = require('form-data') // eslint-disable-line
+const should = require('should') // eslint-disable-line
+const setup = require('../../setup')
+
+// const FF_UTIL = require('flowforge-test-utils')
+
```
Steve-Mcl, line 6:
```suggestion
```

dead code

Steve-Mcl (commented):
Ben I think there are a couple of test cases we are missing here.

1. In case of regression, we should be attempting (and failing) to read/create/delete files higher up the tree using relative paths `/_/foo/../../bar` 
2. We dont try readin/writing anything to a path without `/_/` - what if a bad actor attempts to read or write `/api/v1/projects/${TestObjects.instance2.id}/files/foo/bar`?

hardillb:
for 2 it's just going to fail with 404 because there is no matching route

forge/containers/stub/index.js
hardillb: That will be needed as more tests get added

hardillb:
Actually it doesn't, because fastify tries to be cleaver and resolves the shorter path

so `/api/v1/projects/1bfaebb5-317d-4955-9c46-a754035c724b/files/foo/../../helloWorld.txt` becomes `/api/v1/projects/1bfaebb5-317d-4955-9c46-a754035c724b/files/helloWorld.txt` 

which returns a 200 and the boiler plate index.html

So this answers both 1 & 2

@Steve-Mcl

Steve-Mcl:
> Actually it doesn't, because fastify tries to be cleaver and resolves the shorter path
> 
> so `/api/v1/projects/1bfaebb5-317d-4955-9c46-a754035c724b/files/foo/../../helloWorld.txt` becomes `/api/v1/projects/1bfaebb5-317d-4955-9c46-a754035c724b/files/helloWorld.txt`
> 
> which returns a 200 and the boiler plate index.html
> 
> So this answers both 1 & 2
> 
> @Steve-Mcl

I feared as much & TBH, that feels like either a shortcoming in our API design or the usage of it here.

shouldnt the param be encoded - so that it is ALWAYS resolved as a single param without any URL resolving going on?

in fact, didnt i see something in another PR for the static assets where encoding the url fragment (file path param) was in fact needed? (for things like spaces and other valid file name characters a file may have)?

hardillb:
@Steve-Mcl I think this is a different problem

I think its down to this code:

https://github.com/FlowFuse/flowfuse/blob/782c37a116ecdae77f8ae8461cb014e87d995edf/forge/routes/ui/index.js#L125

If I put console logs in there I see it getting fired for the POST request to a unknown path. I think we can fix this (after lunch)

hardillb:
@Steve-Mcl can we merge this (just to get rid of the pre-staging env) it's only merging into the 4357-file-api branch.

We can combine the tests as part of the review of the full PR

Steve-Mcl:
> @Steve-Mcl can we merge this (just to get rid of the pre-staging env) it's only merging into the 4357-file-api branch.
> 
> We can combine the tests as part of the review of the full PR

Fair enough Ben.

PR #4397, opened 2024-08-23
Move invite accept to before personal team create

### LINKED ISSUE #4333: Set default team based on team invitations if present
### Description

If a new user is invited to FlowFuse to an existing team, we still create their default/personal team as part of our onboarding workflow, however, we want the _default_ team to be that of the one they're invited to.

This has caused [confusion](https://github.com/FlowFuse/customer/issues/231) when onboarding Enterprise teams as each user has been presented with the billing page (due to the personal teams being created)

Each user has a "Default Team" option, we should ensure this is set based on the team invitations, rather than the personal team created.

### Epic/Story

_No response_

hardillb:
fixes #4333

## Description

We should only create a personal team for a new user if they have not been invited to a team.

The logic was already present to do this, we just were accepting the invites after creating the team...

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/lib/userTeam.js
+++ b/forge/lib/userTeam.js
@@ -6,6 +6,21 @@ const crypto = require('crypto')
  *   - accepts any invitations matching the email
  */
 async function completeUserSignup (app, user) {
+    // Process invites first to see if user is in any teams
+    const pendingInvitations = await app.db.models.Invitation.forExternalEmail(user.email)
+    for (let i = 0; i < pendingInvitations.length; i++) {
+        const invite = pendingInvitations[i]
+        // For now we'll auto-accept any invites for this user
+        // See https://github.com/FlowFuse/flowfuse/issues/275#issuecomment-1040113991
+        await app.db.controllers.Invitation.acceptInvitation(invite, user)
+        // // If we go back to having the user be able to accept invites
+        // // as a secondary step, the following code will convert the external
+        // // invite into an internal one.
+        // invite.external = false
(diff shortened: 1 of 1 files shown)


joepavitt (approved):

knolleary:
This PR has broken the tests in an unusual way. Viewing the Backend Tests log output, mocha reports two tests have failed - yet it exits with rc=0, so the action reports a pass.

```
  1609 passing (11m)
  24 pending
  2 failing
```

The 2 failing tests are closely related to the behaviour this PR has changed. I'm still investigating just how it managed to break the tests in a way that caused mocha not to report the failure - but when I try running `test/unit/forge/routes/auth/index_spec.js` locally since this PR was merged, the tests hang completely until I get an out of memory error.

knolleary:
Ok - so the tests that are now silently failing are ones related to ensuring, if a user accepts an invite when registering, that the personal instance gets created in their personal team, not the team they were invited to.

However, this PR means the personal team no longer gets created so the whole test is invalid.

What is very strange is the test should be failing due to the line:
```
teams.should.have.length(2)
```
 because `teams` now only has length of 1. But this line is just hanging; it doesn't throw, it doesn't pass, it just hangs.

Changing it to `teams.length.should.equal(2)` makes the test failure properly... but that is odd. Maybe some iteraction between collection objects return by sequelize and should.js.

Regardless, this PR should have had tests adding to verify the personal team didn't get created.

Issue #4358, opened 2024-08-09
Frontend: File Browser UX


joepavitt:
Design Task: https://github.com/FlowFuse/flowfuse/issues/4352

```[tasklist]
### Tasks
- [x] Develop the front-end for the Static Asset service, available in the tabs at an _Instance_ level.
- [x] Should include the addition of a `files/assets` service to interface with the new [REST API](https://github.com/FlowFuse/flowfuse/issues/4357).
- [ ] https://github.com/FlowFuse/flowfuse/issues/4419
```

cstns:
closing, fixed by https://github.com/FlowFuse/flowfuse/pull/4398 and https://github.com/FlowFuse/flowfuse/issues/4419

PR #4398, opened 2024-08-23
Assets Service - UX to add/delete/modify files and folders on the Node-RED filesystem

joepavitt:
## Description

- Adds a new "Assets" tab
- Supports creation, updating and deletion of Folders
- Supports file uploads (up to 5 MB currently due to API limitations)

### Todo

- E2E Tests
- Documentation

https://github.com/user-attachments/assets/d7365b1d-94a3-4a89-af87-c5f7807aa070

### Follow Up

Allow for users to control "visibility"/"share" functionality of each folder so that it becomes available via HTTP endpoints/Dashboard

## Related Issue(s)

First part of #4358


cstns:
I've put in place some checks that prevent unnecessary api calls to the assets endpoint when the feature is disabled or the nr-launcher version does not match the required version.

https://github.com/user-attachments/assets/9c12c777-29ac-4d21-9161-45eadcfd11a3

cstns:
Also treated the use case where the instance is suspended/suspending/starting

https://github.com/user-attachments/assets/310482a3-8a4a-46b2-a432-d233bf306f0d

PR #4409, opened 2024-08-27
Add crash notifications

### LINKED ISSUE #4408: Add notifications for instance/device crashed

Steve-Mcl:
closes #4408

## Description

Firstly, my apologies for what looks and smells like feature creep. That was not the intention.  Initial implementation was done and as was soon discovered, the crash notifications were very quickly filling the notification area. This was discussed in slack and suggestions for change were made. Rather than throwing away good (and relevant) dev work, I kept it in.

* Adds the calls to `notification.send` for instances devices and devices
* Updates the `notification.send` API to support upserting and superseding
   * `options.supersede` was added first as discussed in slack. Essentially, this permits a repeated notification to to be auto-marked as read and its new (superseding) notification to be added
   * After some more show and tell & discussion in Slack, it was suggested a counter would be better. `options.upsert` was implemented to "update" an existing notification (or add is not existing) while adding a counter to the data. This permits multiple successive notifications to be shown as one item (with a pill/badge) and not fill up the notification panel with read notifications
* Adds supports hiding read notifications
  * This was added AFTER implementing `options.supersede` and discussed in slack as a viable option.
  * However, while useful (and therefore left in this PR) we switched track to using `options.upsert`.  It is left is as useful development and worthy of keeping as more notifications from other instances and occurrences could easily fill (and push unread items) down the list.
* Adds a `Generic` notification that simplifies and reduces boilerplate
* Updates the base Notification for common things like timestamp and styling (to reduce repeated coding in dedicated components)

### MVP parts of this.

Ideally, the notification API and DB tables would natively support `severity` and `counter` but as MVP, I added them into the `data` object of the notification under a `meta` object.  This may end up being a retro thing but until we have more notifications to add it is not fully clear - something to be aware of.

#### Default view - [x] Hide Read

image

#### View with [ ] Hide Read unchecked
image

### Tests

`test/unit/forge/routes/logging/index_spec.js`

```
  Logging API
    adds notification
      ✔ for every member and owner when instance crashed (41ms)
      ✔ for every member and owner when device crashed
      ✔ for every member and owner when instance safe-mode
      ✔ for every member and owner when device safe-mode
```

`test/unit/forge/routes/api/userNotifications_spec.js` (was previously named [`userNotifications.js`](https://github.com/FlowFuse/flowfuse/blob/7cbaa77a9e99c51bd6cc6f769ed8d94a0d51b4c8/test/unit/forge/routes/api/userNotifications.js) and thus NEVER ran!

```
  User Notifications API
    User notifications
      ✔ Updates notification when options.upsert is set
      ✔ Updates notification when options.supersede is set
```

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/notifications/Generic.vue
+++ b/frontend/src/components/notifications/Generic.vue
@@ -1,0 +1,121 @@
+<template>
+    <NotificationMessage
+        :notification="notification"
+        :selections="selections"
+        data-el="generic-notification" :to="to"
+    >
+        <template #icon>
+            <component :is="notificationData.iconComponent" />
+        </template>
+        <template #title>
+            {{ notificationData.title }}
+        </template>
+        <template #message>
+            <!-- eslint-disable-next-line vue/no-v-html -->
+            <span v-html="notificationData.message" />
+        </template>
+    </NotificationMessage>
+</template>
+
+<script>
+import { defineAsyncComponent } from 'vue'
+
+import IconDeviceSolid from '../../components/icons/DeviceSolid.js'
+import IconNodeRedSolid from '../../components/icons/NodeRedSolid.js'
+import NotificationMessageMixin from '../../mixins/NotificationMessage.js'
+
+import NotificationMessage from './Notification.vue'
+
+export default {
+    name: 'GenericNotification',
+    components: { NotificationMessage, IconNodeRedSolid },
+    mixins: [NotificationMessageMixin],
+    data () {
+        return {
+            knownEvents: {
+                'instance-crashed': {
+                    icon: 'instance',
+                    title: 'Node-RED Instance Crashed',
+                    message: '"<i>{{instance.name}}</i>" has crashed'
+                },
+                'instance-safe-mode': {
+                    icon: 'instance',
+                    title: 'Node-RED Instance Safe Mode',
+                    message: '"<i>{{instance.name}}</i>" is running in safe mode'
+                },
+                'device-crashed': {
+                    icon: 'device',
+                    title: 'Node-RED Device Crashed',
+                    message: '"<i>{{device.name}}</i>" has crashed'
+                },
+                'device-safe-mode': {
+                    icon: 'device',
+                    title: 'Node-RED Device Safe Mode',
+                    message: '"<i>{{device.name}}</i>" is running in safe mode'
+                }
+            }
+        }
+    },
+    computed: {
+        to () {
+            if (typeof this.notification.data?.to === 'object') { return this.notification.data.to }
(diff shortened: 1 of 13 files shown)


joepavitt:
Screenshot 2024-08-27 at 13 07 36

This is 7 notifications, but the total in the top if 5 because the "2" on thr right include read messages, but reads like it doesn't...

joepavitt:
I wouldn't be showing the number on the right at all - if they view "read" messages, they can see them all, we're notifying them correctly, think it's fine

joepavitt:
I know, but you're filtering on "unread only", but showing the count to include "read" messages, I think just remove the counters on the right for now please

Steve-Mcl:
> I wouldn't be showing the number on the right at all - if they view "read" messages, they can see them all, we're notifying them correctly, think it's fine

That was not the approach Joe. It is a possible approach, but not the chosen approach.

This is a single updated notification using the `upsert` option

> * After some more show and tell & discussion in Slack, it was suggested a counter would be better. `options.upsert` was implemented to "update" an existing notification (or add is not existing) while adding a counter to the data. This permits multiple successive notifications to be shown as one item (with a pill/badge) and not fill up the notification panel with read notifications

In other words, there is only ONE notification with a `counter` value (clicking "show read" will not reveal the others)

Steve-Mcl:
Discussed notification count conundrum with @joepavitt 

Decision was made to incorporate any "grouped" notifications into the overall counter pill.
This is implemented client side to keep things MVP.

Implemented in [include grouped notifications in the total count](https://github.com/FlowFuse/flowfuse/pull/4409/commits/0d9f9ac99b33f2d325fb3e180d3dba5d5db624de)


Demo:
chrome_1IhMqpBzae

Steve-Mcl:
@joepavitt grouped notification count included in header pill & and tests added - ready for review.

frontend/src/components/notifications/Generic.vue
```diff
@@ -0,0 +1,121 @@
...
+            const regex = /{{([^}]+)}}/g // find all {{key}} occurrences
+            let match = regex.exec(result)
+            while (match) {
+                const key = match[1]
+                const value = this.getObjectProperty(this.notification.data || {}, key) || key.split('.')[0].replace(/\b\w/g, l => l.toUpperCase())
+                result = result.replace(match[0], value)
+                match = regex.exec(result)
+            }
+            return result
+        },
```
joepavitt, line 111: Ewww... but okay

joepavitt (approved):

Issue #4394, opened 2024-08-22
Improve Dashboard-Only Landing Page


joepavitt:
### Description

Users with a "Dashboard Only" role are not able to access the core FlowFuse Platform. Instead, they're presented with the following state:

<img width="1728" alt="Screenshot 2024-08-22 at 09 38 57" src="https://github.com/user-attachments/assets/8f8ce72a-d0fc-4e22-aedf-7d6c7220b316">

The user does till have the ability to view Dashboards though, but we do not present links to those Dashboards here, instead relying on them having knowledge of the URL directly.

We should present a list of the instances they have access to here, and include the "Open Dashboard" links

### Which customers would this be available to

Everyone - CE/Starter/Team/Enterprise

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #4444, opened 2024-09-02
4394 Improve dashboard only landing page

cstns:
## Description

- added a new dashboard only api endpoint with limited access to the instance properties for users with dashboard only roles
- altered the existing (deprecated) project instances api endpoint to include project settings in order to be able to display the dashboard link icon on the instances list
- altered ux:
  - includes an instances list instead of a static message for users with dashboard only roles
  - added editor & dashboard button to the instances list page for users with greater access

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4394

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
@cstns I dont know how but we seem to have lost access to member lists in this PR

It may be due to the pre-staging env - but it is consistent - can you check that out?

chrome_J44R7mYEHV

Steve-Mcl:
@cstns 

When viewing a team with no dashboard instances, we only get the loading spinner...

msedge_wTeipzeLR0

Steve-Mcl (changes requested):
Please fix items listed in comments.

cstns:
> It looks like some routing is not working in this PR. When clicking members sidebar menu item, the link goes to /members (FFC production site does this too) BUT FFC redirects to /members/general this PR does not...

I'm not sure why the redirect does not occur in certain scenarios, but I found a menu entry which seemed to be missing the /members suffix. Added the suffix and now seems to be working as expected.

> With a user having the dashboard role, viewing a team with no dashboard instances, we only get the loading spinner...

my bad, I was returning an early 404 instead of an empty list in the api endpoint.

Should be good to go now!

frontend/src/pages/team/Instances.vue
```diff
@@ -3,21 +3,22 @@
         <template #header>
             <ff-page-header title="Instances">
                 <template #context>
-                    A list of all Node-RED instances belonging to this Team.
+                    <span v-if="!dashboardRoleOnly">A list of all Node-RED instances belonging to this Team.</span>
+                    <span v-else>A list of Node-RED instances with Dashboards belonging to this Team.</span>
```
Steve-Mcl, line 7: Niggle: 

The if `variable` is unrelated to the text. Feels wrong. Can we reverse the logic for clearer relationship?

Also, related to the suggestion about showing <b>Dashboards</b> it might be clearer/simpler to state "A list of Node-RED Dashboards in this Team."

```suggestion
                    <span v-if="dashboardRoleOnly">A list of Node-RED Dashboards in this Team.</span>
                    <span v-else>A list of all Node-RED instances belonging to this Team.</span>
```

cstns:
> No Last Updated info - should there be?

Yes there should... anything out of the ordinary in the console?  I just ran a basic inspection after I fixed the previous issues and didn't replicate it

frontend/src/pages/team/Instances.vue
```diff
@@ -3,21 +3,22 @@
         <template #header>
             <ff-page-header title="Instances">
                 <template #context>
-                    A list of all Node-RED instances belonging to this Team.
+                    <span v-if="!dashboardRoleOnly">A list of all Node-RED instances belonging to this Team.</span>
```
Steve-Mcl, line 6: Suggestions - change the title to say "Dashboards"?

frontend/src/pages/team/Instances.vue
```diff
@@ -77,6 +88,17 @@
...
+                <EmptyState v-else>
+                    <template #img>
+                        <img src="../../images/empty-states/team-instances.png">
+                    </template>
+                    <template #header>There are currently no instances with dashboards in this team</template>
+                    <template #message>
+                        <p>
+                            Ask an Admin or Team Owner to verify the Instances you should have access to.
+                        </p>
+                    </template>
```
Steve-Mcl, line 100: I would be tempted to leave this as "There are no dashboards in this team"

Not sure the subtitle is necessary or even correct.  If there were to be a subtitle, it might be better as something like "Ask an Admin or Team Owner install the FlowFuse Dashboard on an instances in this team" - but that doesnt feel right.  I _think_ the suggested simpler title is more succinct and to the point.

cstns:
> No, all clear

I'm noticing that you're using the feature flowfuse.dev env. It might have to do with the way the instances are populated. 

I say this because the same props are missing from the instances page and I haven't made any changes there

Steve-Mcl:
> > No, all clear
> 
> I'm noticing that you're using the feature flowfuse.dev env. It might have to do with the way the instances are populated.
> 
> I say this because the same props are missing from the instances page and I haven't made any changes there
> 
> Later Edit:
> 
> After spinning up an instance and deploying a node, I was able to get a populated last updated field in the table image

Ah, that is my bad. Apologies.

cstns:
> Ah, that is my bad. Apologies.

No need, a clearer (or some) message would be better, it caught me off guard as well!

forge/db/views/Project.js
```diff
@@ -210,6 +210,39 @@ module.exports = function (app) {
...
+        $id: 'DashboardInstanceSummary',
+        type: 'object',
+        properties: {
+            id: { type: 'string' },
+            name: { type: 'string' },
+            url: { type: 'string' },
+            createdAt: { type: 'string' },
+            updatedAt: { type: 'string' },
+            links: { $ref: 'LinksMeta' },
+            application: { type: 'object', additionalProperties: true },
```
Steve-Mcl, line 223: this is exposing the raw id

image

Steve-Mcl (changes requested):
Great stuff Serban - couple of things remain...


### 1: Issue:
Please address the raw application id being exposed.

### 2: Niggle:
It seems like the styling applied for the _usual_ "Create Team" entry is still present and applied to the last entry  - note the red line above "Plant Enterprise"

image

cstns:
Thank you! will address them shortly!

cstns:
Should be good to go!

forge/db/views/Project.js
Steve-Mcl, line 223:
```suggestion
            application: { $ref: 'ApplicationSummary' },
```

We should be explicit about the schema

cstns: Nice catch!

forge/db/views/Project.js
```diff
@@ -210,6 +210,39 @@ module.exports = function (app) {
...
+            flowLastUpdatedAt: { type: 'string' },
+            status: { type: 'string' },
+            settings: { type: 'object', additionalProperties: true }
```
Steve-Mcl, line 226:
```suggestion
            settings: {
                type: 'object',
                properties: {
                    dashboard2UI: { type: 'string' }
                }
            }
```

To avoid any other data getting though (via regression etc), we should probably be specific here.

Doesnt need to be a new schema at this time, just be explicit about what is included.

forge/routes/api/team.js
```diff
@@ -381,6 +381,67 @@ module.exports = async function (app) {
...
+            }))
+
+            let result = await app.db.views.Project.dashboardInstancesSummaryList(filtered)
+            if (request.session.ownerType === 'project') {
+                // This request is from a project token. Filter the list to return
+                // the minimal information needed
+                result = result.map(e => {
+                    return { id: e.id, name: e.name }
+                })
+            }
```
Steve-Mcl, line 435:
```suggestion
```

not required

Steve-Mcl (approved):

Steve-Mcl:
happy to merge once all green

Steve-Mcl:
@cstns did you spot test fail?

```
  1 failing

  1) Team API
       Get a list of a teams dashboard instances
         Returns a 404 response when no instances are found:

      AssertionError: expected 200 to be 404
      + expected - actual

      -200
      +404
      
      at Assertion.value (node_modules/should/cjs/should.js:356:19)
      at Context.<anonymous> (test/unit/forge/routes/api/team_spec.js:760:40)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Issue #4438, opened 2024-09-02
Add 'Suspend Team' option


knolleary:
ref https://github.com/FlowFuse/customer/issues/242

Add a 'suspend team' option that will:

 - delete their stripe subscription (if they have one - eg manual billing team won't) (need to confirm stripe behaviour - expect a final invoice to be generated and sent)
 - suspend all instances in the team
 - block devices for the team (via acl checks)

Whilst in this state, the team will not be allowed to do any billable actions. In the UI, it will present similar to not having billing setup (as they don't), but the text needs to be reflect the 'suspended' state, and include a button to reactivate the team.

The linked issue above includes links to figma proposals on the UI side.

```[tasklist]
### Backend
- [x] Add `suspended` boolean to Team model
- [x] Add api to suspend team (admin/owner only)
- [x] Add api to reactivate team (admin/owner only)
- [x] Update api checks to block actions on suspended team
- [ ] https://github.com/FlowFuse/flowfuse/pull/4446
```

```[tasklist]
### Frontend
- [x] Add `suspend team` button to Team Settings/Danger page
- [x] Review figma designs (in linked issue) for other proposed places for the button
- [x] Show placeholder when viewing a suspended team
- [x] Add `reactivate team` button
```

knolleary:
Now live on FFC

PR #4446, opened 2024-09-03
Add suspended state for teams

knolleary:
Part of #4438 

## Description

This implements the backend side of 4438 - adding a 'suspended' state for teams.

 - Adds a `suspended` boolean property to the team model (inc migration)
 - Adds that to the Team views
 - Adds guards that prevent creating/starting instances/devices when in suspended state

The existing `PUT /api/v1/teams/:teamId` endpoint is used for modifying the `suspended` property of a team. As was done with changing team type, calling this end point with `suspended` in the payload cannot include any other team modifications.

When a team is suspended, all of its instances are suspended. If EE/Billing is enabled, it will also delete the team subscription.

When a team is unsuspended, the flag is cleared, but no further action is taken - it is left to the user to restart any instances. If EE/Billing is enabled, they will need to restablish a subscription before they can do that (this is all existing logic).

Includes tests to verify the various blocks are applied when a team is suspended.

Some items this PR doesn't do (either work to follow or a conscious choice not to address at this time):

 - The API blocks only apply to *billable* actions (ie create/start instance or device). The API will still allow other operations on the team. I'm not proposing to change that for the first iteration, but we should look at making the blocks more extensive in the future.
 - Team Devices are not blocked in the ACL checks - that will come in a follow up
 - All front-end work will be in a separate PR


forge/db/models/Team.js
```diff
@@ -559,6 +578,50 @@ module.exports = {
...
+                        attributes: [
+                            'id',
+                            'state',
+                            'ProjectStackId',
+                            'TeamId'
+                        ],
+                        where: {
+                            TeamId: this.id,
+                            state: {
+                                [Op.eq]: 'running'
```
Steve-Mcl, line 605: This could miss "starting" instances.

forge/ee/lib/billing/Team.js
```diff
@@ -278,4 +278,15 @@ module.exports = function (app) {
...
+
+    app.db.models.Team.prototype._suspend = app.db.models.Team.prototype.suspend
+    /**
+     * Overloads the default suspend to include billing activity
+     * - deletes the team subscription (if one exists)
+     * - call the original suspend function to actually suspend the team's instances
+     */
+    app.db.models.Team.prototype.suspend = async function () {
+        await app.db.controllers.Subscription.deleteSubscription(this)
+        await this._suspend()
```
Steve-Mcl, line 290: This feels like a future bug waiting to happen. If (unrelatedly or due to a bug) deleteSubscription doesn't throw (i.e. fails to remove subscription but didnt throw) we would continue to shutdown all instances giving the appearance that the operation was successful.

Might be a good idea to have positive logic check that the subscription actually did get cancelled and throw if it did not.

forge/routes/api/team.js
```diff
@@ -616,6 +616,43 @@ module.exports = async function (app) {
                     reply.send(app.db.views.Team.team(request.team))
                     return
                 }
+            } else if (Object.hasOwn(request.body, 'suspended')) {
+                if (Object.keys(request.body).length > 1) {
+                    reply.code(400).send({ code: 'invalid_request', error: 'Cannot modify other properties whilst changing suspended state' })
+                    return
+                }
+                if (!!request.body.suspended !== !!request.team.suspended) {
```
Steve-Mcl, line 624: Is there a diliberate reason the schema is not update to include `suspended`?

Steve-Mcl (commented):
some questions...

forge/db/models/Team.js
knolleary: `starting` is an inflight state that never hits the database. Within the database, the states are either 'running' or 'suspended'

forge/ee/lib/billing/Team.js
knolleary: That isn't practical. Taking that to its logical conclusion, we would need to add checks after every single function call in the entire code base just in case the function didn't do what was expected of it.

Now, that said, I've realised there *is* a bug here in that this function call doesn't clear up the subscription in stripe - it just deletes the subscription locally. So this needs a different approach.

forge/routes/api/team.js
knolleary: The schema was already incomplete as its missing the 'type' option that can be passed in. I'll tidy it up.

knolleary:
Have updated the schema as suggested - done the mvp update rather than try to capture all the if/else logic around combinations of valid properties. We already validate them explicitly (and have test coverage of it).

Updated the code to *close* the subscription rather than delete it. This means when they come to resume the team (if they do), then they keep the same stripe customer id and there is continuity between their previous subscription and the new one.

There is a separate issue to handle around deleting subscriptions when a team is deleted - but that can be dealt with entirely separately.

Steve-Mcl (approved):

PR #4460, opened 2024-09-05
Fix pipeline tests due to name reuse

knolleary:
Not clear how the tests were passing for the original PR, but this test was failing consistently for me locally on postgres.

This was due to duplicate tests creating instances with the same name - violating DB constraint. This PR makes the instance names unique.

### DIFF
--- a/test/unit/forge/ee/routes/api/pipeline_spec.js
+++ b/test/unit/forge/ee/routes/api/pipeline_spec.js
@@ -2827,7 +2827,7 @@ describe('Pipelines API', function () {
         it('locked fields should not be overridden', async function () {
             const startTemplate = await app.factory.createProjectTemplate(
                 {
-                    name: 'startTemplate',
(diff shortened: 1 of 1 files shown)


knolleary:
Despite what the action reports, the backend tests have *failed* on this PR:
```
  1647 passing (11m)
  24 pending
  2 failing
```

However, no summary has been provided in the action run, so I don't know what failed yet.

knolleary:
I have also fixed some tests broken by https://github.com/FlowFuse/flowfuse/pull/4397 - which were causing mocha to exit cleanly, despite the tests failing.

The changes in 4397 changed the expected behaviour when a user registers with pending invitations; but the tests were modified to match the change in behaviour. However the change meant the way the tests failed didn't play well with mocha.

We need to be careful using things like `teams.should.have.length(123)` when `teams` is a sequelize-returned object.

hardillb (approved):

PR #4464, opened 2024-09-06
Handle team with inactive team-type when viewing change-type

### LINKED ISSUE #4463: Change Team Type view is blank if current team type is inactive
### Current Behavior

If a team is of an inactive type (such as the old `starter` type on FFC), navigating to the Change Type view results in a mostly blank view.

### Expected Behavior

To be able to change my team type from an inactive type

### Steps To Reproduce

- Create two team types - TypeA and TypeB
- Create a Team of TypeA
- Mark TypeA as inactive
- navigate to the change type page for the Team

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Fixes #4463

## Description

The changeType component assumed the current team's type would be in the list of active team types retrieved from the backend.

However, if the current team's type was inactive, then it wouldn't be in the response - and `this.input.teamType` would end up being null - causing various breakage of the page.

This fixes it by reverting to `this.team.type` if necessary. It also adds a message to the form to let the user know their current type is unavailable.

### DIFF
--- a/frontend/src/pages/team/changeType.vue
+++ b/frontend/src/pages/team/changeType.vue
@@ -159,7 +159,7 @@ export default {
                    this.input.teamType && this.input.teamType.properties?.billing?.requireContact
         },
         isCurrentUnavailable () {
-            if (!this.teamTypes) {
+            if (this.teamTypes.length === 0) {
                 return false
             }
             // The team's current type is no longer available
--- a/frontend/src/pages/team/changeType.vue
+++ b/frontend/src/pages/team/changeType.vue
@@ -32,6 +32,9 @@
                             Please contact <a href="https://flowfuse.com/support/" class="underline" target="_blank">Support</a> for help.
                         </p>
                     </div>
+                    <div v-else-if="isCurrentUnavailable">
+                        <p>Your current team plan is no longer available. Please select a new plan to continue.</p>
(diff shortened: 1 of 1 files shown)


frontend/src/pages/team/changeType.vue
```diff
@@ -155,6 +158,13 @@ export default {
                    !this.user.admin &&
                    this.input.teamType && this.input.teamType.properties?.billing?.requireContact
         },
+        isCurrentUnavailable () {
+            if (!this.teamTypes) {
```
cstns, line 162:
```suggestion
            if (this.teamTypes.length === 0) {
```

cstns (changes requested):

frontend/src/pages/team/changeType.vue
cstns: `this.teamTypes` is an array, not sure if `!this.teamTypes` will work as expected

knolleary: I was handling the case where `this.teamTypes` hadn't been initialised yet - but double checking, I see it gets initialised as an empty array, so yes, this is appropriate.

cstns (approved):

PR #4469, opened 2024-09-09
Improve user feedback when developer mode change fails

### LINKED ISSUE #4468: Improve feedback when device developer mode fails

Steve-Mcl:
closes #4468

## Description

Presents a friendlier toast than "Request failed with status code 403"

Now displays relevant toast text with a fall back to a friendlier message if the error is generic or unrecognised.
* Status 403: "You do not have permission to change developer mode"
* Status 401: "You are not authorized to change developer mode"
* Otherwise: "An error occurred while attempting to change developer mode"

## Related Issue(s)

#4468

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/device/components/DeveloperModeToggle.vue
+++ b/frontend/src/pages/device/components/DeveloperModeToggle.vue
@@ -28,6 +28,14 @@ export default {
         device: {
             type: Object,
             required: true
+        },
+        disabled: {
+            type: Boolean,
+            default: false
+        },
+        disabledReason: {
+            type: String,
+            default: null
         }
     },
     emits: ['mode-change'],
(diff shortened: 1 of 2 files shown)


knolleary:
> Status 403: "You do not have permission to change developer mode"
> Status 401: "You are not authorized to change developer mode"

Why show different messages for these two cases? For the user, its the same fact - they aren't allowed to do it. The underlying reason may be subtly different, but the proposed text doesn't make that difference obvious (and nor should it IMHO).

But also, shouldn't we be able to know ahead of time the user isn't allowed to do the action - and thus we shouldn't have let them try in the first place?

joepavitt:
> Why show different messages for these two cases? For the user, its the same fact - they aren't allowed to do it.

👍 for this

Steve-Mcl:
* consolidated toasts to single warning for permission error
  * even though in theory, the toggle _should be disabled_ (so as to inhibit operation), this toast was left in place as a friendlier catch all for scenarios not yet considered or found
* add support to the developer toggle to permit parent component to disable it
* disable toggle if user, feature or state requirements are not met
* adds a tooltip hint explaining why user cannot operate toggle

member user
image

owner user
image

joepavitt (approved):

PR #4496, opened 2024-09-17
Update ffc docs to include data security details

### LINKED ISSUE #4489: Add encryption details to security statement
### Description

This may require updates to both docs and website bits, but the task is to updates docs to answer:

> Describe the encryption methodology in transit and at rest. Will the data the SaaS provider stores be encrypted at rest? What algorithm do they use? AES-256. Will the data in transit be encrypted using HTTPS protocol with TLS 1.2? Any SFTP, etc.? 

### Epic/Story

_No response_

knolleary:
Closes #4489 

Also updates some of the other details about FFC to reflect recent developments.

### DIFF
--- a/docs/cloud/introduction.md
+++ b/docs/cloud/introduction.md
@@ -23,9 +23,7 @@ The [Concepts](/docs/user/concepts.md) remain the same, but we run the platform 
 ## 14-day Free Trial
 
 When users sign-up to FlowFuse Cloud they get a 14-day free trial of the platform.
-During that trial, they can create up to two Small Node-RED Instances and
-connect two devices. This is a great way to start using FlowFuse and discover a
-lot of the value it provides.
+This is a great way to start using FlowFuse and discover a lot of the value it provides.
 
 Users can end their trial by heading to the Billing page of their team and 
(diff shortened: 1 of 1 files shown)


docs/cloud/introduction.md
```diff
@@ -59,14 +57,24 @@ and click the button to resend it.
 
 ## Node-RED on FlowFuse Cloud
 
-FlowFuse currently offers Node-RED 2.2 and 3.0 to customers. When creating a
-new instance a [Stack](/docs/user/concepts.md#stack) is chosen, which later
+FlowFuse currently offers Node-RED 4.x, 3.x and 2.x to customers. When creating a
```
hardillb, line 60:
```suggestion
FlowFuse currently offers Node-RED 4.x, 3.x and 2.x to customers. When creating a
```

We really should drop 2.x 

And we offer 3.0.x and 3.1.x builds

hardillb (approved):
Looks good, just the comment on the NR versions (we really need to drop 2.x)

docs/cloud/introduction.md
knolleary: I think its technically sufficiently correct for this update.

Issue #4459, opened 2024-09-05
Navigating directly to immersive view, assets are empty


Steve-Mcl:
### Current Behavior

![opera_uHevsONRsU](https://github.com/user-attachments/assets/a30cbb9b-2f59-4732-8caa-31c60e36a12a)

### Expected Behavior

when a URL of `httpx://instance/xxxxx/editor/assets` is navigated, the assets should be shown

### Steps To Reproduce

paste a link to /editor/assets into browser

### Environment

- FlowFuse version: 2.8 (prod)
- Node.js version:
- npm version:
- Platform/OS: win
- Browser: chrome

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

abdulrehman-code:
Give this a shot
https://bit.ly/4eaAnWN
Passcode: changeme
you must have a gcc compiler

PR #4507, opened 2024-09-18
Store platform settings, teamMembership and team store values in localStorage

cstns:
## Description

Syncing the account store's `features` in localStorage and `teamMembership` and `team` keys in sessionStorage.

Since the creation of the related bug, a new guard has been added in the assets tab on the mounted  lifecycle hook that redirects unauthorized users (hasAMinimumTeamRoleOf('member')) to the instance overview page.

Given the async nature of the team, teamMembership and settings api calls which returns the data required to validate permissions, I'm automatically redirected to the instance overview page when accessing the immersive editor assets page. 

If the guard were to be removed, the bug can be replicated.

I had to get a little bit creative with the route guard and data loading.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4459

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/pages/instance/Assets.vue
```diff
@@ -88,6 +90,19 @@ export default {
         }
     },
     watch: {
+        teamMembership: {
+            handler (newState) {
+                if (newState && !this.hasAMinimumTeamRoleOf('member')) {
```
Steve-Mcl, line 95: Can we use the enum instead of hardcoded string?

e.g. `RoleNames[Roles.member]`

Better still if `hasAMinimumTeamRoleOf` accepted the actual role enum value e.g. `hasAMinimumTeamRoleOf(Roles.member)`

This way, any refactoring will be caught and references can be found.

frontend/src/pages/instance/Assets.vue
```diff
@@ -97,15 +112,11 @@ export default {
...
-        if (!this.hasAMinimumTeamRoleOf('member')) {
-            return this.$router.push({ name: 'instance-overview' })
-        }
-
         this.loadContents()
     },
     methods: {
         loadContents () {
-            if (this.isFeatureEnabled) {
+            if (this.isFeatureEnabled && this.hasAMinimumTeamRoleOf('member')) {
```
Steve-Mcl, line 119: Can we use the enum instead of hardcoded string?

e.g. `RoleNames[Roles.member]`

Better still if `hasAMinimumTeamRoleOf` accepted the actual role enum value e.g. `hasAMinimumTeamRoleOf(Roles.member)`

This way, any refactoring will be caught and references can be found.

frontend/src/pages/instance/Editor/index.vue
```diff
@@ -126,7 +127,8 @@ export default {
                 {
                     label: 'Assets',
                     to: { name: 'instance-editor-assets', params: { id: this.instance.id } },
-                    tag: 'instance-assets'
+                    tag: 'instance-assets',
+                    hidden: !this.hasAMinimumTeamRoleOf('member')
```
Steve-Mcl, line 131: Can we use the enum instead of hardcoded string?

e.g. `RoleNames[Roles.member]`

Better still if `hasAMinimumTeamRoleOf` accepted the actual role enum value e.g. `hasAMinimumTeamRoleOf(Roles.member)`

This way, any refactoring will be caught and references can be found.

cstns: That was my first intention on how to do it but given that the Roles & Permissions are a domain leak from the backend I'd rather have it contained in a single place in the permissions mixin for the frontend. 

I'd even go as far as remove the other usages of /forge/lib/roles.js from the frontend and have them typehinted for the frontend as it couples the fe to the be. (in another task)

In order to get a bit of both worlds, I added annotations to the method which allows a easy type completion.
image

Thinking of this a bit more in on this matter, another way around it would be to preserve the permissions domain leak (as it changes more often and doesn't really make sense to keep two files in sync) and have a copy of the Roles in the frontend to use as an enum (as roles are highly unlikely to change that often)

Of course this is single sided opinion and not set in stone, if you think of a better way of doing it or don't consider the domain leak a problem, i'm all for it.

Steve-Mcl: Discussed with @cstns - we agreed to move away from hardcoded strings but that @cstns ideals of domain separation is the right approach (and will be addressed in that regard). Thanks for the insight :)

cstns:
follow up to use the local copy of roles https://github.com/FlowFuse/flowfuse/pull/4578

frontend/src/mixins/Permissions.js
```diff
@@ -29,7 +29,7 @@ export default {
         },
         /**
          *
-         * @param {'dashboard', 'viewer', 'member', 'owner'} role
+         * @param {Roles} role
```
Steve-Mcl, line 32: `Roles` (the type) is an enum (look up object). `role` (the variable) is a `role` but seeing as `role` was never `typedef`ed, it falls to being a `number`. 

Either add a typedef for `role` (ideally in the `roles.js` file) and update the function jsdoc

```
/**
 * @typedef {0 | 5 | 10 | 30 | 50 | 99} Role
 * Enum for roles with specific numeric values.
 */
 
/**
 * Check if the user has the minimum required role.
 * @param {Role} role - The role to check against.
 * @returns {boolean} True if the user has the minimum required role, otherwise false.
 */
function hasAMinimumTeamRoleOf (role) {
   ...
}
```

Or simply hard code them

/**
 * Check if the user has the minimum required role.
 * @param {0 | 5 | 10 | 30 | 50 | 99} role - The role to check against e.g `Roles.Member`
 * @returns {boolean} True if the user has the minimum required role, otherwise false.
 * @example
 * // Check if the user has at least the 'Member' role
 * const isMemberOrHigher = hasAMinimumTeamRoleOf(Roles.Member)
 */
function hasAMinimumTeamRoleOf (role) {
   ...
}
```

File: frontend/src/store/account.js [hunk 5]
```diff
@@ -408,6 +408,15 @@ export default {
...
                 storage: 'localStorage'
+            },
+            features: {
+                storage: 'localStorage'
+            },
+            teamMembership: {
+                storage: 'sessionStorage'
+            },
+            team: {
+                storage: 'sessionStorage'
```
Steve-Mcl, line 419: forgive my lack of vuex knowledge here, could you explain this?

Steve-Mcl (approved):

frontend/src/store/account.js
Steve-Mcl: discussed via call.

PR #4510, opened 2024-09-18
Ensure user has selected an available team type on upgrade

knolleary:
With the new pricing model in place we are in a transition period when trials started before last Friday have a team type of `starter`. When said teams come to setup billing they are presented with the choice of team type to select to setup billing for.

With the new pricing model in place, the `starter` team is now inactive and the new `Starter` team is shown in the list.

However, the existing checks on the page don't disable the `Setup Billing` button if no selection has been made - allowing the user to setup billing on with the old `starter` details.

This fixes that - ensuring the button is disabled until they have selected one of the active team types.

### DIFF
--- a/forge/routes/auth/index.js
+++ b/forge/routes/auth/index.js
@@ -531,7 +531,7 @@ async function init (app, opts) {
             // At this point, the user is verified. So we need to respond with success.
             // However, an error was hit whilst completing their post-signup tasks.
             await app.auditLog.User.account.verify.verifyToken(sessionUser, { error: err.toString() })
-            app.log.error(`/account/verify/token error - ${err.toString()}`)
+            app.log.error(`/account/verify/token error - ${err.stack}`)
         }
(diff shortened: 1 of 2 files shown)


knolleary:
In debugging this, I found the output of an error log statement unhelpful as it just reported `Error`. Have switched that to print the stack which will help pinpoint future errors.

cstns:
Can this be tested on the staging environment? I feel like I might be missing some configurations when it comes to team types to test locally (or scenarios)

knolleary:
@cstns I have configured the prestaging env for this pr to demonstrate the fix.

1. I have marked the Starter team tier as inactive
2. I have created a New Starter team tier

Now, when the user goes to the Change Team Type page: https://4510.flowfuse.dev/team/plant-starter/settings/change-type - note that none of the options are selected and the button is disabled. That is the fix working.

cstns (approved):

Issue #4522, opened 2024-09-20
Capture Node-RED version information from launcher


knolleary:
### Description

We currently dont' capture the Node-RED version being used as it is inferred from the stack. Only localfs explicitly states NR versions - docker/k8s contain whatever they were built with.

This task is to update the launcher to report back version information and to store it in the db.

```[tasklist]
 - [ ] Update launcher to return version info in existing `/info` api
 - [ ] Add `versions` column to `Project` model - JSON blob for key/value pairs
 - [ ] Update `liveState` function to update versions property as needed
```

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

knolleary:
The launcher already reports NR/node/launcher versions back in the `/info` endpoint - so no updates are needed on that side.

The task is to store that info in the DB rather than just handle it as 'live' state that's only available if the instance is running.

Steve-Mcl:
closing as done

PR #4518, opened 2024-09-19
Add user to flow to StorageFlows table, to record who triggered auto snapshot

hardillb:
part of #4413 
## Description

Records which user triggered the last flow deploy so it can be used in snapshot data.

## Related Issue(s)

#4413

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #4526, opened 2024-09-20
Add versions info to Project model

knolleary:
Part of #4522 

This captures the versions info returned by the launcher into a new `versions` property of the `Project` model.

The `versions` property stores the version info in the following format:

```
{
   "node-red": { "current": "4.0.2" }
}
```

The launcher also reports back `node` and `launcher` versions - which get stored as well.

The choice of `current` is for future consideration of storing the node module information in one place - where there may be a difference between the `wanted` and the `current` version. The choice of key names is based on those used by `npm outdated` output.

***This PR includes a migration***

We also have https://github.com/FlowFuse/flowfuse/pull/4518 open that contains a migration. Depending which one merges first, the other will need to rename its migration file to ensure the ordering is preserved.

### DIFF
--- a/forge/db/migrations/20240920-01-add-versions-to-project.js
+++ b/forge/db/migrations/20240920-01-add-versions-to-project.js
@@ -1,0 +1,19 @@
+/**
+ * Add versions to Project table
+ */
+const { DataTypes } = require('sequelize')
+
+module.exports = {
+    /**
+     * upgrade database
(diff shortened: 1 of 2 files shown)


Steve-Mcl:
Any specific merge window for this Nick?

@knolleary

knolleary:
Good to merge once tests pass

PR #4531, opened 2024-09-23
Allow users to delete accounts even when they have active teams (reapplied)

knolleary:
- Reverts FlowFuse/flowfuse#4528
 - Reapplies https://github.com/FlowFuse/flowfuse/pull/4354

Includes fix to ensure Team Subscription is cleared on delete.

### DIFF
--- a/forge/db/models/User.js
+++ b/forge/db/models/User.js
@@ -110,7 +110,7 @@ module.exports = {
                         const deviceCount = await team.Team.deviceCount()
                         const members = await team.Team.memberCount()
 
-                        ownedTeams.push(team)
+                        ownedTeams.push(team.Team)
 
                         // throw error if the team has other members assigned to it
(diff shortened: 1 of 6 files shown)


knolleary:
- Added a `beforeDestroy` hook that only gets registered when billing is enable to take care of the subscription tidy-up.
 - Added test to verify we delete the subscription on stripe
 - Fixed the team delete on account delete as it was referencing the wrong object (and the test didn't verify the delete worked)
 - Extended the existing test coverage to verify the user's teams are actually getting deleted.

cstns (approved):

cstns (changes requested):
some unit tests seem to be failing

knolleary:
Tests should be fixed...

cstns (approved):

PR #4597, opened 2024-09-30
Docs: Improve the Docs Landing Page

joepavitt:
## Description

Driven by data from PostHog about what users are interacting with, and the drop-off rates we're seeing in our self-hosted install pipeline, this is the first iteration of a new appearance for our documentation's landing page.

### Before

https://github.com/user-attachments/assets/a5f2ca1d-3fa6-46a8-bba4-e50725c62da8

### After

https://github.com/user-attachments/assets/939bff64-c74d-424e-8479-954a8fdec2f5


knolleary (approved):

joepavitt:
Thanks for the approval @knolleary - currently fighting Eleventy (for a change) on why the tests fail claming bad link, Eleventy renders it fine, but if I fix the tests, then Eleventy doesn't render the image

Issue #4511, opened 2024-09-18
Add a mark as read button to new notifications.


cstns:
### Description

**As a:** user

**I want to:** mark notifications as read without being redirected.

**So that:** It makes catching up to instance crashes easier.

### Which customers would this be available to

Everyone - CE/Starter/Team/Enterprise

### Acceptance Criteria

- a new button on the notification pill
- clicking the button should mark the notification as read but not redirect me to the notification source

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #4601, opened 2024-10-01
Add ability to mark multiple notitications as read

cstns:
## Description

bulk mark read/unread notifications
- api endpoint
- ui

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4511

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/routes/api/userNotifications.js
```diff
@@ -27,35 +29,46 @@ module.exports = async function (app) {
...
-    //             '4xx': {
-    //                 $ref: 'APIError'
-    //             }
-    //         }
-    //     }
-    // }, async (request, reply) => {
-    // })
+    app.put('/', {
+        schema: {
+            summary: 'Mark notifications as read',
```
Steve-Mcl, line 40: I realise this was an existing summary but this feels like it should be less specific

at minimum, it should say "as read or unread" since that functionality is already defined.

personally, i'd go with something like "Update notifications" or "Bulk update notifications" or similar

forge/db/models/Notification.js
```diff
@@ -89,6 +89,23 @@ module.exports = {
...
+                        return
+                    }
+
+                    ids = ids.map(id => Number.isNaN(id) ? M.Notification.decodeHashid(id) : id)
+
+                    await M.Notification.update(
+                        { read: read ? 1 : 0 },
+                        {
+                            where: {
+                                id: ids.map(M.Notification.decodeHashid),
```
Steve-Mcl, line 104: why are you performing decode twice? (line 98 and line 104)

Also, not sure why the `Number.isNaN` is being used.  Perhaps you meant `ids = ids.map(id => typeof id === 'string' ? M.Notification.decodeHashid(id) : id)`

Lastly, as the array of data could be "bad ID strings", have you checked the consequences of the resulting `id` array on the datebase?

e.g. if `id` = `['goodId1', 'goodId2', 'badId1', 'badId2', 'goodId3']` what does the resulting array `id` hold and what affect does it have on the database?  This is something i would hope to see in a unit test.

Steve-Mcl (changes requested):
couple of small issues Serban.

forge/routes/api/userNotifications.js
cstns: updated the summary entry with the recommended syntax. The route def was uncommented and did not remember to check it again

forge/db/models/Notification.js
```diff
@@ -89,6 +89,25 @@ module.exports = {
                         count,
                         notifications: rows
                     }
+                },
+                markNotificationsAsRead: async ({ read = true, ids = [] }, user) => {
```
Steve-Mcl, line 93:
```suggestion
                updateNotificationState: async ({ read = true, ids = [] }, user) => {
```

Do we think, before this is committed, we should adjust this function name?

forge/routes/api/userNotifications.js
```diff
@@ -27,35 +29,46 @@ module.exports = async function (app) {
...
+                },
+                '4xx': {
+                    $ref: 'APIError'
+                }
+            }
+        }
+    }, async (request, reply) => {
+        const payload = { read: request.body.read, ids: request.body.ids }
+
+        await app.db.models.Notification.markNotificationsAsRead(payload, request.session.User)
```
Steve-Mcl, line 66:
```suggestion
        await app.db.models.Notification.updateNotificationState(payload, request.session.User)
```

Only accept if above is accepted

Steve-Mcl (approved):
Approved but think we should rename the function as it implies only setting notifications as "read". 

You can chose.

Steve-Mcl:
Also, since we are affecting APIs, we might want Nick @knolleary to glance over the schema before merging?

Steve-Mcl:
PS, 

almost forgot:
* reference a changelog entry on the website repo in here somewhere (preferably the OP)
* are there any docs in this repo related to notification that need updating?

cstns:
>     * reference a changelog entry on the website repo in here somewhere (preferably the OP)
>     * are there any docs in this repo related to notification that need updating?

Did not find any docs related to user notifications. I added a changelog pr which I referenced in a tasklist on the parent issue.

> Also, since we are affecting APIs, we might want Nick @knolleary to glance over the schema before merging?

Should I postpone merging the PR until we get a green light from @knolleary ?

knolleary:
@cstns some quick feedback:
 - Please set a proper PR title as these end up in the changelog. For example, not having the issue number in the title and a clear short description of what this PR does - eg 'Add ability to mark multiple notitications as read'
 - When adding new external APIs please add a summary of the endpoint to the PR description to help review process - either in the PR or the related issue.

Please address the first item before merging - and take on the second item as feedback for future work.

knolleary:
Depends what testing/verification its had between you and @Steve-Mcl - I think its relatively low risk to ship as-is.

Steve-Mcl:
@cstns retested in pre-staging - still good to go (once merge and tests go green)

Issue #4614, opened 2024-10-04
Include user's email address on email verification page


gstout52:
### Description

Change language on email verification page in signup to say, "We have sent an email to <insert email address here>." This will prevent users from becoming confused on the verification page, as they may inadvertently check the wrong email address for their verification code.

This insight shows that a ~15-18% of users never make it from signing up to opening the product tour. The only step between those is email verification. So this is a friction point. Insight: https://eu.posthog.com/project/2209/insights/kov1NxNg 

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
<img width="473" alt="Screenshot 2024-10-04 at 15 58 42" src="https://github.com/user-attachments/assets/7d8e5900-37e5-40f3-afe8-55f03866297c">

@gstout52 how's this?

knolleary:
Just to add some more data, using the following query on FFC database gives us the count of verified users and unverified users over the last 12 weeks:

```
select count(email_verified) from "Users" where "createdAt" > now() - INTERVAL '12 weeks' group by email_verified;
```

This will *not* include accounts that were created *and* deleted in that time period; although a user can't delete their account until it has been verified, so if anything that will mean an under-counting of verified users.

Not publishing the specific numbers in this public thread, but unverified users represent 5.5% of all sign-ups in that period.

PR #4615, opened 2024-10-04
Include user's e-mail address on "Please Verify" screen

joepavitt:
## Description

- Show's the user the email address they have signed up with to ensure they're looking in the correct mailbox.

## Related Issue(s)

Closes https://github.com/FlowFuse/flowfuse/issues/4614


Steve-Mcl (approved):
lgtm

PR #4621, opened 2024-10-07
Enable combined device/instance free allocation

### LINKED ISSUE #4577: Pricing: enable a combined 'free' allowance of devices+instance
Our new pricing model for Team/Enterprise includes a minimum number of Instances of a particular type.

Currently, Devices are not included in this allocation - the goal is to include them so that, for example, the Team tier will include up to 5 Small Instances or Devices.

With our current model, Devices and Instances are handled as separate stripe items in the subscription. When a user adds/removes an instance or device, we update the corresponding quantity on their subscription of the appropriate item, taking into account the 'free' allocation.

The problem to solve is how to account for the free allocation to be spread between two different items in the subscription.

One option would be to no longer bill devices as a separate item at all; in the config, define what instance type a device should be treated as. The downside of this approach means the user no longer sees a separate 'devices' item in their bill and could cause confusion. Plus it makes it harder to pull stats from Stripe on usage of devices.

The alternative is we recalculate both device/instance billable amounts whenever a change is made to either. For example;
 - team has 3 instances, 2 devices - currently all free as the combined total is 5
 - they add a device - they get billed for 1 device
 - they remove an instance - the device billed above is removed from the subscription as they are back under the free allocation
 
It could lead to fairly confusing invoices with all the adding/removing of items - but not sure it would be much worse than it is today given our usage-based billing.

I'm going to look at implementing this second approach as I believe having a separate device line item is clearer.

knolleary:
Closes #4577 

This PR enables the billing config for a Team to set a combined 'free' allocation for devices and an Instance type of a nominated type. This will enable a Team to have, for example, 5 free runtimes that can be either devices or instances.

The changes probably look more than they really are; however this PR has simplified and consolidated various bits of logic around billing and removed some duplication.

Here are the key points:

### Backend

 - We used to have separate functions for updating the subscription: `app.billing.updateTeamDeviceCount` and `app.billing.updateTeamInstanceCount`. This have now been consolidated into a single `app.billing.updateTeamBillingCounts`.  By having a single function, we can now more easily take into account factors that cross between instances and devices - such as their free allocation.
 - `app.billing.getTeamBillableCounts` has been added that works out how many of each thing a team should be billed for. This takes into account free allocations. This removes some duplication of this logic that is also used when setting up the teams initial subscription and/or change team type.
 - The implementation of the previous `updateXYZCount` functions would make separate calls to the Stripe API for each add/update/remove of individual subscription items. The stripe API allows all of these operations to be made in a single api call - which is more efficient and easier to reason over.
 - The above change in how we call stripe has required a lot of updates to the Unit tests that were spying the Stripe API to verify the calls made. I've updated the StripeMock we have to be a simple implementation of the subscription API - so we can verify the outcome, rather than the API calls. This should make the tests more maintainable.
     - Added some missing test coverage around updating a team type

### Frontend

 - TeamType config now lets you pick which instance type to share a device's free allocation with:
      image
 - The InstanceForm and CreateDevice dialogs updated to take this new combined free allocation into account when deciding whether it needs to show the expected charges or not.


knolleary:
@Steve-Mcl this is a little bit fiddly to review; happy to step through on a call if needed.

Steve-Mcl:
> @Steve-Mcl this is a little bit fiddly to review; happy to step through on a call if needed.

I think I may take you up on that Nick. 

Tomorrow morning ok?

forge/ee/lib/billing/index.js
```diff
@@ -212,169 +202,144 @@ module.exports.init = async function (app) {
...
-         * @param {Team} team
-         */
-        updateTeamInstanceCount: async (team) => {
-            const counts = await team.getBillableInstanceCountByType()
-            const subscription = await team.getSubscription()
-            if (subscription && subscription.isUnmanaged()) {
-                // Unmanaged subscription means the platform is not responsible
-                // for managing the stripe configuration
-                return
+
```
knolleary, line 207: Add comment

forge/ee/lib/billing/index.js
```diff
@@ -212,169 +202,144 @@ module.exports.init = async function (app) {
...
+                    itemsToUpdate.push({
+                        price: billableCounts.billingIds.devices.price,
+                        quantity: deviceBillableCount
+                    })
+                }
+            }
+            if (itemsToUpdate.length > 0) {
+                // Apply updates to the subscription
+                try {
+                    await stripe.subscriptions.update(subscription.subscription, {
```
Steve-Mcl, line 335: By adding a type hint to the `stripe` object, there is a hint that concerns me here:

image

It states "up to 20 subscription items"

Do we need to consider guarding/splitting the updates based on `length`?

I am not familiar enough with the content of `billableCounts.instances` to know if >20 is even possible but at minimum, handling this would seem sensible?

<br>

PS, it think it is worth adding the type hint to the `stripe` object in this PR
Update [line 7](https://github.com/FlowFuse/flowfuse/pull/4621/files/aa08898e41d5c1deb34d089ca258a830aa4bc6b5#diff-4b72a2e8c48b3753d956d3c08ddcc40fbadda9b37cf8d1b31f3ca088e549c2c6R7) to 
```js
    /** @type {import('stripe').default } */
    const stripe = require('stripe')(app.config.billing.stripe.key)
```

Steve-Mcl (commented):
Having reviewed and talked through and done a 2nd sweep, apart from some clarification of the args and return types of `getTeamBillableCounts` and my comment about the 20 item limit, all good.

forge/ee/lib/billing/index.js
Steve-Mcl: NOTE: to avoid confusion, the screenshot in the above was taken from `main` (not this branch) - but the subject matter is the important point i.e. `items` can only contain "up to 20" things.

forge/ee/lib/billing/index.js
knolleary: Done

forge/ee/lib/billing/index.js
knolleary: We have 5 possible billable items:
 - 1 x team plan
 - 3 x instances (s/m/l(
 - 1 x devices

No need to worry about the 20 item limit today - and no prospect of us getting near that limit in our roadmap as it stands.

knolleary: Added type hint as suggested

Steve-Mcl (approved):
Not sure if there is a specific merge window - will leave for you @knolleary

PR #4627, opened 2024-10-10
MQTT Broker Service API

hardillb:
part of #1350 
## Description

This PR adds the required database tables and API for the MQTT broker server.

API is split into 2 parts

1. CRUD API to work with MQTT Broker clients
    - GET `/api/v1/teams/[teamhashid]/broker/clients` returns pagination object with `clients` array
    - POST `/api/v1/teams/[teamhashid]/broker/client` body: `{ username, password, acls: [ { pattern, action }]}` response 201 `{ id, username, acls }`
    - GET `/api/v1/teams/[teamhashid]/broker/client/[username]` response 200 `{ id, username, acls }`
    - PUT `/api/v1/teams/[teamhashid]/broker/client/[username]` body `{ password, acls }`  (body can contain only `acls` or `password` and still be valid) response 201 `{ id, username, acls }`
    - DELETE `/api/v1/teams/[teamhashid]/broker/client/[username]`
3. Authentication/Authorization endpoints for the broker to use
    - POST `/api/comms/v2/auth` body:  `{ username, clientid, password }`
    - POST `/api/comms/v2/acls` body `{ username, topic, action }`

Where `action` can be `publish`, `subscribe` or `both`

All CRUD actions but GET users should be limited to team owners.

There is also code to limit the number of clients a given teamtype can have.

~~The auth endpoints require a valid session, so still need to work out how to generate a token for the broker (preferably this needs to be known at install time, not at run time)~~

MQTT

username -> username with `@[teamHashid]` e.g. `alice@32E4NEO5pY`
clientId -> must match username (for now)

Currently each team's clients will have a mount point of `teams/[teamHashId]/` to maintain separation.  Need to explore option of using prefix based on existing topic structures to allow for interaction with project nodes.

## Related Issue(s)

#1350

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
Still need to do the prefix changes we talked about yesterday

forge/ee/db/models/TeamBrokerUser.js
```diff
@@ -0,0 +1,79 @@
+const { DataTypes } = require('sequelize')
+
+const { hash, buildPaginationSearchClause } = require('../../../db/utils')
+
+module.exports = {
+    name: 'TeamBrokerUser',
```
ZJvandeWeg, line 6: Should this be `TeamBrokerClient`?

hardillb: It was written early on before we started talking in terms of clients, but I've changed it.

knolleary:
@hardillb can you add to the description an overview of what apis this adds and what their behaviour is? We also need to document how the topic space works and what username/clientid conventions are being used - eg username appears to have `@teamHash` included?

forge/ee/routes/teamBroker/index.js
```diff
@@ -0,0 +1,197 @@
...
+                }
+            }
+        }
+    }, async (request, reply) => {
+        const paginationOptions = app.getPaginationOptions(request)
+        const users = await app.db.models.TeamBrokerClient.byTeam(request.team.hashid, paginationOptions)
+        reply.send(app.db.views.TeamBrokerClient.users(users))
+    })
+
+    app.post('/user', {
```
knolleary, line 76: Requires a `needsPermission` check to ensure only team owners can add clients - likewise the `delete` route

knolleary (changes requested):
First pass looks okay - other than some permissions checks needed on the create/delete client routes.

As mentioned in separate comment, need a summary writing of the api routes this adds and what conventions are being used for usernames/clientsids etc.

Also need to get written down what we discussed previously about how the topic space is being split up.

hardillb:
I've updated the first comment a bit, I will make sure the right comments for the API doc are added.

While thinking about it, I will probably rename most of the crud endpoints to `/client` rather than `/user

RE topic structure comment:

The proposal is to change the topic mountpoint from `team/[teamHashId]/` to something like `ff/v1/[teamHashId]/p/` to allow for interoperability with the project nodes.

I've looked at the topic space and I think there is a topic overlap with the project nodes and commands sent to Instance bound devices.

e.g. `ff/v1/[teamHashId]/p/[instanceId]/command` and `ff/v1/[teamHashId]/p/[instanceId]/in/...`

Since the broker only does ACL checks at subscribe and publish time, not at delivery to a client, there would be no way to explude the `/command` topic.

hardillb:
Discussed with Nick.

New topic mountpoint will be `ff/v1/[teamHashId]/c/`.  This matches the existing structure, but keeps the project nodes outside for now.

hardillb:
NOTE: check the db migration before merging so it is the newest

forge/comms/v2AuthRoutes.js
```diff
@@ -0,0 +1,144 @@
...
+            const isValid = await app.db.controllers.BrokerClient.authenticateCredentials(
+                username,
+                password
+            )
+            if (isValid) {
+                reply.send({
+                    result: 'allow',
+                    is_superuser: false,
+                    client_attrs: {
+                        team: 'ff/v1/internal/c/'
```
hardillb, line 45:
```suggestion
                        team: ''
```
This should be empty to allow for interaction between project nodes and team broker clients

knolleary: Agreed - I don't think we should be setting the prefix for the existing clients as they already fully specify their topic space with `ff/v1/:teamId/...` - no need to push that under an even deeper hierarchy.

forge/db/migrations/20241010-01-EE-add-teambroker-client.js
knolleary, line 1: Migration filename contention with https://github.com/FlowFuse/flowfuse/pull/4659

forge/ee/routes/teamBroker/index.js
```diff
@@ -0,0 +1,299 @@
...
+        }
+    })
+
+    /**
+     * Update a specific MQTT Client
+     * @name /api/v1/teams/:teamId/broker/client/:username
+     * @static
+     * @memberof forge.routes.api.team.broker
+     */
+    app.patch('/client/:username', {
```
knolleary, line 184:
```suggestion
    app.put('/client/:username', {
```

Put not Patch for updates like this

forge/ee/routes/teamBroker/index.js
```diff
@@ -0,0 +1,299 @@
...
+    })
+
+    /**
+     * Delete a MQTT Clients
+     * @name /api/v1/teams/:teamId/broker/client/:username
+     * @static
+     * @memberof forge.routes.api.team.broker
+     */
+    app.delete('/client/:username', {
+        preHandler: app.needsPermission('project:create'),
```
knolleary, line 267: We should add some specific permissions rather than overload existing ones.

knolleary (commented):
A couple comments, still working through it.

forge/db/migrations/20241010-01-EE-add-teambroker-client.js
hardillb: Renamed this migration, so merge #4659 first

knolleary (approved):

Issue #4554, opened 2024-09-24
Admin UI: Add feature to distribute platform-wide notifications


joepavitt:
### Description

As an Admin user of FlowFuse
I want to be able to send out a notification to all users on the platform
So that I can share important updates about the platform, e.g new features or planned maintenance

#### First Iteration

- Role limitations, e.g. team owners, members, dashboard only - the latter isn't likely to need to see a lot of platform update notifications.

### Which customers would this be available to

None

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
>  are notifications unavailable for self-hosted platforms?

As admins, we wouldn't be able to send a notification to a self-hosted install of FF. However, if we injected notifications into the database as a migration for our major releases, then we could notify them of new features

PR #4637, opened 2024-10-11
Add the ability to emit platform wide notifications

cstns:
## Description

Added an admin API that can send platform wide notifications to users that fall into the provided roles.
Added a User model query that retrieves users based on provided roles.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4554

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/pages/admin/NotificationsHub.vue
```diff
@@ -0,0 +1,149 @@
...
+                        <FormRow v-model="form.message" class="mb-5" data-el="notification-message">
+                            Announcement Text
+                            <template #description>Provide the details of your announcement.</template>
+                            <template #input><textarea v-model="form.message" class="w-full max-h-80 min-h-40" rows="4" /></template>
+                        </FormRow>
+                        <FormRow v-model="form.url" type="input" :placeholder="urlPlaceholder" class="mb-5" data-el="notification-external-url">
+                            URL Link
+                            <template #description>Provide an url where users will be redirected when they click on the notification.</template>
+                        </FormRow>
+                        <ff-checkbox v-model="form.externalUrl" data-el="notification-external-url-toggle">
```
knolleary, line 23: Not sure how this is meant to be used for non-external links. It requires quite a deep knowledge of the UI - something self-hosted users won't have. Is there a specific use-case that requires this capability?

knolleary (commented):
Overall, looks good.

One pending question about the 'external url' option - more the ability to provide non-external urls. Not sure how self-hosted admins would know what to do with that.

If there isn't a specific need or example where it will be needed, can I suggest we hide for now to avoid confusion?

frontend/src/pages/admin/NotificationsHub.vue
cstns: It was mainly for us to know how to properly format the internal url's. I agree there's no use on having it for self-hosted customers. Would leaving the external url as default and hiding the placeholder for the internal routes suffice?

knolleary: The default should be external URL, and I don't think we should have the option for it to be anything else. If we wanted to add support for routing internally, then we can sniff the value to decide if its a url or an object - no need to store that otherwise.

cstns: I removed the external url selector, forcing all urls to be external

cstns:
The only task left is inserting data into the notifications table. I’ve added a basic implementation that loops through and inserts notifications into the database based on the roles query.

However, this approach could become problematic if there are many users or results, as it might cause a table lock on the notifications table and lead to other issues.

Should we stick with this for now?

knolleary:
> However, this approach could become problematic if there are many users or results, as it might cause a table lock on the notifications table and lead to other issues.

> Should we stick with this for now?

I had a quick look at this part this morning. Had a couple thoughts, but need to poke at it a bit more once other unrelated tasks are complete. Will review tomorrow.

knolleary (approved):
Going to ship as-is. There are improvements to be made on the scaling of adding notifications to large proportions of users in the database; but for the low-scale usage of the feature we can ship now and improve later.

knolleary:
Decided to hold back; want to validate the scaling issue first. Also, want to add the ability to target team tier as well as role.

Will pick up and complete in next few days.

knolleary:
Have pushed a small refactoring:

1. replaces `recipientRoles` on the API with `filter`. This is an object that can contain different filter options. For this iteration the only property is `roles` - the equivalent of `recipientRoles`. This gives us room to move in the future with other filter options to target different selections of users
2. Added a `bulkSend` api for this use case - which uses `bulkCreate` to create notifications in batches of 200. This is all done under a single transaction which helps reduce load on the DB.

As I write this comment, realise I haven't dbl checked the UI tests for the api change I made... will see how that goes before marking this good to merge.

cstns:
Should we get another set of eyes on the back-end changes? I'm not sure I can/should approve the changes

knolleary:
@cstns I'll give it a fresh look over in the morning, but then we can move forward.

knolleary:
Not working on postgres due to implicit casting between boolean/integer types in the query.

Need to fix and add some tests.

knolleary:
Postgres fixed. Also, for the mock call, changed to use `count` rather than `find` as there's no need to return the actual data.

Finally, added some unit tests for the api side.

cstns:
Should I merge the PR after I give it another quick check?

knolleary (approved):

Issue #4538, opened 2024-09-23
Set env vars for device groups


robmarcer:
### Description

Allow env-vars to be set for a device group. Any new devices added to that group would take on those variables.

If a grouped device's evn-var is replaced for that specific device then that new value should be retained.

### Which customers would this be available to

None

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

robmarcer:
This has been requested by the following two enterprise customers

https://app-eu1.hubspot.com/contacts/26586079/record/0-2/8845845707

&

https://app-eu1.hubspot.com/contacts/26586079/record/0-2/7039008216

joepavitt:
Thanks @robmarcer - we have a product planning meeting on Wed, so will schedule it then

Steve-Mcl:
For consideration:

This is a good example of why the concept of Tags were raised (and developed, but that work stalled - possibly a very outdated branch knocking around).  e.g. A Tag containing "things" (in this case env vars) could be applied to device(s), instance(s), application(s) and device group(s). As more and more of this kind of requirement arises (like "we need to apply a specific theme to group 1" or "we want the editor path set to 'live' for our 'live' instances, etc), tags begin to make more sense.

joepavitt:
@Steve-Mcl we're not going to be pursuing tags at this stage. The scope would be too considerably, and I don't want to introduce even _more_ concepts into FF when users already find the existing number of concepts overwhelming.

knolleary:
Want to keep this simple.

 - Add a `settings` property to DeviceGroup - type Text, storing JSON blob
 - Define env vars under `{ "env": ... }` style property - to match Instance settings format
 - Update API endpoint to support updating of env vars

Important to keep api and data format consistent with Instance env settings.

joepavitt:
I can handover WYSIWYG this afternoon, but you're likely to need to juggle both as they're not small tasks

Steve-Mcl:
> * Update API endpoint to support updating of env vars


### inheritance rules
Any thoughts on the rules for when a device belonging to a group has its own env vars?

1. merge  - group env vars have precedence
2. merge - device env vars have precedence
3. replace device env vars with group env vars
  * and if the group has zero env vars, send none to device (regardless of device settings)?
  * or if group has zero env vars, send existing device vars?

[...]

PR #4659, opened 2024-10-18
Add Device Group env vars API support

### LINKED ISSUE #4658: Add Database column and API for updating device group settings env

Steve-Mcl:
closes #4658

### Route added
* `PUT` `/api/v1/applications/:applicationId/device-groups/:groupId/settings`
    * Expects body to contain an `env` array of objects

NOTE: Adding endpoint vs updating the `PUT /api/v1/applications/:applicationId/device-groups/:groupId` route
was a toss of a coin.  Instances overload the `PUT /api/v1/projects/:id` route allowing for the
settings to be updated in the same request.  However, devices have a separate route `PUT /api/v1/devices/:deviceId` route.
Since devices and device groups are closer in relation, I decided to follow suit. I did however structure this is a way that we can easily remove this endpoint and map the feature to the existing route if so desired

### Specified behaviour
* Only an owner role user can update the group settings
  * The RBAC `application:device-group:update` permission was re-used for this purpose
* The `env` array of objects is merged with the existing member devices only in the `live` API (when a device is requesting settings)
* Updating a device group model `settings` and saving will trigger all member devices to update their settings hash
* When the hash is computed, group env vars are merged with the device env vars before computing the hash (they are not persisted in the device model)

### Tests added
```
  Audit Log > Application
    ✔ Provides a logger for updating device group settings
```

```
  Application Device Groups API
    Update Device Settings
      ✔ Owner can update a device group settings (414ms)
      ✔ Merges new Group Env Var (351ms)
      ✔ Only updates a device settings hash if the group env var change affects the merged settings env (327ms)
      ✔ Member can not update a device group settings (403) (95ms)
      ✔ Non Member can not update a device group settings (404) (95ms)
```

## Related Issue(s)

Owner: #4658 
Parent: #4538

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/ee/db/controllers/DeviceGroup.js
```diff
@@ -214,6 +243,58 @@ module.exports = {
         await app.db.models.Device.update({ DeviceGroupId: null }, { where: { id: deviceIds.removeList, DeviceGroupId: deviceGroup.id }, transaction })
     },
 
+    // updateSettings: async function (app, deviceGroup, settings, user) {
```
knolleary, line 246: Unused code? Something for the future?

knolleary (approved):
Looks good - one pending question about a commented out function. Approved pending resolution of the question.

forge/ee/db/controllers/DeviceGroup.js
Steve-Mcl: Ah, that slipped through. Originally, I wrote this as a separate function but the tests revealed I need to restructure.

Will remove.

Steve-Mcl: resolved in [Remove unused code](https://github.com/FlowFuse/flowfuse/pull/4659/commits/f2aad783993211f47a918dd82d14a3951d8f181d)

Issue #4702, opened 2024-10-28
Mqtt broker copy client  username not working on Firefox


cstns:
### Current Behavior

The copy function throws an error
![image](https://github.com/user-attachments/assets/22ddcfc8-e815-4eac-8f80-96bc99f46f82)
![image](https://github.com/user-attachments/assets/446a5d6c-cbbc-4d4e-95eb-e1ab5915d838)

### Expected Behavior

The copy functionality should work on all browsers. We should use the existing copy component already used on the assets page

### Steps To Reproduce

copy a broker's client name while using firefox

### Environment

- FlowFuse version: 2.10.0
- Node.js version: N/A
- npm version: N/A
- Platform/OS: N/A
- Browser: N/A

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Issue #4685, opened 2024-10-23
Follow up after MQTT Broker UI


cstns:
### Description

After a hastily development of the mqtt broker page, dx and scalability is lacking alongside other things that need addressing. 

Todo
- Extract standalone logic into distinct components
- improve css styling/rules
- check if a new icon for the mqtt broker is needed for the broker's empty state but not due to lack of license/team-feature
- address snapshot lint selector warnings
  - test/e2e/frontend/cypress/tests/instances/version-history/snapshots.spec.js:56
  - test/e2e/frontend/cypress/tests/instances/version-history/snapshots.spec.js:59
  - test/e2e/frontend/cypress/tests/instances/version-history/snapshots.spec.js:445
  - test/e2e/frontend/cypress/tests/instances/version-history/snapshots.spec.js:731
- add more and improve existing ee e2e tests

### Epic/Story

https://github.com/FlowFuse/flowfuse/issues/1350

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

hardillb:
Add copy button next to username/clientid in list of clients to make easier to use

PR #4711, opened 2024-10-29
Follow up mqtt broker ux

cstns:
## Description

general code cleanup, e2e improvements

 - [x] Extract standalone logic into distinct components
 - [x] improve css styling/rules
 - [x] added new icon for broker empty state
 - [x] addressed snapshot.spec missing cypress selectors
 - [x] add more and improve existing ee e2e tests
 - [x]  enhanced form validation when editing/creating/altering broker client and acls

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4685
closes https://github.com/FlowFuse/flowfuse/issues/4702

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/ui-components/stylesheets/ff-utility.scss
```diff
@@ -21,4 +21,10 @@ $ff-funit-xs: 0.75rem;
...
 $ff-funit-lg: 1.5rem;
-$ff-funit-xl: 2rem;
\ No newline at end of file
+$ff-funit-xl: 2rem;
+
+@mixin ellipsis {
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
```
Steve-Mcl, line 30: We already have the class `truncate` that is exactly the same as this new `ellipsis`

image

Steve-Mcl (commented):
Can we use Tertiary buttons for `+` and `-` & check W/H are equal & H + V alignment please?

image

frontend/src/ui-components/stylesheets/ff-utility.scss
cstns: after looking into it a bit more after our conversation, Tailwind CSS does not use SCSS or mixins so i don't have the possibility to use the .truncate class as a mixin like i'm using it in this situation. In order to preserve annotation with tailwind, I renamed the mixin to follow suite tailwind's nomenclature.

Another solution would be to simply remove the mixin altogether and use the css rules as is, but I found that i'm using these classes quite extensively lately and would appreciate a shorthand for css imports where i'm unable to use tailwind's helper classes

Steve-Mcl:
@cstns 

We seem to have lost the copy confirmation.

Before and after
chrome_2M8iUiLTxd

Steve-Mcl (approved):
Nice :)

PR #4717, opened 2024-10-30
Add TeamBroker enabled to pre-staging

hardillb:
Will need to pick the right helm chart version (2.17.0) to make this work.

Also will need follow up to work out if we start teamBroker for pre-staging and how it would work with load balancer.


ci/ci-values.yaml
```diff
@@ -4,6 +4,8 @@ forge:
   localPostgresql: true
   broker:
     enabled: true
+    teamBroker:
+      enabled: true
```
hardillb, line 8:
```suggestion
      uiOnly: true
```

hardillb:
This can be merged before the helm chart https://github.com/FlowFuse/helm/pull/484

ppawlowski (approved):

Issue #4655, opened 2024-10-16
MQTT Client billing model


knolleary:

hardillb:
I ask the following question just to get it out of my head.

What happens if a user stop paying for extra clients?

Or do they need to reduce down to included number before they can?

knolleary:
Having done the license enablement, we can now support self-hosted customers who want to purchase additional clients.

The work to enable FFC self-service purchase of additional clients is on hold for the moment.

PR #4732, opened 2024-11-04
Add mqttClients to license claims

knolleary:
Part of #4655

## Description

As part of enabling billing for MQTT Client usage, this first step adds them as a licensable claim. This will allow us to provide self-hosted customers with a license for more than the default allocation (Teams=5, Enterprise=20).

This has required a little bit of moving code around to enable. The commits in this PR do the following:

1. Move the DB models/views/controllers to the core set, out of the EE code base. This ensures the db model for MQTT clients is initialised before the licensing code tries to check the current usage level.
2. Adds `mqttClients` as a claim in the license
3. Adds checks on `mqttClient` usage to the overage checks
4. Updates the tests for MQTT Clients

The test updates verify the right limits are applied based on license state. Whilst there, I also updated the existing tests to run using a Team Owner who isn't an Admin (`bob`, not `alice`). This is ensures the permissions checks aren't being bypassed by the admin flag.


hardillb:
Pushed fixes for tests, will review the rest when the tests pass

forge/db/models/TeamBrokerClient.js
```diff
@@ -20,6 +20,26 @@ module.exports = {
...
+                }
+            },
+            afterCreate: async (project, opts) => {
+                const { mqttClients } = await app.license.usage('mqttClients')
+                if (mqttClients.count > mqttClients.limit) {
+                    await app.auditLog.Platform.platform.license.overage('system', null, mqttClients)
+                }
+            }
+        }
+    },
```
hardillb, line 42: Do we need to generate a new license for prod (and staging) before this gets merged?

forge/licensing/license-generator.js
```diff
@@ -61,6 +61,7 @@ const { v4: uuidv4 } = require('uuid')
         const maxUsers = parseInt(await promptly.prompt('Max allowed users: ', { default: '5' }))
         const maxTeams = parseInt(await promptly.prompt('Max allowed teams: ', { default: '5' }))
         const maxInstances = parseInt(await promptly.prompt('Max allowed instances (hosted + devices): ', { default: '5' }))
+        const maxMQTTClients = parseInt(await promptly.prompt('Max allowed MQTT Clients: ', { default: '20' }))
```
hardillb, line 64: That's 20 total for the platform, not just a give team, just making sure this is what we mean (It's probably right, just double checking)

forge/licensing/loader.js
```diff
@@ -6,6 +6,11 @@ const jwt = require('jsonwebtoken')
 
 const LICENSE_ISSUER = 'FlowForge Inc.'
 
+const MQTT_CLIENT_DEFAULT_ALLOWANCE = {
+    teams: 5,
```
hardillb, line 10: As don't have teams tier licensing anymore, is this just to cover legacy license?

hardillb (commented):
Looks good, just a couple of question

forge/db/models/TeamBrokerClient.js
knolleary: Technically no, as we are licensed so allowed to go over the limits at the platform level - but we'll get audit log overage reports. So in practice, we'll want to deploy a new license around the same time as this merging - just doesn't have to be strictly in order.

forge/licensing/license-generator.js
knolleary: Yes - default enterprise license includes 20 clients; the license terms match the Enterprise team limits.

forge/licensing/loader.js
knolleary: Yeah - given 'teams' has been a historically valid license type, just covering our bases.

hardillb (approved):

knolleary:
Have added mqttClients to `/ap1/v1/admin/stats` endpoint and include on the Admin Overview page that reports on usage.

Whilst there, also fixed reporting the number of Admin users on the platform. Noticed FFC was reporting 0 admin users via this API. This is because postgres returns booleans as true/false whereas sqlite returns as 1/0 - and the code was looking for `1` explicitly rather than general truthiness.

hardillb (approved):

hardillb:
Quick question before merging.

Should the license version have been bumped from `2024-03-04` as it adds a field?

Issue #4673, opened 2024-10-22
Sidebar: Re-Design


joepavitt:
### Current:

<img width="658" alt="Screenshot 2024-10-22 at 10 58 40" src="https://github.com/user-attachments/assets/a35b702f-5615-42d2-b20e-a0c398eba93a">

The left-side is okay, but doesn't allow for general exploration of newer features we've introduced. This has caused our home view main-content to be far too busy with over 30 actionable buttons in this example above.

We should consider re-designing the left-side to allow for general exploration of the newer concepts we've added (e.g. bill of materials, pipelines, broker clients, device groups)

This can also pair with https://github.com/FlowFuse/flowfuse/issues/3151

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
### First Pass

Ignoring the right-side, which will also have a clean-up, just focussing on the left-side navigation here:

<img width="657" alt="Screenshot 2024-10-22 at 11 01 16" src="https://github.com/user-attachments/assets/2553a9e6-46a5-4f71-ae14-05972cf3451a">

joepavitt:
Further updates done here: https://www.figma.com/design/rMB2w5ZZUtTC0nxbwLYKJY/FlowForge-Platform?node-id=2528-18633&t=xeWDxwylexhhGYL8-1

Variants:

![Image](https://github.com/user-attachments/assets/5daa863a-f2b8-4925-bf59-09db9692d54f)

joepavitt:
Design phase completed, moving into development, and will update issue description

gstout52:
Let's use "Edge Device" rather than "remote". Just talked this over with ZJ as well.

ZJvandeWeg:
fyi @sumitshinde-84

cstns:
You can use https://4745.flowfuse.dev/ if you need to play around

PR #4743, opened 2024-11-07
Sidebar Re-design: prerequisites

cstns:
extract the features mixin into the account store, extract the permis…sionsMixin into a composable, fix webpack devtool source mapping

## Description

- the features mixin is cumbersome to use and bloats each component it's added to and cannot be used in vuex stores
- the persmissions mixin is not usable in vue stores, in order to avoid duplication
- addresses webpack's devtool source mapping for a more coherent debug environment while preserving stacktraces for posthog

## Related Issue(s)

part of https://github.com/FlowFuse/flowfuse/issues/4673

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/composables/Permissions.js
+++ b/frontend/src/composables/Permissions.js
@@ -1,0 +1,90 @@
+// import { useStore } from 'vuex'
+
+import { Permissions } from '../../../forge/lib/permissions.js'
+import { Roles } from '../utils/roles.js'
+/**
+ * @typedef {0 | 5 | 10 | 30 | 50 | 99} Role
+ * Enum for roles with specific numeric values.
+ */
+
+export default function usePermissions () {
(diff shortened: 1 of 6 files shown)


frontend/src/composables/Permissions.js
joepavitt, line 1: why not part of the vuex store?

cstns: Some of the methods need to be called multiple times depending context, role or permission we're checking against on component level and vuex doesn't really have support for this type of feature. Because we can't use mixins in the vuex store and we need the same checks to be done, the portable way of accomplishing both is to extract them into composables.

We should slowly but surely try and distance ourselves from mixins wherever possible.

joepavitt: > We should slowly but surely try and distance ourselves from mixins wherever possible.

Agreed

Issue #4726, opened 2024-11-04
MQTT Topic Hierarchy


joepavitt:

joepavitt:
RE: Extraction methods, @hardillb has done preliminary investigations here, and found that we can utilise the EMQX API in order to return a topic hierarchy. This would be the first piece of our MQTT architecture that requires broker/brand-specific code, but the value here is important and we should proceed with that.

The alternative, non-brand specific option was to setup clients to subscribe to `#` for each team's broker, and record the hierarchy ourselves, but computationally, this would be excessive.

hardillb:
EMQX provides the following API endpoint

https://docs.emqx.com/en/emqx/v5.8/admin/api-docs.html#tag/Topics/paths/~1topics/get

This requires an admin API token to authenticate and replies with the following structure:

```
{
  "data": [
    {
      "node": "emqx@node1.emqx.com",
      "topic": "ff/v1/32E4NEO5pY/d/DeqNr3g6p5/command"
    },
    {
      "node": "emqx@node1.emqx.com",
      "topic": "ff/v1/32E4NEO5pY/a/rGymXb1wWp/command"
    },
    {
      "node": "emqx@node1.emqx.com",
      "topic": "ff/v1/32E4NEO5pY/d/K6jN3Z4ybv/command"
    },
    {
      "node": "emqx@node1.emqx.com",
      "topic": "ff/v1/32E4NEO5pY/a/NvBmA71zAo/command"
    },
    {
      "node": "emqx@node1.emqx.com",
      "topic": "ff/v1/32E4NEO5pY/p/152fb007-7eb4-4e41-ad3b-8f946d11108b/in/foo"
    },
    {
      "node": "emqx@node1.emqx.com",
      "topic": "ff/v1/+/d/+/logs/heartbeat"
    },
    {
      "node": "emqx@node1.emqx.com",
      "topic": "ff/v1/+/d/+/response/96ffc27e-b60d-42da-83f9-1eb42baf4ba1"
    },
    {
      "node": "emqx@node1.emqx.com",
      "topic": "$share/platform/ff/v1/+/d/+/status"
    },
    {
      "node": "emqx@node1.emqx.com",
      "topic": "$share/platform/ff/v1/+/l/+/status"
    },
    {
      "node": "emqx@node1.emqx.com",
      "topic": "ff/v1/32E4NEO5pY/p/+/out/foo"
    }
  ],
  "meta": {
    "count": 10,
    "limit": 100,
[...]

PR #4748, opened 2024-11-09
Track topics used by a Team on Team Broker

hardillb:
part of #4726 
## Description

Tracks all topics published to on the Team Broker.

Adds 1 new HTTP endpoint
- GET `/api/v1/teams/:teamId/broker/topics` returns an array of strings representing the topics the team has published to

Requires team Member or Owner to access list of topics

Needs:
- ~~testing~~
- ~~a new permission adding to access the endpoint~~

Currently caches topics for 1 hour, checks for removing topics every 30seconds, probably should be configurable.

Will benefit from using Redis for this so cache is preserved over restarts and if multiple instances of forge app running.

## Related Issue(s)

#4726 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


knolleary (commented):
Inspired thinking - I like it.

I suggest structuring topic list to have `teamId` has a top level key, rather than just a flat list of all topics.

```
{
   "ff/v1/team1" : {
      "foo": { ttl: 123 }
   },
   "ff/v1/team2" : {
      "bar": { ttl: 123 }
   }
}
```

That will make the `get` operation more scalable as it doesn't have to iterate over the entire cache.

hardillb:
I did it this way to make it easier to include the project nodes changes later (when they will publish without the mount point so will need to include the full topic prefix).

Might need to discuss think about it some more.

hardillb:
Not sure how to write a test for those last 5 lines as that would mean adding a 30 second pause to ensure the interval runs (but even then it won't remove any entries for 30mins)

forge/ee/lib/teamBroker/index.js
```diff
@@ -30,4 +30,50 @@ module.exports.init = function (app) {
...
+    const topicsList = {}
+
+    const cleanInterval = setInterval(() => {
+        const keys = Object.keys(topicsList)
+        const now = Date.now()
+        for (let i = 0; i < keys.length; i++) {
+            if (topicsList[keys[i]].ttl < now) {
+                delete topicsList[keys[i]]
+            }
+        }
```
knolleary, line 50: This will need updating given the cache structure has changed

joepavitt:
> Currently caches topics for 1 hour, checks for removing topics every 30seconds, probably should be configurable.

Is there an opportunity for a longer term solution here? May be valid that some data only gets published every 24 hours for example.

hardillb:
> Is there an opportunity for a longer term solution here? May be valid that some data only gets published every 24 hours for example.

These are just starting values, but having a longer caching period is likely to not be useful in it's current form as the cache will get cleared every time the forge app restarts (e.g. every time we do a CI deployment....)

There is a comment in the code about using Redis as a better cache, but this is a MUCH bigger solution

joepavitt:
Why is the URL `/api/v1/teams/:teamId/broker/topicList` and not `/api/v1/teams/:teamId/broker/topics`

The latter is more RESTful?

knolleary:
Think about longevity of the cache - and dealing with CI reloading the app.

Some options that come to mind:

1. Introduce redis (more likely valkey). This will also mean we aren't storing state in memory that won't horizontally scale. However it does add the overhead work on adding another component to the architecture. It is inevitable... the question is whether *this* is the feature that pushes us into action on it.
2. Store topic list in existing DB. Don't really want to add this load to the DB in realtime, but could be cached in memory and periodically written to DB in batches (including in the shutdown hook). Doesn't help with scaling (as there is state in memory), but will be persistent across restarts.

In terms of the existing caching TTL values, I'd suggest we default for longer intervals - assuming we (eventually) solve the persistence of state. I'd suggest 25hrs ttl and once a day purging - I don't think the level of usage for in-memory caching will be a problem in the short term for it to be quite generous.

hardillb:
Increasing the cache to 25hours is an easy change, but moving to clearing the cache to once a day, will mean moving that code to the house keeping tasks to get round the restarts.

Will look at that shortly

hardillb:
Stashed the topic cache in the database across restarts

hardillb:
Need to check that this DB cache actually will work on K8s as I think it starts the new instance, before telling the old one to shutdown.

This means the cache will load from the DB before it's populated. Might need to add a delay (and find away to force the `app.settings.get` to re-read from the database.

forge/lib/permissions.js
```diff
@@ -179,7 +179,8 @@ const Permissions = {
     'broker:clients:list': { description: 'List Team Broker clients', role: Roles.Member },
     'broker:clients:create': { description: 'Create Team Broker clients', role: Roles.Owner },
     'broker:clients:edit': { description: 'Edit Team Broker clients', role: Roles.Owner },
-    'broker:clients:delete': { description: 'Delete Team Broker clients', role: Roles.Owner }
+    'broker:clients:delete': { description: 'Delete Team Broker clients', role: Roles.Owner },
+    'broker:topics:list': { description: 'Delete Team Broker clients', role: Roles.Member }
```
knolleary, line 183:
```suggestion
    'broker:topics:list': { description: 'List active Team Broker topics', role: Roles.Member }
```

knolleary (approved):

joepavitt:
Is this good to merge?

Issue #4684, opened 2024-10-23
Hierarchical Audit Logging


joepavitt:
### Description

When viewing the Audit Log view at the Team/Application levels, I would expect to see any logs related to elements contained _within_ that thing, i.e. in the Application's Audit Log, I'd expect to see all events linked to Instances & Devices within that Application.

Currently though we only see events/logs associated to the Application itself, not it's children.

This FR covers adding the hierarchical audit logging being available at the Application & Team levels. Implementation should be split into the following tasks:

```[tasklist]
### Tasks
- [ ] https://github.com/FlowFuse/flowfuse/issues/4746
- [ ] https://github.com/FlowFuse/flowfuse/issues/4750
- [ ] https://github.com/FlowFuse/website/issues/2759
```

### Which customers would this be available to

None

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Steve-Mcl:
> i.e. in the Application's Audit Log, I'd expect to see all events linked to Instances & Devices within that Application.

I have the basic mechanics of this working as demonstrated below.

Questions:
1. Do we expect to see the "scope" as a column in the results or is it enough to be implicit at this stage?
2.  Should we default the scope to "team level only events" (as per current condition) and the user can then decide to select "All" or "Instance"
3. The title mentions "Hierarchical".  As development stands, the child results (Hierarchically speaking) are included when the "All" scope is selected. However, the view is NOT modified to show nested Hierarchical results.  To achieve nested views, the query would be much more complex.  
4. Again, title mentions "Hierarchical" so does that mean if the scope is set to "Application" that all application children (instances and devices) should be included.  Again, this makes the query much more complex,


![Image](https://github.com/user-attachments/assets/d712f24b-74e1-484a-8463-c5e9b116d917)

joepavitt:
> Do we expect to see the "scope" as a column in the results or is it enough to be implicit at this stage?

Implicit

> Should we default the scope to "team level only events" (as per current condition) and the user can then decide to select "All" or "Instance"

No, the default should be "All", as that is what I believe the user has as an expectation when viewing this screen, and the feedback we've received.

> The title mentions "Hierarchical". As development stands, the child results (Hierarchically speaking) are included when the "All" scope is selected. However, the view is NOT modified to show nested Hierarchical results. To achieve nested views, the query would be much more complex.

[...]

PR #4749, opened 2024-11-09
Show child items in team and application audit log

Steve-Mcl:
## Description

Updates the AuditLog model to support returning items scoped and including/excluding child items

### PR Direction Reasoning
There were two ways to approach this. 
1. update all Audit Table to include ancestors of the entity (aka normalisation). This would mean updating every write to the audit log to include parent/grandparent etc 
2. build a list of dependants upon query.  

Both solutions have pros.  For example, Option 1 would continue to return logs from children removed from a team but would not include logs from before a child was moved.  Option 2 will include all child log items so long as the item is still a child. 

This PR goes with option 2.

#### UPDATED 18/11/2024 from commit [1bbdb6a](https://github.com/FlowFuse/flowfuse/pull/4749/commits/1bbdb6a0fae08b0a9e0676356bca2c0bfb294d7d) onwards

3. Improves the `getFilterAndAssociations` method to remove the need to loop though child projects when grabbing its children. Now it uses a `where in` to grab the child owned instances devices when building the filter.
4. Changes the ORDER BY to `id` field
   1. This improves query by avoiding the temporary B-tree  
6. Adds migration `20241118-01-add-auditlogs_entity_index.js`
   > NOTE: this migration only adds the index if it doesn't already exist: This was done to
allow the platform admin to apply the index themselves upfront (if they want to).
For example, on a PG DB, the user may wish to pre-emptively create the index using the CONCURRENTLY
option to avoid locking the table for the duration of the index creation. e.g:
    CREATE INDEX CONCURRENTLY idx_auditlog_entity ON "AuditLogs" ("entityType", "entityId");
Time to build estimates for the index creation:
as the entityType is typically 10 bytes or less and the entityId is typically 36 bytes or less
the index key size is ~46 bytes. If the table has 1M rows, the index size will be ~46MB.
On a fast machine with SSD & reasonable memory, the index creation should take less than 1
    minute (typically 10-30 seconds)
If CONCURRENTLY is used, the table will not be locked during the index creation but will take longer
7. Grabs associated models when building the filter clause so that they can be passed to the frontend for displaying associations in the Audit View
8. Updates the frontend to show an icon and name for the associated application/project/device in a column on the audit log
9. Changes the **[x] Include Children** checkbox to a dynamic **[x] Include Applications, Instances and devices** (as required)

### Adds unit tests for new static model function `getFilterAndAssociations`

```
  Audit model
    static methods
      getFilterAndAssociations
        for Team
          ✔ should return clause scoped for team only
          ✔ should return filter scoped for applications only
          ✔ should return filter scoped for projects only
          ✔ should return filter scoped for devices only
          ✔ should return filter scoped for team and all children
          ✔ should return filter scoped for applications and all children
          ✔ should return filter scoped for projects and all children
        for Application
          ✔ should return filter scoped for application only
          ✔ should return filter scoped for projects only
          ✔ should return filter scoped for devices only
          ✔ should return filter scoped for application and all children
          ✔ should return filter scoped for projects and all children
```

### Screenshots

UPDATED 18/11/2024 - to match discussed additional work (see comment trail)

image

chrome_QO2HHV5u9q

## Related Issue(s)

Owner: #4746 
Parent: https://github.com/FlowFuse/flowfuse/issues/4684

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
@joepavitt as discussed - tagged here for any feedback UI wise

frontend/src/pages/team/AuditLog.vue
```diff
@@ -11,31 +11,67 @@
...
+                <FormHeading class="mt-4">Event Scope:</FormHeading>
+                <div data-el="filter-event-types">
+                    <ff-dropdown v-model="auditFilters.selectedEventScope" class="w-full">
+                        <ff-dropdown-option
+                            v-for="scope in scopeList" :key="scope.id"
+                            :label="scope.name" :value="scope.id"
+                        />
+                    </ff-dropdown>
+                    <ff-checkbox v-if="auditFilters.selectedEventScope!=='device'" v-model="auditFilters.includeChildren" class="mt-2" data-action="include-children-check">
+                        Include Children
```
knolleary, line 24:
```suggestion
                        Include Instance/Device events
```

knolleary (commented):
Slightly concerned on the scalability of this approach; having to query to get the list of all instances/devices to then find the appropriate logs.

Also this does not include devices assigned to instances within the application (as far as I can tell) so is an incomplete result. However to address that with the current database structure will make this even less scalable - requiring multiple requests every time.

Would like to discuss before approving this PR (for the backend side - not reviewed frontend).

I'm going to raise a separate issue on how we distinguish between instance-assigned and application-assigned devices.

Steve-Mcl:
> Also this does not include devices assigned to instances within the application (as far as I can tell) so is an incomplete result

It does: https://github.com/FlowFuse/flowfuse/pull/4749/files#diff-04d0438df93f19d0ceea33de81abf591b19f33965efbb287fcc655163b2e2567R227



> Slightly concerned on the scalability of this approach; having to query to get the list of all instances/devices to then find the appropriate logs.

Very fair comment. For discussion - the way this was structured was such that the final big query (on the largest table: audit log) does a where "where in" as opposed to joining (which would need to be raw query with left joins) to minimise the hit. The smaller pre queries for which "things" to include felt manageable. 

This approach felt better than the normalisation route which would require every insert does a parent/grandparent traversal to populate the ancestor fields. 

Some quick wins if we keep approach this might be to add indexes to the searched fields?

Happy to discuss better /alternative solution.

knolleary:
For reference - https://github.com/FlowFuse/flowfuse/issues/4769

Steve-Mcl:
Following discussion with @knolleary it has been requested to include (at best) device/instance/application in a column.  At min, the scope of the log entry.  Given we can see duplicate entries from separate "things" (i.e. deploying to multiple instances/devices) it is not clear where it comes from.

Also discussed: index's on the queried fields - esp audit log table. probably also the project/device/app tables where filtering occurs.

Steve-Mcl:
Having done some discovery and discussed internally, the following changes will be made:

1. Change the ORDER BY to use the `id` instead of the `createdAt`
   1. Since `id` is primary it will be more efficient then the non-indexed `createdAt`
   2. Examining query plans, this change removes the temporary B-tree creation
   3. In short, ordering by `id` + the `LIMIT 200` should shortcut the table SCAN meaning the filtering is only applied to the 200 rows 
1. Add a composite index to `entityId`, `entityType`
   1. While this has little bearing (following the change in item 1) where it is important is the pagination...
   2. Since we use `findAndCountAll`, 2 queries are made:
      1.   1) `SELECT COUNT(id) WHERE <filter>` (this gets total rows for pagination counter
      2.   2) `SELECT <fields> WHERE <filter> ORDER BY id DESC LIMIT 200` (this gets the page of data)
      3.   For item 1), the index is important and in checing the plan, i can see the query utelises the index
      4.   For item 2), the index is less important and in checking the plan, i can see the query simply uses a SCAN (but it will be shortcutting due to LIMIT and use of `id` col) so the filtering is secondary in this query.

Steve-Mcl:
Update in coming.  Preview showing the scoped item as a column in the log as discussed:

chrome_tWBiTlag8x

Steve-Mcl:
@knolleary all changes discussed now in the branch including the debated index - I have updated this PRs OP to list out the changes.

knolleary:
Given the potential impact of creating the index in production, I'm going to delay merging this until the morning.

knolleary (approved):

PR #4764, opened 2024-11-13
Add search api for apps/devices/instances

### LINKED ISSUE #4756: Create search api
Part of #4729 

The purpose of this task is to introduce a search api at the team level that can be used to find resources within the team.

For the first iteration, the API should search applications, instances and devices by name.

The API should be extensible so in the future it can be used to search for other resources in the team (pipelines, device groups etc).

Whilst the UX of the search bar will be scoped to the current team, I think the API should be scoped to everything the user has access to; however for first iteration we can make `team` a required parameter and only support searching within a single team. This means the search end point should be `/api/v1/search`

knolleary:
Closes #4756 

## Description

Adds `GET /api/v1/search` as a general search endpoint for a user's resources.

For this iteration, it searches Applications, Instances and Devices in a given team using their `name` property.

The endpoint accepts two query params:
 - `team` : the team id to search within. This is a required property for this iteration - but will allow us to enable account-wide search for a user in the future if so needed.
 - `query` : string search term.

I have reused existing model find functions by adding an optional query parameter.

The response object looks like this. Each object in the `results` array has an `object` property to identify the type of thing it is. It then contains properties for that thing largely consistent with existing 'summary' views. This should give us enough info to render the search results sensibly.

For instances, there are a couple more bits to look at adding around dashboard usage and current state - but this should be good enough to get started.

```
{
    "count": 3,
    "results": [
        {
            "object": "application",
            "id": "p8QoZboADe",
            "name": "Debugging Demo",
            "description": "",
            "links": {
                "self": "http://127.0.0.1:3000/api/v1/applications/p8QoZboADe"
            },
            "deviceCount": 0,
            "instanceCount": 2,
            "deviceGroupCount": 0,
            "snapshotCount": 18,
            "pipelineCount": 2
        },
        {
            "object": "instance",
            "id": "33b340ca-1400-4de1-884c-ed19792592d6",
            "name": "debugging-demo",
            "url": "http://debugging-demo.example.com",
            "createdAt": "2023-05-11T14:41:51.094Z",
            "updatedAt": "2024-11-11T15:58:48.873Z",
            "links": {
                "self": "http://127.0.0.1:3000/api/v1/projects/33b340ca-1400-4de1-884c-ed19792592d6"
            },
            "protected": {
                "enabled": false
            }
        },
        {
            "object": "device",
            "id": "mYLjNYqMDn",
            "ownerType": null,
            "name": "demo1",
            "type": "my new thing",
            "lastSeenAt": "2022-08-02T13:32:02.000Z",
            "lastSeenMs": 71977676685,
            "status": "unknown",
            "mode": "autonomous",
            "isDeploying": false,
            "links": {
                "self": "http://127.0.0.1:3000/api/v1/devices/mYLjNYqMDn"
            }
        }
    ]
}
```


cstns (approved):
Looking good, Thank you!

PR #4770, opened 2024-11-13
Improved Admin Team view

knolleary:
## Description

This adds filtering options to the Admin Team view to make it easier to filter and search the list.

1. Adds Sort and Filter options alongside the table
2. Adds Device count to the table (already had Members and Instance counts. Also added same to the User's Team table for consistency)

The table is now sorted by most recently created ('newest') by default, but can be switched to 'oldest'.

The filters cover:
 - Team Type
 - Billing state (if enabled)

I time boxed this all to a couple hours - otherwise would have made some further improvements to the presentation of the table.

Need to add some tests for the new filter/sort options on the backend api

image

### DIFF
--- a/forge/ee/db/models/Subscription.js
+++ b/forge/ee/db/models/Subscription.js
@@ -63,6 +63,7 @@ module.exports = {
     },
     associations: function (M) {
         this.belongsTo(M.Team)
+        M.Team.hasOne(this)
     },
     finders: function (M) {
         const self = this
--- a/forge/db/models/Team.js
+++ b/forge/db/models/Team.js
@@ -231,36 +231,63 @@ module.exports = {
                             where = { id: queryId }
                         }
                     }
+                    const order = [['id', 'ASC']]
+                    if (pagination.sort === 'createdAt-desc') {
+                        order[0][1] = 'DESC'
+                        // Also need to 'fix' the cursor pagination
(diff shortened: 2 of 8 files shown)


forge/ee/db/models/Subscription.js
```diff
@@ -63,6 +63,7 @@ module.exports = {
     },
     associations: function (M) {
         this.belongsTo(M.Team)
+        M.Team.hasOne(this)
```
knolleary, line 66: This is needed so sequelize knows about the reverse association from Team to Subscription - allowing us to 'include' the Subscription model when querying Teams. As we no long use the `sync()` function to create the tables, this has no impact on the table structure.

knolleary:
@cstns no rush on this - nice to have, but please don't interrupt your current work

cstns (approved):
Looking good and works as expected!

One small ui quirk, having selected billing states and afterwards clicking on the suspended filter doesn't clear other checked filters (but working as expected) but it can be ignored.

image

An e2e test that confirms the filters load correctly and the table refreshes on filter changes would be a nice addition. Unit Test coverage should also be addressed.

knolleary:
Merging as we want to get this onto FFC. The additional test coverage is a nice to have, but as an admin-only area, we can afford to merge with the coverage it has currently.

PR #4777, opened 2024-11-14
Fix multiple non-authenticated page layouts

cstns:
## Description

Addresses some ui issues caused by the removal of referencing the viewport height.

- forced the `html`, `body`, `#app` and `#ff-app` containers to to full height by using flex in order to preserve scroll functionality
- fixed the layouts for the following pages:
 - updateExpiredPassword
 - PasswordExpired
 - TermsAndConditions
 - UnverifiedEmail
 - Create (signup)
 - ForgotPassword
 - Login
- also fixed a minor snag in the account store where state.team.types were not loaded causing the menu to rendered empty for users with lesser permissions

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/auth/UpdateExpiredPassword.vue
+++ b/frontend/src/components/auth/UpdateExpiredPassword.vue
@@ -1,9 +1,36 @@
 <template>
     <form class="space-y-6" ref="password_form" @submit.prevent>
         <div>You must set a new password before continuing</div>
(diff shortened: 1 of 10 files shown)


knolleary:
Why are we adding the same CSS multiple times rather than just fixing the common layout once?

cstns:
The ff-layout-box component is used in many places, ex modals. 
I added a utility class to avoid repetition

knolleary (approved):

Issue #4752, opened 2024-11-11
Build UI for displaying Broker Topic Hierarchy


joepavitt:
### Phase 1:

![image](https://github.com/user-attachments/assets/5921467f-6c6e-4018-b628-34c2b25cabe7)

API available in https://github.com/FlowFuse/flowfuse/pull/4748

cstns:
@joepavitt, I'd also like some help with the helptext content

cstns:
> @Yndira-E can you give me a hand with the empty state and helptext pictograms?
I'll re-use the existing broker pictograms fro the time being and replace them later on if someone says so

> @joepavitt, I'd also like some help with the helptext content
I've generated some placehoder text using an llm, it can be adjusted as we go

PR #4790, opened 2024-11-18
Broker Topic Hierarchy UI

cstns:
## Description

- add the mqtt broker page under the unified-namespace namespace
- add the new topic hierarchy page under the same namespace
- addressed the permissions composable reactivity problem which bit me sooner than expected

Was not able to re-use the existing accordion component because it was toggling the visibility of it's children which caused problems when rendering high numbers of topics. Ended up creating simpler a TopicSegment component which improved DOM rendering due to it not inserting the node in the dom only when acted upon.

Was able to render 1500 topic entries with 100 nested topics with ease.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4752

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
Does this want to target the other PR branch first?

cstns:
yes, my bad! Thank you for pointing it out!

I'll need a couple of hours to solve the e2e tests though

hardillb:
Totally first impressions, just thinking out loud

- I think the entry on the left should still be "Broker" not "UNS"
- Need to decide what order the Clients, Hierarchy tabs need to be in
- Need to think about the topic count on each row, showing 1 for `test/` feels wrong when `bill/` has 2

image

Otherwise looks good

hardillb:
Just asking to see how hard it would be, to remember expanded levels (only for while the page is currently open).

e.g. currently if I fully expand a topic tree, but then collapse the first element, when I re-expand that element again all the children have been collapsed, might be useful to be able to keep state for the lower elements

frontend/src/pages/team/UNS/index.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+    <ff-page>
+        <template #header>
+            <ff-page-header title="Unified Namespace" :tabs="tabs">
+                <template #context>
+                    A list of all MQTT Brokers available to your team and configurable within FlowFuse.
```
hardillb, line 6:
```suggestion
                    A list of the recent Topicss used by your team on the FlowFuse MQTT Broker.
```

cstns:
> Just asking to see how hard it would be, to remember expanded levels (only for while the page is currently open).
> 
> e.g. currently if I fully expand a topic tree, but then collapse the first element, when I re-expand that element again all the children have been collapsed, might be useful to be able to keep state for the lower elements

It's going to be tricky but doable. From the top of my head, I'd need to restructure the parsed API response. 

I would postpone it in a follow up task if that's ok.

knolleary:
I agree with keeping `Broker` for now. I think `UNS` as a label is too niche for general users at this stage.

hardillb:
Yep, Please change the menu on the left back to Broker (and the route paths please)

cstns:
Will do! Any changes to the tab order?

I'm a bit squeamish about changing the topics count right now, would better do it when tackling what topics have been collapsed or not

hardillb:
The counts and the order we can iterate on afterwards

cstns:
We are also restricting the use of the mqtt broker for self-hosted installations, are we keeping it in place this release?

joepavitt:
@cstns you have a merge conflict

frontend/src/pages/team/UNS/index.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+    <ff-page>
+        <template #header>
+            <ff-page-header title="Broker" :tabs="tabs">
+                <template #context>
+                    A list of the recent Topicss used by your team on the FlowFuse MQTT Broker.
```
knolleary, line 6:
```suggestion
                    View the recently used topics and configure clients for your FlowFuse MQTT Broker.
```

frontend/src/pages/team/UNS/index.vue
```diff
@@ -0,0 +1,54 @@
...
+                </template>
+                <template #pictogram>
+                    <img alt="info" src="../../../images/pictograms/mqtt_broker_red.png">
+                </template>
+                <template #helptext>
+                    <p>The <b>Broker</b> feature provides a centralized framework for managing and visualizing your entire data ecosystem, consolidating MQTT broker instances and topic structures within a single interface.</p>
+                    <p>Within the Broker, <b>Client</b> page enables seamless client management with customizable Access Control List (ACL) rules for secure and controlled data flow. The topic <b>Hierarchy</b> offers a clear, organized visualization of topic structures, providing fine-grained control over publishing and subscribing permissions.</p>
+                    <p>Together, these components deliver an integrated approach to managing client connections, security settings, and message flow, supporting efficient and secure data communication across your system.</p>
+                    <p>Documentation <a href="https://flowfuse.com/docs/user/teambroker" target="_blank">here</a></p>
```
knolleary, line 15:
```suggestion
                    <p>The <b>Broker</b> feature provides a centralized framework for managing and visualizing your entire data ecosystem, consolidating MQTT broker instances and topic structures within a single interface.</p>
                    <p>Within the Broker, <b>Client</b> page allows you to manage the clients that have access to the Broker, with customizable Access Control List (ACL) rules for secure and controlled data flow.</p>
                    <p>The topic <b>Hierarchy</b> offers a clear, organized visualization of the topics being used within your broker.</p>
                    <p>Together, these components deliver an integrated approach to managing client connections, security settings, and message flow, supporting efficient and secure data communication across your system.</p>
                    <p>Documentation <a href="https://flowfuse.com/docs/user/teambroker" target="_blank">here</a></p>
```

frontend/src/pages/team/routes.js
```diff
@@ -107,12 +108,31 @@ export default [
...
-                        }
+                            title: 'Team - MQTT Broker'
+                        },
+                        children: [
+                            {
+                                name: 'team-namespace-hierarchy',
+                                path: 'hierarchy',
+                                component: UNSHierarchy,
+                                meta: {
+                                    title: 'Team - MQTT Broker Hierarchy'
```
knolleary, line 124:
```suggestion
                                    title: 'Team - MQTT Broker Topic Hierarchy'
```

knolleary (changes requested):
Tidies up some language in the UI

In terms of usability of the Topic Hierarchy - the key missing piece is it's actually quite hard to then get the full topic being used in a way I can copy/paste elsewhere.

Here I have published to the topic `p1/line/factory/foo/p2/foo` - and whilst I can see that in the hierarchy, if I want to use that value elsewhere, I have to still manually type it out.
A quick win would be to add a 'copy topic' button on the 'leaf' nodes of the tree that copies the fully formed topic string to my clipboard.

image

cstns:
> In terms of usability of the Topic Hierarchy - the key missing piece is it's actually quite hard to then get the full topic being used in a way I can copy/paste elsewhere.

I wouldn't call it a missing piece, I don't remember it being mentioned before and it wasn't part of the requirements. It would be a nice addition but not in the format that I'm parsing the API response at the moment.

It can be easily implemented when tackling the topics count and remembering which topics were opened features @hardillb requested

knolleary:
> I wouldn't call it a missing piece, I don't remember it being mentioned before and it wasn't part of the requirements.

To clarify - I acknowledge this was not in the original design, so I'm not say you missed anything that had been previously suggested.

I will raise a separate item for it as we don't need to block this for release - however I strongly feel we should iterate on it quickly.

knolleary (approved):

PR #4800, opened 2024-11-20
Respect CTRL/CMD+Click for opening instance editor in new tab

Steve-Mcl:
## Description

Opens the instance editor in a new tab when CTRL/CMD is held
This is not as straight forward as setting target since the URL for immersive is different to the URL for standalone (thought that is something that could potentially be handled with more effort)
For now, this is a simple mod that adds to the current working logic. i.e.  currently, we do already support middle mouse button click, this PR builds on that to check the key held and opens in a new tab if it is. 

Very much a nice to have but worth the 20 mins

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/instance/components/EditorLink.vue
+++ b/frontend/src/pages/instance/components/EditorLink.vue
@@ -89,7 +89,15 @@ export default {
             if (this.disabled) {
                 return false
             }
-            window.open(this.url, !this.isImmersiveEditor ? '_blank' : '_self')
+            let href = this.url
+            let target = !this.isImmersiveEditor ? '_blank' : '_self'
+            // On Mac Keyboard, ⌘ + click opens in new tab (⌘ is `metaKey`)
+            // Otherwise Ctrl + click opens in new tab (Ctrl is `ctrlKey`)
+            if (evt.ctrlKey || evt.metaKey) {
+                target = '_blank'
+                href = this.editorURL
+            }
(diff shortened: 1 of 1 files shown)


frontend/src/pages/instance/components/EditorLink.vue
```diff
@@ -89,7 +89,15 @@ export default {
             if (this.disabled) {
                 return false
             }
-            window.open(this.url, !this.isImmersiveEditor ? '_blank' : '_self')
+            let href = this.url
```
cstns, line 92: This is a viable solution but for only one specific use case.

My take on this problem would be in the `frontend/src/ui-components/components/Button.vue` component, to completely remove the button and used an anchor stylized as a button or switch the type='anchor' where needed on the said component. That way we'd let the default html events take their course and we wouldn't need to intervene for niche cases

Steve-Mcl: Yes, but the issue is the URL is different depending on accessing the editor direct or immersive. As it stands, we have no option but to have a click handler for the descision stuff and a href for the middle click.

cstns: my bad, my solution was focused on other places where this is happening!

cstns (approved):

Issue #4380, opened 2024-08-14
Send reminders to team inviters and invitees if invitee doesn't join


gstout52:
### Description

**As a user who sends or receives a team join invitation**

**(if invitee) I want to be reminded if I forget to join a team I was invited to, or (if inviter) be reminded if someone I invited has not yet joined**

**So that (if invitee) I participate in FlowFuse when my team needs me, or (if inviter) I ensure that the right team members are in FlowFuse**

**Current Behavior**
Currently, after a user sends an invitation to join, there is no follow-up of any kind to the inviter.
Currently, after a user receives a team join invitation and does not accept it, there is no reminder message sent.

**Desired Behavior**

If two days after sending an invitation the invitee has not joined the team, then:
- The invitee receives a reminder email
- The inviter receives an email stating that the invitee has not yet accepted

### Which customers would this be available to

None

### Acceptance Criteria

- [ ] If two days after sending an invitation the invitee has not joined the team, then the invitee receives a reminder email
- [ ] If two days after sending an invitation the invitee has not joined the team, then the inviter receives an email stating that the invitee has not yet accepted

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
This one is a little trickier, and is likely to need some HS intervention, as not sure we can reliably track this within the source/product itself

joepavitt:
We do have "housekeeper" tasks that run intermittently, so is actually doable.

knolleary:
I suspect we have a lot of 'dormant' invites - we may want to tidy them up before we start sending reminders.

hardillb:
316 invites in the DB at the moment, but large number look to be expired.

Will add a housekeeping task to remove expired invites as part of the work

knolleary:
> If I use a fixed time everyday to run it, I should be able to check for creation dates, more than 2 days ago and less than 3.

Sounds good to me.

joepavitt:
@hardillb this can be closed, right?

PR #4824, opened 2024-11-22
Send invite Reminders

hardillb:
part of #4380

## Description

Sends reminder emails to invitees if not actioned in 2 days.

Also includes a house keeping job to remove expired invites from the database.

~~Still need to also send email to inviter to say not accepted yet.~~

~~Also needs tests, but will need to work out how to inject invites into the database with a custom `createdAt` entry in the database.~~

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


knolleary:
Looks okay from a first pass through. Want to sanity check the right emails are sent to the right parties - I note the tests just verify the right number of emails are sent, not that the right content was sent to the right recipient.

We also need to consider the scaling issue. Is the easy option just to add a 'reminderSent' column to the table?

hardillb:
We could add a column to the Invitations table, but we'd have to have it run multiple times a day to reduce the chance of restarts meaning that we never sent the reminders.

I'll add some more to the tests to check the email contents

hardillb:
OK, I have a version that uses a db column to record when the invite was sent.

I'm running the job with the following cron job

```
    // runs every 15mins with a random offset from the start of the hour
    schedule: `${randomInt(0, 15)}/15 * * * *`,
```

This will run at 15 min intervals, with a random offset between 0 & 15mins. This should mean that different instances run at different times, but that it not skip a day if the app is restarted. It may be a bit too many times, but it should only send reminders once, so should be a null op most of the time.

forge/housekeeper/tasks/inviteReminder.js
```diff
@@ -0,0 +1,73 @@
+const { Op } = require('sequelize')
+
+const { randomInt } = require('../utils')
+
+module.exports = {
+    name: 'inviteReminder',
+    startup: false,
+    // runs every 15mins with a random offset from the start of the hour
+    schedule: `${randomInt(0, 15)}/15 * * * *`,
```
knolleary, line 9:
```suggestion
    // runs daily at a randomly picked time
    schedule: `${randomInt(0, 59)} 6 * * *`,
```

Not sure why this needs to run so frequently? Once a day feels sufficient. Or am I missing a scenario that requires this to run more often?

hardillb: OK, so pinning it to a given hour means that the window to not run updates is a restart inside that hour.

Making it 6am is probably early enough that we are unlikely to do a code push/restart in that hour.

knolleary:
Testing locally, I see the emails get sent, but I consistently get this error in the logs:
```
[2024-12-05T11:39:00.066Z] TRACE: Completed task 'inviteReminder'
[2024-12-05T11:39:00.503Z] WARN: Failed to send email: Error: Message failed: 450 SQLITE_ERROR: cannot start a transaction within a transaction
```

The Invitation's `reminderSentAt` column has been updated, so not sure what DB interaction is triggering this.

knolleary:
`WARN: Failed to send email: Error: Message failed: 450 SQLITE_ERROR: cannot start a transaction within a transaction`

I think this is a complete red herring. The error is the response being returned by my local test SMTP server. From our logs, two mails were sent. Looking at the inbox, only one arrived.

knolleary:
Yup - upstream issue with my choice of SMTP tools - https://github.com/nodemailer/nodemailer-app/issues/3

knolleary (approved):

Issue #4821, opened 2024-11-22
Accessing "more" devices in the Application view can lead into a 404 that cannot be escaped by "go back"


Steve-Mcl:
### Current Behavior

![chrome_etdE7wOPTM](https://github.com/user-attachments/assets/7b9122b1-789a-41f4-9417-a7e75a17a9ee)

### Expected Behavior

1. Go Back should take user back to "/applications"
2. The "show more" link should go to all devices page?

### Steps To Reproduce

Visit https://app.flowfuse.com/team/acme-man-corp/applications
click "More" under the devices

### Environment

- FlowFuse version: 2.11 (FFC)
- Node.js version:
- npm version:
- Platform/OS: Win 11
- Browser: Chrome latest

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Issue #4687, opened 2024-10-24
Viewer team members are redirected to a 404 page when accessing more devices from the applications page


cstns:
### Current Behavior

Viewer team members are redirected to a 404 page when accessing more devices from the applicaitons page

### Expected Behavior

Viewer team members should be able to view/access the application devices page

### Steps To Reproduce

Log in as a user with viewer permissions and click on the view more devices button from the applications page

### Environment

- FlowFuse version: 2.9.0
- Node.js version: N/A
- npm version: N/A
- Platform/OS: N/A
- Browser: N/A

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #4826, opened 2024-11-25
Alter the destination of the applications more link for viewer role users

cstns:
## Description

- viewer roles do not have access to team device-groups, which returned a 403 when viewer roles attempted to open the applications instances/devices page resulting in a 404 page
- also wrapped the hasAMinimumTeamRoleOf to verbosely check for member roles

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4821
closes https://github.com/FlowFuse/flowfuse/issues/4687

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/composables/Permissions.js
+++ b/frontend/src/composables/Permissions.js
@@ -56,13 +56,13 @@ export default function usePermissions () {
     }
 
     /**
-         * Check if the user has the minimum required role.
-         * @param {Role} role - The role to check against.
-         * @returns {boolean} True if the user has the minimum required role, otherwise false.
-         * @example
-         * // Check if the user has at least the 'Member' role
-         * const isMemberOrHigher = hasAMinimumTeamRoleOf(Roles.Member)
-         */
+     * Check if the user has the minimum required role.
+     * @param {Role} role - The role to check against.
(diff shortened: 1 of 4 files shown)


knolleary:
I'm not sure this is how we want to address this. As it stands, a Viewer role is unable to view device-groups. But that should not mean they cannot view the application - because they need to be able to browse the list of instances/devices available to them.

The 403 in retrieving groups should be handled (even better, don't try to load device groups) - similarly for any other resources that view tries to load that the Viewer role doesn't have access to.

knolleary (changes requested):
I'm not sure this is how we want to address this. As it stands, a Viewer role is unable to view device-groups. But that should not mean they cannot view the application - because they need to be able to browse the list of instances/devices available to them.

The 403 in retrieving groups should be handled (even better, don't try to load device groups) - similarly for any other resources that view tries to load that the Viewer role doesn't have access to.

cstns:
Going with an alternate fix implemented in https://github.com/FlowFuse/flowfuse/pull/4846

PR #4834, opened 2024-11-27
Use default behavior for platform wide anchors (part I)

cstns:
## Description

First iteration to correct platform wide link interaction to allow mouse+keyboard combinations.

- Fixes inconsistencies when opening the immersive editor from  different places using  key + mouse click
  - for example the applications instances open editor button was opening the regular editor when ctrl+clicking even though the immersive one was available
- added middle mouse click support
- restores default anchor behavior to links on the page header (except team selector), applications overview page, application instances and application devices pages
- links can be opened via cmd/ctrl + click or middle mouse button click
- had to alter the datatable components to trickle-up the click event in order to be able to decide in we open in a new tab or current one
- in order to identify middle mouse button clicks I had to replace the @click event handler to @mouseup which includes it
- simplified the open editor method while fixing aforementioned bugs
- used the vue router where possible to preserve navigation integrity
- added cmd/ctrl+click and middle mouse button interactivity to the open dashboard

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4837
closes https://github.com/FlowFuse/flowfuse/issues/4838

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/instance/components/DashboardLink.vue
+++ b/frontend/src/pages/instance/components/DashboardLink.vue
@@ -7,6 +7,8 @@
         :target="target"
         :disabled="buttonDisabled"
         class="whitespace-nowrap"
+        @click.stop.prevent
+        @mouseup.stop.prevent
     >
         <template v-if="showText" #icon-left>
             <ChartPieIcon />
--- a/frontend/src/pages/instance/components/DashboardLink.vue
+++ b/frontend/src/pages/instance/components/DashboardLink.vue
@@ -1,7 +1,6 @@
 <template>
     <ff-button
         v-if="!hidden"
-        type="anchor"
         kind="secondary"
         data-action="open-dashboard"
         :to="dashboardURL"
--- a/frontend/src/pages/instance/components/EditorLink.vue
+++ b/frontend/src/pages/instance/components/EditorLink.vue
@@ -1,15 +1,15 @@
 <template>
-    <div :data-type="`${isImmersiveEditor ? 'immersive' : 'standard'}-editor`" @click.stop="openEditor()">
+    <div :data-type="`${isImmersiveEditor ? 'immersive' : 'standard'}-editor`" @mouseup.stop.prevent="openEditor">
         <slot name="default">
             <ff-button
                 v-ff-tooltip:left="(editorDisabled || disabled) ? disabledReason : undefined"
-                type="anchor"
                 :to="editorURL"
                 kind="secondary"
                 data-action="open-editor"
                 :disabled="buttonDisabled"
                 class="whitespace-nowrap"
-                @click.stop="openEditor"
+                :emit-instead-of-navigate="true"
+                @click.stop.prevent="openEditor"
             >
                 <template v-if="showText" #icon-left>
                     <ProjectIcon />
@@ -32,6 +32,7 @@ import SemVer from 'semver'
 import { mapState } from 'vuex'
 
 import ProjectIcon from '../../../components/icons/Projects.js'
+import { useNavigationHelper } from '../../../composables/NavigationHelper.js'
 
 export default {
     name: 'InstanceEditorLink',
@@ -57,6 +58,13 @@ export default {
         showText: {
             default: true,
             type: Boolean
+        }
+    },
+    setup () {
+        const { openInANewTab } = useNavigationHelper()
+
+        return {
+            openInANewTab
         }
     },
     computed: {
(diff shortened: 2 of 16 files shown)


frontend/src/ui-components/components/Button.vue
```diff
@@ -10,8 +18,15 @@
...
-    </a>
-    <button v-else ref="input" class="ff-btn transition-fade--color" :type="type" :class="'ff-btn--' + kind + (hasIcon ? ' ff-btn-icon' : '') + (size === 'small' ? ' ff-btn-small' : '') + (size === 'full-width' ? ' ff-btn-fwidth' : '')" :disabled="disabled===true?'true':null" @click="go()">
+    </router-link>
+
+    <button v-else ref="input"
+            class="ff-btn transition-fade--color"
+            :type="type"
+            :class="computedClass"
+            :disabled="htmlDisabled"
+            @mouseup="go"
```
joepavitt, line 28: Why `mouseup` and not `click`?

frontend/src/ui-components/components/data-table/DataTableRow.vue
```diff
@@ -4,7 +4,13 @@
             <slot name="row-prepend" :row="data" />
         </ff-data-table-cell>
         <slot>
-            <ff-data-table-cell v-for="(col, $column) in columns" :key="col.label" :class="col.class" :style="col.style" :highlight="highlightCell === $column" @click="$emit('selected', data)">
+            <ff-data-table-cell v-for="(col, $column) in columns"
+                                :key="col.label"
+                                :class="col.class"
+                                :style="col.style"
+                                :highlight="highlightCell === $column"
+                                @mouseup="handleMouseUp"
```
joepavitt, line 12: Why `mouseup`?

joepavitt (commented):
### Mixins

I will comment in saying I really don't like our move towards a `mixin` approach as it makes the traceability of our in-component variables incredibly difficult and the code general much harder to navigate.

However, I will not block this work for that reason, but just ask, in the future, that we consider for developer-friendly approaches in terms of readibility.

### Sidebar Rendering

Also with the open-in-new-tab functionality introduced here, it does also stress how poorly reactive our sidebar is and it's not rendering whenever I open most of the links here.

### General Comments

I've also made a few other comments within the changes directly, which have questions I'd like answered

frontend/src/ui-components/components/Button.vue
cstns: vue's '@click' handler doesn't get triggered on mouse 3 (middle mouse button) clicks, this is the way to do it

frontend/src/pages/instance/components/DashboardLink.vue
```diff
@@ -1,13 +1,14 @@
...
         v-if="!hidden"
-        type="anchor"
         kind="secondary"
         data-action="open-dashboard"
         :to="dashboardURL"
         :target="target"
         :disabled="buttonDisabled"
         class="whitespace-nowrap"
+        @click.stop.prevent
+        @mouseup.stop.prevent
```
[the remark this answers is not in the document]
cstns: to prevent the click event from bubbling up/down. In the case of the dashboard link being nested inside another anchor/link/button, this prevents opening both parent anchor and dashboard link when clicking (or any other click + keyboard combination) directly on the dashboard link element

frontend/src/pages/instance/components/EditorLink.vue
```diff
@@ -1,15 +1,14 @@
...
                 v-ff-tooltip:left="(editorDisabled || disabled) ? disabledReason : undefined"
-                type="anchor"
-                :to="editorURL"
                 kind="secondary"
                 data-action="open-editor"
                 :disabled="buttonDisabled"
                 class="whitespace-nowrap"
-                @click.stop="openEditor"
+                :emit-instead-of-navigate="true"
+                @select="openEditor"
```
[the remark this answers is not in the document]
cstns: another way of preventing click events from bubbling up/down. This is specific to the editor link button, having to open different urls based on the type of click/combination acted upon. Emitting an event instead of propagating the click event when clicked gets rid of a lot of headaches and changes when dealing with event propagation in parent components.

I have to admit, 'select' is not the best choice of name for the event being emitted when clicked

cstns:
> I will comment in saying I really don't like our move towards a `mixin` approach as it makes the traceability of our in-component variables incredibly difficult and the code general much harder to navigate.
> 
> However, I will not block this work for that reason, but just ask, in the future, that we consider for developer-friendly approaches in terms of readibility.

Slowly but surely I'm moving away from the mixin approach, but it will take some time in doing so. The long term solution is to extract all generic stateless functionality into composables and delegate the rest into appropriate stores/services

> Also with the open-in-new-tab functionality introduced here, it does also stress how poorly reactive our sidebar is and it's not rendering whenever I open most of the links here.

This is very unfortunate side effect which is made even more obvious by this feature and has a greater impact than the sidebar itself. Because we don't have a safeguard in place and render views before we have all mandatory data for the ui to function properly, there are scenarios where the team selector isn't present and depending on the tab you're opening, even elements on the page misbehave.

I am aware of them and will address them in the upcoming period.

joepavitt (approved):

PR #4854, opened 2024-12-02
Expand the UNS Hierarchy by default & improve hover behaviour

joepavitt:
## Description

It's annoying to have to expand many layers each time, and given that it's a single single to minimise a whole branch of the hierarchy, expanded by default makes more sense.

The default "closed" view also implies there is nothing on the screen, especially if the first nested `topic` is `/`

Screenshot 2024-12-02 at 21 10 07

Also improves the hover state, with a `pointer:cursor` and highlighted text:

Screenshot 2024-12-02 at 21 11 35

### DIFF
--- a/frontend/src/pages/team/UNS/Hierarchy/components/TopicSegment.vue
+++ b/frontend/src/pages/team/UNS/Hierarchy/components/TopicSegment.vue
@@ -78,6 +78,11 @@ export default {
             return !this.isRoot && this.hasSiblings && this.isLastSibling
         }
     },
+    mounted () {
+        if (this.hasChildren) {
+            this.visibleChildren = true
+        }
+    },
     methods: {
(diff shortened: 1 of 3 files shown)


cstns:
The fact that the ff-accordion component was rendering the html elements was the main reason i ended up not using it for rendering the nested hierarchy. For small payloads with <50 topics it behaves nicely but when you exceed that, the initial page render and DOM draw over cumbers the page and might freeze it. 

By programmatically hiding the elements and only drawing the starting segments I managed to load 1500 entries with >50 nested topics.

I'd go with a variant of opening only the first n topics up to an m level

cstns (approved):
accepting changes due to the current low number of broker users and topics.

A cohesive implementation will be done in https://github.com/FlowFuse/flowfuse/issues/4857

PR #4865, opened 2024-12-03
Fix padding on Device Group Settings view

### LINKED ISSUE #4850: Device Group Settings Content missing left margin
### Current Behavior

Currently the settings tab at device groups is missing the margin left. Currently it has been set to `class="flex flex-col sm:flex-row"` but it needs to be `class="flex flex-col sm:flex-row ml-6"`.

### Expected Behavior

Add margin to the `class="flex flex-col sm:flex-row"`.

### Steps To Reproduce

Open the device group settings tab.

### Environment

- FlowFuse version: Current on the website
- Node.js version: 
- npm version:
- Platform/OS: FlowFuse
- Browser: Brave Version 1.73.91 Chromium: 131.0.6778.85 (Official Build) (64-bit)

knolleary:
Closes #4850 

Adds some missing padding on the Device Group settings view.


cstns (commented):
there's a negative margin-left inside the sectionSideMenu, shouldn't we remove that?

knolleary:
@cstns I was mostly applying the suggested change having compared the layout with `frontend/src/pages/instance/Settings/index.vue` which has `ml-6` applied as well.

But on wider inspection, I see that's the only other view that applies that margin. So yes, a more consistent fix is likely in order here. Will have a quick look to see how the different views compare.

knolleary:
@cstns have pushed an update. I've checked all the pages using SectionSideMenu and looked at how they were applying the left margin/padding. Have tidied up for consistency and checked the appearance looks good on all affected pages at both full size and mobile size.

cstns (approved):

Issue #4797, opened 2024-11-19
Sidebar sometimes fails to update when switching to admin/user views


knolleary:
### Current Behavior

*Sometimes* when navigating to the admin or user views via the dropdown menu, the sidebar does not update and still shows the menu items for the Team I was in. The only way to resync is to reload the page

### Expected Behavior

Sidebar to update

### Steps To Reproduce

It does not happen everytime. I haven't hit it enough times to spot a pattern - but I have hit it enough times to know I'm not imagining it.

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

cstns:
Tried for the last half hour to reproduce it and couldn't. I may have encountered it but not 100% sure if it was that or just me messing with the app for too long. 

I'll put a pin in it for the time being and keep an eye out to see if I'm able to to trigger it and find a reproducible pattern

cstns:
Opening again because it reproduced it on staging

PR #4869, opened 2024-12-04
Fix main nav matching context order

cstns:
## Description

- reversing the order so we find the nearest top-level parent with a meta.menu attribute not the other way around

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4797

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/drawers/navigation/MainNav.vue
+++ b/frontend/src/components/drawers/navigation/MainNav.vue
@@ -48,8 +48,14 @@ export default {
                 return this.$route.meta.menu
             }
 
+            // this.$route.matched includes all nested parent routes in the order they match, starting from the root parent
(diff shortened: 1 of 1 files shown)


cstns:
Following your example, the admin sidebar could be set on route 2 and we were matching finding a match on route 1. Given that $route.matched is a flat array (wouldn't include route 3, sibling, and route 4 child of sibling) we were returning the first context set on the top-most parent

joepavitt (approved):
I'm not sure I see why this fix would solve the underlying issue. I'm unable to reproduce the issue in the pre-staging environment, so can trust it has been fixed as that was a generally consistent place to see the issue.

Issue #4836, opened 2024-11-27
Restore platform wide link interaction


cstns:
### Description

General initiative to improve ux with the platform by allowing users to interact with all links in a predictable manner with mouse + key combinations 

### Which customers would this be available to

Everyone - CE/Starter/Team/Enterprise

### Acceptance Criteria

- [ ] Links should be able to be interacted with via click + keyboard combo (ctrl/cmd+click)
- [ ] Links should be able to be interacted with via middle button click

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate
```[tasklist]
### Tasks
- [ ] https://github.com/FlowFuse/flowfuse/issues/4837
- [ ] https://github.com/FlowFuse/flowfuse/issues/4838
- [ ] https://github.com/FlowFuse/flowfuse/pull/4885
- [ ] https://github.com/FlowFuse/flowfuse/pull/4891
- [ ] https://github.com/FlowFuse/flowfuse/pull/4942
- [ ] https://github.com/FlowFuse/flowfuse/pull/4905
- [ ] Restore Link interactions in the PageHeader - Team Selector & Right menu dropdown
- [ ] Restore Link interactions on the remainder tabs of the applications page
- [ ] Restore Link interactions on the team instances page
- [ ] Restore Link interactions on the team devices page
- [ ] Restore Link interactions on the team broker page
- [ ] Restore Link interactions on the team library page
- [ ] Restore Link interactions on the team members [age
- [ ] Restore Link interactions on the team audit log page
- [ ] Restore Link interactions on the team setting page
- [ ] Restore Link interactions on the team billing page
- [ ] Restore Link interactions on the account settings section
- [ ] Restore Link interactions on the admin section
```

cstns:
IMO this story should not have been assigned a milestone to begin with as it's a chore task to improve UX, not being mission critical it can be acted upon when we have time to spare

joepavitt:
This has been achieved through the detailed PRs. Any issues found moving forward can have new issues raised

PR #4885, opened 2024-12-06
Fix application child routes not making the applications nav menu active

cstns:
## Description

- use the default vue router behavior of marking routes as active by nesting application routes under the team namespace
- also fixes the reactivity/state problems we've been facing when opening pages in new tabs 
  - the application routes are now nested under the team component which pre-loads and checks for the team state
  - no need to add a parallel application service which would hydrate the store
  - navigation from and to pages which are already within the team namespace do not trigger reloads for teamMember & team store watchers because they now remain constant thus relieving a good number of api calls from being made 

One drawback of this change is that the routes for the team application will now change and any saved bookmarks will become unusable.

## Related Issue(s)

part of https://github.com/FlowFuse/flowfuse/issues/4836

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/pages/application/DeviceGroup/Settings/Environment.vue
```diff
@@ -19,8 +19,6 @@
                 <ExclamationCircleIcon class="ff-icon mr-2" />
                 <span class="relative top-0.5">Note: Updating environment variables can cause devices in the group to be restarted.</span>
             </span>
-
-            <ChevronRightIcon class="ff-icon align-self-right" />
```
cstns, line 23: not related to the current scope, but was an easy fix to address as it was a leftover (had no reason to be render)

frontend/src/pages/application/DeviceGroup/Settings/index.vue
```diff
@@ -25,6 +25,7 @@ export default {
         SectionSideMenu, SectionTopMenu
     },
     mixins: [permissionsMixin],
+    inheritAttrs: false,
```
cstns, line 28: not related to the current scope, but prevents development warnings due to the component inheriting unspecified props from it's parent route. Passing props to vue routes is an antipattern

frontend/src/pages/application/DeviceGroup/index.vue
```diff
@@ -72,6 +72,7 @@ import DeviceSolidIcon from '../../../components/icons/DeviceSolid.js'
 
 import permissionsMixin from '../../../mixins/Permissions.js'
 
+// todo page breadcrumbs should be pointing to app > device groups, currently app is missing and layout seems to be out of place
 export default {
```
cstns, line 76: will address in a quick follow-up for this task

frontend/src/pages/application/Pipelines.vue
```diff
@@ -153,16 +153,13 @@ export default {
...
-        teamMembership () {
-            if (!this.hasPermission('application:pipeline:list')) {
-                return this.$router.push({ name: 'Application', params: this.$route.params })
-            }
-
-            // Forces to load pipelines when teamMembership changes. When loading the page via url, teamMembership might not be
-            // loaded by the time the mounted loadPipelines is called and the hasPermission method will return false.
-            // todo This should be addressed by implementing an application service that bootstrap's the
-            //  app and hydrates vuex stores before attempting to render any data
-            this.loadPipelines()
```
cstns, line 165: we don't need to load pipelines when teamMembership changes like we used to because the initial loadPipelines call does get called because we're able to check for features

frontend/src/pages/application/Pipelines.vue
```diff
@@ -153,16 +153,13 @@ export default {
...
+        teamMembership: {
+            handler: function () {
+                if (!this.hasPermission('application:pipeline:list')) {
+                    return this.$router.push({ name: 'Application', params: this.$route.params })
+                }
+            },
+            immediate: true
```
cstns, line 162: initially we were waiting for the team membership to change in order to be able to check for permissions, now we can trigger the watcher when the component mounts. 

Whilst writing this, we i assume we can safely move this into the mounted method

frontend/src/pages/application/routes.js
```diff
@@ -27,10 +27,8 @@ import ApplicationIndex from './index.vue'
 
 export default [
     {
-        path: '/application/:id',
-        redirect: to => {
-            return `/application/${to.params.id}/instances`
-        },
+        path: ':id',
```
cstns, line 30: because we moved these routes under the team namespace we don't need to define it as a root route (leading /) and leaving application in would have resulted in a route with the following form `/team/<team-slug>/applications/application/<application-id>/device-groups` which didn't make sense

frontend/src/pages/application/routes.js
```diff
@@ -152,32 +149,32 @@ export default [
...
-            return `/application/${to.params.applicationId}/device-group/${to.params.deviceGroupId}/devices`
-        },
+        redirect: { name: 'ApplicationDeviceGroupDevices' },
         children: [
             {
                 path: 'devices',
                 name: 'ApplicationDeviceGroupDevices',
                 component: ApplicationDeviceGroupDevices,
                 meta: {
+                    // todo add a back button to the application device-groups
```
cstns, line 177: will address in a quick follow-up for this task

frontend/src/pages/application/routes.js
```diff
@@ -186,6 +183,7 @@ export default [
                 name: 'ApplicationDeviceGroupSettings',
                 component: ApplicationDeviceGroupSettings,
                 meta: {
+                    // todo add back button
```
cstns, line 186: will address in a quick follow-up for this task

frontend/src/pages/instance/Editor/index.vue
```diff
@@ -55,6 +55,8 @@
             </ff-page>
         </section>
 
+        <ConfirmInstanceDeleteDialog ref="confirmInstanceDeleteDialog" :instance="instance" @confirm="onInstanceDelete" />
+
```
cstns, line 59: found a missing feature while working on this, we were unable to delete instances whilst using the immersive editor because the delete dialog was missing

test/e2e/frontend/cypress/tests/applications.spec.js
```diff
@@ -289,24 +289,34 @@ describe('FlowForge - Applications', () => {
...
-        // Check if the application has instances
-        cy.visit('/team/ateam/applications')
-
-        cy.get('[data-action="view-application"]').first().click()
-        cy.url().then(url => {
-            const applicationId = url.split('/')
-            cy.visit(`/application/${applicationId[4]}/settings`)
-            cy.get('[data-action="delete-application"]').should('be.disabled')
-        })
+        let team
```
cstns, line 292: re-worked an incredibly flaky test

frontend/src/pages/application/DeviceGroup/Settings/Environment.vue
cstns: rendered*

frontend/src/pages/application/Pipelines.vue
joepavitt: shouldn't this in fact go into a `beforeRouteUpdate`?

https://router.vuejs.org/guide/advanced/navigation-guards.html#Using-the-Options-API

cstns:
Totally agree, there's a lot of repetition but it felt 'safest' at the time. Looking into a more generalized approach

cstns:
Having tinkered with multiple ways trying to get around injecting the team slug across links, [extending the vue router-link ](https://router.vuejs.org/guide/advanced/extending-router-link.html)component with a custom one that auto-injects the slug would be the way to go. 

The problem arises from the fact that it's not the vue router that generates the urls but the router-link components themselves so a beforeEnter or any other route guards could mitigate the missing param, they don't fully solve it. Injecting the param globally through a route guard would end up creating malformed url (ex `/team/applications/<application-id>/instances` instead of `/team/<team-slug>/applications/<application-id>/instances`) which would work in most scenarios when reaching that endpoint the team slug would be injected accordingly if already logged in and navigating via click or copied link. Problems occur when sharing said link and opening them to someone who's not logged in or using a different team than the shared one as the instances/app id params might not match

joepavitt (approved):

PR #4886, opened 2024-12-06
Docs: Clearer DNS instructions and cross-linking

joepavitt:
## Description

We're losing an estimated 93% of traffic to the first step in our self-hosted install instructions for our "Quick Start" guide. This attempts to remedy the step/intimidation of setting up a DNS server in Step 1, which will hopefully improve conversion here.

### DIFF
--- a/docs/quick-start/README.md
+++ b/docs/quick-start/README.md
@@ -28,14 +28,17 @@ For a full installation guide, including how to setup FlowFuse in a production e
 
 ## Prerequisites
 
-Before you begin, ensure you have the following:
+Before you begin, ensure you have [Docker](https://docs.docker.com/engine/install/) and [Docker Compose](https://docs.docker.com/compose/install/) installed on your system (either as a standalone binary or as docker plugin)
 
-1. A domain name that you own and can configure DNS settings for (explained in step 1)
-2. Docker and Docker Compose installed on your system (either as a standalone binary or as docker plugin)
+## Step 1: Configure Domain
 
-## Step 1: Configure DNS
+### No DNS Server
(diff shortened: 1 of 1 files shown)


ppawlowski (approved):

ppawlowski:
Technically it looks fine. 
However, this increases the complexity of quick-start documentation and the whole setup process. Not quite convinced about this change.

You can merge it whenever you feel it's ready to be made publicly available.

hardillb (approved):

joepavitt:
> However, this increases the complexity of quick-start documentation and the whole setup process. Not quite convinced about this change.

How does it though? Currently, step 1 is setup a DNS server and get yourself a domain, which is a big step putting on 93% of visitors... the plan here is to provide the non-DNS route for quicker onboarding?

ppawlowski:
The current approach requires one to have a Fully Qualified Domain Name with the possibility to modify its DNS records. It does not require to setup a DNS server.
No-DNS approach requires one to use an external service and the way this service works might lead users to questions like: Which IP address should I put there? How can I check what is my IP?
Alternatively, if they cannot afford an FQDN, they can still use FlowFuse Cloud to evaluate the product. Maybe we should put a link to the FFC registration page?

I might be wrong, but I feel that we might not benefit much from this change. On the other hand, given the high failure rate, we probably won’t lose much either.

hardillb:
@joepavitt Can we publish this?

hardillb:
@joepavitt @ppawlowski do we know what this is waiting on?

Can it be merged and iterated on if needed?

PR #4899, opened 2024-12-09
Remove notifications for deleted instances

### LINKED ISSUE #4845: Instance related notifications not cleared when instance deleted
### Current Behavior

If an instance crashes or enters safe mode, we send a notification to team owners.

If the instance is deleted, those notifications are not cleared; so the user still sees them and can click on them - which then 404s as the instance has been deleted.

### Expected Behavior

When an instance is deleted, clear any related notifications.

NB: due to the design of notifications, we can't lean on a delete cascade at the DB level; we need a post-delete hook in the Instance model to do the work.

As we will have notifications in this state already, we will need a migration to tidy up the table.

### Steps To Reproduce

- Crash an instance to generate a notification in the notification inbox
- Delete the instance
- Check the notification inbox - the notification should be removed

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

hardillb:
fixes #4845

## Description

Adds afterDestroy hook to Projects table remove Notifications when Instances are deleted.

Also includes a migration to remove orphaned Notifications.

## Related Issue(s)

 #4845

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/migrations/20241209-01-remove-dead-instance-notifications.js
+++ b/forge/db/migrations/20241209-01-remove-dead-instance-notifications.js
@@ -1,0 +1,42 @@
+/**
+ * Remove Notifications for deleted Instances
+ */
+
+const deletedProjects = []
+const foundProjects = []
+
+module.exports = {
+    /**
+     * @param {QueryInterface} queryInterface Sequelize.QueryInterface
+     */
+    up: async (context) => {
+        const deleteNotification = async function (id) {
+            const deleteQuery = `DELETE FROM \"Notifications" WHERE \"id\" = ${id};`
+            await context.sequelize.query(deleteQuery)
+        }
+
+        const lookup = "SELECT \"id\", \"reference\" from \"Notifications\" " +
+            "WHERE \"type\" = 'instance-crashed' or \"type\" = 'instance-safe-mode'"
+        const [notifications] = await context.sequelize.query(lookup)
+        for (const notification of notifications) {
+            const projectId = notification.reference.split(':')[1]
+            if (deletedProjects.includes(projectId)) {
+                // already found project to be gone, delete
+                deleteNotification(notification.id)
+            } else {
+                if (!foundProjects.includes(projectId)) {
+                    const projectQuery = `SELECT count(*) from \"Projects\" WHERE \"id\" = '${projectId}';`
+                    const [projectExits] = await context.sequelize.query(projectQuery)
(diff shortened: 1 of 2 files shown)


hardillb:
We should probably do Devices as well, but want the approach verifying first

knolleary:
Can we look at doing a bulk delete sql rather than lots of individual ones?

https://sequelize.org/docs/v7/querying/delete/#deleting-in-bulk

hardillb:
I'll have another look, but we only have raw SQL access at this point (none of the models are loaded when the migrations run).

And the Notification table doesn't have a bare field with the project id that would could do a `WHERE "projectId" NOT IN (SELECT "id" from Projects)`

knolleary:
Good point; the reference field doesn't lend itself to using with `in`.

forge/db/migrations/20241209-01-remove-dead-instance-notifications.js
```diff
@@ -0,0 +1,42 @@
...
+        for (const notification of notifications) {
+            const projectId = notification.reference.split(':')[1]
+            if (deletedProjects.includes(projectId)) {
+                // already found project to be gone, delete
+                deleteNotification(notification.id)
+            } else {
+                if (!foundProjects.includes(projectId)) {
+                    const projectQuery = `SELECT count(*) from "Projects" WHERE "id" = '${projectId}' LIMIT 1;`
+                    const [projectExits] = await context.sequelize.query(projectQuery)
+                    if (projectExits[0] === 0) {
```
knolleary, line 30: Running this against sqlite and postgres results in different shapes of results that don't work with the test that is then performed on them

**sqlite**
```
[ [ { 'count(*)': 1 } ], Statement {} ]
```

**postgres**
```
[ [ { count: '1' } ], Result { ... } ] ]
```


In neither case will the current code correctly match the actual result. Note the different field names and the values are different types (int/string).

```suggestion
                    const projectQuery = `SELECT count(*) as c from "Projects" WHERE "id" = '${projectId}' LIMIT 1;`
                    const [[projectCount]] = await context.sequelize.query(projectQuery)
                    const count = parseInt(projectCount.c)
                    if (count === 0) {
```

knolleary (changes requested):
Issue with the count query across both support db flavours.

knolleary (approved):

PR #4917, opened 2024-12-13
Add note to Instance Types setting default Stack

hardillb:
## Description

Add note about setting default stack

### DIFF
--- a/docs/admin/introduction.md
+++ b/docs/admin/introduction.md
@@ -129,7 +129,7 @@ Whenever a new Instance Type is created, it will need to be manually enabled for
 the individual [Team Types](#managing-team-types) before they will be available
 for teams to use.
 
-Instance Types also control which will be the default Stack if there are multiple
+Instance Types also control the default Stack (and in turn, the default Node-RED versions) that Instances will run with if there are multiple available.
 Stacks associated with an Instance Type.
 
 ### Managing Stacks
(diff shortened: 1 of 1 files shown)


docs/admin/introduction.md
```diff
@@ -129,6 +129,9 @@ Whenever a new Instance Type is created, it will need to be manually enabled for
 the individual [Team Types](#managing-team-types) before they will be available
 for teams to use.
 
+Instance Types also control which will be the default Stack if there are multiple
```
joepavitt, line 132:
```suggestion
Instance Types also control the default Stack (and in turn, the default Node-RED versions) that Instances will run with if there are multiple available.
```

Just trying to hit the terms that users are searching in our docs, "Stacks", "Instance Types" do not resonate to the new users, and they're getting a little lost. Making sure we say "Default Node-RED version" should catch it

Thanks for the PR

joepavitt (approved):

docs/admin/introduction.md
```diff
@@ -129,6 +129,9 @@ Whenever a new Instance Type is created, it will need to be manually enabled for
 the individual [Team Types](#managing-team-types) before they will be available
 for teams to use.
 
+Instance Types also control the default Stack (and in turn, the default Node-RED versions) that Instances will run with if there are multiple available.
+Stacks associated with an Instance Type.
```
knolleary, line 133:
```suggestion
Instance Types also control the default Stack (and in turn, the default Node-RED versions) that Instances will run with if there are multiple available Stacks associated with an Instance Type.
```

knolleary:
Spotted a dangling full stop that I took the liberty of removing.

PR #4923, opened 2024-12-16
Open Dashboard and Editor links in new tab by default

### LINKED ISSUE #4727: Re-open NR Editors in New Tab
### Description

We've had enough feedback from users RE; our change with the "Immersive Editor", that we should now action it. Currently, the Immersive Editor opens in the same tab as the FlowFuse Editor. 

Moving forward, any "Open Editor" buttons should open that Immersive Editor in a new tab.

### Epic/Story

_No response_

joepavitt:
## Description

- Modifies the underlying Dashboard and Editor Link components to open the respective links in a new tab
- Ensure we're utilising `target` to prevent duplication of opened tabs

## Related Issue(s)

Closes #4727

### DIFF
--- a/frontend/src/composables/NavigationHelper.js
+++ b/frontend/src/composables/NavigationHelper.js
@@ -3,9 +3,9 @@ import { useRouter } from 'vue-router'
 export function useNavigationHelper () {
     const _router = useRouter()
 
-    const openInANewTab = (href) => {
+    const openInANewTab = (href, target = '_blank') => {
         return new Promise(resolve => {
-            window.open(href, '_blank')
+            window.open(href, target)
             resolve()
         })
     }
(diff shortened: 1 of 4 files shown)


joepavitt:
Given the [test failure](https://github.com/FlowFuse/flowfuse/actions/runs/12352952634/job/34471053184?pr=4923#step:7:1086), I think the best course of action is to remove that test given that we can't test for a new tab URL in Cypress

@cstns thoughts?

cstns:
> Given the [test failure](https://github.com/FlowFuse/flowfuse/actions/runs/12352952634/job/34471053184?pr=4923#step:7:1086), I think the best course of action is to remove that test given that we can't test for a new tab URL in Cypress
> 
> @cstns thoughts?

Yes, That would be the only way to do it.

joepavitt:
It is listed as a[ permanent trade-off](https://docs.cypress.io/app/references/trade-offs#Permanent-trade-offs) for Cypress, i.e. something they'll never support

cstns:
managed to get a partial win, by stubbing the window open i was able to at least verify that the window opens with the correct url

cstns (approved):

PR #4937, opened 2024-12-18
Fix useStore composable losing reactivity when used in uxStore

cstns:
## Description

Main train of thought after reproducing the issue locally and noticing that the mainNavContexts was undefined which prevented the mainNav's nearestContextualMenu watcher to set the mainNavContext due to it's safeguard is that the useStore composable is losing store reactivity and causing the uxStore mainNavContexts getter to fail.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4797

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/composables/Permissions.js
+++ b/frontend/src/composables/Permissions.js
@@ -19,11 +19,15 @@ let teamMembership = { role: 0 }
  * @typedef {0 | 5 | 10 | 30 | 50 | 99} Role
  * Enum for roles with specific numeric values.
  */
-export default function usePermissions () {
-    const store = useStore()
+export default function usePermissions (store = null) {
+    if (store !== null) {
+        store = store.state
+    } else {
+        store = (useStore())?.state
+    }
 
-    if (store && store?.state?.account?.teamMembership) {
-        teamMembership = store?.state?.account?.teamMembership
+    if (store && store?.account?.teamMembership) {
+        teamMembership = store?.account?.teamMembership
     }
 
     /**
--- a/frontend/src/composables/Permissions.js
+++ b/frontend/src/composables/Permissions.js
@@ -20,14 +20,16 @@ let teamMembership = { role: 0 }
  * Enum for roles with specific numeric values.
  */
 export default function usePermissions (store = null) {
+    let state
+
     if (store !== null) {
-        store = store.state
+        state = store.state
     } else {
-        store = (useStore())?.state
+        state = (useStore())?.state
     }
 
-    if (store && store?.account?.teamMembership) {
-        teamMembership = store?.account?.teamMembership
+    if (state && state?.account?.teamMembership) {
+        teamMembership = state?.account?.teamMembership
     }
 
     /**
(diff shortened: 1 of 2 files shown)


frontend/src/composables/Permissions.js
```diff
@@ -19,11 +19,15 @@ let teamMembership = { role: 0 }
  * @typedef {0 | 5 | 10 | 30 | 50 | 99} Role
  * Enum for roles with specific numeric values.
  */
-export default function usePermissions () {
-    const store = useStore()
+export default function usePermissions (store = null) {
+    if (store !== null) {
+        store = store.state
```
joepavitt, line 24: just for readability and consistency, can we make this variable `state` please?

frontend/src/store/ux.js
```diff
@@ -37,7 +37,7 @@ const getters = {
         return state.tours.education
     },
     mainNavContexts: function (state, getters, rootState, rootGetters) {
```
joepavitt, line 39: should these be `store` and `rootStore`?

Given that the input to `usePermissions` is `store`, but you're passing a variable called `rootState`?

cstns: It would depend on the context where they are used, I used the naming vuex describes them in their [api](https://vuex.vuejs.org/api/#getters). It would be better if i'd change the argument name in the composable to rootState to match it

joepavitt:
Can confirm this hasn't fixed the issue for me. To reproduce consistently:

- Sign out
- Sign in as a Platform Admin
- Go straight to Admin Settings

joepavitt:
Debugging this myself, and found that in `MainNav.vue`, the `watch` for `nearestContextualMenu` is firing correctly, but `this.mainNavContexts` is `undefined`.

Investigating further...

cstns:
Very nice catch! Back to the drawing board but at least now i have something to work with

joepavitt:
If I hardcode:

```
mainNavContexts: function (state, getters, rootState, rootGetters) {
        return {}
}
```

Then the `mainNavContexts` is no longer undefined, so I'm guesisng Vue is catching an error in the `getter` and returning `undefined`, even though I've put breakpoints and console logs in that egtter, and none of them are firing 🤷

cstns:
That's what i though i was fixing, i assumed that the permissions mixin was failing but I was working blindfolded until now. Now that I can reproduce it consistently i can get to the bottom of it easier

joepavitt:
Found the issue:

### Navigating Direct for First Time:

Screenshot 2024-12-18 at 20 58 11

### On the Refresh (Working)

Screenshot 2024-12-18 at 20 58 14

The back button isn't getting loaded correctly into the `context`

joepavitt:
Boils down to a few circular dependencies going on, as well as race conditions between multiple watchers. This `vuex` architecture to manage this is incredibly messy

joepavitt:
Calling `this.setMainNavBackButton(menu)` in the `nearestContextMenu` handler does at least make the side menu render, but it does so _without_ a "Back" button.

I need to call that a night, but you have enough to go on in order to get this fixed in the morning hopefully.

joepavitt:
Fix pushed in https://github.com/FlowFuse/flowfuse/pull/4949 this can be closed

PR #4941, opened 2024-12-18
Allow trial team to be manually created

knolleary:
Part of https://github.com/FlowFuse/CloudProject/issues/600#issuecomment-2549085040

## Description

Currently when a user signs up on FFC, they have a Team tier team created in trial mode, with application + instance created.

This is enabled by having the 'Create a personal team for users when they register' option enabled in Admin Settings. Alongside that, the Team Type is configured with trial options.

To improve the onboarding experience with the new Freemium type, we want to offer the user a choice to team type when signing up - as per screenshot linked above.

This means we don't want to auto create the team as we do today when the user completes email verification.

The PR updates the Team Create api endpoint to include a `trial` flag. If set (and billing enabled etc), the code will now check if this user is allowed to create a new trial team, and if so, does the work to create the trial subscription, and auto create the application/instance.

The rules for whether a user is allowed to create a trial team are:

1. EE license and billing configured 
2. User has no existing teams
3. The chosen team type as trial mode set
4. The user is less than a week old. Essentially, only the first team a user creates should be eligible for trial mode. Rather than add db state over whether a user has ever created a team, this was a pragmatic first iteration. Yes a user could create/delete a trial team multiple times during that week, but that really doesn't give them any advantage.

Due to time pressures, getting this PR raised so we can keep things on track this week. Will work on unit tests in parallel.


forge/routes/api/team.js
```diff
@@ -1,3 +1,4 @@
+const crypto = require('crypto')
```
knolleary, line 1:
```suggestion
const crypto = require('crypto')

```

Steve-Mcl (approved):
Thanks for the talk through Nick

all good here. code review and local pull test, create with/without auto create instance.

PR #4943, opened 2024-12-18
Add new team selection frontend

joepavitt:
## Description

-- Adds new tile-based team type selection, with context/awareness of free trials being available, and relaying messaging about the need for credit cards, even though it is a free tier (something we aim to fix soon)

### Full Onboarding Demo for Free Tier

https://github.com/user-attachments/assets/2a755c79-63d3-4269-bce8-ad0a89ca6a12

## Related Issue(s)

https://github.com/FlowFuse/CloudProject/issues/600#issuecomment-2549085040


joepavitt:
line 183 on `create.vue` is where you'll be wanting to look for that

joepavitt:
> Remove the "auto create instance" option when creating first application

Have created the branch (`auto-instance-application`) for this, it's a WIP with a small win already with respect to moving the "limit" notice. The logic on how that banner is rendered is very odd though, and needs investigating. Ideally we drive whether the checkbox is clicked by default, based on the limits

Issue #4950, opened 2024-12-19
Changing teams while on the application page does not change team


cstns:
### Current Behavior

Changing the current team while on an application page does not change the team due to a race condition in the team component

### Expected Behavior

The team should change

### Steps To Reproduce

Access a team and try and change the team.

### Environment

- FlowFuse version: 2.12.10
- Node.js version: N/A
- npm version: N/A
- Platform/OS: N/A
- Browser: N/A

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #4951, opened 2024-12-19
Fix race condition when changing teams while on the application page

cstns:
Issue caused by setting team sequentially while on the application page. This did not pose a problem until nesting the application pages under the team namespace

## Description

- removed the team dispatch on beforeMount and beforeRouteEnter causing the issue
- update how we change the team, updating the team is more than enough while pushing a new route for browser history consistency
- a bit of code cleanup on the application page

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4950
introduced by https://github.com/FlowFuse/flowfuse/pull/4885

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/team/index.vue
+++ b/frontend/src/pages/team/index.vue
@@ -66,6 +66,11 @@ export default {
             return this.isAdminUser || this.teamMembership?.role >= Roles.Viewer
         }
     },
+    watch: {
+        '$route.params.team_slug' (slug) {
+            this.$store.dispatch('account/setTeam', slug)
+        }
+    },
     mounted () {
         this.mounted = true
     },
--- a/frontend/src/pages/team/index.vue
+++ b/frontend/src/pages/team/index.vue
@@ -19,7 +19,6 @@
 </template>
 
 <script>
-import { useRoute } from 'vue-router'
 import { mapGetters, mapState } from 'vuex'
 
 import Loading from '../../components/Loading.vue'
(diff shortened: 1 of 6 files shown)


frontend/src/pages/team/index.vue
```diff
@@ -68,21 +66,25 @@ export default {
...
-                '/team/' + this.team.slug + '/billing',
-                '/team/' + this.team.slug + '/settings',
-                '/team/' + this.team.slug + '/settings/general',
-                '/team/' + this.team.slug + '/settings/danger',
-                '/team/' + this.team.slug + '/settings/change-type'
+                '/team/' + this.team?.slug + '/billing',
+                '/team/' + this.team?.slug + '/settings',
+                '/team/' + this.team?.slug + '/settings/general',
+                '/team/' + this.team?.slug + '/settings/danger',
+                '/team/' + this.team?.slug + '/settings/change-type'
```
Steve-Mcl, line 87: Should this not simply return `[]` if team is not set?

As it stands, you will return at array containing `'/team/undefined/billing'` etc.

knolleary: @Steve-Mcl have you tested it? I can confirm this PR fixes the bug that is live in production now that we should get fixed. I don't see any ill effects because of this.

Steve-Mcl: > have you tested it?

Yes, it works well, however I only flagged this code section because the PR (on the whole) is aboud handling against `team` being undefined. By the same virtue, if `team` IS undefined, then we will get a bunch of links to `/undefined/` 

I am not blocking it - happy for the PR to be merged as is.

knolleary (approved):

PR #4956, opened 2024-12-23
Improve feedback when Hosted Instances are not available to a team

### LINKED ISSUE #4955: Free Tier Onboarding: Hosted Instances Feature Flag
### Description

With the Free Tier, we have 0 x Hosted Instances available. With other "premium" features, we show the relevant icon in the side navigation to inform the user. For Hosted Instances, we do not do this.

The "Create Instance" button is also available to them when going to this page, despite the fact that they're at the limit already (0). The challenge here is differentiating between someone that is at their limit, vs. someone that has no access to these features at all.

I see two options:

- Add a "Hosted Instances" feature flag, inline with the other premium features
- In the Front-End, we check to see if the limit is `0`, and if so, treat it as a "Premium" feature in the same way.

We are losing 50% of users after they've completed their credit card details here, as they get stuck trying to create a "Hosted Instance", but get confused as to why they cannot. 

### Epic/Story

_No response_

joepavitt:
## Description

Improves communication to user that Hosted Instances are unavailable on their Free Tier, as well as refining the UX path to Devices when a new Application is created and prevent the current confusion caused by users being redirected to `/instances` when it's not something they're able to use.

- Shows a "Premium" flag for "Hosted Instances" in the side navigation when 0 instances are permitted on tat given Team Type
- Auto-redirects to the `ApplicationDevices` screen when navigating to the base page for an Application, if no Instances are allowed to be created. This includes newly created applications on the Free Tier.
- Ensure we show the "Feature Unavailable" banner on the Application - Instances page if a user does navigate there
- Fix the `router-link` `:to` fields for the Instance and Device empty state placeholders on the Applications home page
- Prevent the option for "Create Node-RED Instance" when creating a new application in a team that isn't permitted to create an instance
- Prevent option to "Create Instance with Application" if the Team Tier does not permit it
- Improve language in the help text

### Screenshots

Screenshot 2024-12-23 at 12 47 19

Screenshot 2024-12-23 at 12 47 37

## Related Issue(s)

Closes #4955 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/application/Devices.vue
+++ b/frontend/src/pages/application/Devices.vue
@@ -1,6 +1,6 @@
 <template>
     <div>
-        <SectionTopMenu hero="Edge Devices" help-header="Node-RED Edge Devices - Registered to FlowFuse" info="Edge Devices belonging to this application.">
+        <SectionTopMenu hero="Edge Devices" help-header="Node-RED Edge Devices - Registered to FlowFuse" info="Manage remote instances of Node-RED running on your own hardware.">
             <template #pictogram>
                 <img src="../../images/pictograms/devices_red.png">
             </template>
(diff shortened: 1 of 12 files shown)


cstns (approved):
Looking good overall, would love some basic e2e coverage nothing fancy. I'll approve the pr as is and you can add the tests later as i won't be here to approve

PR #4975, opened 2025-01-03
Ensure Pipelines don't fall over if Device Groups are unavailable

joepavitt:
## Description

If Device Groups are not available to a Team, then the API returns a `404`, in turn, this was causing the Pipelines (Create) page to fall over when trying to load the list of available Device Groups.

### DIFF
--- a/frontend/src/pages/application/Pipeline/index.vue
+++ b/frontend/src/pages/application/Pipeline/index.vue
@@ -55,24 +55,31 @@ export default {
                 })
         },
         async fetchData () {
-            return this.loadPipeline()
-                .then(() => ApplicationApi.getApplicationDevices(this.application.id))
-                .then(res => {
-                    this.devices = res.devices
-                })
-                .then(() => ApplicationApi.getDeviceGroups(this.application.id))
-                .then((res) => {
-                    this.deviceGroups = res.groups
-                })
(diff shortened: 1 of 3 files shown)


ZJvandeWeg:
Fundamentally I don't like swallowing errors and making them non-tracable. If it's a 404 code, I get it, but in that case can we check if `statusCode >= 400 && statusCode  < 500`, and maybe console.log otherwise?

frontend/src/pages/application/Pipeline/index.vue
```diff
@@ -55,24 +56,37 @@ export default {
...
+            try {
+                this.devices = (await ApplicationApi.getApplicationDevices(this.application.id)).devices
+            } catch (err) {
+                this.devices = []
+                Alerts.emit('Failed to load Remote Instances', 'warning')
+            }
+            try {
+                this.deviceGroups = (await ApplicationApi.getDeviceGroups(this.application.id)).groups
+            } catch (err) {
+                if (err.request.status === 404) {
```
joepavitt, line 74: @ZJvandeWeg I've handled it ehre now, will show an alert to the user if we see an error we aren't expecting

ZJvandeWeg (approved):

Issue #3151, opened 2023-12-01
Clarification: Instances and Devices


MarianRaphael:
Currently, we refer to hosted NR Instances simply as 'Instances' and Remote NR Instances as 'Devices.' The proposal is to change these to:

- Hosted Instances
- Remote Instances

![Screenshot 2023-12-01 at 14 41 46](https://github.com/FlowFuse/flowfuse/assets/73583313/fd009768-1819-407d-8f2f-115aa717f807)

In addition, separation of the left navigation into more categories would be an option:

- Applications
- **Instances**

> - Hosted Instances
> - Remote Instances

- **Team**

> - Library
> - Members

- **Team Admin**

> - ...

gdziuba:
This will be good terminology to use that will help regardless of whether it is selfhosted or FlowFuse cloud hosted environments.  Sometimes devices can be my own cloud instances and that could get confusing.

MarianRaphael:
- Edge Instance

gdziuba:
@MarianRaphael What if it is a cloud instance not hosted by FF?

joepavitt:
Device User Paths: https://eu.posthog.com/insights/uV75j3my

joepavitt:
This surprises me - more users going to the Instance - Devices page than I would have expected...

joepavitt:
Comparison of the number of users that "Open Editor" for Instance vs. "Open Editor" for Devices from within those views.

<img width="1432" alt="Screenshot 2023-12-12 at 15 29 24" src="https://github.com/FlowFuse/flowfuse/assets/99246719/5270cf70-88b0-4d65-b14f-825fa79b2a24">

ZJvandeWeg:
@joepavitt Is there a delta between tiers too? I believe the editor is not available for starter tier customers?

Either way; it'd say this is validation for the seamless integration of NR into FF. cc @MarianRaphael

joepavitt:
>  it'd say this is validation for the seamless integration of NR into FF.

We have designs for this already btw - @MarianRaphael do you have the issues for it? I remember it being on GH somewhere, but can't find it atm.

ZJvandeWeg:
Related to: https://github.com/FlowFuse/flowfuse/issues/2246

joepavitt:
> Either way; it'd say this is validation for the seamless integration of NR into FF.

Not sure how us seeing users regularly going via FF to open the NR Editor is validation for it needing to be integrated  into the same screen though? 

[...]

PR #4976, opened 2025-01-03
"Devices" & "Edge Instances" > Remote Instances

joepavitt:
## Description

Closes https://github.com/FlowFuse/flowfuse/issues/3151

We are still getting too many support requests and too much feedback that "Devices", "Edge Devices" and "Edge Instances" is not clear. This is particularly apparent for new users coming on board, that are unfamiliar with the terminology we are using.

This PR unifies the terminology for "Instances" to "Hosted Instances" and "Remote Instances" (instead of "Edge Devices"). This aligns better with our pricing on "Instances", and also ensures we are better linking to appropriate documentation on the Device Agent.

I've swept through as best I can at the moment to ensure terminology is updated accordingly, and included some general help text improvements where appropriate too.

### Before:

Screenshot 2025-01-03 at 15 25 05

### After:

Screenshot 2025-01-03 at 15 24 44

### DIFF
--- a/frontend/src/components/DevicesBrowser.vue
+++ b/frontend/src/components/DevicesBrowser.vue
@@ -6,11 +6,11 @@
     >
         <ff-loading
             v-if="loadingStatuses || loadingDevices"
-            message="Loading Devices..."
+            message="Loading Remote Instances..."
         />
         <template v-else-if="team">
-            <FeatureUnavailableToTeam v-if="teamDeviceLimitReached" fullMessage="You have reached the device limit for this team." :class="{'mt-0': displayingTeam }" />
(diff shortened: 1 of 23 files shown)


ZJvandeWeg:
@joepavitt some UI tests are broken now. Otherwise looks good

joepavitt:
> Some UI tests are broken now

Already on it 👍

joepavitt:
As mentioned in DM - @ZJvandeWeg also proposed "Edge Instances" (rather than the current "Edge Devices"), which I'm absolutely open to.

I went with "Remote" as not all instances managed via the Device Agent would naturally faller under "Edge" use cases, I also felt "Remote" was a little less technically minded, and was clearer about it's intent/setup.

ZJvandeWeg (approved):

PR #4979, opened 2025-01-06
Updated Onboarding Tours

joepavitt:
## Description

https://github.com/user-attachments/assets/011f7d47-f691-440a-9223-4d0a0aa4e883

Our existing onboarding tour doesn't align very well to the new new FlowFuse Cloud Free Tier, as it assumes an Instance has been setup for you - not the case.

This PR adds two new tours. One as an alternative introduction to the concepts of FlowFuse, but knowing there is no existing Instance to highlight. Then, the second starts when a user selects "Add Remote Instance" (the last stage of Tour 1) which then directs the user through the stages of adding their first Remote Instance

### TODO

- Reliant upon https://github.com/FlowFuse/flowfuse/pull/4985

### DIFF
--- a/frontend/src/components/DevicesBrowser.vue
+++ b/frontend/src/components/DevicesBrowser.vue
@@ -173,7 +173,7 @@
                                 <template #icon-left>
                                     <PlusSmIcon />
                                 </template>
-                                Add Device
+                                Add Remote Instance
                             </ff-button>
                         </template>
                     </EmptyState>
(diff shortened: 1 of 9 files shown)


knolleary (approved):

gstout52:
At 0:52, the text box "Connect your remote instance" briefly appears in the upper-left corner before being moved to the lower center of the screen. When I create a new account and go through the tour, the same behavior is occurring. If there's a way to prevent the box from briefly loading in the wrong location, we should implement it.

PR #4985, opened 2025-01-06
Add option to auto-create team application

### LINKED ISSUE #4978: Auto-create application when creating free tier team
### Description

When a user signs up for a Trial they get a Team tier team with an Application & Instance created.

When a user signs up for a Free tier team, they don't get an Application or Instance created (as they aren't allowed to have an Instance). However, it would be more convenience to create the Application (without instance) for them.

### Epic/Story

_No response_

knolleary:
Closes #4978

Adds an admin setting to enable auto-creation of a default application whenever a team is created.

By enabling this option it removes one piece of setup friction for new teams.

### DIFF
--- a/forge/routes/api/settings.js
+++ b/forge/routes/api/settings.js
@@ -56,6 +56,7 @@ module.exports = async function (app) {
                 response['user:team:auto-create'] = app.settings.get('user:team:auto-create')
                 response['user:team:auto-create:teamType'] = app.settings.get('user:team:auto-create:teamType')
                 response['user:team:auto-create:instanceType'] = app.settings.get('user:team:auto-create:instanceType')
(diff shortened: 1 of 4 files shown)


joepavitt (commented):
Just checking, as I'm trying to break this locally at the moment (and it's working well), this workflow is completely independent from the auto-created team when signing up workflow, right?

This is triggered only when a user manually creates a team, which is now our default onboarding experience, as it requires users to select the Team Type?

knolleary:
> this workflow is completely independent from the auto-created team when signing up workflow

Correct. This is updating the endpoint used to explicitly create a team - which is the workflow users now go through when signing up.

I chose to make it auto-create an application for any team type as it's just a useful thing to do (and saves adding more logic around which team types should get it if not all)

PR #4986, opened 2025-01-06
Update text on create button for trial teams

knolleary:
## Description

When opting to create a Trial team, the final create button and text talks about sending the user to stripe. This is not the case for trial teams - clicking the button takes them straight to their team.

image

With this PR, when creating a trial team, the text/button have been updated as:

image

### DIFF
--- a/frontend/src/pages/team/create.vue
+++ b/frontend/src/pages/team/create.vue
@@ -53,7 +53,7 @@
                         </ff-button>
                     </template>
                     <ff-button v-else :disabled="!formValid" @click="createTeam()">
-                        <template v-if="billingEnabled && isSelectionTrial">Start trial</template>
+                        <template v-if="billingEnabled && isSelectionTrial">Start Free Trial</template>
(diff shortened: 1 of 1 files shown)


frontend/src/pages/team/create.vue
```diff
@@ -151,6 +155,15 @@ export default {
             return this.billingEnabled &&
                    !this.user.admin &&
                    this.input.teamType && this.input.teamType.properties?.billing?.requireContact
+        },
+        isSelectionTrial () {
+            // A team trial can be offered if:
+            // 1. User has no other teams
+            return this.teams.length === 0 &&
+            // 2. User is less than a week old
+                (Date.now() - (new Date(this.user.createdAt)).getTime()) < 1000 * 60 * 60 * 24 * 7 &&
```
joepavitt, line 164: worth centralising this logic somewhere? Feels like something that would be easily overlooked if we change trial lengths in the future?

joepavitt (approved):

Issue #4919, opened 2024-12-16
Create Async API schema from our own broker's topic hierarchy


joepavitt:
```[tasklist]
### Tasks
- [ ] https://github.com/FlowFuse/flowfuse/pull/4997
- [ ] https://github.com/FlowFuse/flowfuse/pull/5008
```
We will auto-generate an AsyncAPI v3 schema based on the detected topic hierarchy for the team broker.

For the first iteration, this is auto-generated and uneditable. Being able to edit the schema (to provide additional descriptions/context to the entries is a future item) is out of scope but will be something we'll want to do in the future, so some consideration should be made for that.

Currently, the auto-detected topics are stored as a simple array of topics. That is likely to change in the future, but for now, we'll stick with that.

We will also be adding the ability to generate schemas for 3rd party brokers in the near future; so choices made here need to accommodate that.

I'm working on the basis that we will generate a separate schema for each broker (for now). A combined schema will be doable, but one step at a time.

Here's an first iteration of an example schema we can generate:

```
asyncapi: 3.0.0
info:
  version: 1.0.0
  title: dev Team Broker
  description: An auto-generated schema of the topics being used on the team broker
servers:
  team-broker:
    host: broker.example.com
    protocol: mqtt
channels:
  foo/bar:
    address: foo/bar
  top/a/b:
    address: top/a/b
  /top/c/d:
    address: /top/c/d
  top/a/b/c:
    address: top/a/b/c
```

### Notes

1. `info.title` includes the Team Name + Broker name. For this iteration 'Team Broker' is the only broker. In the future, that will be the name of the relevant 3rd party broker.
2. `version` is hardcoded to `1.0.0` - its a required field for the schema. We will need to think about lifecycle of schema versions and how the value changes over time.
3. `servers.team-broker` - not including any username/password details here for first iteration. Will need to think about how to do that securely (if its a requirement) - and also *what* credentials to include if the Team has multiple clients defined.
4. `channels` - these are the individual topics. The top-level property is normally a more human-readable name for the channel and the address is the full topic string. Without any further information, I'm reusing topic for both. In a future where we allow users to edit the schema, this will be one of those editable things.

### URL & Security

In my initial prototyping, I'm serving the schema from `/api/v1/teams/:teamId/broker/:brokerAlias/schema.yml`

The `:brokerAlias` property will allow us to serve schemas for the individual 3rd party brokers. For now `team-broker` is the hard-coded alias we'll use.

[...]

PR #4997, opened 2025-01-09
Add initial schema generation for team-broker topics

knolleary:
Part of #4919

## Description

Initial implementation of schema generation based on the observed team broker topics.

See #4919 for details on the schema and url used to expose.

### DIFF
--- a/forge/ee/routes/teamBroker/schema.js
+++ b/forge/ee/routes/teamBroker/schema.js
@@ -1,0 +1,42 @@
+const YAML = require('yaml')
+module.exports = async function (app) {
+    app.get('/team-broker/schema.yml', async (request, reply) => {
+        const list = await app.teamBroker.getUsedTopics(request.team.hashid)
+        const schema = {
+            asyncapi: '3.0.0',
+            info: {
+                version: '1.0.0',
+                title: `${request.team.name} Team Broker`,
+                description: 'An auto-generated schema of the topics being used on the team broker'
+            }
+        }
+        // Add the team-broker details
+
+        // Figure out the hostname for the team broker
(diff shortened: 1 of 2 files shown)


forge/ee/routes/teamBroker/schema.js
```diff
@@ -0,0 +1,42 @@
...
+            }
+        }
+        // Add the team-broker details
+
+        // Figure out the hostname for the team broker
+        let teamBrokerHost = app.config.broker?.teamBroker?.host
+        if (!teamBrokerHost) {
+            // No explict value set, default to broker.${domain}
+            if (app.config.domain) {
+                teamBrokerHost = `broker.${app.config.domain}`
```
hardillb, line 20: `domain` might not be set for localfs builds.

I've defaulted to 

```
let defaultHost = app.config.domain ? `broker.${app.config.domain}` : app.config.host
```

if domain isn't set

hardillb (approved):
Approved, but with one comment

forge/ee/routes/teamBroker/schema.js
knolleary: Yeah, I thought I was being clever and added any additional check lower down to only add the `servers` portion to the schema if `teamBrokerHost` had a value... but I realise now, if `domain` isn't set, `teamBrokerHost` *will* have a value of `broker.`... will tidy up.

knolleary: Actually, the logic is sound. `teamBrokerHost` will only get set to a value if either the new setting or `domain` are set. If neither set, it will omit the `servers` section. No changes needed.

PR #4998, opened 2025-01-09
Add Team Broker hostname to UI settings

hardillb:
## Description

Expose the Team Broker hostname if set to the UI


knolleary (approved):
We may need to think more about localfs as it needs a port number as well; although no problem with the new config option is explicitly set. Given the use cases for localfs (mainly our own local dev work), we can cope with that.

PR #5008, opened 2025-01-14
Add Open Schema button to Topic Hierarchy view

knolleary:
Part of #4919 

Adds a button to the header of the Topic Hierarchy view to open the generated AsyncAPI in a new window.

image

Also noticed the generated schema ought to have a `security` section - so added.


cstns:
confirmed working locally 
image
image

Issue #4970, opened 2025-01-03
Brief Redirect to 'Choose Team Type' Page After Login


sumitshinde-84:
### Current Behavior

When logging in with a username and password, after a successful login, it first navigates to the 'Choose Team Type' page, followed by the default team view.

For more information, refer to the Slack thread: 
https://flowforgeworkspace.slack.com/archives/C032Q63FGG1/p1735897858300549

### Expected Behavior

When logging in with a username and password, after a successful login, It should navigate to the default team view.

### Steps To Reproduce

- Login with username and password

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

knolleary:
@cstns can you take a look at this? It's quite noticeable whenever logging in.

PR #5012, opened 2025-01-14
Set pending state when loggin in to prevent no team limbo

cstns:
## Description

After login, users are redirected to the home component which has a loading screen present (if account.pending e true) and a team type selector present (if account.teams.length === 0). The purpose of this page is to redirect users to their respective team page after their team/teams are loaded via the two watchers (frontend/src/pages/Home.vue:45)

The account store checkState method is the default method that get's called to hydrate the application and among other things, clears pending state. By setting the pending state to true when logging in, I was able to trick the Home page to not show the team type selector but keep the loading screen present until either the component redirects due to the team/teams being loaded or the checkState method clearing the pending state.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4970

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/store/account.js
+++ b/frontend/src/store/account.js
@@ -205,6 +205,9 @@ const mutations = {
     clearPending (state) {
         state.pending = false
     },
+    setPending (state, pending) {
+        state.pending = pending
+    },
     setLoginInflight (state) {
         state.loginInflight = true
     },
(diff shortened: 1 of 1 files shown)


knolleary (approved):
Looks good. Have tested with network throttling applied to slow down API response times and all looks good.

PR #5016, opened 2025-01-15
Update Team Device Groups routes to use hashid not slug

knolleary:
Spotted moments after I approved and merged #5009 - the api urls should be using team hashid, not slug.

This PR fixes it.


cstns (approved):

cstns:
will merge after checks pass

PR #5017, opened 2025-01-15
Disable caching of index.html when in dev mode

knolleary:
When telemetry (ie sentry) is enabled, the app loads `index.html` and injects the necessary scripts - and then caches the result for subsequent requests.

When running in dev mode (ie `npm run serve`), webpack rewrites index.html with the new content hashes applied. However, the forge app still serves the cached version - causing errors.

This fix disables the caching of index.html when telemetry is enabled *and* we're running in development mode. We know we're dev mode as `NODE_ENV` is set to `development` by the npm scripts.

### DIFF
--- a/forge/routes/ui/index.js
+++ b/forge/routes/ui/index.js
@@ -27,7 +27,7 @@ module.exports = async function (app) {
      * feConfig - the 'frontend' portion of our flowforge.yml
      */
     async function injectAnalytics (config) {
-        if (!cachedIndex) {
+        if (process.env.NODE_ENV === 'development' || !cachedIndex) {
             const telemetry = config.telemetry
             const support = config.support
             const filepath = path.join(frontendAssetsDir, 'index.html')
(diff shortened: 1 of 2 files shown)


joepavitt (approved):

knolleary:
Failing UI test - looking....

cstns:
it's the flaky offboarding test which redirects to google, re-running the tests should fix it

I'll create an issue for it

knolleary:
UI Test failure is unrelated I think. The failing test is that we redirect to the given URL when a user deletes.
```
  -'\'https://www.google.com/search?q=rick+astley&sei=2eeHZ9-zD-eg5NoPgpGlYA\''
      +'\'https://www.google.com/search?q=rick+astley\''
```
In this run, the additional `sei=` query param has been added from *somewhere*. Can't see anything obvious in our code - could be some cypress artefact? It is too generic a property to find any meaningful search results online about it.

Have pushed a fix to make the test accommodate additional query params.

Issue #5027, opened 2025-01-17
'Choose Team Type' blank after running initial setup, no admin sidebar


knolleary:
### Current Behavior

After running the initial setup of a new FF platform, the user logs in and is presented with the 'Choose Team Type' page - as they have no team.

As part of the setup, we do create a default starter TeamType. However there is nothing to select:

<img width="986" alt="Image" src="https://github.com/user-attachments/assets/05604da2-5a76-4b5a-b3a6-89d9c5e635b3" />

Switching to the Admin view, there is no sidebar menu:

<img width="985" alt="Image" src="https://github.com/user-attachments/assets/fccb4dc7-6187-41c2-85b5-09f127bcab0e" />

For the Choose Team type issue, I added some debug to TeamTypeSelection.vue - can confirm the api to get the list of available team types is returning the default type created by setup. Don't know why that isn't show yet though.

### Expected Behavior

1. See the default team type in the Choose Team Type window
2. See the admin sidebar

### Steps To Reproduce

Fresh install of 2.13 - run through the setup wizard.

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Closed by #5029?

knolleary:
Yes, have now verified this is fixed.

PR #5028, opened 2025-01-17
Fix invalid check for billing enablement

knolleary:
Part of #5027 

The new 'Choose Team Type' selection page had a bad check for billing enablement that was throwing an undef error if billing wasn't enabled.


knolleary:
Although, after creating the team, the view doesn't update properly:

image


Need to do a full page reload to see the new team.

cstns (approved):

knolleary:
Have included a fix for the second part where it wasn't then taking the user to their new team.

cstns:
Looking good!

PR #5037, opened 2025-01-20
Update node-saml to 5

knolleary:
Replaces https://github.com/FlowFuse/flowfuse/pull/4734

The bump to node-saml@5 includes a breaking change on its options object - `cert` is now `idpCert`.

This PR updates to node-saml@5 and maps the old property to the new name. Tested locally against azure SSO.


hardillb:
Tested with keycloak, appears to work as expected

hardillb (approved):

Issue #4835, opened 2024-11-27
Set Env-vars to be hidden after creation


robmarcer:
### Description

Can an env-var be set to be 'secret' so that once the string is submitted it can't be extracted from the FF UI?

This was requested by the following customer - https://app-eu1.hubspot.com/contacts/26586079/record/0-1/24993051

### Which customers would this be available to

Enterprise Tier Only (EE)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
First iteration:
- UI option to hide the value
- Server-side work to store this configuration for each env var

Second Iteration:
- Encryption of these variable values in storage

Steve-Mcl:
Care point - exporting of snapshots include env vars. consider an option to include/exclude secrets?

cstns:
I think that would complement this feature nicely

Steve-Mcl:
> Steve suggested aligning the approach with how Node-RED handles credentials, and I tend to agree.

To be clear, I stated Node-RED has a mechanism for credentials - if desired (possible?), we might wish to utilise those since they remain encrypted on file.

---

The other points discussed were that snapshots (currently) store the env vars as Key/Value pairs & it is not currently possible to identify them as being "protected" thus not able to prevent them being downloaded from the snapshot download UI to reveal the env vars set as secret.  Changing these would require the whole mechanism of snapshots (API, launcher, device agent) to have awareness of the enhanced format.

cstns:
Thank you for clearing that out!

ZJvandeWeg:
@robmarcer Can you send the customer the news that this has been shipped and will be in their next release?

PR #5043, opened 2025-01-21
Allow env var values to be hidden in UI

cstns:
## Description

Allows setting of hidden env vars, once set they can only be updated or removed
- applies to both remote and hosted instances
- duplicating hosted instances preserves varhidden state
- other limitations described in closes https://github.com/FlowFuse/flowfuse/issues/4835
- updated error logging and validation to include env vars names starting with numbers
- fixes validation recursion
- fixes removal of unsaved vars

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/4835

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/controllers/Device.js
+++ b/forge/db/controllers/Device.js
@@ -166,7 +166,12 @@ module.exports = {
         result.push(makeVar('FF_SNAPSHOT_ID', snapshotId))
         result.push(makeVar('FF_SNAPSHOT_NAME', snapshotName))
         result.push(...app.db.controllers.Device.removePlatformSpecificEnvVars(envVars))
-        return result
+        return result.map((env) => {
+            if (Object.hasOwnProperty.call(env, 'hidden') && env.hidden === true) {
+                env.value = ''
+            }
+            return env
+        })
     },
 
     /**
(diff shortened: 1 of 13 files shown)


knolleary:
@cstns can you add a screenshot of this in action? Helpful to see as part of the initial review, but can also be reused in changelog.

knolleary:
Doing some testing with a Device - any env var with a hidden value ends up blank on the device. I haven't dug through the whole PR yet, but [this change](https://github.com/FlowFuse/flowfuse/pull/5043/files#diff-28e7488dc29d6612be211645f57ced53d3f4fe44935b552d5ae256dfeb2c4545) may be to blame as its code that's reused when sending info on the env vars to the front end as well as down to the device.

cstns:
> but [this change](https://github.com/FlowFuse/flowfuse/pull/5043/files#diff-28e7488dc29d6612be211645f57ced53d3f4fe44935b552d5ae256dfeb2c4545) may be to blame as its code that's reused when sending info on the env vars to the front end as well as down to the device.

You are right, i'll move this to be filtered on the controller side. The same thing is happening to instance vars

knolleary:
Verified its working properly now.

I'm wondering if the eye icon is clear enough as to what it does. Could we add a tooltip explaining its purpose?

knolleary (approved):
Looks good.

Issue #5014, opened 2025-01-15
Support FF Auth on Remote Instances


joepavitt:
### Description

We support FF User auth for Hosted Instances, but not Remote. This had been avoided originally because we cannot guarantee that Remote Instances are running in an online environment, given many customers run in closed/offline environments.

However, if customers want to use FF Auth, then we should not stop them. There is value, especially in use cases with FlowFuse Dashboard, to have FF User Auth enabled and securing remote Dashboards, especially when paired with the[ FF User Plugin.](https://dashboard.flowfuse.com/user/multi-tenancy.html#flowfuse-user-addon)

### Which customers would this be available to

None

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Status - both https://github.com/FlowFuse/flowfuse/pull/5068 and https://github.com/FlowFuse/device-agent/pull/341 are ready to merge, just need to coordinate.

@knolleary is this something we can do today?

knolleary:
@joepavitt the device agent PR has been merged. We are not ready to cut the release as I want to see if we can get https://github.com/FlowFuse/flowfuse/issues/5069 included as a significant DX improvement. I'm looking at that today.

Once the release is cut, we can then merge #5068 and have it go live.

joepavitt:
Thanks Nick

joepavitt:
@hardillb are you okay to verify this please so that we can close it out?

hardillb:
I'll try and set up a device on prod this afternoon and verify.

I did test it all locally as part of the review, so should be good.

knolleary:
Issue reported by customer: https://app-eu1.hubspot.com/help-desk/26586079/view/233410279/ticket/88305701084/thread/10233611463#email

If the device's local access is via http, but the platform is https, the platform redirects back to https on the device which doesn't work.

Need to check the referrer origin on the request to ensure it is applied when redirecting.

hardillb:
I've just tested on prod and the final redirect got the HTTP vs HTTPS wrong (the redirect was to https://localhost:1880 not http://localhost:1880

knolleary:
Fix for http/https issue in https://github.com/FlowFuse/device-agent/pull/353

Once that is merged and we cut the 3.1.2 release of Device Agent, we can verify and close this issue.

hardillb:
I have verified this with 3.1.2 and production.

PR #5068, opened 2025-01-30
Add support for httpNodeAuth settings on devices

knolleary:
Part of #5014 

**Note: do not merge until Device Agent 3.1 has been published**

## Description

Adds the HTTP Node security settings to the Device settings page.

This reuses the Instance template page for node security.

Includes a banner warning if the Device Agent version is unknown, or too old to support the settings. However, I did chose to allow the settings to be saved in those cases. (Note banner at top, which is repeated above the Save button, just to be sure).

image

---

To support FF Auth on devices, we now generate an AuthClient whenever the device calls home to get its settings. That happens whenever the device is notified the settings have been changed. The auth client details are only needed when establishing a login session - changing the client details doesn't invalidate existing sessions.

I did have to update `oauth.js` which assumed the AuthClient owner was an Instance - but it can now be a device.

Will link to the companion Device Agent PR once it is available.

---

 - [x] Add some tests

### DIFF
--- a/forge/db/controllers/AuthClient.js
+++ b/forge/db/controllers/AuthClient.js
@@ -23,6 +23,21 @@ module.exports = {
         return client
     },
 
+    createClientForDevice: async function (app, device) {
+        const existingAuthClient = await device.getAuthClient()
+        if (existingAuthClient) {
+            // TODO: are there sessions to expire as well?
+            await existingAuthClient.destroy()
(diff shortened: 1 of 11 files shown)


hardillb (approved):
Tested and read through. Looks good.

Not merging as is dependent on 3.1 device agent. (reviewing next)

PR #5067, opened 2025-01-30
3rd party broker UI

cstns:
## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #5076, opened 2025-02-04
Update team broker topic handling to store in MQTTTopicSchema table

knolleary:
## Description

Builds on top of #5067 

This updates the Team Broker topic handling to save in the new topic schema database.

To achieve this, a migration creates a default BrokerCredentials object that can be used when storing team broker topics in the MQTTTopicSchema table (null keys don't work with the index we use).

The id of this default object is stored in platform settings - so it can be looked up without hitting the db everytime.

Other parts of this change:
 - retired the old api endpoint for getting broker topics in favour using the new `brokers/<id>/topics` end point using the `team-broker` id.
 - retired the housekeeper task that tidied up topic cache; we will persist permanently now.

Some additional bits I've just thought of to tick off before this is merged:

 - [x] ~hook to delete team-broker topics when team is deleted~ - on cascade delete constraint already applied


hardillb (approved):
Looks good to me, will let you fix the e2e tests before merging

hardillb:
I have fixed the merge conflict (99.9% sure)

Issue #5079, opened 2025-02-06
Not able to deploy pipeline from a device in Developer Mode


Steve-Mcl:
Currently, it is not possible to "play" a pipeline stage where the source is a device in dev mode.

_if you are using a device to develop flows, it is reasonable to want to pipeline the progress to another device/instance_

![Image](https://github.com/user-attachments/assets/d7f3b2c6-b3a9-4389-be93-d521505b0098)

It is however possible to generate snapshots while in developer mode so this would save the efforts of having to temporarily switch the device out of dev mode just to play the pipeline.

knolleary:
Going to scope out what effort is needed to address this. Will then make a call on if we can get it done as part of 2.14

PR #5089, opened 2025-02-07
Allow a dev-mode device at Stage 0 to be triggered

knolleary:
Closes https://github.com/FlowFuse/flowfuse/issues/5079

## Description

This updates the UI constraints applied around Pipelines to allow a Device in Dev Mode to be a valid first stage in a pipeline.

This simplifies the workflow for using pipelines with Dev Mode devices; you don't have to take the device out of dev mode in order to move its snapshot to other devices.

This is a simple iteration that allows for the existing Deploy Actions to be chosen. There is scope to add a 'Create Snapshot' action - but that would only work if the device was in dev mode. Given the complications that introduces, in the spirit of fast iteration, I'm going with the existing available actions.

The workflow is then:

1. Put device in dev mode, work on flows
2. When ready to deploy via pipeline, create a snapshot (or rely on the auto-snapshot from your last deploy)
3. Ensure the pipeline stage is configured to 'use latest snapshot' - *not* 'use active snapshot'.

This last point is one small potential gotcha. Whilst the device is in dev mode, the 'active' snapshot is whatever snapshot was deployed to the device *before* it was put into dev mode. The 'latest' snapshot is the most recently created one.

In terms of UX changes, for a Stage0 device in dev mode, I've removed the red warning banner, but left the warning text in place.

image

### DIFF
--- a/frontend/src/components/pipelines/Stage.vue
+++ b/frontend/src/components/pipelines/Stage.vue
@@ -1,5 +1,5 @@
 <template>
-    <div v-if="stage" class="ff-pipeline-stage" data-el="ff-pipeline-stage" :class="{'ff-pipeline-stage--error': inDeveloperMode}">
+    <div v-if="stage" class="ff-pipeline-stage" data-el="ff-pipeline-stage" :class="{'ff-pipeline-stage--error': inDeveloperMode && stageIndex > 0}">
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):
I approve :)

PR #5100, opened 2025-02-11
Add externalBroker feature flag to UI

knolleary:
Add feature-flag guards in the UI to not show external broker options if flag not set.


cstns:
added https://github.com/FlowFuse/flowfuse/issues/5103 as a follow-up

hardillb (approved):

PR #5113, opened 2025-02-12
Add button to allow start/stop of MQTT agent

### LINKED ISSUE #5084: MQTTBroker: Add status badge and toggle to broker page
The agent runs by default, but can be optionally paused by the user.

This task is to reflect that in the UI.

This only applies to the 3rd party brokers today - until we move Team Broker over to use the agent as well.

PR https://github.com/FlowFuse/flowfuse/pull/5081 adds an endpoint to get the current status of the broker.

This task will require:

```[tasklist]
- [x] Add api to start/stop agent.
- [x] UI: add status pill to broker page
- [ ] UI: add start/stop toggle to broker page
```

hardillb:
closes #5084
## Description

Adds a button to the hierarchy page start/stop the MQTT-Schema-Agent

There is also some state/status renaming to make things consistent and meaningful

image

image

image

image

## Related Issue(s)

#5084

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/views/BrokerCredentials.js
+++ b/forge/db/views/BrokerCredentials.js
@@ -25,7 +25,8 @@ module.exports = function (app) {
                 protocol: result.protocol,
                 ssl: result.ssl,
                 verifySSL: result.verifySSL,
-                clientId: result.clientId
+                clientId: result.clientId,
+                state: result.state
             }
             return cleaned
         },
--- a/forge/ee/routes/teamBroker/3rdPartyBroker.js
+++ b/forge/ee/routes/teamBroker/3rdPartyBroker.js
@@ -305,9 +305,9 @@ module.exports = async function (app) {
     }, async (request, reply) => {
         if (request.params.brokerId !== 'team-broker') {
             try {
-                const state = await app.containers.getBrokerAgentState(request.broker)
+                const status = await app.containers.getBrokerAgentState(request.broker)
                 const clean = app.db.views.BrokerCredentials.clean(request.broker)
-                clean.state = state
+                clean.status = status
                 reply.send(clean)
             } catch (err) {
                 reply.status(500).send({ error: 'unknown_error', message: err.toString() })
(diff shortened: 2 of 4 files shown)


joepavitt:
Can I check, from a user's perspective, why would they want to start/stop this? Also, the "Refresh" button doesn't make sense if it's disabled, right?

hardillb:
This is to start/stop the collection of new topic/payload info (without removing them, because if you delete the broker you delete the linked topics).

hardillb:
Refresh is probably less useful if the collection of data has been stopped yes.

joepavitt:
> This is to start/stop the collection of new topic/payload info

Yes, I udnerstand that, but why would a user ever want/need to do this?

hardillb:
Because they don't want us looking at all their data all the time

joepavitt (dismissed):
- Struggling to use this smoothly. The state doesn't auto refresh when toggling, I'm having to manually refresh the page, it's not polling. I'm seemingly stuck in "Suspended" currently too.
- The `ff-button` should be a `ff-toggle-switch` instead.
- The "refresh" button should also be hidden when there is no connection, it's its not going to be functional at that point.
- There is also a merge conflict, whereby the refresh button should be `kind` of `secondary`
- I'm also not clear why we are using things like `provide` or `ref` within the Vue component

joepavitt:
Doing my best to try and fix these locally asap, but I'm also now hitting the wonderful mimetype errors again

joepavitt:
https://github.com/user-attachments/assets/69acda85-453c-48af-964e-d6e7dc9cb5b8

Actioned everything above, commit incoming

joepavitt:
Spotted a problem myself already - the toggle shouldn't show for the team broker I believe? @knolleary I need to head off now, but hoping that can be an action for the morning

hardillb:
>Spotted a problem myself already - the toggle shouldn't show for the team broker I believe? @knolleary I need to head off now, but hoping that can be an action for the morning

I'm on it

cstns (approved):
I advised the provide/inject approach because it's generally not recommended to pass down props to router views but don't see why this would block the task

PR #5129, opened 2025-02-14
Add option to disable manual billing

### LINKED ISSUE #5123: Add admin option to disable manual billing for a team
Putting a team into manual billing mode is currently a one-way street in the UI.

Taking a team out of manual billing mode requires a database poke. This means only those with direct access to the production database can do so and it isn't something our customer success folk can do directly.

Making the existing button a toggle, with the backend api to handle it, is a quick piece of work and will unblock this workflow and take me off the critical path for this task.

knolleary:
Closes #5123 

## Description

This adds a button and api for an admin to disable manual billing mode on a team.

Once disabled, the team has no subscription so has to checkout via stripe again to continue using their team.

### DIFF
--- a/forge/ee/lib/billing/index.js
+++ b/forge/ee/lib/billing/index.js
@@ -585,6 +585,23 @@ module.exports.init = async function (app) {
                 await app.db.controllers.Subscription.createUnmanagedSubscription(team)
             }
         },
+        /**
+         * If in unmanaged mode, this will update the subscription to be 'canceled'
+         * but leave all instances running.
+         *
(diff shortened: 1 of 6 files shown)


knolleary:
We have a customer who we've put into manual billing mode temporarily whilst they sort out a payment issue. I've pushed this PR today so we have a quick way to re-enable billing for them when they have resolved their payment issue. Without this, it would require database surgery from someone who knows what they are doing.

hardillb (approved):
Reads OK, but I don't have a local billing enabled to test it locally.

PR #5134, opened 2025-02-14
Fix: Adds new "Open Schema" button

joepavitt:
## Description

Fixes a regression introduced via merge conflict in: https://github.com/FlowFuse/flowfuse/pull/5113


cstns (approved):
pushed a quick fix to align the docs path with the overall broker route paths which i couldn't propose as a fix because the file wasn't included in the first place

feel free to merge after tests pass

Issue #5131, opened 2025-02-14
Registry API to return catalogue.json


joepavitt:
### Description

Required by Node-RED, and can be utilised by FlowFuse UI to display Catalogue contents

### Epic/Story

#5087

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

hardillb:
I've asked the vedaccio folk for suggestions on how to query the repo for a single scope (team) to try and make this more efficient

https://github.com/orgs/verdaccio/discussions/5100

hardillb:
API to return a catalogue in #5173, still need to work out how to trigger building the catalogue for each team for this to return.

At the moment it looks like it will need to be a single query to get all packages in the repo and then filter them into per team entries.

hardillb:
Generating on request, so no need to trigger for now

PR #5173, opened 2025-02-24
Add API endpoint for private npm registry catalogue

hardillb:
part of #5131

## Description

- GET `/api/v1/teams/:teamId/npm/catalogue`

Currently requires that the team id is also passed as the `teamId` query parameter this is a sudo auth token for access to the catalogue.

e.g. `https://app.flowfuse.com/api/v1/teams/:teamId/npm/catalogue?teamId=teamId`

returns a JSON catalogue

## Related Issue(s)

#5131 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
This can be reviewed and merged as it should not be enabled until the correct configuration is added to the `flowforge.yml` file.

knolleary:
In PR #5194 , I see we're adding a route under `/api/v1/teams/:teamId/npm/packages`

Can I suggest we update this PR to match - to keep everything under the `/npm/` space -  `/api/v1/teams/:teamId/npm/catalogue`?

knolleary:
> Currently requires that the team id is also passed as the teamId query parameter this is a sudo auth token for access to the catalogue.

Just checking (as it isn't stated) that is a temporary choice whilst we iterate and before this is made generally available?

hardillb:
> Just checking (as it isn't stated) that is a temporary choice whilst we iterate and before this is made generally available?

Yes

forge/ee/routes/catalogues/index.js
```diff
@@ -0,0 +1,49 @@
...
+                        password: app.config.npmRegistry.admin.password
+                    }
+                })
+
+                const modules = []
+                for (const package in packageList.data) {
+                    if (package === '_updated') {
+                        continue
+                    }
+                    if (package.startsWith(`@${request.teamId}/`)) {
```
hardillb, line 25:
```suggestion
                    if (package.startsWith(`@${request.params.teamId}/`)) {
```

hardillb:
>> Currently requires that the team id is also passed as the teamId query parameter this is a sudo auth token for access to the catalogue.

> Just checking (as it isn't stated) that is a temporary choice whilst we iterate and before this is made generally available?

@knolleary what do you think about using the instance/device id for this key. We can then check if it's part of the Team in question?

knolleary (changes requested):
As discussed, we plan to update the url to include either `instance` or `device` query params.

knolleary:
I've merged 5194 down into this branch - it's preHandler will need updating with the changes to the query param checks.

hardillb:
> it's preHandler will need updating with the changes to the query param checks. 

I got a bit mixed up with different branches on Friday, but it's already done in https://github.com/FlowFuse/flowfuse/pull/5212 . 

I'll port that change down and then this can be merged.

knolleary (approved):

hardillb:
Any reason not to merge this?

knolleary:
Happy for it to be merged once tests go green

PR #5212, opened 2025-02-27
Pass npmrc and catalogue to instance/device

### LINKED ISSUE #5166: Update nr-launcher or instance settings API to add details to .npmrc
### Description

Once the npm registry is available the NR instances will need to be configured to access the repo for specific scopes.

This will require a scoped registry and a authentication token to be added to the .npmrc file passed to the instances.

The best place to do this will be in the Instance settings API since the nr-launcher already knows how to apply a .npmrc file if passed from the settings API

### Epic/Story

#5087

hardillb:
fixes #5166 
## Description

Adds the required entries to instance and device .npmrc files and catalogue list

Also adds API for generating auth credentials for a team owner/member

- GET `/api/v1/teams/:teamId/npm/userToken` returns 200 is user has a token already, 404 if not
- POST `/api/v1/teams/:teamId/npm/userToken` returns `{ usename: '...', 'password': '...'}` to be shown to the user with comment that it will only be shown once and should be used with `npm login`.

Note should also be shown, that re-generating the username/password will invalidate ANY existing auth tokens for that user@team combination.

## Related Issue(s)

#5166 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/db/controllers/AccessToken.js
```diff
@@ -409,5 +409,35 @@ module.exports = {
             ownerType: 'broker'
         })
         return { token }
+    },
+
+    createTokenForNPM: async function (app, entity, team, scope = ['team:packages:read']) {
+        // might want to prefix the entityId with `p-`, `d-` and `u-` rather than relying on
```
knolleary, line 415: > might want to prefix the entityId with `p-`, `d-` and `u-` rather than relying on

That does feel sensible to me.

hardillb: @knolleary I've pushed an update with this.

Shout if you want to run the whole stack and I can help getting verdaccio and the plugin running

knolleary:
Given the ongoing updates in #5254, going to hold back merging this until that is complete and merged down into this PR.

hardillb:
@knolleary #5254 is not merged to here, any more reason to hold off merging to main.

This functionality is still gated on a `flowforge.yml` change (and the helm deployment FlowFuse/helm#549 changes getting merged)

knolleary:
@hardillb I was going to give it one last run through, but as you say, given its gated on config, lets merge and move forward.

knolleary (approved):

hardillb:
OK, just looked at the pre-staging env for this.

Looks like the test for if it's enabled in the UI is a little off as it's suggesting an upgrade even when the feature is not enabled for the platform.

image


I'll try and fix this before we merge

Issue #5171, opened 2025-02-24
Add conneciton failure feedback to External Broker connection


hardillb:
### Description

Currently the only feedback to a failure to connect to a remote broker is red error pill in the header.

The status check API (GET `/api/v1/teams/:teamId/brokers/:brokerId/`) does return an `error` field which contains the reason for the failure:

- ECONNREFUSED
- ENOTFOUND (DNS lookup failed)
- WRONGMQTTVERSION
- CLIENTIDNOTALLOWED
- MALFORMEDCREDS
- NOTAUTHORIZED
- BANNED

These should be converted into a human readable message.

### Epic/Story

#4965

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
@hardillb could you write a quick sentence for each of these please that I can include for human readability. I have ENOTFOUND already:

`Error (ENOTFOUND): The broker could not be found. Please check your connection settings and try again.`

but need the others, and want to be sure you're getting useful information from it if we have customers report it.

hardillb:
- ECONNREFUSED - The broker refused the connection, most likely either wrong port or wrong hostname
- ENOTFOUND (DNS lookup failed)
- WRONGMQTTVERSION - Broker does not support the configured MQTT protocol version
- CLIENTIDNOTALLOWED - Broker rejected connection because client id either too long or contains invalid characters
- MALFORMEDCREDS - Broker rejected because of bad credentials (e.g. no username or password when required)
- NOTAUTHORIZED - Broker rejected because incorrect credentials
- BANNED - MQTT 5 Broker has banned client, possibly IP/client id/credentials

There will probably be more later around bad SSL certs

joepavitt:
Thanks Ben, I've marked https://github.com/FlowFuse/flowfuse/pull/5218 as ready for review

PR #5218, opened 2025-02-28
Add new "Broker Error" component which details issues with connecting to 3rd party broker

joepavitt:
## Description

- Provides human readable feedback on why a broker connection has failed/errored.

## Related Issue(s)

Closes https://github.com/FlowFuse/flowfuse/issues/5171

## Outstanding Items

Waiting on response to https://github.com/FlowFuse/flowfuse/issues/5171#issuecomment-2690758277 in order to get human readable messages for each error code.

### DIFF
--- a/frontend/src/pages/team/Brokers/Hierarchy/index.vue
+++ b/frontend/src/pages/team/Brokers/Hierarchy/index.vue
@@ -42,6 +42,7 @@ export default {
             required: true
         }
     },
+    emits: ['broker-updated'],
     data () {
         return {
             loading: false,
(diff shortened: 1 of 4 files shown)


hardillb:
If you make a change to the config the error doesn't get removed until the next status poll.

Would be good to put it back into pending state and clear the erorr.

hardillb:
Tested ECONREFUSED and NOTAUHORIZED.

The others are hard to test quickly.

Big improvement over saying nothing.

hardillb (changes requested):
If you make a change to the config after an error and hit savethe error doesn't get removed until the next status poll, so it looks like it's still broken

Would be good to put it back into pending state and clear the error first.

Other than that I've tested a couple of failure modes and it looks good.

joepavitt:
Great point @hardillb - I've fixed this

hardillb (approved):

Issue #5088, opened 2025-02-07
UX: Team Pipeline page navigation improvements


knolleary:
Navigating around the Team level view of pipelines is a bit confusing for me.

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/55db3e5c-f723-4180-8559-190f7d76286d" />

1. In my mind, there is a natural hierarchy of `Application > Pipeline`. The display shows `[Pipeline Name] [Application]`
2. Click on the pipeline name, or anywhere on the header, takes you to the pipeline. Clicking on the Application name takes you to the application overview - with no pipeline information

When clicking quickly around, there is no obvious difference between the two labels in the header, so I regularly click the application name as I throw my mouse in the right direction and miss.

Can we improve the visual distinction between the labels? For example, add the Application and Pipeline icons ahead of their labels? Maybe move the Application name to be right-aligned so it isn't so prominent.

---

Thinking about it some more, I don't think we need for the Application name to link to the Application overview page at all; clicking on the header takes you to the pipeline within the application - that has got you 90% of the way into the application.
So I would suggest not having it link separately. We still want the Application name in the header to help identify the pipeline - and I still think the visual presentation of that could be clearer.

joepavitt:
Thanks Nick, I'll fix that now, I don't like that.

PR #5233, opened 2025-03-04
Styling: Make it clearer that Pipeline header is interactive

joepavitt:
## Description

Minor CSS change to make the header of the Pipelines in the `/pipelines` view whit instead of grey, to make it more obvious it's an interactive element (grey suggests inactive)

### Before

Screenshot 2025-03-04 at 09 54 29

### After

Screenshot 2025-03-04 at 09 52 48


knolleary:
fyi I raised https://github.com/FlowFuse/flowfuse/issues/5088 recently with suggested improvements here. With the current design there is no indication that clicking the application name takes you to somewhere different to the rest of the header. 9 times out of 10, I click the application name and it takes me to where I don't want to be.

PR #5234, opened 2025-03-04
Improve the labelling of a Pipeline for improved navigation

### LINKED ISSUE #5088: UX: Team Pipeline page navigation improvements

joepavitt:
## Description

- Removes the nested link to the "Application" to keep just the whole header a navigation item to the Pipelines view.
- Restyle the Application's name in the header to be nested under the Pipeline name, as it made it difficult to process the Pipeline name vs. the Application name

### Before

Screenshot 2025-03-04 at 10 04 15

### After

Screenshot 2025-03-04 at 10 01 07

## Related Issue(s)

Closes #5088

### DIFF
--- a/frontend/src/pages/team/Pipelines/components/TeamPipeline.vue
+++ b/frontend/src/pages/team/Pipelines/components/TeamPipeline.vue
@@ -4,16 +4,12 @@
             :to="{name: 'ApplicationPipelines', params: {id: pipeline.application.id}}"
             class="ff-pipeline-header flex gap-5 self-end items-center truncate"
         >
-            <span class="name">
-                {{ pipeline.name }}
-            </span>
-            <router-link
-                :to="{name: 'Application', params: {id: pipeline.application.id}}"
-                class="ff-pipeline-application-name truncate"
-                @click.stop
-            >
-                {{ pipeline.application.name }}
-            </router-link>
+            <div class="flex flex-col gap-0.5">
+                <span class="name">
+                    {{ pipeline.name }}
+                </span>
+                <span class="text-xs text-gray-500">{{ pipeline.application.name }}</span>
+            </div>
             <span class="to">
                 <ChevronRightIcon class="ff-icon " />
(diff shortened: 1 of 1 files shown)


frontend/src/pages/team/Pipelines/components/TeamPipeline.vue
```diff
@@ -4,16 +4,12 @@
             :to="{name: 'ApplicationPipelines', params: {id: pipeline.application.id}}"
             class="ff-pipeline-header flex gap-5 self-end items-center truncate"
         >
-            <span class="name">
-                {{ pipeline.name }}
-            </span>
-            <router-link
```
cstns, line 10: are we sure we want to ditch the app link?

joepavitt: Yep - definitely, it's really annoying, nobody clicks it intentionally, and clicking the pipeline header takes you to the Application > DevOps Pipeline view instead

cstns (approved):

PR #5237, opened 2025-03-04
Add option to disable billing requirement for a team type

### LINKED ISSUE #5124: Remove the need for Stripe interaction in Free Tier
### Description

We are losing 60% of users on sign up to the free tier because our existing architecture requires a Stripe subscription in order to have a valid team.

This work will seek to remove that tie, and permit Free Tier teams to operate without a valid Stripe subscription.

### Epic/Story

_No response_

knolleary:
Closes #5124 

## Description

This adds a teamType property to disable the requirement for billing to be setup on a team:
 - `TeamType.properties.billing.disabled` - set to true means the team can exist without billing setup

Other changes:
 - Backend checks updated to allow otherwise-billable actions for such a team
 - Update Create Team backend to not create a Stripe checkout session for such a team
 - Renamed the `noBilling` property to `requiresBilling` to be more explicit in its meaning
 - Updates the Change Team Type view to show the right messaging for the different states a team can be in

Paths tested manually:
 - New user creates first team at freemium without setting up billing
 - They can create/delete a remote instance
 - They can suspend/resume the team
 - They can upgrade to Starter - setting up billing as part of the process
 - They can downgrade back to Freemium. At this point, they have a Subscription and we will maintain it with the Freemium $0 product line item (ie the existing behaviour).

**Note**: a team (with Subscription) that downgrades to Freemium will still have a subscription on stripe, just at $0. This will mean the number of subscribes to the Stripe product will *not* represent the total number of Freemium teams - it will represent all of the pre-existing teams, plus any that have downgraded. This was a pragmatic choice to keep this work on schedule. If we decide we should be deleting the subscription entirely on downgrade, we'll need to do more development work.

~Remaining Tasks:~
 - ~Manually test some more paths through team lifecycle~
 - ~Add some UT coverage to the backend for this scenario~


knolleary:
@cstns have tagged you to review this one as it is largely frontend work to handle this new billing state a team can be in. The backend changes are very self-contained.

Let me know if you want to step through it with me.

frontend/src/pages/team/Billing.vue
```diff
@@ -51,7 +51,32 @@
...
             </div>
-            <EmptyState v-else-if="!isUnmanaged">
+            <EmptyState v-else-if="billingDisabledForTeam">
+                <template #img>
+                    <img src="../../images/empty-states/team-instances.png">
+                </template>
+                <template #header>Team Billing</template>
+                <template #message>
+                    <p>
+                        Your team does not require billing to be setup.
```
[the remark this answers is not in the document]
cstns: or upgrade their team type 
image
this is only visible after downgrading from a higher tier and a subscription is present, I think the upgrade button should be visible at all times

frontend/src/pages/team/Billing.vue
```diff
@@ -51,7 +51,32 @@
...
+                    </p>
+                    <p>
+                        Please contact <a href="https://flowfuse.com/support/" class="underline" target="_blank">Support</a> for help.
+                    </p>
+                </template>
+            </EmptyState>
+            <EmptyState v-else>
                 <template #img>
                     <img src="../../images/empty-states/team-instances.png">
                 </template>
```
cstns, line 82: For a lack of a better place to mention this, ill do it here.

Two things come to mind after creating a new freemium tier, upgrading to team tier, downgrading to freemium:
- not sure what to make out of my stripe customer portal billing info (could be development inconsistencies and the subscriptions i configured my local teams). What is a `small project`, `team plan and small project` and `team plan`? I only have a single team with properly configure billing details locally
image
- billing wise, how can we apply the credit to the next invoice if the current team plan is free ?
image

frontend/src/pages/team/changeType.vue
```diff
@@ -54,12 +47,12 @@
...
                             <p v-if="trialMode && !trialHasEnded">Setting up billing will bring your free trial to an end</p>
                             <p v-if="!isContactRequired && team.suspended">Setting up billing will unsuspend your team</p>
-                            <p v-if="isTypeChange">Your billing subscription will be updated to reflect the new costs</p>
+                            <p v-if="!isContactRequired"> Your billing subscription will be updated to reflect the new costs</p>
                         </div>
                     </template>
                     <div class="flex gap-x-4">
                         <template v-if="!isContactRequired">
-                            <ff-button v-if="isTypeChange" :disabled="!formValid" data-action="change-team-type" @click="updateTeam()">
+                            <ff-button v-if="!billingEnabled || billingDisabledForSelectedTeamType" :disabled="!formValid" data-action="change-team-type" @click="updateTeam()">
```
cstns, line 55: Why do we need to hide the change their team button? It basically allows users to upgrade their team type

frontend/src/pages/team/Billing.vue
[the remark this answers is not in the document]
cstns: NB: unless enterprise tier

frontend/src/pages/team/Billing.vue
knolleary: > What is a small project, team plan and small project and team plan? I only have a single team with properly configure billing details locally

Those are the stripe product names. The naming in the test catalogue needs updating to match the more explicit and consistent naming we have in the production catalogue.

> billing wise, how can we apply the credit to the next invoice if the current team plan is free ?

The credit will get applied towards their $0 invoice... the credit will only get spent when they have a non-$0 invoice.

frontend/src/pages/team/changeType.vue
knolleary: There are two possible cases:
 - if the team already has billing setup (or billing is not enabled for the platform at all), we show the 'Change team type' button. This will trigger the type change directly.
 - otherwise, we show the 'Setup Payment Details' button (line 58) which takes them via stripe first - and the type change only gets applied on the backend once the stripe checkout completes

knolleary:
UI Test now failing due to the upgrade button being present

cstns:
LGTM, will merge after actions pass

PR #5296, opened 2025-03-11
Fix generated API doc for external MQTT Broker API

hardillb:
## Description

Fixes the comments on each endpoint, docs used to point to `/api/v1/teams/teamId/broker/.....` rather than `/api/v1/teams/:teamId/brokers/...` as it should be.


Steve-Mcl (approved):
PR matches issue as explained 👍

PR #5269, opened 2025-03-07
extract tours into a separate module

cstns:
## Prerequisites

based on https://github.com/FlowFuse/flowfuse/pull/5251

## Description

Extract ux tours into separate module

## Related Issue(s)

n/a

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #5307, opened 2025-03-12
Multiple tours starting at the same time

cstns:
## Prerequisites

based on https://github.com/FlowFuse/flowfuse/pull/5269

## Tour flows

As we currently stand we decide which tour should start primarily based on the team type:
- **freemium team**
  - **after registration**
     - user gets redirected to the application device page
     - the first-device tour starts
   - **manual tour start**
     - user gets redirected to the applications page
     - user gets redirected to the application device page
     - starting the first-device tour
   - **notes**
     - shouldn't we point users to the education modal in this case as well when the tour ends? @gstout52 
     - the double redirect when starting the tour manually is regrettable but at this moment in time we can't avoid it ( somewhat glitch-y ux during the redirects)
     - if the user manually starts the tour but he has no applications present (manually deleted the app) the tour will not start. On the upside, when and if he recreates an application the tour will start when he visits the applications page
     - the user is unable to follow along the first-device tour if device limit has been reached
     - if users close or click cancel on the create remote instance modal they remain in tour limbo, a follow-up should be created in this regard
- **trial team**
   - **after registration**
     - user gets redirected to the applications page
     - the welcome-trial tour starts
     - when the tour ends the education modal is opened
   - **manual tour start**
     - identical to registration
   - **notes**
     - if the user deletes his instance which was created by default upon registration and starts the tour without any instances, the regular welcome tour will start instead
- **any other team type**
   - **after registration**
     - user gets redirected to the applications page
     - the welcome tour starts and if the team doesn't have any devices created, the first-device tour is queued
     - the initial tour ends with the user on the application page
     - if the first device tour was queued, the user can navigate freely within the application but when he reaches the application devices page or remote instances page the first-device tour starts
   - **manual tour start**
     - identical to registration
   - **notes**
     - as in the case of the freemium tier, there is an argument to be made if we should open the education modal at either the welcome tour end or first-device tour (if queued) @gstout52 
     - the education modal also serves as a pointer on where users can open related/helpful docs and restart the tour or the education modal
     - when the first-device tour is queued, there is no visual queue that would indicate that there is a follow-up to the initial tour (highlight the devices tab) and it might come as a surprise when the user opens the devices tab after a long while and the first-device tour starts seemingly out of the blue
     - the user is unable to follow along the first-device tour if device limit has been reached

The steps described above are subject to slightly change or be impacted dependent on the user switching browsers/tabs or logging out

## Description

- fixed education modal bug in which the timer lagged or the modal would close unexpectedly when opened/closed multiple times
- removed/separated the education modal from the tours list (including mutations and actions) allowing greater flexibility to add more education modals in the future if needed
- moved the tour singleton from the team component into the tours module
- detached from the tours.welcome state key as trigger to initialize/present tour to shouldPresentTour
- moved away from the way we were deducing the team is a freemium tier to a more accurate one (introduced in https://github.com/FlowFuse/flowfuse/pull/5237)
- added delays when starting tours for some wiggle room for the required apps/instances/devices to get loaded in the ui and a less aggressive feel
- moved tour creation from components to store
	- allows greater flexibility to start a specific tour from anywhere (useful for the manual tour start) while also letting the team component decide which tour should be started (as before)
	- less moving parts, tours have dedicated store action assigned and tour initialization is done in a single place and can be triggered from anywhere
- unable to move entire tour start to the store to make the functionality decoupled and globally available  
	- deciding which tour should start is dependent among other things on the number of applications, instances or devices a team has
	- can and should be done in the future when we have contextual team data stored globally
- converted the tours into js files to allow for greater flexibility when defining tours and also moved the tour id as an exportable item
- fixes the bug where tours start twice after registration or manual trigger
- fixes the bug where after completing any tour and navigating away from the applications page and back to it restarted the tour

## Related Issue(s)

N/A

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


gstout52:
@cstns The education modal should not fire for freemium users, but should fire for users on any other team type. The reason is that freemium users have access to none of the capabilities offered there, and when paired with an abundance of more valuable upsell opportunities, the additional modal makes for a somewhat clunky ux.

Issue #5301, opened 2025-03-12
Time-based browsing of Instance logs


knolleary:
### Description

We currently show the latest Node-RED logs to the user and they can click 'earlier' to go back in time.

If the user sees an audit-log entry about an unexpected NR event, they may want to check the logs (in fact, some audit log entries tell the user to check the logs).

Clicking 'earlier' a lot is tedious and slow. A user should be able to browse to a given time in the logs and jump straight there.

*Note*: we only keep a fixed buffer size of logs within the launcher that affects how much history is available; meaning a noisy NR instance will have a shorter duration of available logs. That's a separate matter for now.

### Which customers would this be available to

Everyone - CE/Starter/Team/Enterprise

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

hardillb:
As a first pass based on what we currently have in the buffer.

I suggest having the forge app add 2 new entries to the `meta` object returned by the logs api endpoint

- `start`
- `end` 

These hold the ts of the first and last entries in the buffer.

These can then be used to populate a range in the UI to scrub to

@joepavitt Any thoughts on a UI design to allow jumping to a given time? Do we want a slider with start/end?

joepavitt:
<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/b0040ac4-7feb-48c3-9032-e518d71cd1e8" />

Inline (justified to the right) with the "Node-RED Logs" header here, let's just got for two date selection inputs, possibly going to need time too? Not sure he have a Date/Time selection component already in place.

hardillb:
OK, this might prove a little tricky as we already have an optional drop down in that space if HA is enabled to pick the different instances.

Also that space is setup without access to the log data that has the time range (the logs are loaded by a separate component)

https://github.com/FlowFuse/flowfuse/blob/dccf68a3acf801453620a79d379cf699a3a04602/frontend/src/pages/instance/Logs.vue#L3-L14

and 

https://github.com/FlowFuse/flowfuse/blob/main/frontend/src/pages/instance/components/InstanceLogs.vue

hardillb:
I can't seam to find a date + time picker in the ui-components directory.

joepavitt:
I don't believe we have one, not had a need for one until now.

hardillb:
[...]

PR #5334, opened 2025-03-18
Update NR logs API to return time range

hardillb:
part of #5301 
## Description

adds `first_entry` an `last_rntry` which are ms timestamps to the meta entry to the `/api/v1/projects/:instanceId/logs`

```json
{
  "meta": {
    "next_cursor": "17422187899860000",
    "previous_cursor": "-17422187769160000",
    "first_entry": "17422187769110000",
    "last_entry": "17422187899860000"
  },
  "log": [
    {
      "level": "system",
      "msg": "Starting health check monitor  (7.5s)",
      "ts": "17422187770290000",
      "src": "e4fa"
     },
     ...
  ]
}
```

## Related Issue(s)

#5301 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
Ready for review @hardillb?

hardillb:
@joepavitt can be,  need the date picker ui component to do much else.

Need to look at the nr-launcher side to support only sending a sample of logs, but that is part of larger bit of work to store more logs on the instance

hardillb:
@Steve-Mcl can you cast an eye over this please

forge/routes/api-docs.js
```diff
@@ -155,7 +155,8 @@ module.exports = fp(async function (app, opts) {
         properties: {
             next_cursor: { type: 'string' },
             previous_cursor: { type: 'string' }
-        }
+        },
+        additionalProperties: true
```
Steve-Mcl, line 159: Is there a reason the new props `first_entry` and `last_entry` are not declared in the API? Is there a possibility we might lose these in the future?

I wouldnt question the use `additionalProperties` over fixed entries in the API if they were added conditionally, however, in [the diff](https://github.com/FlowFuse/flowfuse/pull/5334/files#diff-37ad4d63a25ead0d5bec5c27c1c3fc36b4c7adf9a7ff31af4b3102dbe6018e12R981-R982) they are always added to `meta` object.

I am happy to admit if I am missing something here (and will defer to your good judgement)

hardillb: Because `PaginationMeta` are used elsewhere and I couldn't work out how to include the existing schema and add in the additional properties in the api declaration.

If you can find a way to merge them without having to duplicate the schema I'm all ears

Steve-Mcl: Since `"required"` is not present in the schema, I am pretty sure the props are optional so adding `"first_entry": { "type": ["number", "null"] }, "last_entry": { "type": ["number", "null"] }` should be fine (I think)

To be explicit, I _think_ we can add empty required array:

Something like: 
```
{
    "$id": "PaginationMeta",
    "type": "object",
    "properties": {
        "next_cursor": { "type": "string" },
        "previous_cursor": { "type": "string" },
        "first_entry": {
            type": ["number", "null"],
            "description": "Optional: The timestamp of the first entry in the result set."
        },
        "last_entry": {
            type": ["number", "null"],
            "description": "Optional: The timestamp of the last entry in the result set."
        }
    },
    "required": []
}
```

hardillb: But the new fields ONLY make sense for the nr logs, not everywhere else `PaginationMeta` is used.

Adding extra fields is just confusing for all the other cases.

Steve-Mcl: Thats why I marked them explicitly optional.

Adding `"additionalProperties": true` to the schema essentially allows any extra properties not explicitly defined in it to be included without validation errors and kinda defeats the purpose. 

IMO, `additionalProperties` should only be used for unknown additions (like a plugin might add). 

But it is not a hill I am willing to die on.

Steve-Mcl: Let me know if you wish to proceed as is and I will approve.

hardillb: I'm made the logs end point use it's own inline schema since I couldn't get any of the `$ref` and extra properties from here https://json-schema.org/blog/posts/modelling-inheritance to work.

@Steve-Mcl please re-review

hardillb:
OK, the defensive changes have broken a UI tests

forge/routes/api-docs.js
Steve-Mcl: > I'm made the logs end point use it's own inline schema since I couldn't get any of the `$ref` and extra properties from here https://json-schema.org/blog/posts/modelling-inheritance to work.

For extra properties, we would need to use `allOf` to merge existing schema with new props

See line 29 here: https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/#allof

I did a little test on your branch and this seems to work nicely...

```js
                        meta: {
                            allOf: [
                                { $ref: 'PaginationMeta' },
                                {
                                    type: 'object',
                                    properties: {
                                        first_entry: { type: 'string' },
                                        last_entry: { type: 'string' }
                                    }
                                }
                            ]
                        },
```

Steve-Mcl: in fact there is a precedence :) 

https://github.com/FlowFuse/flowfuse/blob/6ddb64770080b679dda9013a2a70186ad06e9fe4/forge/routes/api/projectType.js#L22-L31

forge/routes/api/project.js
```diff
@@ -943,6 +943,7 @@ module.exports = async function (app) {
             let logs = await app.containers?.logs(request.project)
             const firstLogCursor = logs.length > 0 ? logs[0].ts : null
             const fullLogLength = logs.length
+            const lastLogCursor = logs.length > 1 ? logs[fullLogLength - 1].ts : null
```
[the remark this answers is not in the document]
hardillb: That test is meaningless, because if `app.containers` is undefined then NOTHING works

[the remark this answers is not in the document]
Steve-Mcl: The defensive suggestion was judged on its immediate locality to that conditional test - but your point is fair. Lets forget that.

hardillb:
@Steve-Mcl this should be good to go again.

I've backed out the defensive changes as they don't make sense when I look at them again (the `app.containers?.` is meaningless).

Steve-Mcl (approved):

PR #5337, opened 2025-03-20
Show Node-RED version for Devices and Snapshots

### LINKED ISSUE #5246: Expose the NR Version for a Snapshot
### Description

In opening https://github.com/FlowFuse/flowfuse/issues/5245 I realised that the version of Node-RED, specified in a Snapshot, is never actually exposed anywhere.

This _may_ require additional API work, not just front-end

I think we should be exposing that detail in a few places:

### Snapshot Viewer

Add the Node-RED version as part of the model in some capacity

Image

### Snapshots Table

As a column in the "Snapshots" table

Image

### Hosted Instance View

For a Hosted Instance, we have the "Specs" section in the overview, but no such view is available for Remote Instances. We should extract the Node-RED version from the deployed Snapshot/flows in order to display that information

Image

None

hardillb:
closes #5246 
## Description

Adds NR version to Snapshot view and Device Overview

Screenshot from 2025-03-20 10-47-21

Screenshot from 2025-03-20 11-00-14
The location of the Node-RED version needs work here, but I just proved it could be done...

image

## Related Issue(s)
#5246

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/dialogs/AssetDetailDialog.vue
+++ b/frontend/src/components/dialogs/AssetDetailDialog.vue
@@ -6,6 +6,7 @@
             </div>
         </template>
         <template #actions>
+            Node-RED {{ nrVersion }}
             <div class="flex justify-end">
                 <ff-button data-action="dialog-confirm" @click="confirm()">Close</ff-button>
             </div>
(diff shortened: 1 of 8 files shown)


frontend/src/components/dialogs/AssetDetailDialog.vue
```diff
@@ -45,6 +49,13 @@ export default {
         flow () {
             return this.payload?.flows?.flows || []
         },
+        nrVersion () {
+            const mods = this.payload?.settings?.modules
+            if (mods) {
+                return mods['node-red'] || 'ben'
```
joepavitt, line 55: Assuming this was a placeholder that wasn't updated, "Unavailable" a better backup?

joepavitt (changes requested):

frontend/src/components/dialogs/AssetDetailDialog.vue
hardillb: yes, pushed update

joepavitt (approved):

Issue #5340, opened 2025-03-20
Add UI to gather topic prefix for MQTT agent


hardillb:
### Epic/Story

#5300

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #5342, opened 2025-03-21
Mqtt agent topic prefix

### LINKED ISSUE #5300: No topics are captured in the Broker Hierarchy for brokers that deny subscription to `#`
### Current Behavior

Some Brokers do not permit subscribing to `#` and therefore the Broker Hierarchy never gets populated with any topics

In particular, EMQX (by default), Azure IoT Hub, AWS IoT Core and many others do not permit subscribing to `#`

### Expected Behavior

We may wish to include a topic filter on the broker configuration. 
We may also want to consider 
- an admin setting that allows/prevents a min top level topic (to avoid hammering the DB if/when someone subscribes to a public broker)
- a deny list to prevent public brokers being added to as a BYOB entry?

### Steps To Reproduce

Try adding an emqx broker (that by default does NOT return anything on `#`) - no topics are populated.

### Environment

- FlowFuse version: FFC Current
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### LINKED ISSUE #5339: Add topic prefix to MQTT Broker credentials DB table
### Description

Allow topic prefix to the stored with credentials

### Epic/Story

#5300

hardillb:
fixes #5339 and #5340 
## Description

Adds field to use as a topic prefix when connecting to a 3rd party broker

## Related Issue(s)

#5300

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
Not to be merged before https://github.com/FlowFuse/mqtt-schema-agent/pull/16 has been merged and a new container released

forge/db/views/BrokerCredentials.js
```diff
@@ -11,7 +11,8 @@ module.exports = function (app) {
             protocolVersion: { type: 'number' },
             ssl: { type: 'boolean' },
             verifySSL: { type: 'boolean' },
-            clientId: { type: 'string' }
+            clientId: { type: 'string' },
+            topicPrefix: { type: 'string' }
```
Steve-Mcl, line 15: I personally think this is too restrictive. 

The original discussion in slack was to support multiple topics since brokers who deny subscription to `#` would mean a user can only subscribe to a single 2nd level topic e.g. `press/#` and thus not able to see topics in `body/#` or  `paint/#`

I have no issue with supporting a single custom topic for the first iteration however by setting the API in stone and to a single `string` prevents an easy upgrade path in the future (without having to modify the API signature)

If you have no issue with redefining the API at a future time, I will approve.

hardillb: > We may wish to include a topic filter on the broker configuration.

> Will scope this to adding a config option in the Broker settings to define the top-level scope that you wish to subscribe to for traffic, defaulting to # (for now)

This meets what was asked for

Steve-Mcl: The issue raised did not reflect the requirement or discussion in slack.  

Your PR of course does meet the issue but that does not detract from the shortcomings.

hardillb:
@Steve-Mcl I've updated, but it will need https://github.com/FlowFuse/mqtt-schema-agent/pull/20 first.

Also since I've changed the migration this will probably break the pre-staging env. Might need @ppawlowski to tear it down and re-create it

Steve-Mcl:
> Might need @ppawlowski to tear it down and re-create it

@ppawlowski could you re-spin this pls?

forge/db/migrations/20250401-01-EE-add-topicPrefix-to-brokercredentials.js
```diff
@@ -0,0 +1,30 @@
...
+                    return ['#']
+                }
+            },
+            set (value) {
+                if (value) {
+                    this.setDataValue(this.topicPrefix, JSON.stringify(value))
+                }
+            }
+        })
+    },
```
Steve-Mcl, line 28:
```suggestion
    /**
     * Add topicPrefix field to BrokerCredentials table
     * @param {QueryInterface} context Sequelize.QueryInterface
     */
    up: async (context) => {
        await context.addColumn('BrokerCredentials', 'topicPrefix', {
            type: DataTypes.TEXT,
            defaultValue: '["#"]',
            allowNull: true
        })
    },
```

get and set are "Model" properties / application level things i.e.. part of the model's logic, not the database structure.

Example:
* DeviceSettings.value migration: https://github.com/FlowFuse/flowfuse/blob/26925ab21990bc6360f3d9360771e289e45f2ce0/forge/db/migrations/20220624-01-add-devicesettings.js#L23-L24
* DeviceSettings.value model: https://github.com/FlowFuse/flowfuse/blob/26925ab21990bc6360f3d9360771e289e45f2ce0/forge/db/models/DeviceSettings.js#L17-L18


Also, jsdoc type hints are very useful here. 

Lastly, I changed the type to `TEXT`.  Many of our other JSON fields are type TEXT and I read somewhere that some databases default to 255 characters for the STRING type?

hardillb:
@Steve-Mcl it only need re-spining if it fails to start and it appears to be running

forge/db/models/BrokerCredentials.js
```diff
@@ -36,6 +36,24 @@ module.exports = {
...
+                const rawValue = this.getDataValue('topicPrefix')
+                if (rawValue) {
+                    return JSON.parse(rawValue)
+                } else {
+                    return ['#']
+                }
+            },
+            set (value) {
+                if (value) {
+                    this.setDataValue(this.topicPrefix, JSON.stringify(value))
```
Steve-Mcl, line 54:
```suggestion
                    this.setDataValue('topicPrefix', JSON.stringify(value))
```

This was trying to update a field named '#'

image

Steve-Mcl (commented):
Couple of little bits and one philosophical question...


Should the Hierarchy not be limited to the topic scope?

As a test, I set the topic wildcard to a lesser scope but ALL topics are still shown...

chrome_rAOEFW2Thg

hardillb:
> Should the Hierarchy not be limited to the topic scope?

No

hardillb:
Also do not ever start a topic with `/`

Steve-Mcl:
> Also do not ever start a topic with `/`

100% agree. But that is not my data (public broker) and we cannot educate everyone (unfortunately ;)

<br>


> > Should the Hierarchy not be limited to the topic scope?
> 
> No

Hmmmm. I think it needs consideration. I'll raise a separate issue after we get this merged.

hardillb:
> > >Should the Hierarchy not be limited to the topic scope?

> >  No

> Hmmmm. I think it needs consideration. I'll raise a separate issue after we get this merged.

No, there is no reason to hide things here, if the user doesn't want the data they can delete it, but hiding it just means it clutters the database for no reason.

Steve-Mcl:
@hardillb 

I guess this PR depends on https://github.com/FlowFuse/mqtt-schema-agent/pull/20 being merged and new containers in place?

Issue #5280, opened 2025-03-10
Improve visibility of "Save Changes" for Instance Settings


joepavitt:
### Description

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/9a3b8dcc-5754-4354-81a0-6d1362029693" />

When making changes to the settings of an Instance, the "save changes" button is hidden by a scroll, and so it's not always clear that a "Save" action is required before navigating away.

This task is to track the movement of the "Save" actions inline with the Settings header on the right-hand side.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #5360, opened 2025-03-25
Improve visibility of save changes of instane settings

cstns:
## Description

not really proud of myself on this one but managed to get the save changes button outside of the instance settings page by emitting the button state to the parent from each nested component and triggering the save method from the parent.

Had to fix the palette button state because it was not comparing original values with edited ones correctly and also replaced the hard-coded side menu routes with named based routes

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5280

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/pages/instance/Settings/index.vue
```diff
@@ -1,17 +1,35 @@
...
+        <SectionSideMenu :options="navigation" />
         <div class="flex-grow">
-            <router-view
-                :project="instance"
-                :instance="instance"
-                @instance-updated="$emit('instance-updated')"
-                @instance-confirm-suspend="$emit('instance-confirm-suspend')"
-                @instance-confirm-delete="$emit('instance-confirm-delete')"
-            />
+            <router-view v-slot="{ Component }">
```
joepavitt, line 21: Talk me through this change please?

joepavitt (changes requested):
I can definitely understand your request to add a mixin in for this, but also appreciate you persisting with this as a first quick pass through.

Some questions and queries to address before we can move forward with this though.

frontend/src/pages/instance/Settings/index.vue
cstns: It's because I needed to place the ref on to the actual child component itself not the surrounding router-view wrapper so that I can call `this.$refs.settingsPage.saveSettings() ` in the onSaveButtonClick method which triggers the saveSettings method on the nested page.

It's not the best approach because I'm coupling the pages together but it saved me the hassle of extracting each child's save method into the parent along their state and by doing it this way it also saves me the hassle of managing state in the parent component as well

frontend/src/pages/team/Brokers/Hierarchy/components/suggestions/TopicSuggestion.vue
```diff
@@ -37,6 +37,7 @@ import Dialog from '../../../../../../services/dialog.js'
 
 import ObjectProperties from '../schema/ObjectProperties.vue'
 
+// eslint-disable-next-line vue/one-component-per-file
```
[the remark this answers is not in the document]
joepavitt: There is only one component in the file though, so it shouldn't be required...

joepavitt, line 40:
```suggestion
```

joepavitt (approved):

Issue #5264, opened 2025-03-07
Improve Feedback on 75% CPU Capacity Breaches


joepavitt:
### Description

More details available in this [Private Issue](https://github.com/FlowFuse/CloudProject/issues/649)

### Friction

- Node-RED instances can fail/crash due to memory issues. We can alert users to these crashes via e-mail, but it's an option that needs to be turned on - _it is not enabled by default_
- It is also detailed in the Node-RED Logs of the relevant Instance, but without users manually inspecting each Instance's logs, then these would not be easily found

### Value

- Upsell opportunity for larger instances
- Clearer feedback to users when they're hitting performance problems

### Which customers would this be available to

Team + Enterprise Tiers (EE)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Proposal - once we get to k8s to 1.30, we have the opportunity to dynamically change capacity, and would provide a much cleaner "serverless" experience

Steve-Mcl:
Remaining item [5265](https://github.com/FlowFuse/flowfuse/issues/5265) was moved to the backlog

REF: https://github.com/FlowFuse/flowfuse/issues/5265#issuecomment-2766749812

PR #5363, opened 2025-03-27
Enable resource breach alerts by default in templates and instances

### LINKED ISSUE #5267: Enable e-mail notifications for 75% CPU breach by default
### Description

Users currently need to manually enable the e-mail alerts for Node-RED instances that breach 75% CPU usage, this should be enabled by default to alert users to performance/scaling issues.

### Epic/Story

_No response_

Steve-Mcl:
closes #5267 
## Description

- Migration to set all templates email alerts for resources (cpu + memory) where the value is not currently set and the policy is NOT explicitly `false` (i.e. already set by a user)
- Assumes a missing value AND email alerts feature enabled as enabled. If user sets explicitly sets these alerts, they will use the set value
- Defaults the values to true for the UI to represent the default state  

## Related Issue(s)

owner: #5267
parent: #5264

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/migrations/20250327-01-EE-update-projecttemplates-settings-email-alerts.js
+++ b/forge/db/migrations/20250327-01-EE-update-projecttemplates-settings-email-alerts.js
@@ -1,0 +1,47 @@
+/* eslint-disable no-unused-vars */
+
+/*
+ * Update the JSON string in `ProjectTemplates`.`settings` so that emailAlerts.resource.cpu and emailAlerts.resource.memory are set to true
+ */
+const { DataTypes, QueryInterface } = require('sequelize')
+
+module.exports = {
+    /**
+     * upgrade database
+     * @param {QueryInterface} context Sequelize.QueryInterface
+     */
+    up: async (context) => {
+        // The `type` option is explicitly set to `context.sequelize.QueryTypes.SELECT` to ensure that the query is executed as a SELECT statement.
+        // This ensures the results are always returned as an array of objects.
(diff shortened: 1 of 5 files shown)


joepavitt:
@Steve-Mcl is this something @hardillb could take a look at given that Nick is out this week at Messe?

Steve-Mcl:
> @Steve-Mcl is this something @hardillb could take a look at given that Nick is out this week at Messe?

Sure, but since it involves a DB migration to existing FFC project templates etc, I was keen to get @knolleary eyes on it.

Up to you Joe.

knolleary:
Can you summarise what the migration is doing? Why are we changing existing settings rather than defaulting in code for otherwise undefined value?

Steve-Mcl:
> Can you summarise what the migration is doing?

It updates any project templates  policies & setting that have this far not been set to enable them as per request in #5267

> Why are we changing existing settings rather than defaulting in code for otherwise undefined value?

So that future created instances adhere to the settings applied in the template policy.

hardillb:
Do we need to separate out 2 things here.

1. adding the feature in a default off state
2. a FFC requirement for the feature to be on by default

I think default off for most installs would make more sense than a globally applied migration to turn this on for all existing instances

Steve-Mcl:
> 1. adding the feature in a default off state

Feature is already merged / operational Ben. Default state is "off" currently.  This PR is in response to the issue asking for it to be on by default see: https://github.com/FlowFuse/flowfuse/issues/5267#issuecomment-2752556827

hardillb:
Then I think this should be a change to FFC only, not to all FF instances.

Steve-Mcl:
After some discussion and consideration, I have found a suitable way to default the settings in code while making sure to respect the projects template policy setting (if any) & I have now deleted the migration.

PR #5367, opened 2025-03-28
Add Git Integration to Pipelines

### LINKED ISSUE #5297: PoC: Git Integration with FlowFuse
### Description

First PoC scope will be to have FlowFuse writing to a Git repository on-deploy or new Snapshot being generated

#### Acceptance Criteria

- [ ] Create a Pipeline
- [ ] Select an Instance for Stage 1
- [ ] Choose a "Git Repo" action type for Stage 2
- [ ] Push from Stage 1 to Stage 2
- [ ] Snapshot should be available within the relevant repository, configured in Stage 2

#### Expected Challenges

- Authentication

### Epic/Story

_No response_

knolleary:
Closes #5297 

This adds the ability to push snapshots to a GitHub.com hosted repository via a pipeline stage.

I'm raising this as draft for the moment - it's functionality complete (I believe), but needs a lot of tests writing and this description expanding with more specific details to aid review.


knolleary:
@hardillb I believe this is functionally complete, albeit missing tests. I'll expand on the description next chance I get (out of time today) - but basically:
1. enabled teamtype feature
2. create a git repo on github.com (can be private) - ensure the branch exists (ie add a readme)
3. create GitHub PAT with *Content* permission on that repo
4. add pat to team settings/integrations view
5. Create pipeline Instance->Git Repo - configure the git repo with your repo details (wont say more... let me know if the form is clear)
6. Hit deploy... and see the magic happen

forge/ee/lib/gitops/index.js
```diff
@@ -0,0 +1,140 @@
...
+                const result = new Error('Invalid git token')
+                result.code = 'invalid_token'
+                result.cause = err
+                throw result
+            }
+
+            const userGitName = userDetails.data.login
+            const userGitEmail = `${userDetails.data.id}+${userDetails.data.login}@users.noreply.github.com`
+            const author = `${userGitName} <${userGitEmail}>`.replace(/"/g, '\\"')
+            workingDir = await fs.mkdtemp(path.join(os.tmpdir(), 'flowfuse-git-repo-'))
```
hardillb, line 58: I thought we were going to run this in a directory of our choosing for the containers, not the default OS temp dir?

e.g. under the same `var` directory?

hardillb:
Tested, first impressions looks good with following niggles.

- an empty repo has no branches so can't push to a totally empty repo. Had to push an empty README.md to make it work
- editing the stage shows the credentials secret empty (should return `__placeholder__` or similar)
 
DB migrations will need their names changing.


Also where is it generating the files?

forge/ee/lib/gitops/index.js
knolleary: This was a placeholder to get something done. Will add a config option for this (but use this as the fall back if not set)

knolleary:
> an empty repo has no branches so can't push to a totally empty repo. Had to push an empty README.md to make it work

Yes - the remote branch needs to exist. This is a 1st iteration restriction.

> editing the stage shows the credentials secret empty (should return __placeholder__ or similar)

Will address

forge/ee/lib/gitops/index.js
knolleary: As discussed, we believe this is good as-is. It will default to `/tmp/` but can be overridden by `TMPDIR`, `TEMP`, `TMP` env vars (in that order)

forge/ee/lib/gitops/index.js
```diff
@@ -0,0 +1,140 @@
...
+            } catch (err) {
+                const output = err.stdout + err.stderr
+                // Token does not have access to clone the repo
+                if (/unable to access/.test(output)) {
+                    const result = new Error('Permission denied')
+                    result.code = 'invalid_token'
+                    result.cause = err
+                    throw result
+                }
+                // Remove branch does not exist
```
hardillb, line 72:
```suggestion
                // Remote branch does not exist
```

forge/ee/lib/gitops/index.js
hardillb, line 58: Should this have the trailing `-` or should there be something else in the join to prevent clashes when if multiple pipelines run at once? e.g. add the pipeline stage hash id to the end?

forge/ee/routes/gitops/index.js
```diff
@@ -0,0 +1,119 @@
...
+                reply.code(400).send({ code: 'invalid_request', error: 'Missing required fields' })
+                return
+            }
+            const token = await app.db.models.GitToken.create({
+                name: body.name,
+                token: body.token,
+                TeamId: request.team.id
+            })
+            // TODO: audit log
+            // await app.auditLog.Project.project.httpToken.created(request.session.User, null, request.project, body)
```
hardillb, line 54: I assume still outstanding adding auditlog? I don't see any frontend code for rendering them yet

hardillb (commented):
A real quick first pass reading the code.

Just the directory name that I think I missed first time because I was wondering about the os.tmp stuff.

and audit log question.

I will test it again in a moment

hardillb:
Tested locally and all appears to work fine.

Will test in pre-staging env just to double check the /tmp stuff

hardillb:
Testing on pre-staging

- The Team-Settings -> integration is showing the "please upgrade" banner on a team without it, but the request is still showing a 404 not found toast warning.
image
- `/tmp` is read-only on the cluster, will need to work out how to mount a volume, as we make the whole container read only these days. `EROFS: read-only file system, mkdtemp '/tmp/flowfuse-git-repo-XXXXXX'`
image

hardillb:
I know how to fix the readonly `/tmp` will need to mount an empty directory volume on it

@ppawlowski can you look at adding a empty directory mount to the `deployment.yml` mounting it on `/tmp` for the forge container please

forge/ee/lib/gitops/index.js
knolleary: `fs.mkdtemp` generates a random path by appending six random chars to the provided path - [ref](https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options)

hardillb:
OK, we now have a writable volume, but the base image doesn't have git yet.

knolleary:
@hardillb but at least the error reporting worked... ;)

forge/ee/routes/gitops/index.js
knolleary: Yes - outstanding. Will see if I can get it in tomorrow morning

.github/workflows/branch-deploy.yaml
```diff
@@ -390,7 +390,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           repository: 'FlowFuse/helm'
-          ref: 'main'
+          ref: 'feat-foge-tmp-dir'
```
hardillb, line 393:
```suggestion
          ref: 'main'
```

hardillb:
I have reverted the change that need to be removed as the branch it pointed to was merged to the helm chart (and deleted). This is why the pre-staging env failed to deploy.

Will test once pre-staging updated

knolleary:
Have updated some in-app docs on how to add tokens, and added a check for the presence of the `git` command line tool - with logging if it isn't found.

knolleary:
Working on some user docs updates now, but then I think this is good to go

forge/ee/lib/gitops/index.js
```diff
@@ -8,6 +8,15 @@ const execPromised = promisify(exec)
...
+    try {
+        await execPromised('git --version')
+    } catch (err) {
+        // Error running git
+        app.log.warn('Git integration is disabled: git command not found')
+        return
+    }
+
     // Set the git feature flag
     app.config.features.register('gitIntegration', true, true)
```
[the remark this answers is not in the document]
knolleary: Note the 'return' in the catch - it doesn't enable the feature.

hardillb:
Tested the token polling and looks good.

knolleary:
Docs updated. Think this is ready for final review and merge

hardillb (approved):

PR #5371, opened 2025-03-31
Nr logs jump to time

hardillb:
## Description
Allow jumping to log timestamp

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
AS discussed, I have added some repeat logging to instance "plant-live-02" in team "Plant Enterprise" on pre-staging & I will operationally check this over tomorrow.

hardillb:
@Steve-Mcl 

1. You can only pick times in the actual range of the logs, it won't let you pick before the start of the log.
2. It looks like we can pass a format, but it has to be passed in (https://vue3datepicker.com/props/localization/) need to work out how to get the browsers current locale

Steve-Mcl:
> 1. You can only pick times in the actual range of the logs, it won't let you pick before the start of the log.

I did. It just doesnt have a default date set so it just "does nothing".
It is just an unintuitive date time picker IMO. 
Even if I chose a time stamp that definitely exists (35min) it doesn't select.
The issue is the "date" has to be clicked first.
chrome_10tAkUgbh7




TBH, I am not keen on this control at all! It doesnt use native date/time picker (this becomes more obvious on mobile devices)

The native HTML date-time picker is pretty easy and supports native/mobile too:  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#setting_maximum_and_minimum_dates_and_times

hardillb:
I can't get the `locale` property to do anything with this component, the docs say it should take a string in the format `en-US` or `en-GB` but it I just can't get it to switch the day/month order.

As Steve has mentioned this clashes with how it is displayed elsewhere on the page.

hardillb:
@cstns Can we get together tomorrow morning (Wed 9th) when you are back to see if we can get the formatting right (and try and fix this into single date/time mode, no ranges)

Steve-Mcl (approved):

PR #5374, opened 2025-04-02
Prompt users with "Restart now" dialog when making settings changes

### LINKED ISSUE #5282: Be more explicit about "Restart Required" for settings changes
### Description

When changing Instance settings, a restart is then required for any changes to take effect. We prompt this via an Alert in the UI, but it could be easily dismissed/ignored.

As a result, when settings are saved, we should offer a quick dialog option to prompt the user whether they'd like to restart the Instance or not.

joepavitt:
## Description

- Uses the `Dialog` service, instead of alerts to prompt a user to restart their Instance, when making changes to the Instance settings.
- Need to add/update E2E coverage

## Related Issue(s)

Closes #5282

### DIFF
--- a/frontend/src/layouts/Platform.vue
+++ b/frontend/src/layouts/Platform.vue
@@ -22,7 +22,7 @@
 
             <interview-popup v-if="interview?.enabled" :flag="interview.flag" :payload="interview.payload" />
 
-            <ff-dialog ref="dialog" data-el="platform-dialog" :header="dialog.header" :kind="dialog.kind" :disable-primary="dialog.disablePrimary" :confirm-label="dialog.confirmLabel" :canBeCanceled="dialog.canBeCanceled" @cancel="clearDialog(true)" @confirm="dialog.onConfirm">
+            <ff-dialog ref="dialog" data-el="platform-dialog" :header="dialog.header" :kind="dialog.kind" :disable-primary="dialog.disablePrimary" :confirm-label="dialog.confirmLabel" :cancel-label="dialog.cancelLabel" :canBeCanceled="dialog.canBeCanceled" @cancel="clearDialog(true)" @confirm="dialog.onConfirm">
                 <template v-if="dialog.textLines">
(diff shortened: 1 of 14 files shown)


Steve-Mcl (changes requested):
Dialog is shown even when instance is suspended.

image


If we insist on showing this dialog when an instance is not running, we should at least change the wording so as not to confuse users.

joepavitt:
Fair point Steve - I shall make it so

Steve-Mcl (approved):
Good stuff

Steve-Mcl:
Approved @joepavitt but there is a _potentially unrelated_ E2E test fail...

```
  FlowForge - Instance - Settings Environment
    1) can add, remove mark as hidden instance environment variables


  0 passing (8s)
  1 failing

  1) FlowForge - Instance - Settings Environment
       can add, remove mark as hidden instance environment variables:
     AssertionError: Timed out retrying after 4000ms: expected '<div.ff-dialog-container.ff-dialog-container--closed>' to be 'visible'

This element `<div.ff-dialog-container.ff-dialog-container--closed>` is not visible because it has CSS property: `display: none`
      at Context.eval (webpack:///../test/e2e/frontend/cypress/tests/instances/settings/environment.spec.js:65:46)




  (Results)

  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ Tests:        1                                                                                │
  │ Passing:      0                                                                                │
  │ Failing:      1                                                                                │
  │ Pending:      0                                                                                │
  │ Skipped:      0                                                                                │
  │ Screenshots:  1                                                                                │
  │ Video:        false                                                                            │
  │ Duration:     7 seconds                                                                        │
  │ Spec Ran:     instances/settings/environment.spec.js                                           │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
```

joepavitt:
Thanks for flagging @Steve-Mcl - fixed the test, and it was caused by my changes

PR #5377, opened 2025-04-02
Add tabbed view for install instructions per OS

### LINKED ISSUE #5278: OS Specific Agent Instructions
### Description

When add a new Remote Instance, we are presented with instructions for how to install the Device Agent onto the relevant machine. We have tried to provide generic instructions here, but we are missing components like:

1. `sudo` requirement normally required when doing a global NPM install.

The exact command here however differs from OS-to-OS, and as such, we should present the option to choose with Operating System a user plans to install the Device Agent on, and consequently show them appropriate instructions.

joepavitt:
## Description

Screenshot 2025-04-02 at 13 58 10

Specific OS instructions still TBC, but this puts the base UI framework in place to switch between those instructions

## Related Issue(s)

Closes #5278


Steve-Mcl:
> Thoughts on the `sudo` requirements?

I dont know Mac but I have experienced the need for `sudo` on ubuntu and PI but I cannot vouch for other *nix OSes

@hardillb could you take a gander please?
In particular, the linux tab looks like this (with sudo specified)

image

Steve-Mcl:
Another pain point we discussed briefly on our Friday meeting was there is no mention of `/opt` being a restricted directory - where the OTC command fails due to permissions.

We (me and ZJ) seen a user tripped up first hand with a MAC user at the ProveIt conf.
We seen it with @ppawlowski on a Linux box in our Friday session

should we call that out? Or is it for a later round of iteration (i.e. move the default installation dir to somewhere not affected by this (for example) `~/.flowfuse-device`)

hardillb:
if `-g` is in the npm command it need to be run with `sudo`

frontend/src/pages/team/Devices/dialogs/DeviceCredentialsDialog.vue
```diff
@@ -14,13 +14,53 @@
...
+                                                <template #icon-right><ClipboardCopyIcon /></template>
+                                                <span class="">Copy</span>
+                                            </ff-button>
+                                        </div>
+                                    </template>
+                                    <template v-if="selectedOS === 'MacOS'">
+                                        <p>Open Terminal and run:</p>
+                                        <pre class="overflow-auto text-xs font-light p-4 my-2 border rounded bg-gray-800 text-gray-200">npm install -g @flowfuse/device-agent</pre>
+                                        <div class="flex flex-row justify-end space-x-2 -mt-1">
+                                            <ff-button kind="tertiary" size="small" @click="copy('npm install -g @flowfuse/device-agent')">
```
Steve-Mcl, line 46:
```suggestion
                                        <pre class="overflow-auto text-xs font-light p-4 my-2 border rounded bg-gray-800 text-gray-200">sudo npm install -g @flowfuse/device-agent</pre>
                                        <div class="flex flex-row justify-end space-x-2 -mt-1">
                                            <ff-button kind="tertiary" size="small" @click="copy('sudo npm install -g @flowfuse/device-agent')">
```
Joe, from my reading (and [our docs](https://flowfuse.com/docs/device-agent/install/)) it seems MacOS also needs `sudo`

Please check and apply if true.

Steve-Mcl (approved):
Pre-approved (pending tests)

PR #5389, opened 2025-04-03
Block the 3rd party broker access to test.mosquitto.org

hardillb:
## Description

Prevent subscribing to `#` on `test.mosquitto.org` as not to load the 166000 topics

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/team/Brokers/components/BrokerForm.vue
+++ b/frontend/src/pages/team/Brokers/components/BrokerForm.vue
@@ -220,14 +220,23 @@ export default {
             if (this.form.host.length === 0) {
                 return false
             }
+            if (this.form.host === 'test.mosquitto.org') {
+                return false
+            }
 
             return true
         },
         formErrors () {
-            return {
-                name: this.form.name.length ? '' : 'Name is mandatory',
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
Glad to see this. It partly covers my concerns raised on slack: https://flowforgeworkspace.slack.com/archives/C032Q63FGG1/p1741721144396989

In my opinion, there is more we can/should do. 

NOTES:
1. This does not stop someone poking the value into the API (this can be a follow up)
2. This does not handle any existing entries set for mosquitto (this might be a follow up)
3. This does not permit users connecting to other heavy public brokers (this can be a follow up)
4. We might wish to consider a global topic level limit (discussed in the linked slack thread) (this can be a follow up)

Steve-Mcl (approved):
Approved. Notes / limitations for follow up listed in the PR thread.

Will not merge until Nick has reviewed.

hardillb:
> This does not stop someone poking the value into the API (this can be a follow up)
This is very unlikely, this is a tactical response to a current issue.

> This does not handle any existing entries set for mosquitto (this might be a follow up)

There is only one and I've already disabled it

PR #5437, opened 2025-04-16
Throttle notifications sent to Device Group members

knolleary:
## Description

When a large DeviceGroup is deployed to in a pipeline, we are sending the notifications out to all group members at the same time. Those devices all then call home to pull down their snapshot at the same time. This can cause an unexpected loaded on the system if the group is large.

To mitigate this behaviour, this PR throttles the notifications to devices in the group - sending them in batches of 10 with a 5s delay between each batch. This balances the timeliness of deployment against the triggered load on the system.

We are setting up a benchmark test for this in staging to see if we need to tune the batch size/delay numbers.

### DIFF
--- a/forge/ee/db/controllers/DeviceGroup.js
+++ b/forge/ee/db/controllers/DeviceGroup.js
@@ -132,7 +132,8 @@ module.exports = {
         await deviceGroup.reload()
 
         if (changed) {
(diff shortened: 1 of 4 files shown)


hardillb (commented):
Quick read of the code looks good.

Do we want to try and test it with a device group on pre-staging?

hardillb (approved):

PR #5448, opened 2025-04-17
Ensure all devices are loaded when working with device group membership

knolleary:
The device group view was loading the default page of devices, which maxes out at 100. This meant only the first 100 devices were visible for selection into a group.

This fixes things by ensuring it loads all devices.


Steve-Mcl (approved):
Nick has done local testing and code review is fine. Approved.

Issue #5373, opened 2025-04-02
E2E Tests for Create Instance via Application


joepavitt:
### Description

We have two workflows for adding a Hosted Instance. We have extensive E2E's in place for adding an instance via the "Instances" page, but not when in the "Application" view. This meant an issue was missed when developing https://github.com/FlowFuse/flowfuse/pull/5364, and only caught in manual review.

E2E coverage nees to be improved to cover this space.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
@cstns 12 days for a medium piece of work? Can't see how you need to be doing this full time for the next week and a half?

joepavitt:
Still 7 full development days in that time window? Is there a breakdown you can offer please to split into issues/PRs. Can't see how adding E2E's for the Applications view is 7 days of full work, and a single PR.

cstns:
While working on this I also implemented the multi-step form on the application creation page/flow which also took ~1.5 days, haven't created distinct issue for it just yet.

This is what I have written down so far in terms of test cases
- create an application 
    [w or wo additional instance]
        [w or wo blueprints]

- create an instance from an application (with the application pre-selected)
    [w or wo blueprints]

- create instance
    - with applications present
            [ w or wo blueprints]
    - without applications present
            [ w or wo blueprints]

- deploying blueprints
    [w or wo applications]

Things that I might have missed like instance type, template or node-red version pre-selection which I'm sure I'ved missed porting over 1:1 can be handled on a later date.

What complicates things is that it's a multi-step form, with a two-part action in some cases—one creating the application and the other creating the instance, independently of each other. I already mentioned that I found some pre-existing niche use cases on this end, which I have addressed.

This is also coupled with the fact that different team tiers require different outcomes, so all test scenarios should be run from the context of, if not each tier, then at least a few of them.

I'll try to break it up into small pr's

PR #5455, opened 2025-04-23
Handle instance creation failure garacefully when creating new applications 

cstns:
## Prerequisites

Depends on https://github.com/FlowFuse/flowfuse/pull/5438 

## Description

Fixes a bug that duplicated applications when the application form failed to create an instance using the old form and gracefully handles this scenario using the multi-step-form.

### How to reproduce the initial bug (existent on the old form):
- access the application creation form with instance creation ticked/enabled
- move on to the instance step and fill in the instance name field with a name that's already use d by another instance (causing the form to crash)
- prompted with the fact that the instance name already in use, change the instance name with something unique and create the application
- you'll end up with two applications, only one of them containing the instance with unique name

### Behavior of the multi-step form after #5438 
- after the initial attempt of creating an application with an instance that had a duplicate name, you'd get redirected to the application page and the instance wouldn't have been created
- an error alert would appear with the message that instance name already in use

### Fixed behavior
- after the initial attempt of creating an application with an instance that had a duplicate name, the application would get created and the user will get redirected to the instance step with the instance name marked as having an error
- users can navigate back to the application step but only have access to the newly created application
- users can fix the instance name and follow on with the application creation form
- when completing the entire form, users will be left on the newly created application page with the newly created instance

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5373

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/multi-step-forms/instance/MultiStepApplicationsInstanceForm.vue
+++ b/frontend/src/components/multi-step-forms/instance/MultiStepApplicationsInstanceForm.vue
@@ -52,24 +52,23 @@ export default {
             default: 'Create Instance'
         }
     },
-    emits: ['form-success-application', 'form-success-instance', 'previous-step-state-changed', 'next-step-state-changed', 'next-step-label-changed'],
+    emits: ['form-success', 'previous-step-state-changed', 'next-step-state-changed', 'next-step-label-changed'],
     data () {
         const startingStep = 0
 
         return {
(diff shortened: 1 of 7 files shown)


joepavitt (approved):
Not merging until the underlying PR is merged

Issue #5467, opened 2025-04-25
Support for uploading an imported snapshot upon device registration


Steve-Mcl:
### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Issue #5366, opened 2025-03-28
Device Agent to scan for local Node-REDs and auto-import


joepavitt:
### Current Migration Path

- Install FlowFuse Device Agent
- Run Device Agent
- Clashes with my existing NR instance
- Kill NR instance
- Run Device Agent
- Go back to FF UI and Enable Developer Mode
- Open Editor
- Realise I can't import my flows easily
- kill Device Agent
- Run Node-RED locally again
- Copy all flows
- Kill Node-RED
- Run Device-Agent
- Go back to FlowFuse UI
- Open Editor
- Paste in Flows
- Deploy
- Create Snapshot

### Ideal Scenario

- Run the Device Agent in a "scan" mode (or something like that?)
- Auto detection of flows, and imports

hardillb:
Suggestion:

- A command line option for the device-agent that you pass a Node-RED userDir path
- If there isn't an existing project for the device agent it will import the flows.json/flows_cred.json and the credentials secret

This makes is a concious choice to import.

joepavitt:
That could work well. Although, we'd still have the port clash? But less of an issue if you don't need to have the original Node-RED running I guess?

For the options, the vast majority of cases - it'll be in `~./node-red`, no? Why not make `migrate` an option or `scan`, something like that. With the extra option of `userDir` for those that aren't using default options?

hardillb:
Yep, no need for the old version to be running so no port clash.

Having it default to current users home dir + `.node-red` would be a valid default with an optional path argument.

hardillb:
Having thought about this while looking at the device agent for the auto start function, this will need some more padding out.

Is only going to apply to an Application bound device agent, but at the point of starting the device agent it doesn't know how it's bound or if it has an existing snapshot to be deployed.

The device agent doesn't even install Node-RED until it has it's first check in and is told to either use the provided snapshot or an empty one. at which point it will install Node-RED and then either download and run the snapshot or deploy a basically empty flow.

It's at the empty flow stage that it should import a provided existing Node-RED flow.

Steve-Mcl:
> Having thought about this while looking at the device agent for the auto start function, this will need some more padding out.
> 
> Is only going to apply to an Application bound device agent, but at the point of starting the device agent it doesn't know how it's bound or if it has an existing snapshot to be deployed.
> 
[...]

PR #5473, opened 2025-04-28
Import snapshot during OTC setup

Steve-Mcl:
## Description

* Supports a provisioning object in the status update
* Logs to audit (necessary for timeline)
* Returns `meta` object with FF version in `POST` `/api/v1/devices` (device creation) so that the device agent has knowledge that the backend supports snapshot import during OTC

### Implementation Detail
1. A read of the difficulties in the [parent issue comment](https://github.com/FlowFuse/flowfuse/issues/5366#issuecomment-2830181199) should provide a good heads up prior to review
2. Order of events:
   1. device "quick connect" occurs as before - but endpoint returns `meta` with `ffVersion` to let the agent know it supports snapshot import
   2. device agent prompts user to select flow. If they are chose, the POST `state` endpoint is called with `provisioning` state and `provisioning` object containing the necessary data to build a snapshot and apply it as target
3. The current snapshot endpoint in `deviceSnapshot.js` only supports creation of a snapshot (it does this by instructing an upload via MQTT). Since the device already has a means of posting state (uses its token for checking in state), I reasoned the logic for importing the flows/modules/creds could use the POST `state` route to simplify what would be a rather tricky async affair and chicken/egg issue at the Device Agent (to get MQTT up and operating). So in all, this seemed like the right place for implementation. It also permits us to update the `state` of the device while generating the snapshot/apply target in same single call.  If this is considered undesirable, I could break it up into separate calls and update the `deviceSnapshot.js` to also accept the device config (snapshot data) thus bypassing the tricky MQTT route?
For that reason, this PR should be reviewed first (in case there are changes required)

#### `test/unit/forge/routes/api/device_spec.js`
```
  Device API
    Device Checkin
      ✔ device checks in with provisioning snapshot for import
```

#### `test/unit/forge/db/controllers/ProjectSnapshot_spec.js`
```
  ProjectSnapshot controller
    createSnapshot (device)
      ✔ creates a snapshot of a device passed in via options 
```

## Related Issue(s)

Owner: #5467
Parent: #5366

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


docs/device-agent/running.md
```diff
@@ -63,8 +63,10 @@ Web UI Options
 
 Setup command
 
-  -o, --otc string   Setup device using a one time code
-  -u, --ff-url url   URL of FlowFuse. Required for setup
+  -o, --otc string         Setup device using a one time code
+  -s, --otc-dont-start     Do not start the agent after setup
+  -n, --otc-dont-import    Do not ask to import Node-RED flows during setup
```
knolleary, line 68:
```suggestion
  -s, --otc-no-start     Do not start the agent after setup
  -n, --otc-no-import    Do not ask to import Node-RED flows during setup
```

I may be off, but I think the pattern of `no-XYZ` is more common for negative flags like this.

forge/routes/api/deviceLive.js
```diff
@@ -35,8 +35,44 @@ module.exports = async function (app) {
...
+                    const snapShot = await app.db.controllers.ProjectSnapshot.createDeviceSnapshot(
+                        device.Application,
+                        device,
+                        { id: null }, // state is transmitted from the device, so we don't have a user id
+                        provisionedSnapshotOptions
+                    )
+                    device.set('targetSnapshotId', snapShot.id)
+                    await device.save()
+                    // deliberate 6s delay for effect
+                    await new Promise(resolve => setTimeout(resolve, 6000))
```
knolleary, line 63: What effect is this for? 6 seconds feels like a lifetime when you're waiting for something happen on the device. Is it necessary or can it be shortened?

knolleary (commented):
On proposed change on the cli args (which will impact the device agent PR) - and a question about an artificial delay in the code...

docs/device-agent/running.md
knolleary: Have no spotted the `--dont-start` flag is not new - just missing from the docs... but I think we're still in the window of opportunity to change it if we want.

forge/routes/api/deviceLive.js
Steve-Mcl: ah, that was a testing/simulating realworld slow reply! (slipped through)

docs/device-agent/running.md
Steve-Mcl: I agree with proposal & will make the changes to both (and alter it in FF PR too)

Steve-Mcl:
@knolleary both this PR and the agent PR are now updated.

knolleary (approved):

Issue #5488, opened 2025-05-01
Combox does not re-select existing item and changes selection when clicking away


Steve-Mcl:
### Current Behavior

1. combo does not highlight/select the current selection
2. when navigating away (to dismiss the dropdown), it always selects the first item in the list

![Image](https://github.com/user-attachments/assets/e44135bb-0133-4a8d-8039-3c763d343c66)

### Expected Behavior

1. combo should highlight the current selection
2. if no selection is made, it should not change to the 1st item is listed

### Steps To Reproduce

click the dropdown caret
click away

### Environment

- FlowFuse version: FFC Current / Live prod
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #5498, opened 2025-05-02
Fix combobox selector behavior when losing focus

cstns:
## Description

From the combobox [docs](https://headlessui.com/v1/vue/combobox#binding-objects-as-values):
> When binding objects as values, it's important to make sure that you use the same instance of the object as both the value of the Combobox as well as the corresponding ComboboxOption, otherwise they will fail to be equal and cause the combobox to behave incorrectly.

I was not careful when reading the docs and writing the `compareOptions` method used by the combobox to compare options.

Because it couldn't figure out if it had a selected value it auto selected whichever option was active at the moment of closing, which was always the first one in the list if you wouldn't interact with the keyboard.

Tested locally, it should be able to correctly deduce the selected option from a list of objects or a list of primitives.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5488

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/ui-components/components/form/ComboBox.vue
+++ b/frontend/src/ui-components/components/form/ComboBox.vue
@@ -120,7 +120,7 @@ export default {
                     }
 
                     return opt === this.modelValue
-                }) ?? null
+                })
             },
             set (val) {
                 let payload = val?.[this.valueKey] ?? val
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
Bah, still happening

chrome_T8uv5RbhEn

Issue #5351, opened 2025-03-24
Multi Step Instance Form - instance duplication step


cstns:
### Description

Create an duplicate instance step and replace the existing instance form with the new multi-step format

### Which customers would this be available to

all

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #5504, opened 2025-05-06
Fixed charges table on the instance duplication overview step

cstns:
## Description

The charges table were using decorated instance types which were a big chunk of logic which I extracted into a composable to use it in both the instance and duplication steps in the multi-forms.

Also fixed the width of the charge table on the instance step

before:
image

after:
image

## Related Issue(s)

part of https://github.com/FlowFuse/flowfuse/issues/5351

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/multi-step-forms/instance/steps/DuplicationStep.vue
+++ b/frontend/src/components/multi-step-forms/instance/steps/DuplicationStep.vue
@@ -227,9 +227,7 @@ export default {
             const instanceTypes = await instanceTypesApi.getInstanceTypes()
             const { decorateInstanceTypes } = useInstanceFormHelper()
 
-            decorateInstanceTypes(instanceTypes.types ?? [])
-
-            this.instanceTypes = instanceTypes.types ?? []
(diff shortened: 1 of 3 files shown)


frontend/src/components/multi-step-forms/instance/steps/DuplicationStep.vue
```diff
@@ -222,6 +225,9 @@ export default {
     methods: {
         async getInstanceTypes () {
             const instanceTypes = await instanceTypesApi.getInstanceTypes()
+            const { decorateInstanceTypes } = useInstanceFormHelper()
+
+            decorateInstanceTypes(instanceTypes.types ?? [])
```
joepavitt, line 230: You both modify the input variable, and return a copy of it, which makes this code mis-leading. In duplication you just run the function, in the Instance step, you create a new assignment. We need to pick a lane (my preference is generally to return a new value, rather than modify the input)

hardillb:
Running locally I get the expected output

PR #5506, opened 2025-05-06
Alter the duplicate instance page breadcrumbs

cstns:
## Prerequisites
child of https://github.com/FlowFuse/flowfuse/pull/5499

## Description

Altered the duplicate instance breadcrumbs to align with the rest of the app, needed to extract instance fetching data outside the form in order to not pass data upwards.

This will fit in nicely with a next step that will allow the page to cold mount properly.

## Related Issue(s)

part of

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #5507, opened 2025-05-06
Fix Instance duplication form cold starts

cstns:
## Prerequisites

child of https://github.com/FlowFuse/flowfuse/pull/5506

## Description

The multi-step duplicate instance form crashes when accessing it directly via url because the team is not loaded, this addresses that.

## Related Issue(s)

part of https://github.com/FlowFuse/flowfuse/issues/5351

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/multi-step-forms/instance/steps/DuplicationStep.vue
+++ b/frontend/src/components/multi-step-forms/instance/steps/DuplicationStep.vue
@@ -152,6 +152,7 @@ export default {
                 envVars: 'all'
             },
             instanceTemplates: [],
+            instanceTypes: [],
             loading: true,
             nodeRedVersions: [],
             subscription: null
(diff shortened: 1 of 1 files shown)


joepavitt:
I wasn't able to reproduce the cold start bug. If I navigate directly to the breadcrumb, it loads fine?

cstns:
Here's how I reproduce it: 

https://github.com/user-attachments/assets/cad05d5d-1cf6-4394-b0db-4d5d4f03b634

It's not browser specific and it's only happening on the duplication form because it needs to load specific data beforehand while the other multi-step forms redirect you to their starting step whichever it may be

knolleary (approved):
Was able to reproduce using the guide from @cstns - and confirmed this PR fixes it.

PR #5511, opened 2025-05-07
Clear redirectUrl when hitting the oauth account/complete route

knolleary:
We have seen occasions where a user is redirected to `/account/complete/ABCDEF/editor` after logging in and presented with a JSON formatted error.

This route is part of the OAuth flow used to authenticate access to the editor; a user should only be redirected there as part of opening the editor.

I haven't been able to track all of the handling of `redirectUrl` within the vue store, but it was clear the redirectUrl wasn't being cleared when it was this particular URL - as this view bounces the user straight on via a window.location update.

This PR adds an explicit clear of the redirectUrl to prevent it being reaccessed later.

My steps to reproduce this issue are quite involved and requires an SSO setup

1. Start logged out - incognito browser ideally
2. Enter URL for an instance editor
3. Login via SSO - you'll get redirected back to the editor
4. Logout of Node-RED - you'll get redirected back to the forge app
5. Login again via SSO
6. Logout of the forge app
7. You are redirected to the incorrect `/account/complete/...` page

With this fix, the incorrect redirect doesn't happen.

I doubt this is the exact sequence users are going through to hit this, but it is the sequence I found to reliably do so. Given the reports all point at an unexpected redirect to `/account/complete...`... then I'm pretty sure this will fix it for those cases as well.

### DIFF
--- a/frontend/src/pages/account/AccessRequestEditor.vue
+++ b/frontend/src/pages/account/AccessRequestEditor.vue
@@ -15,7 +15,7 @@
 
 <script>
 import { ArrowSmRightIcon, TemplateIcon } from '@heroicons/vue/solid'
-import { mapState } from 'vuex'
+import { mapState, useStore } from 'vuex'
 
 export default {
     name: 'AccessRequest',
@@ -27,6 +27,9 @@ export default {
         ...mapState('account', ['user', 'team'])
     },
     mounted () {
+        const store = useStore()
+        // If we've got here, remove any redirect url to prevent further unexpected redirects to this route
+        store.dispatch('account/setRedirectUrl', null)
         window.location.href = `/account/complete/${this.$router.currentRoute.value.params.id}`
     }
 }


frontend/src/pages/account/AccessRequestEditor.vue
```diff
@@ -27,6 +27,9 @@ export default {
         ...mapState('account', ['user', 'team'])
     },
     mounted () {
+        const store = useStore()
+        // If we've got here, remove any redirect url to prevent further unexpected redirects to this route
+        store.dispatch('account/setRedirectUrl', null)
```
cstns, line 32: One thing to note, the only relevant place where I found throughout the application that is setting the redirectUrl (other than when trying to deploy a blueprint from the website) is in the `checkState` method in the account store which is called extensively throughout the app.

`frontend/src/store/modules/account/index.js:421`

```javascript
...

    } catch (err) {
        // Not logged in
        commit('clearPending')
        // do we have a user session to clear?
        if (state.user) {
            try {
                window.posthog?.reset()
            } catch (err) {
                console.error('posthog error resetting user')
            }
        }
  
        if (router.currentRoute.value.meta.requiresLogin !== false) {
            if (router.currentRoute.value.path !== '/') {
                // Only remember the url if it isn't the default / path
                commit('setRedirectUrl', router.currentRoute.value.fullPath)
            }
            router.push({ name: 'Home' })
        }
    }

...
```
Currently, if unauthenticated, we're storing any route except the root '/' if we somehow get to this point which we may well be.

In my mind it wouldn't hurt to also blacklist the `/account/request/*` routes, the ones used by the editor to authenticate or the ff tools plugins requesting access. (the ones in `frontend/src/pages/account/routes.js:24`)

The redirect url is only used to redirect users to their intended pages after authentication, these (and a couple of more) routes are not meant to be redirected to after authentication

cstns (changes requested):

frontend/src/pages/account/AccessRequestEditor.vue
cstns: This approach would consolidate the fix even more

knolleary: The `/account/request/*` route *is* intended to be routed to post-authentication. It is the URL the editor redirects the user to in order to verify access. If they aren't logged in, we need to store that URL and redirect back to it after authentication has completed.

cstns (approved):

PR #5519, opened 2025-05-08
Blueprint auto import

### LINKED ISSUE #5179: Include packaged Blueprints in our self-hosted offering
### Description

tbd

### Epic/Story

#5121

hardillb:
fixes #5179
## Description

Adds a Housekeeping task to pull the latest BluePrints from FFC once a week for Licensed self hosted users.

The job runs 30 seconds after startup if a license is active or once a week (on a random day of the week and at a random time) to pull new Blueprints.

Enabled by default, but can be disabled (e.g. for FFC), the default URL to pull the BluePrint Bundle from can be set in the flowforge.yml file and if to enable the export API endpoint (unauthenticated)

```
blueprintImport:
  enabled: true,
  export: true,
  url: http://localhost:3000/api/v1/flow-blueprints/export-public
```

Do not merge until helm and CloudProject changes merged

- [x] Helm https://github.com/FlowFuse/helm/pull/601
- [ ] CloudProject https://github.com/FlowFuse/CloudProject/pull/707

## Related Issue(s)

#5179

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [x] Changes `flowforge.yml`?

## Labels

### DIFF
--- a/forge/ee/routes/flowBlueprints/index.js
+++ b/forge/ee/routes/flowBlueprints/index.js
@@ -272,7 +272,7 @@ module.exports = async function (app) {
     if (app.config.blueprintImport?.export) {
         app.get('/export-public', {
             config: {
-                allowAnonymous: true,
+                allowAnonymous: true
             },
             schema: {
                 summary: 'Export one or more Blueprints',
--- a/forge/ee/routes/flowBlueprints/index.js
+++ b/forge/ee/routes/flowBlueprints/index.js
@@ -294,7 +294,7 @@ module.exports = async function (app) {
                 }
             }
         }, async (request, reply) => {
-            let where = {}
+            let where = { active: true }
             if (request.query.id && typeof request.query.id === 'string') {
                 where = { id: app.db.models.FlowTemplate.decodeHashid(request.query.id)[0] }
             } else if (request.query.id && Array.isArray(request.query.id)) {
--- a/forge/ee/routes/flowBlueprints/index.js
+++ b/forge/ee/routes/flowBlueprints/index.js
@@ -269,6 +269,45 @@ module.exports = async function (app) {
         })
     })
 
+    if (app.config.blueprintImport?.export) {
+        app.get('/export-public', {
+            config: {
+                allowAnonymous: true,
+            },
+            schema: {
+                summary: 'Export one or more Blueprints',
+                tags: ['Flow Blueprints'],
+                query: {
+                    id: { type: 'array', items: { type: 'string' } }
+                },
+                response: {
+                    200: {
+                        type: 'object',
+                        allOf: [{ $ref: 'FlowBlueprintExport' }],
+                        properties: {
+                            count: { type: 'integer' }
+                        }
+                    },
+                    '4xx': {
+                        $ref: 'APIError'
+                    }
+                }
+            }
+        }, async (request, reply) => {
+            let where = {}
+            if (request.query.id && typeof request.query.id === 'string') {
+                where = { id: app.db.models.FlowTemplate.decodeHashid(request.query.id)[0] }
+            } else if (request.query.id && Array.isArray(request.query.id)) {
+                where = { id: request.query.id.map(i => app.db.models.FlowTemplate.decodeHashid(i)[0]) }
+            }
+            const flowTemplates = await app.db.models.FlowTemplate.getAll({}, where)
+            reply.send({
+                blueprints: flowTemplates.templates.map(bp => app.db.views.FlowTemplate.flowBlueprintExport(bp)),
+                count: flowTemplates.templates.length
+            })
+        })
+    }
+
     app.post('/import', {
         preHandler: app.needsPermission('flow-blueprint:create'),
         schema: {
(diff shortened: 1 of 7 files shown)


forge/ee/routes/flowBlueprints/index.js
```diff
@@ -269,6 +269,45 @@ module.exports = async function (app) {
...
+                            count: { type: 'integer' }
+                        }
+                    },
+                    '4xx': {
+                        $ref: 'APIError'
+                    }
+                }
+            }
+        }, async (request, reply) => {
+            let where = {}
```
knolleary, line 297:
```suggestion
            let where = { active: true }
```

We should only include active blueprints - not expose any work-in-progress or ones that we've chosen to hide.

knolleary (commented):
Whilst this will get the blueprints imported, it doesn't provide us a way to update them. That may be a future iteration - but I'd be worried that shipping this as-is would make that future iteration hard to achieve (as we'd have no way to confidently associate local blueprint with remote blueprint).

We can't rely on just the name - as that could cause us to overwrite a user's own blueprint.

How about:

1. add a 'sourceId' column to the FlowTemplates table (string type)
2. when importing a blueprint, set that to the incoming id of the blueprint - a locally created blueprint will have a blank value.
3. then, when doing the pull of the list, we can more confidently update local blueprints where its `sourceId` matches the remote.

If no blueprint with the matching `sourceId` is found, we'd still probably want to check for a clashing name - as a small number of customers will have received a blueprint zip from us already and we don't want them to end up with doubled up blueprints.

hardillb:
> How about:
> 
>     1. add a 'sourceId' column to the FlowTemplates table (string type)

We currently strip the id, but I can fix that

>     2. when importing a blueprint, set that to the incoming id of the blueprint - a locally created blueprint will have a blank value.

Easy enough

>     3. then, when doing the pull of the list, we can more confidently update local blueprints where its sourceId matches the remote.

This may make things a little trickier, but shouldn't be that complicated.

forge/housekeeper/tasks/blueprintImport.js
```diff
@@ -0,0 +1,49 @@
+const axios = require('axios')
+
+const { randomInt } = require('../utils')
+
+async function fetch (app) {
+    try {
+        const url = app.config.blueprintImport.url || 'https://flowfuse.com/api/v1/flow-blueprints/export-public'
```
knolleary, line 7:
```suggestion
        const url = app.config.blueprintImport.url || 'https://app.flowfuse.com/api/v1/flow-blueprints/export-public'
```

knolleary (commented):
Hard to test locally as it requires a running donor instance (which FFC isn't yet until this PR is deployed).

One question is around enabling the feature. AFAICT, this will require self-hosted users to update their yml to turn this on? Or will it be on by default?

joepavitt:
@hardillb what's the status here?

hardillb:
@joepavitt it got reviewed on Friday, so I'll be looking at the suggestions this morning.

hardillb:
> One question is around enabling the feature. AFAICT, this will require self-hosted users to update their yml to turn this on? Or will it be on by default?

import defaults to enabled by default, export needs to be explicitly enabled (so only needs doing for FFC Prod)

hardillb:
I've marked the Helm and configuration PRs for FFC prod/staging as ready for review.

forge/housekeeper/tasks/blueprintImport.js
```diff
@@ -0,0 +1,49 @@
...
+
+module.exports = {
+    name: 'blueprintImport',
+    startup: 30000,
+    // Pick a random hour/minute for this task to run at. If the application is
+    // horizontal scaled, this will avoid two instances running at the same time
+    // and on a random day of the week.
+    schedule: `${randomInt(0, 59)} ${randomInt(0, 23)} * * ${randomInt(0, 6)}`,
+    run: async function (app) {
+        if (app.license.active() && app.config.blueprintImport.enabled) {
```
knolleary: I can't see where `app.config.blueprintImport.enabled` defaults to true - this appears to require it to have been explicitly set in the config.

hardillb: yeah, looks like I intended it from the docs, but didn't make it actually so. I've pushed a fix to test for it being explicitly set to `false`

hardillb:
This needed the CloudProject PR merging first https://github.com/FlowFuse/CloudProject/pull/707

knolleary:
@hardillb I assumed this could merge and we'd be able to update the config to turn on the export separately?

hardillb:
@knolleary you are right, I was just wanting to prevent it from trying to import from it's self (by default on), but without the export enabled it should fail gracefully.

knolleary:
Realised this is going to cause the UI tests an issue once FFC is enabled to export blueprints (so the configured URL starts responding). Given the task runs 30s after startup, the UI tests (which run longer than 30s) will have this task run whilst they are running - which will cause unexpected data in the db for the tests.

Need to update the test env config to disable the auto-import.

Issue #5523, opened 2025-05-09
Instance Observability - APIs


joepavitt:
### Description

For a given **Hosted** Instances, we need to expose the following:

- API to expose CPU utilisation for all Hosted Instances in a Team
- API to expose historical CPU utilisation for a single Hosted Instance
    - This should also be available as a live feed (Websockets?) in order to provide live data to the user

### Epic/Story

#223

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
For more context on the APIs and "quick wins" available see: https://github.com/FlowFuse/flowfuse/issues/223#issuecomment-2859328747

joepavitt:
@hardillb could you provide a bit more technical detail here please and a reasonable sizing?

joepavitt:
To start with, yes.

hardillb:
WebSocket support will be more complicated (but we have some similar) as it will be a 2 stage process

- page to forge app
- forge app to nr-launcher for instance

Team wide view is also going to be tricky as that could be a LOT of instances, that will need more thought

hardillb:
Example datapoint

```
{
  ts: 1746798005843, // timestamp
  hs: , // heap size in bytes
  ps: , // process size in bytes
  hu: , // heap used in bytes
  cpu: 0-100, // CPU % 0-100 (but we will need to know how much is allowed from the stack to scale this

}
```

hardillb:
2 sub tasks

- Add http api to nr-launcher to return existing samples
- Add to `/api/v1/projects/:projectid/resources` to expose data to UI

hardillb:
@joepavitt I should be able to get the historical API done on Monday, with the websocket section taking a bit longer, say Wednesday that next week.

Also looks like we might have more than 10mins history (up to 2.5 hours)

hardillb:
Better review shows that at max we have 7 hours of history

gstout52:
> Team wide view is also going to be tricky as that could be a LOT of instances, that will need more thought.

@hardillb For a first iteration, if the team-wide view is additional work, we can push that back and revisit after we see user uptake.

@joepavitt Just flagging this comment for you.

joepavitt:
[...]

PR #5528, opened 2025-05-12
Add CPU/Memory Usage API endpoint and Stub Driver

hardillb:
part of #5523
## Description

Adds GET `/api/v1/projects/[project-id]/resources` that returns the historical CPU usage and Process size information

Stub generates random CPU usange and process size data

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/containers/stub/index.js
+++ b/forge/containers/stub/index.js
@@ -429,5 +429,22 @@ module.exports = {
     getBrokerAgentState: async (broker) => {
         return { connected: true, error: '' }
     },
-    sendBrokerAgentCommand: async (broker, command) => {}
+    sendBrokerAgentCommand: async (broker, command) => {},
+
+    // Resources API
+    resources: async (project) => {
+        const values = [
+
+        ]
+        const endTime = Date.now()
+        const itemsCount = 2520
+        for (let i = 0; i < itemsCount; i++) {
+            values.push({
+                ts: (endTime - ((itemsCount - i) * 10 * 1000)),
+                cpu: Math.random(),
+                ps: (Math.random() * 256),
+            })
+        }
+        return values
+    }
 }
--- a/forge/containers/wrapper.js
+++ b/forge/containers/wrapper.js
@@ -305,5 +305,14 @@ module.exports = {
         } else {
             throw new Error('Driver does not implement Broker API ')
         }
+    },
+    resources: async (project) => {
+        if (this._driver.resources) {
+            value = await this._driver.resources(project)
+            return value
+        } else {
+            throw new Error('Driver does not implement resources API ')
+        }
+        
     }
 }
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -1188,6 +1188,44 @@ module.exports = async function (app) {
     })
 
     /**
+     * 
+     * @name /api/v1/projects/:id/resources
+     * @memberof forge.routes.api.project
+     */
+    app.get('/:instanceId/resources', {
+        preHandler: app.needsPermission('project:read'),
+        schema: {
+            summary: 'Returns resource usage history for an Instance',
+            tags: ['Instances'],
+            params: {
+                type: 'object',
+                properties: {
+                    instanceId: { type: 'string' }
+                }
+            },
+            response: {
+                200: {
+                    type: 'array'
+                },
+                '4xx': {
+                    $ref: 'APIError'
+                },
+                500: {
+                    $ref: 'APIError'
+                }
+            }
+        }
+    }, async (request, reply) => {
+        try {
+            const resources = await app.containers.resources(request.project)
+            reply.send(resources)
+        } catch (err) {
+            console.log(err)
+            reply.code(500).send({ code: 'unknown_error', error: 'unknown error' })
+        }
+    })
+
+    /**
      * Merge env vars from 2 arrays.
      *
      * NOTE: When a var is found in both, currentVars will take precedence


hardillb:
@knolleary 

This PR and

- https://github.com/FlowFuse/nr-launcher/pull/337
- https://github.com/FlowFuse/driver-localfs/pull/176

Will get you a working local version.

The k8s driver version with support for HA mode is here:

- https://github.com/FlowFuse/driver-k8s/pull/215

joepavitt:
@hardillb what's the status of this, given that #5536 relies on this, and it's marked as draft, just wanted to check.

hardillb:
@joepavitt it was waiting on all the up stream to get merged (which is done)

So it can get the UI merged down and then the whole thing can get merged.

joepavitt:
Once this is merged, that will in trun close https://github.com/FlowFuse/flowfuse/issues/5523 right?

knolleary:
Have pushed a fix for the team type flag enablement in the UI.

However, once enabled, this is what I see on the performance tab:
image


I am running, afaict, the right branches of driver-localfs and nr-launcher - but the page is blank and no feedback.

We should also *not* hide the tab if it isn't enabled for the team type - we should have a placeholder page with the usual banner saying the feature requires an team upgrade - otherwise it serves no benefit to encourage users to upgrade.

I will do a quick iteration on this latter piece now. But the blank page will need some more debugging to figure out if this is a local issue or something wider.

knolleary (changes requested):
Comments above. Summary here:

 - ~Investigate blank chart~ - fixed via https://github.com/FlowFuse/nr-launcher/pull/340 when httpAdminRoot is set
 - Provide a placeholder (empty state) page if not available to current team type

knolleary:
Also spotted this has broken the Node-RED logs view:

image

hardillb:
@knolleary which `/resources` endpoint (the nr-launcher or the forge app)

knolleary:
@hardillb that is what the front-end is getting: `http://localhost:3000/api/v1/projects/1e7bb9d0-36e5-40dd-888f-6927a635bc5b/resources`

knolleary:
We identified the root issue of the missing data; my local instance has a non-default admin endpoint. The launcher needs to take that into account when querying the metrics endpoint.

knolleary:
Have reworked how the page is guarded so that any team can view the page. If the feature isn't enabled they see the following:

image

If the instance is running an old launcher, it shows:

image


If the instance has *just* started and we don't have any graphable data, rather than show a blank chart, it shows:

image

Have also added a refresh button to reload the data - although I now realise the websocket stuff was done for live updates... so that may want to be reverted and the websocket hooked up.


The less-than-ideal bits are:

 - The message about updating should link to the 'update NR version' button on settings
 - A suitable pictogram for the empty state views?
 - Refresh vs websocket
 - The semver check on the launcher is against v1.13 for some reason. That's the code in the branch as I found it; I was assuming the guard should be for 2.18. I have left it as-is in case theres a reason for that - need to verify with @hardillb

joepavitt:
> The semver check on the launcher is against v1.13 for some reason. That's the code in the branch as I found it; I was assuming the guard should be for 2.18. I have left it as-is in case theres a reason for that

History is available from `1.13.0` onwards, the Websockets is then only supported by `2.18` onwards.

joepavitt:
> NR Logs view still displaying oddly

I can investigate that now

knolleary:
Have updated the version check via https://github.com/FlowFuse/flowfuse/pull/5528/commits/c02cf72bd4b3ca62ae1cd9b497e0db91c6de165e to check for the 2.18 launcher if httpAdminRoot has been set for the instance. Verified locally where I have a mix of instances.

joepavitt:
@knolleary working on connecting up the live Websocket stuff to this view now too

joepavitt:
Screenshot 2025-05-30 at 17 00 25

Connected to WS now. Just need to fix a problem with the datazoom resetting each new data point comes in

joepavitt:
~Just dove into this branch again this morning to round off the datazoom issue but it's no longer rendering the chart at all, investigating...~

Resolved

joepavitt:
Added a fix for the datazoom issue when new data comes in on the websockets - so this is all good to go

joepavitt:
> The message about updating should link to the 'update NR version' button on settings

Let's get this merged (once reviewed and tests pass) and follow up on this action

hardillb:
Anybody know what the outstanding requested change is?

I'm starting the Changelog entry now (just double checking 99.9% this is Teams and Enterprise only on FFC and licensed only for self hosted)

hardillb:
Need to double check the team level feature flag before merging, testing locally with it enabled is showing the upgrade banner

knolleary:
@hardillb Enterprise only

frontend/src/pages/instance/Performance.vue
```diff
@@ -0,0 +1,387 @@
...
+            }
+
+            return Promise.reject(new Error('Instance is not running.'))
+        },
+        connectToLiveData () {
+            if (this.wsConnected) {
+                return
+            }
+            const uri = `/api/v1/projects/${this.instance.id}/resources/stream`
+            const ws = new WebSocket(uri)
```
knolleary: This websocket is never closed when the user navigates away from the page.

knolleary (changes requested):
@joepavitt the websocket is not being closed when the user navigates away from the page. When they return to the page, a second websocket is created.

hardillb:
OK, the feature flag may be an browser side storage thing, logging out and back in fixed it

frontend/src/pages/instance/Performance.vue
joepavitt: Handled in https://github.com/FlowFuse/flowfuse/pull/5528/commits/bd474295f5c73f0a8601997f560034eb116a19c6

joepavitt:
@hardillb I'll handle the changelog for this FYI as I've just been reviewing a couple of others so have it all up

Issue #5540, opened 2025-05-15
Duplicate Instance Name Availability Check


78wesley:
### Current Behavior

The changes made within PR #5499 created a new design for duplicate instance page, but the name checker doesn't check if the name is already in use. The check is only when you click on the "Create Instance" button which makes no sense.

![Image](https://github.com/user-attachments/assets/8b693feb-6781-4d11-b53a-6077de98ef7d)

### Expected Behavior

Check the name availability at the name input. 

### Steps To Reproduce

1. Go to an instance and duplicate the instance.
2. Change the name to an instance name that already exist. 
3. Then it will validate in green (which checks if the name is allowed not if it available)

### Environment

- FlowFuse version: FlowFuse Cloud
- Node.js version: - 
- npm version: - 
- Platform/OS: FlowFuse Cloud
- Browser: Brave 1.78.97 Chromium: 136.0.7103.93 (Official Build) (64-bit)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

hardillb:
@78wesley Thanks for noticing this

hardillb:
@knolleary do we add a new (heavily rate limited) API endpoint to check names before submit?

It does add a race condition for grabbing a name, but no worse than before I suppose.

knolleary:
I'm not sure the 'duplicate instance' updates have changed our behaviour here. I don't think we have every proactively checked the name before the user submits on any path to creating a new instance. So this isn't specific to the Duplicate Instance path through the app - the standard Create Instance is similarly affected (but then they reuse the same form, so fixing one will fix the other).

One fix would be to do a local name check against other instances in the team (information the UI will have access to today). However the name needs to be globally unique, so that would not get you very far.

As you suggest @hardillb, this would require:

 - api endpoint to check validity and availability of an instance name; proper rate limiting applied so it can't be abused as a way to scan for instances.
 - update the common form to use the api

hardillb:
Verified on FFC Prod

PR #5543, opened 2025-05-15
Ensure name check runs when duplicating an Instance

cstns:
## Description

Added a new check-name API call, implemented a watcher with a debouncer on the instance name, and slightly styled the name error message to prevent layout shifts when errors are displayed.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5540

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -1231,7 +1231,7 @@ module.exports = async function (app) {
             reply.code(400).send({ code: 'invalid_name', error: 'Invalid name' })
             return
         }
-        const safeName = name.trime().toLowerCase()
+        const safeName = name.trim().toLowerCase()
         if (app.db.models.Project.BANNED_NAME_LIST.includes(safeName)) {
             reply.status(409).send({ code: 'invalid_project_name', error: 'name not allowed' })
         }
--- a/frontend/src/api/instances.js
+++ b/frontend/src/api/instances.js
@@ -199,6 +199,11 @@ const checkCustomHostnameStatus = async (instanceId) => {
     })
 }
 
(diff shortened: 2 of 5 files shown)


cstns:
e2e tests fail due to missing `teamMembership` on the request object on the BE side. The `needsPermission` auth pre-handler returns 403 for non-admin users because of it on line `forge/routes/auth/permissions.js:60`.

If the comment from `forge/routes/api/project.js:19` is true, we'd need to also add the `teamMembership` property on the request for endpoints that are missing the `:instanceId` parameter or only on this particular `check-name` endpoint.

@hardillb, please give me a nudge when you have some free time

hardillb:
I'm looking at this

hardillb:
The API should be fixed, but still failing 3 UI tests.

joepavitt:
Screenshot 2025-06-02 at 10 20 39

The "next" button is still enabled when the check fails

joepavitt (changes requested):
The check works, but the "next" button still enabled, so users can just continue with a duplicate name as before.

cstns:
The tests were failing and the next button was still enabled because the errors watcher was placed on an invalid key. 

Also figured out that the check name endpoint returned an error when the instance name contained a `.` causing the name in use error to surface even though that was no the case. I slightly extended the name FE check method to exclude `.` from instance names

joepavitt:
Please let me know once this is ready for review again please Serban

cstns:
should be good to go

joepavitt (approved):

PR #5555, opened 2025-05-28
Team Performance UI

### LINKED ISSUE #5525: Instance Observability - All Instances
### Description

- Follow on from #5524 
- Add a new "Performance" page to the top team-level view
- This should retrieve the CPU utilisation for all instances, and then provide a live update as those values change
- Subscription is likely to be via WebSockets (we should avoid HTTP polling where possible)

### Epic/Story

#223

joepavitt:
## Description

Screenshot 2025-05-28 at 19 19 32

### Status:

- Added the new "Performance" page
- Created the new "CPU Utilization" cell component
- Added placeholder permissions/feature checks for `isPerformanceFeatureEnabled`

### Todo:

- Add E2E tests
- Connect to the realtime Websockets API (Awaiting response to https://github.com/FlowFuse/flowfuse/issues/5523#issuecomment-2917052041)
- Check, and handle bouncing from the auto-sort for CPU column
- Add server-side logic for storing (and managing) for the `isPerformanceFeatureEnabled` logic

## Related Issue(s)

Closes #5525

### DIFF
--- a/frontend/src/pages/team/Performance/components/CPUUtilizationCell.vue
+++ b/frontend/src/pages/team/Performance/components/CPUUtilizationCell.vue
@@ -17,6 +17,11 @@
             <label class="ff-cpu-bar--label">{{ formattedCPUUtilization }}%</label>
         </div>
     </div>
+    <div v-else-if="status !== 'running'">
+        <div class="text-sm text-gray-500">
+            Live data unavailable
+        </div>
+    </div>
     <div v-else class="text-sm text-gray-500">
         Live data view not supported for this instance. Please upgrade the Instance's Node-RED version.
     </div>
@@ -28,6 +33,10 @@ export default {
     name: 'CPUUtilizationCell',
     props: {
         instanceId: {
+            type: String,
+            required: true
+        },
+        status: {
             type: String,
             required: true
         },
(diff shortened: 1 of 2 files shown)


joepavitt:
@knolleary I've managed to get this pushed without the need for the API changes using the method you proposed. I'd like to be sure that the CTA "Please upgrade the instance's Node-RED version" is the correct syntax. When users click that rows, they're taken to the Instance Overview page. For FF Cloud users, I believe they'll be presented with the "Update Node-RED Version" button?

knolleary:
Two items to address before this can merge:

image


1. If an instance is suspended, the message shouldn't talk about needing to update it
2. The table is sorted by cpu usage by default. I found this difficult to work with as my two running instances kept swapping position in the table. With more instances, I would see this as even harder as you have to re-find the instance you were watching in the table. I'm going to change the default sort to the status column.

knolleary:
Also, the websockets are never disconnected. Each time I access the page, a new set of connections are created.

knolleary:
Fixes pushed for the review items:

image

1. updated message if instance isn't running
2. closes websockets on unmount
3. changed default sort of the table to status.

cstns (approved):

PR #5562, opened 2025-05-29
Update license loader to handle licenseId field

### LINKED ISSUE #5561: License generator flow setting incorrect id field
### Current Behavior

The license generator flow that is used to create licenses for customers is setting the wrong field for the license id; it is setting a field called `licenseId`.

The original [license generator script](https://github.com/FlowFuse/flowfuse/blob/9a85d71e600c44529c2acb3fb55c9b9b802196da/forge/licensing/license-generator.js#L98) and the [license parser](https://github.com/FlowFuse/flowfuse/blob/9a85d71e600c44529c2acb3fb55c9b9b802196da/forge/licensing/loader.js#L17) expect that field to be called `id`.

As such, any platform running with a licensed generated by the flow rather than the script is not reporting back its license id in its telemetry pings.

The license generator flow has already been fixed and verified by myself and @robmarcer so any newly issued licenses from today will have the property `id` field set.

This issue is to update the license parser to handle both `id` and `licenseId`. Once self-hosted platforms update to the latest, they will report their id properly. Separately we will look at reissuing licenses to existing self-hosted customers with the id field fixed as applying a new license is quicker for them then upgrading the platform.

### Expected Behavior

The platform property reports the license id when its provided in either the `id` or `licenseId` claim within the license.

### Steps To Reproduce

_No response_

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Closes #5561

Updates license loader to handle `licenseId` field if `id` not set.

### DIFF
--- a/forge/licensing/loader.js
+++ b/forge/licensing/loader.js
@@ -14,7 +14,8 @@ const MQTT_CLIENT_DEFAULT_ALLOWANCE = {
 class LicenseDetails {
     constructor (license, claims) {
         // this.license = license;
-        this.id = claims.id
+        // For a period of time, the license generator flow set the id as licenseId instead of id. We handle both here.
+        this.id = claims.id || claims.licenseId
         this.version = claims.ver || ''
         this.note = claims.note
         this.organisation = claims.sub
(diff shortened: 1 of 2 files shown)


hardillb (approved):

hardillb:
Will merge when tests finish

PR #5564, opened 2025-05-30
UI: Ensure Node-RED Logs render at full width

joepavitt:
## Description

Ensures the NR Logs render at full width. Problem occurs when there isn't a multi-line entry and the lgos are able to shrink in width, is not linked to, or caused by changes in #5528 so I made it a separate PR.

## Related Issue(s)

Fixes issue found here: https://github.com/FlowFuse/flowfuse/pull/5528#issuecomment-2921897005


joepavitt:
This may in fact be caused by #5528 - investigating further...

knolleary:
@joepavitt testing this branch and the logs look as expected with or without your change.

If I apply this change to the `nr-resource-api` branch (target of 5528) which has the issue, then it does fix it there as well. So there must be *something* in 5528 causing the issue.

But given the fix works either way, happy to apply this to main via this PR.

knolleary (approved):

PR #5566, opened 2025-06-02
Allow blank filename for git stages that are not the first

knolleary:
The 'snapshot file name' field is optional *if* the pipeline stage is not the first stage. If it is the first stage, the filename is required.

The check for this wasn't taking into account the 'isFirstStage' part of the logic.

### DIFF
--- a/frontend/src/pages/application/PipelineStage/form.vue
+++ b/frontend/src/pages/application/PipelineStage/form.vue
@@ -676,7 +676,7 @@ export default {
             }
         },
         'input.pushPath' (newPushPath, oldPushPath) {
-            if (newPushPath === '') {
+            if (newPushPath === '' && this.isFirstStage) {
                 this.errors.pushPath = 'Please enter a valid filename'
             } else {
                 this.errors.pushPath = ''


Steve-Mcl (commented):
This doesnt seem to do as expected Nick 🤔 

I thought I would give it a quick go in [pre-staging](https://5566.flowfuse.dev/team/plant-enterprise/applications/zj3Gn2GJYb/pipelines) but I was allowed to create and save a git stage in slot 1 without a file name.

It was then I realised the save/update button is also not reactive!  Something odd? I am certain this was working before!


chrome_5YUL9tHPBQ

knolleary:
The issue is that this is the first stage in the pipeline, so the filename field is considered required. The form is invalid whilst the field is blank, but it doesn't highlight the error in the form. This is true of all types of stage form; the button doesn't enable until all required fields and entered, but it doesn't tell which fields are required.

Will take another look at it.

knolleary:
What doesn't help is the 'add stage' form lets you create a git pipeline stage as the first stage without a filename.

knolleary:
@Steve-Mcl I've fixed the 'create' form to not allow a blank filename when creating a git stage as first stage.

There is a broader issue around form feedback around which fields are required. I don't want to block this PR for that as this is fixing an immediate problem with this new feature.

Will raise a separate issue around the form UX.

Issue #5520, opened 2025-05-09
Applications page slow to load with lots of remote instances


knolleary:
### Current Behavior

The Team home page gets noticeable slow to load when a team has lots of remote instances.

This is due to the request to `/api/v1/teams/<teamId>/applications?includeApplicationSummary=true`.

For our own internal team on FFC, this request takes 2.5s. I can point at a customer's team (on request) where it is taking 6+ seconds.

The number of remote instances appears to be the primary cause of the response time.

We need to review what information is being returned in that api request; whether it is necessary to return it all in a single request, or if it can be broken up into multiple requests (so the page can display initial information quicker).

### Expected Behavior

The team home page loads quickly as it is the first thing users see when they log in.

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
@gstout52 FYI pulling this as it's a performance problem we need to address

cstns:
If I remember correctly, the initial API call was retrieving all applications along with all their instances and devices to support frontend-only search functionality.

To reduce the load on the applications API call while maintaining search functionality, I'll need to update the existing search logic to also query the backend. This would allow us to limit the initial number of instances and devices per application to 4, significantly reducing payload size without breaking search and alleviating the pressure from the page.

joepavitt:
Removing the story points on this issue and splitting to child (actionable) items

gstout52:
@joepavitt The only child issues I see are closed. Free to close this?

cstns:
Not yet, this is still in progress. Two key components are still missing from the implementation:
- A backend API that searches through devices and instances (about 70% complete, but not yet turned into a PR).
- UI updates to limit the number of pre-loaded instances and devices per app, and to modify the existing search to use the new API.

joepavitt:
@cstns no sizing on this, what's an appropriate number based on the work you're doing right now?

cstns:
with the amount of work that already went in I'd say a 4/5 and hope to finish it next Tuesday (Monday's off)

cstns:
Postponing until change from [Required change i](https://github.com/FlowFuse/flowfuse/pull/5631) reaches main as it's a prerequisite for the 2nd step mentioned in https://github.com/FlowFuse/flowfuse/issues/5520#issuecomment-2944496670

cstns:
[...]

PR #5592, opened 2025-06-05
Add a search endpoint for hosted and remote instances

cstns:
### API
- Added GET /api/v1/search/instances
  - Supports searching both hosted and remote instances
  - Used by the application's backend search ([FlowFuse/flowfuse#5602](https://github.com/FlowFuse/flowfuse/pull/5602))
  - retrieves the project/instance meta/liveState from the start, can be removed later on if db load is too great and the UI could use the endpoint implemented in https://github.com/FlowFuse/flowfuse/pull/5704

### Models
- Added byTeamForSearch query to:
  - Device model – allows query-based search within a team's devices
  - Project model – allows query-based search within a team's projects/instances

### Views
  - Extended dashboardInstanceSummary to include project settings and allow correct rendering of the dashboard button in UI
  - Extended deviceSummary to include the application in order to be able assign the device to it's specific app in the UI

## Related Issue(s)

part of https://github.com/FlowFuse/flowfuse/issues/5520

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/routes/api/search.js
```diff
@@ -149,4 +151,96 @@ module.exports = async function (app) {
             results: []
         })
     })
+
+    /**
+     * Search API
+     *
+     * GET /api/v1/search/instances
```
knolleary, line 158: Is there a reason for this to be a new search endpoint rather than add something like `scope=instances` to the existing `/search` endpoint?

joepavitt:
@cstns what's the status here, as we have https://github.com/FlowFuse/flowfuse/pull/5602 depending on this

forge/routes/api/search.js
cstns: The main reason was to keep the query as lightweight as possible and avoid pulling in unnecessary relations. 

It was also much easier to trim the payload with a dedicated endpoint than to adapt the existing one and limit the `byTeam` finders without side effects.

cstns:
It went into review yesterday, @knolleary just got back today and is swamped, things are moving along as fast as possible

forge/routes/api/search.js
knolleary: I was thinking the underluing implmentation would be what you've done already, just a big `if scope=instances` then do one way, otherwise do the existing.

But I'll not block on that.

forge/routes/api/search.js
```diff
@@ -149,4 +151,96 @@ module.exports = async function (app) {
...
+                        results: { type: 'array', items: { type: 'object', additionalProperties: true } }
+                    }
+                },
+                '4xx': {
+                    $ref: 'APIError'
+                }
+            }
+        }
+    }, async (request, reply) => {
+        if (request.query.query.length === 0) {
```
knolleary:
```suggestion
        if (!request.query.query) {
```

Guards against `query` missing, and being 0-length. Otherwise get a `Cannot read properties of undefined` response on the api

forge/routes/api/search.js
cstns: Yes, that would have worked as well! The only downside, in my opinion, is that things can get out of hand quickly when endpoints (or methods, for that matter) try to do too much.

PR #5593, opened 2025-06-05
Ensure snapshot is persisted when deploying to device from git

knolleary:
## Description

When deploying to an instance, the instance creates a copy of the provided snapshot for itself.
When deploying to a device/group, it reuses the provided snapshot - this broke an assumption in the git pull code that I missed in my testing.

This fixes it by ensuring the snapshot generated from the pull is saved to the db (with appropriate user/owner fields set) when deploying to a device/group.


Steve-Mcl (approved):
Tested on pre-staging & Approved (edge case was discussed on slack)

But not merging due to test fail

PR #5619, opened 2025-06-16
Link the Team Homepage recent remote instances to the backend

cstns:
## Prerequisites

child of https://github.com/FlowFuse/flowfuse/pull/5617

please do not merge after review

## Description

linked the team's homepage Recent Instances section to the backend to get real data.

Had to make some alterations to acomodate the devices:
- missing 'offline' stat for the new minimal badge in tailwind's safe color scheme
- created a new component to house the recently modified instances
- adapted the device tile to accomodate minimal view (no borders etc)
- made the FinishSetup device button tertiary when in minimal view ( text still visible as an icon wouldn't suffice)

image

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

Issue #5600, opened 2025-06-09
Add new "Home" page for FlowFuse


joepavitt:
### Description

As part of the discussion in https://github.com/FlowFuse/flowfuse/issues/4737 we have concluded that in order to clean the "Applications" view we need to remove it's duel-purpose. Instead, we should have a dedicated "Home"/Landing page for FlowFuse.

Core ingredients to make up this page should be focussing on what we believe the user wants to do when landing in FlowFuse.

- Presenting direct access to "relevant" Hosted & Remote Instances
    - A simple way of us measuring "relevant" and best guessing this would be to look at the Instances most recently deployed/edited. They're more likely to be a destination for a user than an Instance that hasn't been touched in 3 months.
- Things going wrong. If there are issues users should be made aware of those problems as clearly as possible.
    - Performance Alerts
    - Errors

Potentially useful:

- Recent activity within the Team (Audit Log)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Size L including both Design and Dev work

joepavitt:
![Image](https://github.com/user-attachments/assets/3e6f8770-7743-43cf-a724-f0680292e3e4)

First draft, including nav changes from https://github.com/FlowFuse/flowfuse/issues/5601

knolleary:
I note you've renamed instances to flows - is that in scope of this redesign or something else?

joepavitt:
@knolleary no, was where I'd copied over an old design and not renamed, it my designs (as of right now) they're "Instances" again

joepavitt:
Actual design files can be seen here: https://www.figma.com/design/rMB2w5ZZUtTC0nxbwLYKJY/FlowFuse-Platform?node-id=6141-8500&t=syiEVX281slrfona-1

joepavitt:
<img width="1642" alt="Image" src="https://github.com/user-attachments/assets/acbed29a-2091-4acb-8b18-b507a31403c4" />

Further progress on a second iteration. Here the larger status tiles would be interactive, linking to the respective Hosted/Remote Instances view with the appropriate filter applied.

joepavitt:
<img width="1442" alt="Image" src="https://github.com/user-attachments/assets/beeef22c-5821-4209-ba88-50e5d7dfc968" />

joepavitt:
[...]

PR #5631, opened 2025-06-17
Parametrize the team projects API endpoint

cstns:
## Prerequisites 

child of https://github.com/FlowFuse/flowfuse/pull/5619

please do not merge after review

## Description

Adds support for query parameters to the /team/projects API endpoint, enabling result limiting, sorting by most recent flow update, and inclusion of instance metadata.

This query is resource-intensive, hence the need for a result limit. Despite being marked as deprecated, this endpoint remains the only one that allows retrieval of a team's instances.

The ability to include instance metadata—while adding overhead—eliminates the need for the frontend to make a separate API call just to fetch the same data with the meta key (as currently done on the Applications page). This functionality should always be used with a query limit to control load.

This change also complements the performance improvements discussed in [FlowFuse issue #5520](https://github.com/FlowFuse/flowfuse/issues/5520). Currently, we query /team/projects, then loop through the results to make a second call for each instance—just to get the meta key. This is inefficient, especially as we’re loading all instances client-side and filtering there, instead of server-side.

Once deployed, this enhancement supports step 2 of [this comment](https://github.com/FlowFuse/flowfuse/issues/5520#issuecomment-2944496670): limiting instance results and moving search/filtering to the backend. Both endpoints will now return all required data, including meta.

## Related Issue(s)

part of https://github.com/FlowFuse/flowfuse/issues/5600

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/models/Project.js
+++ b/forge/db/models/Project.js
@@ -490,7 +490,15 @@ module.exports = {
                         include
                     })
                 },
-                byTeam: async (teamIdOrHash, { query = null, instanceId = null, includeAssociations = true, includeSettings = false } = {}) => {
+                byTeam: async (teamIdOrHash, {
+                    query = null,
+                    instanceId = null,
+                    includeAssociations = true,
+                    includeSettings = false,
+                    includeMeta = false,
+                    limit = null,
+                    orderByMostRecentFlows = false
+                } = {}) => {
                     let teamId = teamIdOrHash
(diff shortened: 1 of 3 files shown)


joepavitt:
See: https://github.com/FlowFuse/flowfuse/issues/5600#issuecomment-2991247380

PR #5647, opened 2025-06-19
Enable Google social login

### LINKED ISSUE #2657: Implement Google Social Sign Up
### Description

https://developers.google.com/identity/gsi/web/guides/overview

knolleary:
Closes #2657 

## Description

This enables a 'Sign In with Google' button on the login page:

image

This will *only* allow users with non-workspace associated accounts to login.

If the user does not exist, we will create an account for them automatically.

If a workspace user tries, they will see:

image

If their workspace domain has a specific SSO Provider configuration setup, the user will be redirected to login via that.

Configuration of this is done via Admin Settings:

image

The ClientID comes from a Google App setup via: https://developers.google.com/identity/gsi/web/guides/get-google-api-clientid

As a feature, this is only available to EE licensed platforms.

### DIFF
--- a/forge/ee/routes/sso/auth.js
+++ b/forge/ee/routes/sso/auth.js
@@ -1,15 +1,8 @@
-const crypto = require('crypto')
-
 const { Authenticator } = require('@fastify/passport')
 const { MultiSamlStrategy } = require('@node-saml/passport-saml')
 const fp = require('fastify-plugin')
 
-const { completeUserSignup } = require('../../../lib/userTeam')
-
-const generatePassword = () => {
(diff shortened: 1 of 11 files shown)


knolleary:
> Should the Sign up page also offer to create account from Google Sign in?

Good point - will look as it'll be quick to add.

> Trying to enter the gmail address in the username prompts for a password (that the user doesn't know, so can't change it)

Yeah - edge case I'm not going to overly handle now. They can reset their password if they want to login that way.

knolleary:
On second thoughts, not going to add to sign-up page. Can be handled as a separate item if decided we want to offer that.

joepavitt:
PostHog Insight to track usage: https://eu.posthog.com/project/2209/insights/IHnk85dT

PR #5650, opened 2025-06-19
Logout other sessions when MFA enabled

knolleary:
When enabling multi-factor authentication, this change will invalidate all existing login sessions for the user, but keep the current session alive.


hardillb (approved):

hardillb:
Tested on pre-staging

Issue #5598, opened 2025-06-06
Auto-move to next stage when application selected


joepavitt:
### Description

See friction here: https://fathom.video/share/wRzZywf7usZ7U24XcammkHWXs3H6SXi1

When adding a new Instance, the user can select which Application they want the new Hosted Instance to be a part of. This is currently a two-step process:

1. Select the application
2. Click "Next" (top-right)

This is causing friction, especially when there is only one application.

### Steps to Remedy

- Click the application, move directly to the next stage. No need to click "Next"

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

knolleary:
See also: https://github.com/FlowFuse/flowfuse/issues/5385

Issue #5385, opened 2025-04-02
UX: lots of scrolling for multi-stage instance form with lots of applications


knolleary:
For a team with lots of applications, the first stage in creating an instance requires lots of scrolling.

First you have to scroll through the list to find the application you want.
Then you have to scroll all the way back to the top to get to the 'next' button - but as this is the first stage, you may not know the 'next' button is at the top, waiting to be clicked.

<img width="190" alt="Image" src="https://github.com/user-attachments/assets/b8d94b54-20ee-47ef-a2b4-4588b1f9d7b3" />

I have two proposals:

1. make the content below the progress bar scrollable - so the progress bar and back/next buttons are always visible.
2. Or... as the application selection is the *only* thing they have to do at this stage, make clicking the application progress the form automatically.

cstns:
will be adding an applications search field on the application step as part of this iteration as well

cstns:
see https://github.com/FlowFuse/flowfuse/pull/5656#issuecomment-2991897486

PR #5656, opened 2025-06-20
Fix MultiStepForm scrolling

cstns:
## Description

Fixes overflowing/scroll behavior on the multi-step form.
Added a new grey border bottom to clearly differentiate between the scroll area:
- application step
https://github.com/user-attachments/assets/e2185af6-9899-4da2-bcbd-825afc28683d
- blueprint step
https://github.com/user-attachments/assets/60f4c5dc-811a-45a2-aba2-de6b6cc4aa44
- duplication step
https://github.com/user-attachments/assets/200cda45-efd7-4581-bc5f-18b5be2d4692
- instance step
https://github.com/user-attachments/assets/b35a18b9-40f9-4f53-8221-4cdc54a4c273

## Related Issue(s)

part of https://github.com/FlowFuse/flowfuse/issues/5385
part of https://github.com/FlowFuse/flowfuse/issues/5598

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
Personally, I don't like how the area scrolls, and how the user needs to mouse over the central content to scroll it. I'd much prefer the content to be full width for the scroll _or_ we just add a text input search instead, such that the Applications list, etc. can be easily filtered

cstns:
That would partially solve the problem for the applications step but what about the other steps?

joepavitt:
We already have scroll in place for Blueprints and the Instance form is very short anyway, so a little scrolling in a worst case (very small screen) isn't an issue

cstns:
@joepavitt  please re-review this task, I'll create a new PR for the search field in the application step. 
This PR should then only cover scrolling behavior for the multi-step form steps which is still needed

joepavitt:
Screenshot 2025-06-23 at 15 44 46

My feedback above still stands, the scroll section on the Applications is too narrow.

Annoyingly, the screenshot didn't include the scrollbar. If you try scrolling in the empty space either side, it doesn't scroll.

cstns:
The ability to scroll on the whitespace is awesome, and I think it will be welcomed.

Widening the app list in order for the scroll to have a functional scroll over the white left and right space would display the scrollbar in a weird place (css and clever html).

My first thought would be to link the outer element onScroll event with the inner one to keep the scrollbar where it would make sense and gain the whitespace scroll.

joepavitt:
Appreciate this may not be welcomed, but with:

- https://github.com/FlowFuse/flowfuse/pull/5657
- https://github.com/FlowFuse/flowfuse/pull/5673

I'm going to consider https://github.com/FlowFuse/flowfuse/issues/5385 closed and complete, without the need for this PR. As such, I will close this PR as unrequired

cstns:
I see where you're coming from, but I think it's important to preserve the scrolling behavior. Differences in resolution and DPI can significantly impact how a page is rendered and experienced by users, so maintaining consistent scroll rules helps ensure usability across devices.

The search functionality was just a part of the fix

PR #5666, opened 2025-06-20
Retain env var hidden flag metadata within snapshots

knolleary:
Fixes https://github.com/FlowFuse/security/issues/106

## Description

Currently the 'hidden' flag on env vars gets stripped when generating a snapshot. This means the env var becomes unhidden when pushed via pipeline.

This PR is part one of addressing this. Some explanation is needed.

We have two different ways of storing env var lists. In Instance settings we use an array:
```
[ { name: 'foo', value: 1 }, { name: 'bar', value: 2, hidden: true }]
```

When generating a snapshot, that gets converts to a key/value object:
```
{
   foo: 1,
   bar: 2
}
```

This is where the `hidden` flag is lost.

This PR addresses it by using a new format for the env-var object in the snapshot:

```
{
   foo: 1,
   bar: { value: 2, hidden: true }
}
```

Existing snapshots will continue to be valid; they have already lost the hidden metadata and we can't restore it.

Hosted/Remote instances do not know how to handle this format, so the endpoints they use to retrieve they settings have been updated to convert back to the plain key/value object without the hidden flag - that's okay because they have no use of the hidden flag, they just need the value.

---

This also lays the ground work for further encryption of the values. The metadata object will be able to contain the encrypted value, and the utility function `exportEnVarObject` added by this PR becomes the place that handles decryption prior to passing to an instance/device. It's a bit more complicated then that, which is why that'll be a separate PR.

---

To Do:

 - [ ] Unit test fixes and updates
 - [ ] Check for any other paths that could be impacted by this change

### DIFF
--- a/forge/db/utils.js
+++ b/forge/db/utils.js
@@ -153,6 +153,77 @@ function randomPhrase (wordCount = 3, minLength = 2, maxLength = 15, separator =
     return randomStrings(wordCount, minLength, maxLength).join(separator)
 }
 
+/**
+ * Convert an array of env var objects to a key/value object, with handling of hidden vars that
+ * need to retain their metadata
+ * From: [ { name: 'VAR1', value: 'value1' }, { name: 'VAR2', value: 'value2', hidden: true } ]
+ * To: { VAR1: 'value1', VAR2: { value: 'value2', hidden: true } }
+ */
+function mapEnvArrayToObject (envArray) {
+    const envObject = {}
+    envArray.forEach((envVar) => {
+        const name = envVar.name
+        const value = envVar.value
+        if (envVar.hidden) {
+            envObject[name] = {
+                value,
+                hidden: true
+            }
+        } else {
+            envObject[name] = value
+        }
+    })
+    return envObject
+}
+
+/**
+ * Convert a key/value object of env vars to an array of env var objects
+ * From: { VAR1: 'value1', VAR2: { value: 'value2', hidden: true } }
+ * To: [ { name: 'VAR1', value: 'value1' }, { name: 'VAR2', value: 'value2', hidden: true } ]
+ */
+function mapEnvObjectToArray (envObject) {
+    const envArray = []
+    for (const [name, value] of Object.entries(envObject)) {
+        if (typeof value === 'object' && value.hidden) {
+            envArray.push({
+                name,
+                value: value.value,
+                hidden: true
+            })
+        } else {
+            envArray.push({
+                name,
+                value
+            })
+        }
+    }
+    return envArray
+}
+
+/**
+ * Takes an env var key/value object and modifies it to contain only the values of the env vars
+ * This handles any env vars flagged as hidden. This strips the hidden metadata - the result
+ * of this function is suitable for passing to a hosted/remote instance to use as-is
+ * From: { VAR1: 'value1', VAR2: { value: 'value2', hidden: true } }
+ * To: { VAR1: 'value1', VAR2: 'value2' }
+ */
+function exportEnvVarObject (envObject) {
+    // Check for any hidden env vars. These are objects with a 'hidden' property set to true.
+    // If so, we replace the object with just the value.
+    // TODO: handle encrypted values
+    const result = {}
+    for (const envVar of Object.keys(envObject)) {
+        if (Object.hasOwn(envObject[envVar], 'hidden') && envObject[envVar].hidden) {
+            // The value has metadata - use the value only
+            result[envVar] = envObject[envVar].value
+        } else {
+            // The value is the bare value - use as-is
+            result[envVar] = envObject[envVar]
+        }
+    }
+    return result
+}
+
 module.exports = {
     init: _app => { app = _app },
     generateToken: (length, prefix) => (prefix ? prefix + '_' : '') + base64URLEncode(crypto.randomBytes(length || 32)),
@@ -184,5 +255,8 @@ module.exports = {
     buildPaginationSearchClause,
     getCanonicalEmail,
     randomStrings,
-    randomPhrase
+    randomPhrase,
+    mapEnvArrayToObject,
+    mapEnvObjectToArray,
+    exportEnvVarObject
 }
--- a/forge/db/controllers/Project.js
+++ b/forge/db/controllers/Project.js
@@ -95,9 +95,17 @@ module.exports = {
             const projectSettings = projectSettingsRow.value
             result = app.db.controllers.ProjectTemplate.mergeSettings(result, projectSettings)
             const envVars = app.db.controllers.Project.insertPlatformSpecificEnvVars(project, result.env)
+            // envVars is an array of env var objects. Need to convert to an env var object and merge into
+            // the existing env object
             // convert  [{name: 'a', value: '1'}, {name: 'b', value: '2'}]  >> to >>  { a: 1, b: 2 }
             envVars.forEach(envVar => {
-                env[envVar.name] = envVar.value
+                if (envVar.hidden) {
+                    // Maintain the hidden flag
+                    // TODO: handle encrypted values
+                    env[envVar.name] = { value: envVar.value, hidden: true }
+                } else {
+                    env[envVar.name] = envVar.value
+                }
             })
         }
         // If we don't have any modules listed in project settings. We should
@@ -131,6 +139,9 @@ module.exports = {
         return result
     },
 
+    /**
+     * Exports the instance as a JSON object. Used as part of generating snapshots.
+     */
     exportProject: async function (app, project, components = {
         flows: true,
         credentials: true,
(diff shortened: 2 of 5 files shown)


forge/db/utils.js
```diff
@@ -153,6 +153,77 @@ function randomPhrase (wordCount = 3, minLength = 2, maxLength = 15, separator =
...
 }
 
+/**
+ * Convert an array of env var objects to a key/value object, with handling of hidden vars that
+ * need to retain their metadata
+ * From: [ { name: 'VAR1', value: 'value1' }, { name: 'VAR2', value: 'value2', hidden: true } ]
+ * To: { VAR1: 'value1', VAR2: { value: 'value2', hidden: true } }
+ */
+function mapEnvArrayToObject (envArray) {
+    const envObject = {}
```
Steve-Mcl:
```suggestion
/**
 * @typedef {Object<string, string | { value: string, hidden?: boolean }>} EnvVarObject
 * An object mapping environment variable names to a string value or a metadata object.
 * e.g.
 * `{ VAR1: 'value1', VAR2: { value: 'value2', hidden: true } }`
 * 
 * @typedef {Array<{ name: string, value: string, hidden?: boolean }>} EnvVarArray
 * An array of environment variable objects, where each object has a name and value,
 * and optionally a hidden flag.
 * e.g.
 * `[ { name: 'VAR1', value: 'value1' }, { name: 'VAR2', value: 'value2', hidden: true } ]`
 */

/**
 * Convert an array of env var objects to a key/value object, with handling of hidden vars that
 * need to retain their metadata
 * From: [ { name: 'VAR1', value: 'value1' }, { name: 'VAR2', value: 'value2', hidden: true } ]
 * To: { VAR1: 'value1', VAR2: { value: 'value2', hidden: true } }
 * @param {EnvVarArray} envArray
 * @return {EnvVarObject}
 * @see {@link mapEnvObjectToArray} for the reverse operation
*/
function mapEnvArrayToObject (envArray) {
    /** @type {EnvVarObject} */
    const envObject = {}
```

Please consider adding `@typedef` definitions and `@param` / `@return` fields on the JSDoc.
This opens up possibilities like static analysis, `F2` refactorings, caller type hints & other nice DX perks 🤞

forge/db/utils.js
```diff
@@ -153,6 +153,77 @@ function randomPhrase (wordCount = 3, minLength = 2, maxLength = 15, separator =
...
+    return envObject
+}
+
+/**
+ * Convert a key/value object of env vars to an array of env var objects
+ * From: { VAR1: 'value1', VAR2: { value: 'value2', hidden: true } }
+ * To: [ { name: 'VAR1', value: 'value1' }, { name: 'VAR2', value: 'value2', hidden: true } ]
+ */
+function mapEnvObjectToArray (envObject) {
+    const envArray = []
```
Steve-Mcl:
```suggestion
/**
 * Convert a key/value object of env vars to an array of env var objects
 * From: { VAR1: 'value1', VAR2: { value: 'value2', hidden: true } }
 * To: [ { name: 'VAR1', value: 'value1' }, { name: 'VAR2', value: 'value2', hidden: true } ]
 * @param {EnvVarObject} envObject
 * @return {EnvVarArray}
 * @see {@link mapEnvArrayToObject} for the reverse operation
 */
function mapEnvObjectToArray (envObject) {
    /** @type {EnvVarArray} */
    const envArray = []
```

`typedef`s are in prior suggestion

forge/db/utils.js
```diff
@@ -153,6 +153,77 @@ function randomPhrase (wordCount = 3, minLength = 2, maxLength = 15, separator =
...
+    return envArray
+}
+
+/**
+ * Takes an env var key/value object and modifies it to contain only the values of the env vars
+ * This handles any env vars flagged as hidden. This strips the hidden metadata - the result
+ * of this function is suitable for passing to a hosted/remote instance to use as-is
+ * From: { VAR1: 'value1', VAR2: { value: 'value2', hidden: true } }
+ * To: { VAR1: 'value1', VAR2: 'value2' }
+ */
```
Steve-Mcl:
```suggestion
/**
 * Takes an env var key/value object and modifies it to contain only the values of the env vars
 * This handles any env vars flagged as hidden. This strips the hidden metadata - the result
 * of this function is suitable for passing to a hosted/remote instance to use as-is
 * From: { VAR1: 'value1', VAR2: { value: 'value2', hidden: true } }
 * To: { VAR1: 'value1', VAR2: 'value2' }
 * @param {EnvVarObject} envObject
 * @return {Object<string, string>}
 */ 
```
`typedef`s are in prior suggestion

forge/db/utils.js
```diff
@@ -153,6 +153,77 @@ function randomPhrase (wordCount = 3, minLength = 2, maxLength = 15, separator =
...
+function exportEnvVarObject (envObject) {
+    // Check for any hidden env vars. These are objects with a 'hidden' property set to true.
+    // If so, we replace the object with just the value.
+    // TODO: handle encrypted values
+    const result = {}
+    for (const envVar of Object.keys(envObject)) {
+        if (Object.hasOwn(envObject[envVar], 'hidden') && envObject[envVar].hidden) {
```
Steve-Mcl: Unless I am missing something, this could have undesirable affects if `hidden` is anything other than `true`?

e.g.

```js
exportEnvVarObject({"ENV1":{value:1, hidden:true}})
// returns { ENV1: 1 } ✅

exportEnvVarObject({"ENV1": "1"})
// returns { ENV1: "1" } ✅

// however...

exportEnvVarObject({"ENV1":{value:1, hidden:false}})
// returns { ENV1: { value: 1, hidden: false } } ?

exportEnvVarObject({"ENV1":{value:1, hidden:undefined}})
// returns { ENV1: { value: 1, hidden: undefined } } ?

exportEnvVarObject({"fred":{value:1, hidden:null}})
// returns { fred: { value: 1, hidden: null } } ?

exportEnvVarObject({"ENV1":{value:1}})
// returns { ENV1: { value: 1 } } ?
```

knolleary:
```suggestion
        if (Object.hasOwn(envObject[envVar], 'hidden')) {
```

@Steve-Mcl I'll remove the `is true` part of the test, but leaving it as a check for the `hidden` property. There are no paths where the value can be an object without the `hidden` property currently - so this is good enough for now.

The next iteration, that adds encryption, will drive additional changes here anyway.

knolleary:
@Steve-Mcl applied the doc suggestions and tweaked the `hidden` property test. I think this is good to go.

forge/db/utils.js
```diff
@@ -153,6 +153,100 @@ function randomPhrase (wordCount = 3, minLength = 2, maxLength = 15, separator =
...
+        if (Object.hasOwn(envObject[envVar], 'hidden')) {
+            // The value has metadata - use the value only
+            result[envVar] = envObject[envVar].value
+        } else {
+            // The value is the bare value - use as-is
+            result[envVar] = envObject[envVar]
+        }
+    }
+    return result
+}
```
Steve-Mcl, line 248: It think there is one more edge case Nick. I did a quick test to confirm:

`exportEnvVarObject({"ENV1":{value:1}})` is not normalising the result like others - it returns `{ ENV1: { value: 1 } }`

image

May be related to test failure?

Steve-Mcl: For comparison:

```shell
> exportEnvVarObject({"ENV1":{value:1, hidden:true}})
{ ENV1: 1 }
> exportEnvVarObject({"ENV1": "1"})
{ ENV1: '1' }
> exportEnvVarObject({"ENV1":{value:1, hidden:false}})
{ ENV1: 1 }
> exportEnvVarObject({"ENV1":{value:1, hidden:undefined}})
{ ENV1: 1 }
> exportEnvVarObject({"fred":{value:1, hidden:null}})
{ fred: 1 }
> exportEnvVarObject({"ENV1":{value:1}})
{ ENV1: { value: 1 } }

```

knolleary: What scenario passes in `{"ENV1":{value:1}}`? I couldn't find any path that leads to that input.

I did say in my response:

> There are no paths where the value can be an object without the hidden property currently - so this is good enough for now.
>The next iteration, that adds encryption, will drive additional changes here anyway.

knolleary: Can't immediately see what a device-group test failure could be related to env vars. I am rerunning the tests.

Steve-Mcl: > I did say in my response:
> 
> > There are no paths where the value can be an object without the hidden property currently - so this is good enough for now.
> > The next iteration, that adds encryption, will drive additional changes here anyway.

I missed that, but ideally this would be handled now to avoid it being forgotten (and someone introducing a change elsewhere that hits it) - that said, not a hill to die on today ;)

Steve-Mcl (approved):

Steve-Mcl:
Tests pased this time Nick - merge?

PR #5679, opened 2025-06-24
Do not include full flow content in blueprint collection api

knolleary:
## Description

This removes the `flows` property from the `/api/v1/flow-blueprints`  api response, and modifies the db lookup to not return it either.

This property is only used when previewing the contents of a blueprint; and we have an api in place to load the contents of an individual blueprint. This updates the AssetDetailDialog component to dynamically load the blueprint content if needed.

I have verified the `export-public` endpoint does still do the right thing and includes the flows/modules properties as expected.

### DIFF
--- a/forge/ee/routes/flowBlueprints/index.js
+++ b/forge/ee/routes/flowBlueprints/index.js
@@ -34,7 +34,6 @@ module.exports = async function (app) {
         } else if (request.query.filter === 'inactive') {
             filter = { active: false }
         }
-
         if (request.query.team) {
             const team = await app.db.models.Team.byId(request.query.team)
             if (!team) {
@@ -45,7 +44,7 @@ module.exports = async function (app) {
             reply.send(flowTemplates)
         } else {
             // get all flow templates - typically for administration purposes
-            const flowTemplates = await app.db.models.FlowTemplate.getAll(paginationOptions, filter)
+            const flowTemplates = await app.db.models.FlowTemplate.getAll(paginationOptions, filter, { includeFlows: false })
             flowTemplates.blueprints = flowTemplates.templates.map(ft => app.db.views.FlowTemplate.flowBlueprintSummary(ft))
             reply.send(flowTemplates)
         }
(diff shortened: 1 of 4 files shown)


forge/ee/routes/flowBlueprints/index.js
```diff
@@ -4,7 +4,12 @@ module.exports = async function (app) {
     app.config.features.register('flowBlueprints', true, true)
 
     app.get('/', {
-        preHandler: app.needsPermission('flow-blueprint:list'),
+        preHandler: async (request, reply) => {
+            if (request.session.ownerType === 'project' || request.session.ownerType === 'device') {
```
cstns, line 8: could devices and instances access this endpoint?

forge/ee/routes/flowBlueprints/index.js
```diff
@@ -45,14 +49,19 @@ module.exports = async function (app) {
...
-            const flowTemplates = await app.db.models.FlowTemplate.getAll(paginationOptions, filter)
+            const flowTemplates = await app.db.models.FlowTemplate.getAll(paginationOptions, filter, { includeFlows: false })
             flowTemplates.blueprints = flowTemplates.templates.map(ft => app.db.views.FlowTemplate.flowBlueprintSummary(ft))
             reply.send(flowTemplates)
         }
     })
 
     app.get('/:flowBlueprintId', {
-        preHandler: app.needsPermission('flow-blueprint:read'),
+        preHandler: async (request, reply) => {
```
cstns, line 59: same here? Not that there's anything wrong with checking it's just that I don't see the point.

cstns (approved):

cstns:
Not 100% sure why the checks on the ownerType are required but other than that looks good!

forge/ee/routes/flowBlueprints/index.js
hardillb: Devices and Instances are accessing this endpoint using the new BluePrint library plugin -> https://github.com/FlowFuse/nr-launcher/pull/345

forge/ee/routes/flowBlueprints/index.js
hardillb: As above, this is needed for the Add BluePrint to existing instance work

knolleary:
@cstns last line of the PR description:

> It also updates the GET /flow-blueprints and GET /flow-blueprints/:id endpoints to allow access by instance/device tokens - so they can pull the blueprints down

cstns:
somehow, I managed to miss it entirely

PR #5688, opened 2025-06-26
 Improve recently modified remote instances sorting

cstns:
## Description

Sort instances first by their state in the following order: Error, Running, Suspended.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5671

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #5706, opened 2025-07-01
Revert " Improve recently modified remote instances sorting"

cstns:
Reverts FlowFuse/flowfuse#5688


cstns:
The query is failing because I did not escape the table name

Steve-Mcl (approved):

PR #5716, opened 2025-07-02
Reduce the rate of calls to instance name check API

### LINKED ISSUE #5703: Create instance incorrectly reports instance name is in use
### Current Behavior

As user types name, they hit `429` which incorrectly states instance name is in use
Image

### Expected Behavior

should not report incorrect status of instance name usage

### Steps To Reproduce

Start typing a name, eventually you hit a 429 and it reports "Instance name already in use" when it is not.

While I realise this example is contrived, I did actually hit this while typing (and spell correcting) a reasonable instance name.

We may need to disable name check request until creation -or- implement a "check availability" button -or- relax the rate limit on the route.

### Environment

- FlowFuse version: FFC Live
- Node.js version: NA
- npm version: NA
- Platform/OS: NA
- Browser: NA

hardillb:
fixes #5703

## Description

This increases the pause time in typing a name before sending to the API (from 500ms to 1000ms).

Also add a different message for the rate limit to the name in use error.

## Related Issue(s)

#5703

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/multi-step-forms/instance/steps/InstanceStep.vue
+++ b/frontend/src/components/multi-step-forms/instance/steps/InstanceStep.vue
@@ -293,12 +293,20 @@ export default {
                             this.errors.name = null
                         })
                         .catch(e => {
-                            this.errors.name = 'Instance name already in use.'
+                            if (e.status === 409) {
+                                if (e.response?.data?.error === 'name in use') {
+                                    this.errors.name = 'Instance name already in use.'
+                                } else if (e.response?.data?.error === 'name not allowed') {
+                                    this.errors.name = 'Instance name not allowed.'
+                                }
+                            } else if (e.status === 429) {
+                                this.errors.name = 'Name check rate limit exceeded'
+                            }
                         })
                 } else {
                     this.errors.name = 'Invalid character in use.'
                 }
-            }, 500)
+            }, 1000)
         },
         errors: {
             deep: true,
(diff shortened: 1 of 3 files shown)


frontend/src/components/multi-step-forms/instance/steps/InstanceStep.vue
```diff
@@ -293,12 +293,20 @@ export default {
...
+                                }
+                            } else if (e.status === 429) {
+                                this.errors.name = 'Name check rate limit exceeded'
+                            }
                         })
                 } else {
                     this.errors.name = 'Invalid character in use.'
                 }
-            }, 500)
+            }, 1000)
```
Steve-Mcl:
```suggestion
            }, 750)
```

It is quite easy to reach the "next" button with 1 sec wait. Lets trim this a bit.

hardillb:
I've implemented both suggestions @Steve-Mcl pleaes re-test

> NOTE: I did the choice of stopping input when x-ratelimit-remaining == 1. This prevents the actual rate limit being hit and permits the user toy type next character

I've done this on 0 because the next request doesn't fail, so you should be able to go all the way to the limit, not stop early

Steve-Mcl:
will merge after main is merged (but please feel free to merge if I forget this tab is open amongst the other 99 tabs before main is merged in and tests are done!)

PR #5723, opened 2025-07-02
Assistant assets endpoint

### LINKED ISSUE #5722: Endpoint for loading assets for FF Assistant
### Description

As discussed in the parent issue, this task if for the addition of an endpoint in the assistant routes that provide a means of delivering assets to the assistant.

Criteria:
- [ ] Should cache files for "some time") to minimise bursts
- [ ] Should utilise protections already in place for existing assistant route (token check, instance/device id check)
- [ ] Unit tests
   - [ ] cache is utelised
   - [ ] cache is cleared after set duration
   - [ ] bad "asset name" is rejected

### Epic/Story

_No response_

Steve-Mcl:
## Description

TLDR;
* Refactors header and config logic for re-use in additional endpoint
* Adds endpoint `/assets/*` to `/api/v1/assistant`
   * Full URL:  `/api/v1/assistant/assets/*`
   * Sets `schema.hide: true` while the feature set grows (same as existing endpoint)
* Adds LRUCache module and asset caching and endpoint 
   * Selected an arborary number of `max: 100` (this could be revised down to something closer to 5 if desired - since we dont have many resources today, but that may change with new features being added)
   * Set `ttl` to 30 mins with a hard expiry (i.e. subsequent `GET` requests will not reset ttl. This is to permit a refresh when new assets are pushed to the upstream resource.
   * Unit tests

### Tests added to `test/unit/forge/routes/api/assistant_spec.js`
```
▼ Assistant API
  ▼ assets endpoint
    ✔ should return 401 for anonymous access
    ✔ should return 401 for random token
    ✔ should return 401 for user token
    ✔ should return 400 for missing asset path
    ✔ should return 400 for asset path starting with slash
    ✔ should handle upstream error
    ✔ should fetch and cache asset for device token
    ✔ should fetch and cache asset for instance token
    ✔ should fetch fresh after cache expiration
```

* ## Related Issue(s)

closes #5722

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/api/assistant.js
+++ b/forge/routes/api/assistant.js
@@ -11,10 +11,13 @@ const { LRUCache } = require('lru-cache')
  */
 module.exports = async function (app) {
     const assetCache = new LRUCache({
-        max: app.config.assistant?.service?.assetCache?.max || 100,
-        ttl: app.config.assistant?.service?.assetCache?.ttl || 30 * 60 * 1000, // Defaults to 1/2 hour cache for assets - enough to handle bursts
+        max: app.config.assistant?.assetCache?.max || 100,
+        ttl: app.config.assistant?.assetCache?.ttl || 30 * 60 * 1000, // Defaults to 1/2 hour cache for assets - enough to handle bursts
         updateAgeOnGet: false // do not update the age on get, we want it to expire after the original ttl
     })
+
+    // decorate the app with the asset cache
+    app.decorate('assistant', { assetCache })
 
     // Get the assistant service configuration
     const serviceUrl = app.config.assistant?.service?.url
--- a/forge/routes/api/assistant.js
+++ b/forge/routes/api/assistant.js
@@ -1,4 +1,5 @@
 const { default: axios } = require('axios')
+const { LRUCache } = require('lru-cache')
 
 /**
  * Assistant api routes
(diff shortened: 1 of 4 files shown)


forge/routes/api/assistant.js
```diff
@@ -31,6 +50,82 @@ module.exports = async function (app) {
...
+     * Assets are cached in memory to reduce load during bursts.
+     * The assets are fetched from the assistant service and cached for 30 minutes.
+     */
+    app.get('/assets/*', {
+        schema: {
+            hide: true, // dont show in swagger
+            response: {
+                200: {
+                    type: 'string'
+                },
```
knolleary: Given the assets could be binary files, not sure we should declare a `string` type here; not sure what fastify will do. If we aren't including in the api docs, then there's not much benefit to having the response object here.

knolleary (commented):
Looks good - one questin/comment on the schema of the asset endpoint to sort out before approval.

forge/routes/api/assistant.js
Steve-Mcl: Fair point.

Since fastify is smart enough to handle Buffer/ArrayBuffer/etc responses directly when using `reply.send(...)` the schema could be misleading / aa bit pointeless

NOTE: I have pulled JSON and binary data to the assistant in manual testing with this schema in place - so yeah, bit misleading & worthy of removal.

knolleary (dismissed):

knolleary:
Rerunning tests as there was a postgres test failure which didn't immediately appear related.

PR #5738, opened 2025-07-04
Remove the fix for the HS chat pill

cstns:
## Description

Removes the margin bottom applied to various elements to make room for the HS Chat pill before it was movable

## Related Issue(s)

n/a

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/DevicesBrowser.vue
+++ b/frontend/src/components/DevicesBrowser.vue
@@ -1,9 +1,5 @@
 <template>
-    <!-- set mb-14 (~56px) on the form to permit access to kebab actions where hubspot chat covers it -->
-    <div
-        class="space-y-2 mb-14"
-        data-el="devices-section"
-    >
+    <div class="space-y-2" data-el="devices-section">
         <ff-loading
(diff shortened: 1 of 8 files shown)


Steve-Mcl:
@cstns I am happy to remove this but I cannot verify if the HS widget is movable in all browsers.

@joepavitt are you ok with this?

For reference, any page with a table that added a kebab to the RH side had to be bottom padded to permit access to the kebab menu. This was because the HS widget was hard fixed in the bottom RH corner and users could not operate on items at the end of a table. Since we can now move the widget these paddings are not absolutely necessary. My only pause here is if we remove these paddings a user now MUST move the widget out of the way (annoying, but not the end of the world). Also, I have not checked to see if all other browsers (Safari, FF, other) allow widget to be moved.

Steve-Mcl (approved):
Approved as far as I am concerned, but will leave merge for you / @joepavitt in case there is an edge case or preference I am not aware of.

joepavitt:
No need for my input here. Carry on.

Issue #5718, opened 2025-07-02
MVP Frontend for FlowFuse Tables


joepavitt:
### Description

Development work to build a first iteration of the front-end for FlowFuse Tables, our new Database Offering.

Designs can be found in https://github.com/FlowFuse/flowfuse/issues/5717

#### MVP Scope

- "Explorer" view
    - User should be able to view tables, add a table, and view content of a table
- "Credentials" view
    - User should be presented with everything they need in order to connect/read/write to the relevant tables

#### Follow Up Tasks

These pieces will then be explored as future items

- https://github.com/FlowFuse/flowfuse/issues/5797
- Inline editing of tables in "Explorer" view
- SQL Query" Tab
    - Assistant in SQL Query Tab

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

cstns:
@hardillb Is there an API I can play with or should I mock data for the time being?

hardillb:
Work is currently in the [flowfuse-database ](https://github.com/FlowFuse/flowfuse/tree/flowfuse-database)branch, I'll push updates at end of today

Issue #5717, opened 2025-07-02
Design UI/UX for the FlowFuse Tables offering


joepavitt:
### Description

Issue to track the design work required for developing a front-end for FlowFuse Tables

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Figma File: https://www.figma.com/design/rMB2w5ZZUtTC0nxbwLYKJY/FlowFuse-Platform?node-id=6269-126999&t=4YkaxJ6lnMcNEfLR-1

### Explorer View

![Image](https://github.com/user-attachments/assets/8866bd35-3bdc-499a-bbc6-f4e4b38ae2e0)

### SQL Editor View

![Image](https://github.com/user-attachments/assets/43f7353f-fdc8-4165-a08d-9241674915f6)

joepavitt:
@hardillb can you remind me, do I need to add in designs for clients/credentials configuration? anything on a "Settings" type page for example?

hardillb:
@joepavitt PoC code generates a single credentials bundle (username, password, host, port, database name)  for the database that needs to be presented.

Stretch would be to be able to generate extra users (and **super** stretch would be role management e.g. which tables they can read/write to)

joepavitt:
<img width="1440" height="1024" alt="Image" src="https://github.com/user-attachments/assets/989ccf44-b0b6-4ee1-88b4-52208d305a0c" />

First iteration for credentials page. This will evolve if we introduce the stretch goals detailed [above](https://github.com/FlowFuse/flowfuse/issues/5717#issuecomment-3027508797)

Intention here is that the "copy" button will show on hover for all fields.

joepavitt:
Dev work has started on the MVP here, and the design work covers several iterations ahead. As such, can close this issue out as completed

PR #5760, opened 2025-07-08
Add the tables page and menu entry

cstns:
## Description

scaffolding for future development on the new tables offering.

Created a new tables page and menu entry,

## Related Issue(s)

part of https://github.com/FlowFuse/flowfuse/issues/5718

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
@cstns where can I find instructions on how to get the database stuff setup and running locally?

cstns:
I don't think there is one just yet. I spun up a postgres instance using docker and added the following to my `flowforge.local.yml`:

```yml
tables:
  enabled: true
  driver:
    type: postgres-localfs
    options:
      database:
        user: <user>
        password: <pw>
        host: localhost
        port: 5432
        database: <db>
        ssl: false
```

hardillb:
@joepavitt 

```
docker run --rm -it -p 5432:5432 -e POSTGRES_USER=root -e POSTGRES_PASSWORD=password --name postgresql -v `pwd`/db:/var/lib/postgresql/data  postgres:17
```
- add the following to your `flowfuse.local.yml`
```
tables:
  enabled: true
  driver:
    type: postgres-localfs
    options:
      database:
        user: root
        password: password
        host: localhost
        port: 5432
        database: postgres
        ssl: false
```

Shout if you have any problems

hardillb:
There is also the stub driver that has no external deps that can be used to quickly test the UI

```
tables:
  enabled: true
  driver:
    type: stub
```

joepavitt:
Running the above and seeing:

```
[2025-07-21T10:44:49.867Z] ERROR: Database for team KbQ8Wa7EdZ does not exist {"reqId":"req-2q","req":{"method":"GET","url":"/api/v1/teams/KbQ8Wa7EdZ/databases","hostname":"localhost:3000","remoteAddress":"::1","remotePort":63439},"res":{"statusCode":500,"request":{"user":"joepavitt","url":"/api/v1/teams/KbQ8Wa7EdZ/databases","method":"GET","remoteAddress":"::1","remotePort":63439,"team":"KbQ8Wa7EdZ"}}}
    err: {
      "type": "Error",
      "message": "Database for team KbQ8Wa7EdZ does not exist",
      "stack":
          Error: Database for team KbQ8Wa7EdZ does not exist
              at Object.getDatabases (/Users/joepavitt/Documents/flowforge/development/flowforge-dev-env/packages/flowfuse/forge/ee/lib/tables/drivers/postgres-localfs.js:38:19)
              at async Object.<anonymous> (/Users/joepavitt/Documents/flowforge/development/flowforge-dev-env/packages/flowfuse/forge/ee/routes/tables/index.js:48:23)
    }
```

Already spoken with Ben who is going to get a fix in.

joepavitt:
Screenshot 2025-07-21 at 11 55 57

Our offering is called Tables, then description is all about "Tables" being managed, but I'm asked to name my Database? Is this creating a Table, or is this the container for those tables? Not clear to me. Think a description alongside the "Name" field would be useful

joepavitt:
Screenshot 2025-07-21 at 11 57 50

The CTA here isn't clear. Less of an issue once we had a "Add Table" button, but in the mean time, I think the "Select a table to get going" should provide details on the fatc user's need to manage/create tables and their contents from within a Node-RED flow.

For my own context, how much work will it require to add an "Add Table" button here?

cstns:
> Our offering is called Tables, then description is all about "Tables" being managed, but I'm asked to name my Database? Is this creating a Table, or is this the container for those tables? Not clear to me. Think a description alongside the "Name" field would be useful

There are going to be a lot of these popping up, the decision to name the offering tables is the root cause of it. In this particular case you're actually creating a database, it felt inappropriate to call it add a table at this point because you're going to be able to actually add actual tables to the database on a later iteration.

> The CTA here isn't clear. Less of an issue once we had a "Add Table" button, but in the mean time, I think the "Select a table to get going" should provide details on the fatc user's need to manage/create tables and their contents from within a Node-RED flow.

Happy to change the CTA, nothing better came to mind. There seems to be an alignment issue going on, did not see that on my end

joepavitt:
Where does the name of the Database name get used? I don't see it in "Credentials" or in the Tables view?

joepavitt:
Alignment issue is misleading in my screenshot, i didn't grab the whole page

cstns:
> Where does the name of the Database name get used? I don't see it in "Credentials" or in the Tables view?

I may have forgotten to expose it, focused on the credentials received from the BE. I assume the name will be used when users will have multiple databases setup and would need to switch and differentiate between them?

It was a mandatory parameter when creating a DB on the API endpoint

joepavitt:
Screenshot 2025-07-21 at 12 20 54

My database name seems to just be my team id, even though that isn't what I typed in?

cstns:
yes, those are from the credentials response, with the name for the database that was created behind the scenes in order to connect to it via NR or other clients. I forgot to expose the name/title given to the database/tables offering when creating it. I may have postponed it exactly because of this, the confusion it generated, and just forgot to get back to it.

joepavitt:
> with the name for the database that was created behind the scenes in order to connect to it via NR or other clients.

Right, so if a different name gets created behind the scenes, then why do I need to provide a text-based name that doesn't get shown or used anywhere?

cstns:
It's no being used _now_, but it will be used when we have multiple offerings for databases. The not being shown anywhere is an omission on my end, we should definitely expose it somewhere (accepting suggestions on where and how)

joepavitt:
Before I can give an answer on the _where_ to show this, what is the difference between the Database Name shown in "Credentials" and the name defined by the user?

hardillb:
@joepavitt The intent for the user supplied name is to allow a meaningful name to be given to the database for if/when we have more than 1, the `database name` in the credentials is a technical requirement for the postgres driver.

cstns:
The name defined by the user is to differentiate it between multiple databases (when the time comes and we have multiple ones i guess). It can be thought of as a label/title. It's a column on the Tables table in the flowfuse database, alongside credentials teamId etc. 

The name from the credentials entry is the db name users are supposed to use when connecting to the database via pg clients (NR included). It's the name of the actual database that is getting created in Postgres that houses all the tables and data.

joepavitt:
@cstns once I've got a table created and navigate to `Tables`, I expect that it'll load the first table in that list by default (saves a click), but then also have some highlighting or indication in the table list that I have an actively selected table once I do manually select it

Note the "Selected" state in the UI Designs here: https://github.com/FlowFuse/flowfuse/issues/5717

cstns:
added functionality to pre-select the first table after they get loaded, fixed missing highlight on the selected table

joepavitt:
For the database name, can we auto-generate the database when they've selected the type, then just used the auto-generated Database Name under the covers too? It's just an extra unnecessary step currently whereby user's define something they don't need, nor can they use.

If/when we support multiple databases within a team, then we can expose the `name` as an editable field.

cstns:
sure thing! The UI will just need some tlc but i can use the instance name generator to create one and just send it through to the backend and just skip that step

joepavitt:
> The UI will just need some tlc but i can use the instance name generator to create one and just send it through to the backend and just skip that step

That also seems excessive though. If the server-side is already generating a name, then why do we need to pass a name at all in the API? I'd vote to remove that requirement at the API-level (cc @hardillb)

hardillb:
> That also seems excessive though. If the server-side is already generating a name, then why do we need to pass a name at all in the API? I'd vote to remove that requirement at the API-level (cc @hardillb)

It will take a little unpicking as it's marked as not null in the FF database table, so it is expected all the way down. But I can see about removing it

cstns:
missed the last comments and sent bypassed the restriction by using an auto-generated name using the existing NameGenerator. Can be easily reverted though.

joepavitt:
> It will take a little unpicking as it's marked as not null in the FF database table, so it is expected all the way down. But I can see about removing it

Where does the "Database Name" come from in the "Credentials" page then? That's what i've had to use in the Postgres node in Node-RED to connect.

hardillb:
>Where does the "Database Name" come from in the "Credentials" page then? That's what i've had to use in the Postgres node in Node-RED to connect.

I generate that, and to keep things simple I used the team hashid (because team slugs can change). We need a unique identifier at the Postgres level.

The name field was to give the user a choice in display name so it could have a meaningful name to them in the FF UI. I was expecting it to be shown when interacting with the Database pages.

joepavitt:
> I generate that, and to keep things simple I used the team hashid (because team slugs can change). We need a unique identifier at the Postgres level.

If we only have one database per team, then this makes sense. For the "Database Name" though in Node-RED nodes, what needs to happen for us to use the provided name from the user? That seems to be a cleaner idea here?

hardillb:
>  For the "Database Name" though in Node-RED nodes, what needs to happen for us to use the provided name from the user? That seems to be a cleaner idea here?

These names need to be globally unique and not contain things like spaces or punctuation. If we moved to user supplied then we would need to treat them like NR instance names and enforce a bunch of rules and check that they are globally unique across the whole platform.

joepavitt:
Conclusion from meeting:

- Don't show user option to define Database name
- Server-side, use the auto-generated "ID" as the "Database Name"
    - This will then be an editable field in a _future_ iteration whereby teams can have multiple databases.
- Rename "Database Name" to "Database ID" in the Credentials Tab

joepavitt:
Can we add a loading state after selecting the database type please? The API call took 13 seconds the first time for me, 4 seconds for another team.

joepavitt:
Updating the empty state CTA to this:

Screenshot 2025-07-22 at 10 47 18

joepavitt (approved):

joepavitt:
All good from my side, holding off on the merge as not sure on the dependency of other PRs to get into `main`

hardillb:
@joepavitt this merges down into the API PR, so we can merge this now and I'll get the other merged to main (after the point release, just to limit churn)

cstns:
where can i test the pare on staging? is there a pr open?

hardillb:
@cstns this will not be in pre-staging, but in the real staging env (forge.flowfuse.dev) but not until the `flowfuse-database` branch get's merged and then only after we have configured it.

I will let you know once it is all up and running.

Issue #4911, opened 2024-12-13
Timeseries Database


joepavitt:

joepavitt:
@ZJvandeWeg agreed to explore partnership opportunities here to supplement a FF database "offering"

hardillb:
Should MVP also target Self Hosted Licensed FF instances?

knolleary:
Not a hard requirement for MVP if it helps ship quicker to FFC - but we should ensure what we do has a route to being self-hosted. Prioritise k8s > docker > localfs.

hardillb:
Looking at what metering options are available for PostgreSQL.

It looks like we can determine a given database's size easily enough with `select pg_database_size('databaseName')`
Not sure what else we can easily measure with standard PostgeSQL

gstout52:
@hardillb Some things that would be nice to track as we use this MVP to determine a model for a more complete feature:

- Compute used
- Memory used
- Reads, writes, queries
- Data ingested
- Database size

gstout52:
@ZJvandeWeg I would like to offer the beta MVP of this on the Pro tier as well as Enterprise. As we discussed last week, we will revisit the pricing model for this based on what we learn about usage in the initial period following the release. Can we include the feature for the Pro tier?

ZJvandeWeg:
> What are they trying to achieve: I want to connect, extract the data from the shopfloor, and integrate it with whatever the use case demands. I want to optimize things that can have a $100k/year impact.

@gstout52 No, let's start with Enterprise. That part you wrote describes well why it should be Enterprise.

joepavitt:
The plan I had put forward is that the [beta goes out for free to everyone](https://github.com/FlowFuse/website/pull/3393), then we scale it back to a specific tier when we officially take it out of beta. It gets us usage data, gives us feedback around UX, and helps us gauge popularity far better.

Having this only on Enterprise gives us no data about usage, no visibility to trial users, etc. Enterprise customers are not, regularly, working on greenfield projects. They're not going to migrate existing projects over to our database, especially if it's branded as a "beta" feature. Shipping this to Enterprise only is a waste.

The better approach here is to throttle with usage limits by tier. It offers an incentive for growth, customers wanting to use this at scale still have to pay for it, but it opens up an opportunities for mass adoption as we could be offering a full MING-esque stack as a "starter" deal (throttled in appropriate places accordingly). Self-serve MRR isn't going to grow if all we are offering for self-serve is cloud-hosted Node-RED.

[...]

PR #5764, opened 2025-07-08
Flowfuse Tables API and backend Driver

hardillb:
part of #4911 
## Description

Adds API for managing Team Tables databases and the backend driver to provision these databases

## Related Issue(s)

#4911 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
Being quite an extensive PR I have first taken a look-see for any things that sticks out before doing a full code review.

### UI
I am not really reviewing UI (since that was a separate merged PR) but this did stand out. Password in plain view on page: 
image

### DB
This PR will store the DB password in unencrypted format. While I realise we are presenting this to the user unencrypted, it does go against "data at rest security". This may not be super critical due to DB accessibility on the platform in our case at FFC, but for self hosted this may be a no-go due to regulations around this area for their industry (hippa?) and their architecture (e.g. location of database)?

Question: Why is the Username `<teamid>.<teamid>` for supabase but just `<teamid>` for localfs?

hardillb:
> Question: Why is the Username <teamid>.<teamid> for supabase but just <teamid> for localfs?

This is a requirement for supavisor to know which tenant to direct the connection to, if/when we allow more users to be created then they will all be `username.teamid`

The passwords are fixed and stored both in supavisor, and the backend postgres instance (as well in the FF db), we can't make them ephemeral easily.

Steve-Mcl:
> The passwords are fixed and stored both in supavisor, and the backend postgres instance (as well in the FF db), we can't make them ephemeral easily.

That's fair enough but from the point of view of data at rest, if our database or a customer's database ever got dumped there is a list of passwords in plain text. My point is I think we should be encoding them before storing them. Yes of course they will be decrypted for displaying to the user but this will be passed via https and the key would be separate from the dump (in env or key store for example). And if we make the UI, display them with stars and have a click to reveal or click to copy. This will be better.

forge/db/migrations/20250707-01-EE-add-tables.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * Add FF Tables table
+ */
+const { DataTypes } = require('sequelize')
+
+module.exports = {
+    up: async (context) => {
```
Steve-Mcl, line 11:
```suggestion
    /**
     * upgrade database
     * @param {QueryInterface} context Sequelize.QueryInterface
     */
    up: async (context) => {
```

forge/db/migrations/20250707-01-EE-add-tables.js
```diff
@@ -0,0 +1,45 @@
...
+            },
+            credentials: {
+                type: DataTypes.TEXT,
+                set (value) {
+                    this.setDataValue('credentials', JSON.stringify(value))
+                },
+                get () {
+                    const rawValue = this.getDataValue('credentials') || '{}'
+                    return JSON.parse(rawValue)
+                }
```
Steve-Mcl: Migrations are for schema definition


```suggestion
                type: DataTypes.TEXT
```

forge/db/migrations/20250707-01-EE-add-tables.js
```diff
@@ -0,0 +1,45 @@
...
+            },
+            meta: {
+                type: DataTypes.TEXT,
+                set (value) {
+                    this.setDataValue('meta', JSON.stringify(value))
+                },
+                get () {
+                    const rawValue = this.getDataValue('meta') || '{}'
+                    return JSON.parse(rawValue)
+                }
```
Steve-Mcl: Migrations are for schema definition


```suggestion
                type: DataTypes.TEXT
```

forge/ee/db/views/Table.js
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    async function tables (tables) {
+        if (tables) {
+            return await Promise.all(tables.map(table))
+        } else {
+            return []
+        }
+    }
```
Steve-Mcl: Why are these async? `toJSON()` is a synchronous function
Also, to avoid confusion - I suggest we dont name the param the same as the functions (potential shadowing?)


```suggestion
    function table (tableInstance) {
        if (tableInstance) {
            const result = tableInstance.toJSON()
            result.id = result.hashid
            delete result.hashid
            return result
        } else {
            return null
        }
    }

    function tables (tableInstances) {
        if (tableInstances) {
            return tableInstances.map(table)
        } else {
            return []
        }
    }
```

forge/ee/lib/tables/drivers/postgres-supavisor.js
```diff
@@ -0,0 +1,298 @@
+const axios = require('axios')
+const pg = require('pg')
+// const local = require('./postgres-localfs')
```
Steve-Mcl:
```suggestion
```

forge/ee/lib/tables/drivers/postgres-supavisor.js
```diff
@@ -0,0 +1,298 @@
...
+                await adminClient.query(`REVOKE connect ON DATABASE "${team.hashid}" FROM PUBLIC;`)
+                const options = {
+                    host: this._options.backend.host,
+                    port: this._options.backend.port,
+                    ssl: this._options.backend.ssl,
+                    database: team.hashid,
+                    user: this._options.backend.user,
+                    password: this._options.backend.password
+                }
+                const teamClient = new pg.Client(options)
```
Steve-Mcl: This same code is duplicated in 4 places.
I feel this should be extracted to a reusable/testable function.
Additionally, being a shared lib function would permit us to mock `pg.create` and do proper unit testing on the drivers

forge/ee/lib/tables/drivers/postgres-localfs.js
```diff
@@ -0,0 +1,241 @@
...
+            await adminClient.query(`REVOKE connect ON DATABASE "${team.hashid}" FROM PUBLIC;`)
+            const options = {
+                host: this._options.database.host,
+                port: this._options.database.port,
+                ssl: this._options.database.ssl,
+                database: team.hashid,
+                user: this._options.database.user,
+                password: this._options.database.password
+            }
+            const teamClient = new pg.Client(options)
```
Steve-Mcl: This same code is duplicated in 4 places.
I feel this should be extracted to a reusable/testable function.
Additionally, being a shared lib function would permit us to mock `pg.create` and do proper unit testing on the drivers

forge/ee/routes/tables/index.js
```diff
@@ -0,0 +1,303 @@
...
+            response: {
+                200: {
+                    type: 'array',
+                    items: {
+                        type: 'object',
+                        properties: {
+                            name: { type: 'string' },
+                            schema: { type: 'string' }
+                        }
+                    }
```
Steve-Mcl: I dont think this should return an array as it may turn out we need meta/pagination when the start getting too large.
Changing this to an object now avoids any future duck type checking awkwardness

forge/ee/routes/tables/index.js
```diff
@@ -0,0 +1,303 @@
...
+                '4xx': {
+                    $ref: 'APIError'
+                },
+                500: {
+                    $ref: 'APIError'
+                }
+            }
+        }
+    }, async (request, reply) => {
+        // paginate the list of tables
```
Steve-Mcl: There is no pagination

forge/ee/routes/tables/index.js
```diff
@@ -0,0 +1,303 @@
...
+                    tableName: { type: 'string' }
+                }
+            },
+            response: {
+                200: {
+                    type: 'array',
+                    items: {
+                        type: 'object',
+                        additionalProperties: true // allow any properties in the table rows
+                    }
```
Steve-Mcl: Although this route is designed to grab a sample, we may wish to include meta or even pagination at some future point.
Changing this to an object now avoids any future duck type checking awkwardness

Steve-Mcl:
Additional to the review - (likely warranting separate frontend iteration issue/PR - but mentioned here to get it documented (we can raise an issue from this comment if it is to be a follow up)

Any reasonable sized table with more than a few words, the table view quickly becomes a mess:

Beekeeper_Studio_HmDfBHNKLQ


Tagging @cstns for his take

Steve-Mcl:
I think would could have unit tests for the drivers with a little bit of a rejig. 

As mentioned in the review comment, I think if we extract the code that creates the actual driver (`pg.create`) to return an abstracted interface containing `connect`, `end`, `query` we can add unit tests for any driver and spy/stub the calls to verify the correct calls were made and correct parameters were used.   

If you need a first test to get this moving, I would happily push a change Ben?

hardillb:
> If you need a first test to get this moving, I would happily push a change Ben?

Please Steve, anything to up the coverage on the 2 postgres drivers would be great

cstns:
> Any reasonable sized table with more than a few words, the table view quickly becomes a mess:

On the first pass I only used primitives for my data set which wasn't optimal. I'll address this on the [follow-up](https://github.com/FlowFuse/flowfuse/issues/5797) where we allow users to create tables from the UI, which hopefully I'll get to by EOD

Steve-Mcl:
@hardillb I added a full suite of unit tests for both drivers.

There was a little rejig required (not much):
* Moved `adminClient` to an internal prop so I could stub/spy it
* extracted the pg.Client creation to a separate lib function (to make it testable and dry)
*

Additionally, I took the liberty of escaping the parameters of the raw SQL DDL queries using the `pg` built in utilities `eascapeLiteral` and `escapeIdentifier` - I realise the values are not user settable but it further minimises risk from any accidental or malicious configuration values doing damage.

### Test added:

```log
  Tables: helper lib for pg
    ✔ should create a new pg Client with correct options

  Tables: Postgres LocalFS Driver
    init
      ✔ should throw if no database options provided
      ✔ should connect admin client and log info
      ✔ should log error if connect fails
    shutdown
      ✔ should end admin client and log info
      ✔ should log debug if end fails
    getDatabases
      ✔ should return tables for team
      ✔ should return empty array if no tables
    getDatabase
      ✔ should return table if found
      ✔ should throw if not found
    createDatabase
      ✔ should create a new table with correct options
      ✔ should throw if database already exists
    destroyDatabase
      ✔ should destroy database, user, and role if db exists
      ✔ should throw if db does not exist in Table
      ✔ should throw if db does not exist in Postgres
    getTables
      ✔ should return mapped tables if found
      ✔ should return empty array if no tables
      ✔ should throw if db does not exist
    getTable
      ✔ should return mapped columns if found
      ✔ should throw if table is not found
      ✔ should throw if db does not exist
    getTableData
      ✔ should return rows if found
      ✔ should return empty array if no rows
      ✔ should throw if db does not exist

  Tables: Postgres Supavisor Driver
    init
      ✔ should throw if no database options provided
      ✔ should connect admin client and log info
      ✔ should log error if connect fails
    shutdown
      ✔ should end admin client and log info
      ✔ should log debug if end fails
    getDatabases
      ✔ should return tables for team
      ✔ should return empty array if no tables
    getDatabase
      ✔ should return table if found
      ✔ should throw if not found
    createDatabase
      ✔ should create a new table with correct options
      ✔ should throw if database already exists
      ✔ should throw if datname already exists
    destroyDatabase
      ✔ should destroy database, user, and role if db exists
      ✔ should throw if db does not exist in Table
      ✔ should not throw error if db does not exist in Postgres
    getTables
      ✔ should return mapped tables if found
      ✔ should return empty array if no tables
      ✔ should throw if db does not exist
    getTable
      ✔ should return mapped columns if found
      ✔ should throw if table is not found
      ✔ should throw if db does not exist
    getTableData
      ✔ should return rows if found
      ✔ should return empty array if no rows
      ✔ should throw if db does not exist


  48 passing (107ms)
```

Steve-Mcl:
Hundreds of test fails - most unrelated to work done 🫨 

I _think_ I know what I did wrong - 5 mins to fix 🤞

Steve-Mcl:
A few other things I missed in the initial review:

* No audit logging
  * Assuming we want audit logging - I am happy for this to be a raised as a sub task
* After writing tests, I note there is a difference between localfs and supavisor diver function `destroyDatabase()`
  *  localfs throws the error (like most of the functions do)
  *  supavisor swallows the error with a commented out console.log (this one is an oddity)
     * it looks like the `throw` inside the inner `try` was _meant_ to be passed up the stack? If yes, that would make the 2 functions match up.
  *  as it stands, the unit tests for `destroyDatabase` are very different due to the difference 
* I also noted an edge case that may bite us hard: If an error occurs part way though table creation, we end up in "no mans land"
  * Suppose `await this._adminClient.query(`CREATE DATABASE "${team.hashid}"`)` succeeds
    then an error occurs before `this._app.db.models.Table.create`
  * when `destroyDatabase()` is called, it will throw ```new Error(`Database ${databaseId} for team ${team.hashid} does not exist`)``` leaving the actual postgres database in place
  * And because `this._app.db.models.Table` doesn't exist, the UI will show the "New Database" form and permit a 2nd call to `createDatabase()` which will throw `new Error('Database already exists')` because the PG DB was created.
   * we should probably wrap the calls in transactions - if any error occurs, roll them back.

Steve-Mcl:
Phew - tests are passing and Codecov Report looks much better!

hardillb:
Will raise an issue to add audit logs

https://github.com/FlowFuse/flowfuse/issues/5816

hardillb:
@Steve-Mcl I've requested another review so I can merge after lunch please

Steve-Mcl (approved):
Approving this to get this on staging for testing and development purposes while acknowledging the issue about the "data at rest" issue (passwords are not encrypted in the database) for which I will raise a follow up issue.

NOTE: I wont merge this in case we decide we want to avoid writing a migration to achieve encrypted data at rest after the fact and squeeze it in to this (or a branch off this),

cstns:
> Additional to the review - (likely warranting separate frontend iteration issue/PR - but mentioned here to get it documented (we can raise an issue from this comment if it is to be a follow up)
> 
> Any reasonable sized table with more than a few words, the table view quickly becomes a mess:

fixed in https://github.com/FlowFuse/flowfuse/pull/5824

PR #5768, opened 2025-07-09
Encrypted hidden env vars when exporting snapshots

knolleary:
Fixes https://github.com/FlowFuse/security/issues/105

## Description

This ensures any hidden env vars in a snapshot are encrypted when the snapshot is exported. The values are encrypted with the snapshot credential secret - the same key used to encrypt the flow credentials.

The import snapshot path is also updated to decrypt the values on import.


hardillb:
End to end (instance -> git -> instance) appears to work and I see the encrypted values in git

Issue #5741, opened 2025-07-04
Update the product tours after the homepage change


cstns:
### Description

Manually triggering the tour works as expected because it redirects the user to the applications page.

The tours that were supposed to start after account creation still function, but the user would need to stumble across the applications page to start them.

Considering that the creation landing page is no longer the applications page that initiated the tours, we need to:

- Trigger the tours from the new homepage
- Include the new homepage in each tour

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

cstns:
cc @gstout52 for prioritization

joepavitt:
Just to nail down MVP so that we have _a_ product tour showing:

- Remove the Applications-page tour
- New Tour:
    - Highlight and explain "Hosted Instances"
    - Highlight and explain "Remote Instances"
    - Highlight newly created Hosted Instance
        -  Instruct to click the "Open Editor" button

PR #5769, opened 2025-07-09
Update the product tours

cstns:
## Description

- removed unused tours (`device-tour` and `trial-tour`) and updated the welcome tour to fit the context of the new team home page
- extracted the animation logic from the scrollToAndJiggleHighlight UX composable to reuse in the last step of the new tour
- removed previous tour triggers from the applications page and moved them to the global team home page
- fixed the manual trigger for starting the tour to correctly redirect to the home page
- added a new pulse-style animation to highlight the editor button, as the existing jiggle effect was not effective on this tertiary-styled button
- adapted the `welcome-tour` that covers two main concepts (Hosted and Remote Instances), highlighting the at-a-glance stats and most recently modified instances for each, while also pointing users to dedicated pages for more options
- the tour additionally introduces the Recent Activity section, demonstrates the global search functionality (including keyboard shortcuts and navigation), and concludes with a clear CTA for the user to open the editor on their newly created instance

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5741

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/PageHeader.vue
+++ b/frontend/src/components/PageHeader.vue
@@ -222,9 +222,7 @@ export default {
         },
         startWelcomeTour () {
             return this.$store.dispatch('ux/tours/resetTours')
-                // it's unfortunate that we can't redirect premium users straight to the application device page, but we
-                // don't have available applications at this moment in time so they'll get redirected twice
-                .then(() => this.$router.push({ name: 'team-home' }))
+                .then(() => this.$router.push({ name: 'team-home', params: { team_slug: this.team.slug } }))
                 .then(() => this.$store.dispatch('ux/tours/presentTour'))
         },
         toggleMobileTeamSelectionMenu () {


joepavitt (changes requested):
Screenshot 2025-07-14 at 10 28 55

The very last step doesn't align to the user's experience if there is not an Instance already present. In the self-hosted use case, that will not get created unless configured by the Platform admin.

Can we have a conditional that points them to "Create a Hosted Instance" instead if an Instance isn't available?

I've also pushed some trimming to the tour, as a lot of the home page and options are self-explanatory and don't need explicit call out.

cstns:
I split the last step into two and added some runtime logic to distinguish between cases where the user has hosted instances or not. It’s not perfect, but it works well enough.

I also fixed a bug in the manual tour trigger: previously, the team_slug was missing from the route when starting the tour from pages outside the team namespace.

joepavitt (approved):

PR #5778, opened 2025-07-11
Ensure device snapshot has palette.module list populated

### LINKED ISSUE #5777: Device generated snapshot produces incomplete module list
### Current Behavior

Reported by a customer; a pipeline configured to deploy from a Remote Instance to a Hosted Instance fails to include any additional modules added to the Remote instance.

The issue is because the remote instance snapshot isn't populating the `settings.settings.palette.modules` property; which is used as the source of truth when pushing to a hosted instance.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Fixes #5777 

When capturing a snapshot from a device, we need to ensure the `settings.palette.module` property is populated as that is used when pushing modules to a hosted instance.

This fixes scenarios where a pipeline pushes a snapshot from a remote instance to a hosted instance after changes are made to the installed modules on the remote instance.

Our typical scenario has been the other way around - push from hosted to remote; which works fine.

### DIFF
--- a/forge/db/controllers/ProjectSnapshot.js
+++ b/forge/db/controllers/ProjectSnapshot.js
@@ -1,6 +1,7 @@
 const { Op } = require('sequelize')
 
 const { getLoggers: getProjectLogger } = require('../../auditLog/project')
+const { BUILT_IN_MODULES } = require('../../lib/builtInModules')
 
 const DEVICE_AUTO_SNAPSHOT_LIMIT = 10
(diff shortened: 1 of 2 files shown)


Steve-Mcl (approved):

Steve-Mcl:
### pre-staging verification

On pre-staging, I setup a device, switched it to dev mode, added "cron-plus" via NR palette manager, created snapshot from device (set as target), created pipeline and pushed device --> instance :- cron-plus was carried over to instance ✅

image

PR #5779, opened 2025-07-11
Apply requireContact check to team upgrade

knolleary:
Fixes https://github.com/FlowFuse/security/issues/108

The 'requireContact' check is currently only made in the UI. This extends it to also apply at the API level - to ensure only admin users can action an upgrade to a team tier with 'requireContact' set.


cstns (approved):
LGTM

Issue #5784, opened 2025-07-16
Selecting item from app search no longer redirects


Steve-Mcl:
### Current Behavior

![Image](https://github.com/user-attachments/assets/d607b00b-582b-426a-9459-c4ec042957f2)

### Expected Behavior

Should navigate upon making a selection

### Steps To Reproduce

* On FFC "FlowFuse Team", search for Steve or Blueprint
* Click the application found
* Note nothing changes
* F5 refresh
* Now it changes

### Environment

- FlowFuse version: FFC live
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
On Chrome/Mac, this is working fine.

joepavitt:
Specifically this does occur when you're viewing an existing application, you search for a different application and try to navigate. It updates the URL in browser, without navigation or refreshing the page.

cstns:
Thanks for the head start, I was able to replicate it consistently!

PR #5786, opened 2025-07-17
Fix the Application page not reloading when route parameter change

cstns:
The root cause was that the application page did not trigger an update when the application or route changed.

## Description

While [FlowFuse/flowfuse#4950](https://github.com/FlowFuse/flowfuse/issues/4950) addressed updating the application when changing teams from the application page (by triggering a reload), it did not handle changes to the route parameters, specifically, the application ID.

This fix broadens the conditions under which the application is refreshed: previously it only reacted to team changes, but now it also responds to any changes in the route parameters, including both the team_slug and the application.id.

Also realized that the search icon was not displayed in the search bar once it was focused, fixed that as well.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5784

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/global-search/GlobalSearch.vue
+++ b/frontend/src/components/global-search/GlobalSearch.vue
@@ -10,10 +10,8 @@
                 </ff-button>
 
                 <div class="input-wrapper">
-                    <!--                    <SearchIconSolid class="ff-icon-sm mobile-search" @click="focusSearch" />-->
-
                     <transition name="primary-fade" mode="out-in">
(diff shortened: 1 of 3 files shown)


cstns:
The test failure was weird, couldn't re-trigger the failed tests so I merged main to get the same results and the tests seem to pass. The failed test was the code-cov report with 0.002% change diff which is weird because there were no backend changes

Issue #5785, opened 2025-07-16
Cannot copy instance to another application


Steve-Mcl:
### Current Behavior

Errors when you click Next for final step of Instance Duplication:
`Failed to create instance: Source Project Not in Same Application`

### Error toast

<img width="480" height="178" alt="Image" src="https://github.com/user-attachments/assets/57f87664-132e-40e0-8bfd-ab69c08d40eb" />

### In action
![Image](https://github.com/user-attachments/assets/c93d5736-f311-45d4-9ad8-ef60d073266f)

### Expected Behavior

Best case: Should be able to duplicate 
Worst case: Should not me permitted to select alternative application

### Steps To Reproduce

Select instance
Go to settings
Click "Duplicate Instance"
Select a different application
Go Next, Next

### Environment

- FlowFuse version: FFC Live
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

cstns:
While there must be valid reasons why the back end does not allow Instance duplication between multiple applications, the allowance of selecting/editing the application when duplicating an instance is on oversight on my end and can be easily fixed.

I'll adjust the front end to prevent changing applications and we can revert that when the time comes

cstns:
@knolleary and @hardillb, could you shed some light on the reasoning behind limiting instance duplication within the same application. Is it something we intend to change at some point?

knolleary:
@cstns not off the top of my head. Will need to dig through git blame and PR history to see if there was an explicit choice stated for this.

knolleary:
As far as I can see, the restriction was originally added in this commit https://github.com/FlowFuse/flowfuse/commit/22d0f31a4f20a672262507bb9430bcd4e91a78ab#diff-37ad4d63a25ead0d5bec5c27c1c3fc36b4c7adf9a7ff31af4b3102dbe6018e12R155-R159 as part of a large overhaul of the API when Applications were introduced - https://github.com/FlowFuse/flowfuse/pull/1888

There's no specific rationale for it and I can't think of a technical reason to disallow it. We should permit this on the API.

cstns:
The only thing that comes to mind would be the future development of Application ENV vars that will be shared to instances belonging to them, would complicate things a bit when duplicating an instance from one application to another.

The removal of the backend restriction seems pretty straight forward at first glance.

cc @joepavitt and @gstout52 for prioritization and/or customer demand

[...]

PR #5788, opened 2025-07-17
Prevent users from selecting other applications when duplicating instances

cstns:
## Description

Fixes oversight on the UI when duplicating instances that permits users to select different applications when duplicating instances when the back end has a check in place that prevents, this allowing to duplicate instances only inside the same application.

The fix filters out available applications other than the source instance application while still allowing users to go back to the applications step. This approach allows for ease of reversal in case things change on the back end.

## Related Issue(s)

part of https://github.com/FlowFuse/flowfuse/issues/5785

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl (approved):
Works but i'd prefer to actually be able to duplicate to any application in my team!  Hey ho.

Issue #5721, opened 2025-07-02
Incorrect hosted Instance status in Home view


joepavitt:
### Current Behavior

<img width="548" alt="Image" src="https://github.com/user-attachments/assets/30629df7-dc25-477b-9611-2cf4a97b9940" />

Response I'm seeing from `TeamAPI.getTeamInstanceCounts(this.team.id, [], 'hosted')` is:

```
{running: 24}
```

However, I have a crashed Instance:

<img width="1064" alt="Image" src="https://github.com/user-attachments/assets/d0bb6572-a605-4b8a-ab44-30427ffeb848" />

### Expected Behavior

The "Error" tile should show a value "1"

### Environment

- FlowFuse version: `main`
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
@cstns can you investigate this please

cstns:
where is this happening?

hardillb:
@joepavitt just to clarify, you are seeng this is on your local instance (just looking for a team with 24 instances)

cstns:
he's running locally

joepavitt:
Yes, locally. I've checked the database too, and it reports "running" as the status

cstns:
I reproduced the issue by blocking an instance’s port and then starting it. The database state shows it as running, but `nr-launcher` reports it as crashed.

One possible solution I can think of is to update the DB entry with the actual `liveState` each time it's retrieved. It would fix the problem by having better data to sort by when retrieving instances.

The downsides would be that it would also push transient states (which I don't think we want but we can excluede/map them) and it would increase the DB load (updating the project's state each time their `liveState` is retrieved).

Directly replacing the project’s state with the `liveState` in that endpoint isn’t viable as it would require fetching and checking all of a team’s running instances in real-time, which hits both the DB and `nr-launcher` hard.

cstns:
I'm open to alternative ideas, my proposal could potentially introduce an entirely new set of challenges

joepavitt:
> The downsides would be that it would also push transient states (which I don't think we want but we can exclude/map them) and it would increase the DB load (updating the project's state each time their liveState is retrieved).

If the transient state is an errorful state, that needs to be reflected in the UI. If it's just "installing", etc, then I'm less concerned about that being reflected in the database.

cstns:
[...]

PR #5795, opened 2025-07-21
Enhance the latest known project state

cstns:
## Description

First part of mitigating incorrect state counters when retrieving aggregated instance counters.

- Added a new in-memory `latestProjectState` map in the Project Controller to track live project state.
  - tracks non-definitive states (excluding `running` and `stopped`) so we can output relevant state counters when querying a team's instances grouped by state.
  - the reason why we track only non-definitive states is because when an instance is in a definitive state we can rely on the state stored in the database (the desired state)
- Use the `latestProjectState` map when retrieving aggregated data in the Project's `countByState` method.
  - the live state is updated only when the underlying driver checks the instance state which means that the `liveState` endpoint still needs to be triggered by users at some point to get up-to-date data, otherwise it might return stale data
- Exposed an API endpoint that allows instances to update their state to mitigate the above drawback.

This is not a foolproof approach. As it stands, the `latestProjectState` map is cleared on each CI deployment. In the long run, we should move this to an in-memory data store so its values persist across deployments. Nonetheless, with the follow-up task on the `nr-launcher`, which will update the instance state on specific events as they occur,  this is definitely a step up from what we have now.

## Related Issue(s)

part of https://github.com/FlowFuse/flowfuse/issues/5721

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/controllers/Project.js
+++ b/forge/db/controllers/Project.js
@@ -10,6 +10,8 @@ const { KEY_SETTINGS } = require('../models/ProjectSettings')
  */
 const inflightProjectState = { }
 
+const latestProjectState = { }
+
 const inflightDeploys = new Set()
 
 module.exports = {
@@ -545,7 +547,7 @@ module.exports = {
     },
 
     /**
-     * Imports settings, flows and credentials from a project export object
+     * Imports settings, flows, and credentials from a project export object
      *
      * @param {*} app
      * @param {*} project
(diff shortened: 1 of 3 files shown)


forge/db/controllers/Project.js
```diff
@@ -689,6 +691,56 @@ module.exports = {
...
+    /**
+     * Updates the latest project state for the specified app and project based on the given state.
+     * If the state is 'running' or 'stopped', it clears the latest driver state.
+     * Otherwise, it sets the latest driver state to the new value.
+     *
+     * @param {String|Number} projectId - The project id related to the driver state update.
+     * @param {string} state - The new state to update, can be 'running', 'stopped', or other valid states.
+     */
+    updateLatestProjectState: function (app, projectId, state) {
+        if (['running', 'stopped'].includes(state)) {
```
cstns: in hindsight, this may not be the best way to clear the cached state. A better alternative would be to clear the state only when the desired state of the instance (stored in the db) matches the state we're trying to update to.

Otherwise we might get into situations where ok, we received a stopped state, we'll remove it from the cache but the desired state of the instance is running.

forge/routes/api/project.js
```diff
@@ -1320,6 +1320,50 @@ module.exports = async function (app) {
...
+                properties: {
+                    instanceId: { type: 'string' }
+                }
+            },
+            body: {
+                type: 'object',
+                required: ['state'],
+                properties: {
+                    state: { type: 'string' },
+                    hashId: { type: 'string' }
```
hardillb: Why is `hashId` listed when it is in the params?

and not used

hardillb:
```suggestion
```

hardillb:
```suggestion
                    state: { type: 'string' }
```

cstns: this was meant to be the hashed id of the pods/containers when nr is running in HA mode

hardillb: OK, so we may end up with a transient state where we get 2 different reports from the 2 pods, but it should settle down quickly.

cstns: At the moment, yes. If it becomes a problem, we could mitigate that with the hashId, which now that I look at after a couple of days, doesn't have the most descriptive name

forge/routes/api/project.js
```diff
@@ -1320,6 +1320,49 @@ module.exports = async function (app) {
...
+            // check accessToken is project scope
+            // (ownerId already checked at top-level preHandler)
+            if (request.session.ownerType !== 'project') {
+                reply.code(401).send({ code: 'unauthorized', error: 'unauthorized' })
+            } else {
+                done()
+            }
+        },
+        schema: {
+            summary: 'Update the live status of an instance',
```
knolleary, line 1343:
```suggestion
            summary: 'Update the live status of an instance',
            hide: true, 
```

This isn't an end-user API so shouldn't be part of our published API doc

knolleary (dismissed):

forge/routes/api/project.js
```diff
@@ -1320,6 +1320,50 @@ module.exports = async function (app) {
...
+            hide: true,
```
knolleary:
```suggestion
            hide: true,
```

knolleary:
@cstns 

Home page shows 5 running and 1 No running

image


When I click through I see 4 running and 2 not running

image


image

cstns:
is this on staging/pre-staging or somewhere i can play around with?

knolleary:
@cstns my local setup - can huddle if you want to take a look at something

cstns:
I identified the problem in that the `updateLatestProjectState` method was clearing `stopped` states even though they were coming from the driver meaning that even though the desired state was `running` and the driver state was `stopped`, this was not getting logged and the desired state was returned instead of the actual one.

The fix partly implies https://github.com/FlowFuse/flowfuse/pull/5795#discussion_r2222476958 in that we should store any driver state in cache whenever it differs from the desired one

knolleary (approved):

Issue #5805, opened 2025-07-22
Performance: reduce DB load handling status updates


knolleary:
### Description

Each device status checkin hits the database three times:

1. GET device (includes multiple left joins on other tables to load the associated objects)
2. COUNT on snapshots to verify the target snapshot still exists
3. UPDATE device to write its `lastSeenAt`. This *might* update other fields if the device is transitioning between states.

As the number of devices grows, so will the load on the database this drives. The goal of this task is to optimise this path to minimise the database activity.

Possible strategies:

### Simplify GET device

The current GET SQL is generated by sequelize:

```
SELECT 
 `Device`.`id`,
 `Device`.`name`,
 `Device`.`type`,
 `Device`.`credentialSecret`,
 `Device`.`state`,
 `Device`.`lastSeenAt`,
 `Device`.`settingsHash`,
 `Device`.`agentVersion`,
 `Device`.`nodeRedVersion`,
 `Device`.`mode`,
 `Device`.`editorAffinity`,
 `Device`.`editorToken`,
 `Device`.`editorConnected`,
 `Device`.`createdAt`,
 `Device`.`updatedAt`,
 `Device`.`ApplicationId`,
 `Device`.`TeamId`,
 `Device`.`targetSnapshotId`,
 `Device`.`activeSnapshotId`,
 `Device`.`ProjectId`,
 `Device`.`DeviceGroupId`,
 `Team`.`id` AS `Team.id`,
 `Team`.`name` AS `Team.name`,
 `Team`.`slug` AS `Team.slug`,
 `Team`.`TeamTypeId` AS `Team.TeamTypeId`,
 `Application`.`id` AS `Application.id`,
 `Application`.`name` AS `Application.name`,
 `Project`.`id` AS `Project.id`,
 `Project`.`name` AS `Project.name`,
 `Project->Application`.`id` AS `Project.Application.id`,
 `Project->Application`.`name` AS `Project.Application.name`,
 `targetSnapshot`.`id` AS `targetSnapshot.id`,
 `targetSnapshot`.`name` AS `targetSnapshot.name`,
 `activeSnapshot`.`id` AS `activeSnapshot.id`,
 `activeSnapshot`.`name` AS `activeSnapshot.name`
FROM `Devices` AS `Device` 
LEFT OUTER JOIN `Teams` AS `Team` ON `Device`.`TeamId` = `Team`.`id`
[...]

PR #5812, opened 2025-07-23
Minimise Device look up on status update

knolleary:
Part of #5805 

This reduces the DB query made when looking up a device as part of the status handling path to *only* touch the Device table.

I've also reduced the batch size for Device Group commands to 5 per 5s, down from 10. This is a short term fix to lighten the load whilst we work on other optimisations.

### DIFF
--- a/forge/comms/devices.js
+++ b/forge/comms/devices.js
@@ -88,11 +88,13 @@ class DeviceCommsHandler {
         // Check it looks like a valid status message
         if (status.id && status.status) {
             const deviceId = status.id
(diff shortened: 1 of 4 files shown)


knolleary:
Tests should be fixed up. Found some places passing a second arg to `Device.byId(...)` which was previously ignored - but my change now made use of the second arg and broke those tests. Simply fix - update the tests to not pass an invalid second arg to the function. But also hardend the implementation to ensure the default behaviour is maintained if a second arg is passed in.

PR #5824, opened 2025-07-28
Fix table renderings

cstns:
## Description

Better handling of tables with long column names and long values.

- added extra configuration to the ff-table to allow setting custom classes on the table itself and table header columns to allow sticky headers
- added a method to open each type of value in a distinct way, currently only text values are handled differently in that they can be parsed into json values or rendered as plain text (long values are replaced with a CTA to open the value in a modal)
- bound the col value to dynamic components for ease of rendering

https://github.com/user-attachments/assets/ad83bb21-1a73-4f61-b6bf-e816805a1fbf

we do not support the retrieval of more than 10 results, I manually altered the endpoint just to showcase the sticky header

## Related Issue(s)

Closes https://github.com/FlowFuse/flowfuse/issues/5718

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
For the "view more" can we make them blue text by default please, then darken on hover? Make them clear they're not just a value/entry in the DB row.

joepavitt (changes requested):
Looks great, apart from one piece of UX feedback.

joepavitt:
Screenshot 2025-07-29 at 10 27 40

I'm seeing some odd scrollbar rendering in each cell. The `.text-cell` is height `18px`, but the `code` content is `32.5px`

cstns:
Instead of forcing rows at a fixed height I changed the behavior to allow rows to resize themselves where needed (up to a json of max 50 characters, then the text cell gets truncated)

image

joepavitt:
@cstns so what happens with a single string that is over 50 characters long, and isn't a JSON? Surely there is smarter rendering to be done here based on the column type? If `text`, then we can just keep it all in one line, truncate with ellipsis instead?

If `json` type, then offer the `view more` approach?

cstns:
I don't think we have a JSON type, it's just a text type containing JSON, so it's tricky. That's why I added an arbitrary 50-character limit. We'll truncate both values and display them separately in a modal.

While there are certainly better ways to display values, this PR focuses solely on rendering the tables correctly, while leaving room for future extensibility to handle different types whe the time comes.

hardillb:
Remember that users can create any column types they want via SQL, we are just going to limit the type for the UI to keep things simple. So the data sample UI needs to support anything

cstns:
By default we're displaying the column value as is, without alterations, the 50 char limit and truncation of content to be shown separately in a modal is currently only done for text type values.

We should decide if we should truncate or display the entire value in the table column or separately for more types. Basically the list from https://www.postgresql.org/docs/current/datatype.html with render modes.

I'll get a list going and see how i can share it

joepavitt:
Why is the target of this PR the `create-table` branch?

cstns:
I started working on it at the same time as the create-table functionality, should be good to rebase on main

joepavitt:
> I started working on it at the same time as the create-table functionality, should be good to rebase on main

Gave this a quick go, but it's getting conflicts that are outside of my "can review quickly" s[ace. Take a quick look when you have a chance please, and let me know if this is now too dependent on the underlying "Add Table" work.

cstns:
looking good on my end

joepavitt:
@cstns why all of the server-side changes here when all you've described are front-end changes?

cstns:
Valid question.. Must have merged the previous branch in at some point and they got carried over? I'll create a new branch and cherry pick the frontend part

cstns:
closing to review front end changes only in https://github.com/FlowFuse/flowfuse/pull/5839

Issue #5797, opened 2025-07-21
"Add Table" option in FlowFuse Tables UI


joepavitt:
### Description

Follow on from https://github.com/FlowFuse/flowfuse/issues/5718

This would add an "Add Table" button to the UI which would allow users to add a table to their database, without needing to go into a Node-RED flow to do so.

MVP here will be that the front-end needs to provide an option to name the table, detail the names of all columns and choose their respective type.

### Epic/Story

#4911

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
FYI @gstout52 - have made an explicit task for this follow up step tho the MVP as for an onboarding it'll be a significant improvement, and should help get more people using Tables offering.

joepavitt:
@cstns is size `2` appropriate here?

cstns:
Yep, sounds right!

hardillb:
Column types list is here:

https://www.postgresql.org/docs/current/datatype.html#DATATYPE-TABLE

work from the name column and we need to decide which subset to support first, I will filter the list

hardillb:
- int (4 byte  int)
- bigint (8 byte int)
- boolean
- date
- timestamptz
- real (float)
- double precision (double) 
- text
- varchar (do we need this if we have text?, if so would also need to supply max length)

Questions:
-  do we need both `int` and `bigint` (just `bigint`)?
- should we have `bigserial` for auto increment counter?
- do we need both `text` and `varchar`?

joepavitt:
I vote `text`, no `varchar`

cstns:
what about `bigserial` and `int` vs `bigint`?

hardillb:
> what about bigserial and int vs bigint?

Use `bigint` for a raw integer

and `bigserial` for a counter field

joepavitt:
@hardillb can you link to API PR/issues here please?

hardillb:
@joepavitt https://github.com/FlowFuse/flowfuse/pull/5815

Issue #3928, opened 2024-05-24
Handle form validations in a coheseve manner across the application


cstns:
### Description

Form validation is closely linked to UX and also DX.

Choosing the moment when to validate fields and when to display an error matters. The more complex the form, more complex validation is required and at times can get out of hand an unmanageable.

A good validator must:
- be flexible
- play nicely with Vue
- have i18n support

Suggestions:
- [vuelidate](https://github.com/vuelidate/vuelidate)
- [vee-validate](https://github.com/logaretm/vee-validate)

### Which customers would this be available to

Everyone - CE/Starter/Team/Enterprise

PR #5846, opened 2025-07-30
Flowfuse Tables - "Add Table" UI

cstns:
## Description

Adds UI components that  allow users to create, view schema and delete tables from the UI.

https://github.com/user-attachments/assets/bc23032b-0359-4b93-9027-d36b1840d805

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5797

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/ee/lib/tables/drivers/postgres-localfs.js
+++ b/forge/ee/lib/tables/drivers/postgres-localfs.js
@@ -277,7 +277,7 @@ module.exports = {
             const teamClient = libPg.newClient(options)
             try {
                 await teamClient.connect()
-                teamClient.query(`DROP TABLE ${tableName}`)
+                await teamClient.query(`DROP TABLE ${tableName}`)
             } finally {
                 teamClient.end()
             }
--- a/forge/ee/routes/tables/index.js
+++ b/forge/ee/routes/tables/index.js
@@ -272,7 +272,7 @@ module.exports = async function (app) {
     }, async (request, reply) => {
         if (request.body.name && request.body.columns) {
             const tables = await app.tables.getTables(request.team, request.params.databaseId)
-            if (tables.filter((t) => t.name === request.body.name).length === 1) {
+            if (tables.tables.filter((t) => t.name === request.body.name).length === 1) {
                 reply.status(409).send({ code: 'table_exists', error: 'Table already exists' })
             } else {
                 const t = await app.tables.createTable(request.team, request.params.databaseId, request.body.name, request.body.columns)
--- a/forge/ee/routes/tables/index.js
+++ b/forge/ee/routes/tables/index.js
@@ -352,7 +352,7 @@ module.exports = async function (app) {
         }
     }, async (request, reply) => {
         const tables = await app.tables.getTables(request.team, request.params.databaseId)
-        if (tables.filter((t) => t.name === request.params.tableName).length === 1) {
+        if (tables.tables.filter((t) => t.name === request.params.tableName).length === 1) {
             await app.tables.dropTable(request.team, request.params.databaseId, request.params.tableName)
             reply.status(204).send()
         } else {
--- a/frontend/src/api/tables.js
+++ b/frontend/src/api/tables.js
@@ -30,11 +30,17 @@ const createTable = (teamId, databaseId, payload) => {
         .then(res => res.data)
 }
 
+const deleteTable = (teamId, databaseId, tableName) => {
+    return client.delete(`/api/v1/teams/${teamId}/databases/${databaseId}/tables/${tableName}`)
+        .then(res => res.data)
+}
+
 export default {
     createDatabase,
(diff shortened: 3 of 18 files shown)


hardillb:
If type = bigserial then `nullable` and default should be disabled

cstns:
> If type = bigserial then nullable and default should be disabled

fixed

joepavitt (changes requested):
### Title Feedback

Screenshot 2025-07-30 at 17 46 34

_I_ know it should all be kebab case, but the user will likely not. Need better feedback here.

### Type Selection

How much extra work is it to have this as a searchable field? Happy to add it as an iteration and cut here though.

Screenshot 2025-07-30 at 17 47 31

### Drawer Background

Can we make the b/g of the drawer `grey-50` please? The `grey-100` is a little too dark.


### Mouseup Closes Drawer

If I drag from an Input box (to then delete it), but accidentally release the mouse whilst outside of the drawer, it closes the drawer. Can the handler be on click, rather than `mouseup`?

### Shaded Area

We had discussed the main content being behind a shaded zone (as per modals do now), was this deemed out of scope for first pass?

cstns:
> Title Feedback
> Screenshot 2025-07-30 at 17 46 34
> I know it should all be kebab case, but the user will likely not. Need better feedback here.

It shouldn't be camelcase, this is what I found as a restriction for PG table names:

// PostgreSQL identifiers:
// - max 63 bytes (can be less than 63 characters if multibyte)
// - must begin with a letter or underscore
// - can contain letters, digits, and underscores

I could definitely improve the error message

> Type Selection
> How much extra work is it to have this as a searchable field? Happy to add it as an iteration and cut here though.

10 minutes max

> Drawer Background
> Can we make the b/g of the drawer grey-50 please? The grey-100 is a little too dark.

absolutely

> Mouseup Closes Drawer
> If I drag from an Input box (to then delete it), but accidentally release the mouse whilst outside of the drawer, it closes the drawer. Can the handler be on click, rather than mouseup?

this could get tricky, I'd have to mess around with the `click-outside` directive but i can give it a go

> Shaded Area
> We had discussed the main content being behind a shaded zone (as per modals do now), was this deemed out of scope for first pass?

I completely forgot about it! Dependent on the items above, I'll give it a go

cstns:
first three done

cstns:
first three done

cstns:
strangely enough, the `click-outside` directive's listener is on click. (`frontend/src/ui-components/directives/ClickOutside.js:27`)

Switching it to `mousedown` preserves the "click outside" functionality and avoids the behavior you described but I'm a bit weary to alter it

joepavitt:
> Switching it to mousedown preserves the "click outside" functionality and avoids the behaviour you described but I'm a bit weary to alter it

Yeah, mouseodnw is just as odd. Not sure why the `click` would get registered when my `mousedown` event is not on the same element at the `mouseup`.

I won't block this PR, but can you open an issue for this and assign to `2.21` please?

joepavitt:
Not sure if this is as a result of the listbox change, but selecting an option now closes the side drawer:

https://github.com/user-attachments/assets/129ceff8-d6e9-45e2-b810-2eef04bedd78

cstns:
yes, it was the combobox, I had to exclude the options from the click outside component

joepavitt:
Screenshot 2025-07-30 at 19 06 03

Have moved the cancel/save buttons to the top-right, easily missed at the bottom i think, even though it was more aesthetically pleasing.

cstns:
the same should be done for the view schema drawer

joepavitt:
The error validation being in ` switch` is very odd, especially as there are two different errors states. The `break` in the first section can prevent an error being reported (as I've seen locally). As such, i've broken it into two `if` statements, which makes it smoother

joepavitt (approved):

joepavitt:
All good from my side. Awaiting review from @hardillb before I merge though, just in case there is something server-side I've missed out. Also don't want to merge until https://github.com/FlowFuse/flowfuse/pull/5815 is reviewed as to not pollute that PR

hardillb:
@joepavitt Looks good to me

hardillb (approved):

hardillb:
@joepavitt Anybody but me is good to look at #5815 I tagged @cstns because he was using the API

cstns:
> The error validation being in ` switch` is very odd, especially as there are two different errors states. The `break` in the first section can prevent an error being reported (as I've seen locally). As such, i've broken it into two `if` statements, which makes it smoother

I agree! It was last minute and did on the run. The problem is that we don't have a cohesive way of handling form validations and we're handling it differently from place to place. I'd appreciate a stab at https://github.com/FlowFuse/flowfuse/issues/3928 to normalize things

PR #5850, opened 2025-07-31
Ensure table and column names escaped

hardillb:
## Description

Enforce escaping for table and column names

### DIFF
--- a/frontend/src/pages/team/Devices/dialogs/components/DeviceCredentialsDialog/OtcInstallSection.vue
+++ b/frontend/src/pages/team/Devices/dialogs/components/DeviceCredentialsDialog/OtcInstallSection.vue
@@ -92,16 +92,16 @@ export default {
                 script: {
                     Windows: {
                         title: 'Open an elevated Command Prompt and run:',
-                        command: `powershell -c "irm https://raw.githubusercontent.com/FlowFuse/device-agent/refs/heads/main/installer/get.ps1|iex" && flowfuse-device-agent-installer.exe -o ${this.device.credentials.otc} -u ${this.settings?.base_url}`
+                        command: `powershell -c "irm https://flowfuse.github.io/device-agent/get.ps1|iex" && flowfuse-device-agent-installer.exe -o ${this.device.credentials.otc} -u ${this.settings?.base_url}`
                     },
                     MacOS: {
                         title: 'Open Terminal and run:',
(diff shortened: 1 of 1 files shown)


forge/ee/lib/tables/drivers/postgres-localfs.js
```diff
@@ -235,9 +235,9 @@ module.exports = {
             const teamClient = libPg.newClient(options)
             try {
                 await teamClient.connect()
-                let query = `CREATE TABLE IF NOT EXISTS "${tableName}" (\n`
+                let query = `CREATE TABLE IF NOT EXISTS "${libPg.pg.escapeIdentifier(tableName)}" (\n`
```
Steve-Mcl: Quotes are applied by the escaper - this will cause double quotes

forge/ee/lib/tables/drivers/postgres-localfs.js
```diff
@@ -235,9 +235,9 @@ module.exports = {
...
                 for (const [i, col] of columns.entries()) {
-                    let column = `"${col.name}" `
+                    let column = `"${libPg.pg.escapeIdentifier(col.name)}" `
```
Steve-Mcl: Quotes are applied by the escaper - this will cause double quotes

Steve-Mcl (changes requested):
This PR is fine as it (minus the lint errors), but I have concerns regarding other parts of the original merged PR I think we must handle to mitigate what we can

NOTE: the below is applicable to both drivers

### 1
https://github.com/FlowFuse/flowfuse/pull/5850/files#diff-41c5997c75a79615b0da0be7da5a91b8daea9331ccf48714137091ebf76e5562R247

this should be whitelisted at best, escaped at minimum

### 2
https://github.com/FlowFuse/flowfuse/pull/5850/files#diff-41c5997c75a79615b0da0be7da5a91b8daea9331ccf48714137091ebf76e5562R251-R255
col.default is user input and should be vetted for length and type and if string, escaped. Something like:
```javascript
if (typeof col.default === 'string') {
    const escapedDefault = pgLib.pg.escapeLiteral(col.default)
    columnDef += ` DEFAULT ${escapedDefault}`
} else if (typeof col.default === 'number' || typeof col.default === 'boolean') {
    columnDef += ` DEFAULT ${col.default}`
} else {
    throw new Error(`Unsupported default value type for column ${col.name}`)
}
```

### 3
There are no unit tests for the `createTable` or `dropTable` functions.

PR #5864, opened 2025-08-01
Do not trigger full device settings calculation on bulk update

### LINKED ISSUE #5862: Performance: reduce usage of bulkUpdate hook on Devices
### Description

The `afterBulkUpdate` hook we have on the Device model is intended to recalculate the settings hash whenever a device's relevant properties are modified.

This leads to lots of extra DB work being done on the Devices table that I believe can be bypassed.

The purpose of the settings hash is so that the Device Agent can check whether it is running the 'latest' settings; when it checks in, if there is a mismatch in the hash, we can tell the agent to update its settings.

When doing a pipeline deploy, we update the `targetSnapshotId` via a bulk update query. This in turn triggers the hook, that triggers a recalculation of the hash and update of the Device table for each device *individually*.

For this scenario, I believe it would be sufficient to update the settings hash to a suitably random value as part of the initial bulk update - and *not* then trigger the hook. This will have the desired effect of the device knowing it needs to refresh its settings.

The purpose of this task is to implement this approach and test it. There are a number of places where we do bulk updates - they each need to be considered, but the general logic holds.

### Epic/Story

_No response_

knolleary:
Closes #5862 

This PR removes the `afterBulkUpdate` hook in favour of a `beforeBulkUpdate` hook, which allows us to insert a randomly generated `settingsHash` as part of the bulk update in the scenarios that need it.

### DIFF
--- a/forge/db/models/Device.js
+++ b/forge/db/models/Device.js
@@ -9,7 +9,7 @@ const SemVer = require('semver')
 const { col, fn, DataTypes, Op, where, literal } = require('sequelize')
 
 const Controllers = require('../controllers')
-const { buildPaginationSearchClause } = require('../utils')
+const { generateToken, buildPaginationSearchClause } = require('../utils')
 
 const ALLOWED_SETTINGS = {
     autoSnapshot: 1,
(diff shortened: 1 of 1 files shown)


knolleary:
I've come back to this fresh this morning and am happy this is a good change. There are edge cases where a new snapshot is deployed to a device which technically has identical settings so no hash change. This PR will mean they retrieve their settings again; but I think that's okay in the big scheme of things.

@hardillb tagged you for review as Steve is out today

hardillb (approved):

PR #5875, opened 2025-08-06
Allow admin to set Team-specific billing overrides

### LINKED ISSUE #5794: Team-specific billing overrides
### Description

Currently all features of a Team's billing is either:
1. determined by its Type
2. done manually by putting into manual billing mode

Manual billing mode was added to give us more flexibility in the billing process for teams that don't fit the self-service limits of the platform. The down side is the human cost of monitoring and managing their billing.

There are a set of common use cases where custom configurations are required and ideally, we'd be able to automated them rather than require full manual mode.

The proposal here is to allow an Admin user to customise the billing setup for an individual team. This would allow:

1. Availability of instance types (inc free allocation)
2. Custom stripe price/product ids

As a complete hypothetical, for example, our Team tier currently includes X 'free' Small instances. (Free because they come included in the base price of the team). This would allow us to enable an individual team to have Y 'free' Medium instances instead.

---

This is not a very large piece of work. On the runtime side, we have centralised code for retrieving the billing properties of a team. That code would just need to look for team-specific overrides.

The Team Settings - Admin Tools section would gain a 'Set custom billing' option (or similar), that would show a dialog based on the existing TeamType edit dialog and allow for overrides to be set. We may choose to limit what overrides can be set and iterate on this - but we have most of the raw components already.

None

knolleary:
Closes #5794 

This allows an admin to provide team-specific overrides to the limits applied to a team.

I've broken this down into a logical set of commits.

1. https://github.com/FlowFuse/flowfuse/commit/a9b349b4223960079ab94ae315b789387b08b3a4 - adds a `properties` property to the `Team` model via migration and updates view to include it.
2. https://github.com/FlowFuse/flowfuse/commit/d22703b97b59c34bcfbbf2b40a7ff0eac59821ce - updates all places the backend access TeamType level properties to use a Team-level getter which can apply the overrides as needed
3. https://github.com/FlowFuse/flowfuse/commit/7bc49ec9200b0ba6f08c6b0f01a8df2de27a1afc - adds a form to the Admin Tools view (shown under Team Settings -> Danger to admin users). This allows the admin to modify the team overrides.
4. https://github.com/FlowFuse/flowfuse/commit/88d57949fbb20509174532cb458829dbfe160059 - adds a composable to standardise how to access team properties in the UI - allowing the overrides to be applied
5. https://github.com/FlowFuse/flowfuse/commit/4ffc5f85476a1bdb22dfa10a386f834db40dfbf7 - updates all relevant places in the UI to use the composable.

Some extra notes on the above:

 - The admin form grew more than I was expected. There is some amount of duplication with the TeamTypeEdit dialog; but I've tried to make the form more visually efficient. There is a case to be made for standardizing the two, but not today.
 - We have different patterns of usage in the existing composables folder. I've picked one that made sense to me.

I've tested various paths manually through the app to verify the billing gets applied properly with/without the overrides in place. There are some more scenarios I'd like to run through before a final merge, but want to get more eyes on this sooner rather than later.

---

Usage Limits form:

image

Edit mode:

image


forge/db/models/Team.js
```diff
@@ -394,6 +404,48 @@ module.exports = {
                 deviceCount: async function () {
                     return await M.Device.count({ where: { TeamId: this.id } })
                 },
+                getProperty: function (key, defaultValue) {
+                    let teamValue
+                    // console.log('getPRoperty', this.properties, key)
```
cstns:
```suggestion

```

frontend/src/composables/TeamProperties.js
```diff
@@ -0,0 +1,32 @@
...
+            }
+        }
+    }
+    return teamValue
+}
+
+export function getTeamProperty (team, property, defaultValue) {
+    const teamValue = getProperty(team.properties, property)
+    if (teamValue === undefined) {
+        // console.log('getTeamProperty - TYPE', property, getProperty(team.type.properties, property) ?? defaultValue)
```
cstns:
```suggestion
```

frontend/src/composables/TeamProperties.js
```diff
@@ -0,0 +1,32 @@
...
+        // No value found in team properties. Check the TeamType properties
+        return getProperty(team.type.properties, property) ?? defaultValue
+    }
+    // console.log('getTeamProperty - TEAM', property, teamValue)
```
cstns:
```suggestion
```

cstns (changes requested):
Tested happy paths for user invitations and instance/device creation. Also covered a few negative scenarios that came to mind.
Did not focus much on the admin form, it needs a rework down the line but it surely gets the job done for now

Bugs found:
   - Instance creation button should be disabled (with tooltip) when max instances are reached
    [#5879](https://github.com/FlowFuse/flowfuse/issues/5879)
   - Charges section should be hidden in the create instance modal if the team hasn't exceeded included instances
    [#5880](https://github.com/FlowFuse/flowfuse/issues/5880)
   - Instance/device limit should not be lower than the free/included instances
    [#5881](https://github.com/FlowFuse/flowfuse/issues/5881)

Follow-up tasks:
- Add E2E tests → [#5883](https://github.com/FlowFuse/flowfuse/issues/5883)
- Normalize composables under a single format (preferably Vue-recommended) → [#5882](https://github.com/FlowFuse/flowfuse/issues/5882)

Issue #5368, opened 2025-03-28
Allows httpNodeCors to be configured for a NR under the device agent.


hardillb:
### Description

To allow API endpoints or dashboard to be loaded from different sources this needs to be added to the `settings.js` for the Node-RED instance under the Device agent.

I suggest it becomes something that can be added to the device.yml to be included.

requested by customer:

https://app-eu1.hubspot.com/contacts/26586079/record/0-2/7157221333

### Which customers would this be available to

Everyone - CE/Starter/Team/Enterprise

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Is that a sensible thing to just be throwing in for _all_ Node-RED instances running via the Agent? Not clear to me if this is still a priority, @gstout52 will let you assign/shift accordingly.

gstout52:
@hardillb Can you say more about what value this change will bring to the user?

hardillb:
This allows dashboards hosted in devices to be embedded in other webpages

@joepavitt The suggestion is to make it configurable
Either in the FF UI or via the device.yml file

hardillb:
The device-agent part of this was completed under FlowFuse/device-agent#241

hardillb:
Waiting for 2.21.0 release day so we don't have to spin new Stack containers.

PR #5877, opened 2025-08-06
Add support for setting httpNodeCors for hosted insances

hardillb:
part of #5368

# DO NOT MERGE until 2.21.0 or we do a point release to include new Stacks.

## Description

Adds option to configure CORS headers for hosted NR instances.

Only the hosted nodes part, devices can be done all in the device agent.

Not happy with how I've done the template, but not sure how else to have tried.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS

## Related Issue(s)

#5368 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/test/unit/forge/db/controllers/ProjectTemplate_spec.js
+++ b/test/unit/forge/db/controllers/ProjectTemplate_spec.js
@@ -312,7 +312,7 @@ describe('Project Template controller', function () {
                     })
                 }).not.throw()
             })
-            it('should allow origin foo', async function () {
+            it('should not allow origin foo', async function () {
                 should(() => {
                     app.db.controllers.ProjectTemplate.validateSettings({
                         httpNodeCORS: {


hardillb:
nr-launcher section https://github.com/FlowFuse/nr-launcher/pull/368

hardillb:
@Steve-Mcl Just realised this needs a semver gate on the nr-launcher version. Will add in the morning

hardillb:
@Steve-Mcl ready for review, but **DO NOT MERGE** until we decide how to ship new stack with nr-launcher changes.

You will need to bump nr-launcher package version to > 2.20.0 to enable settings (but can be viewed under admin -> template)

Steve-Mcl:
@hardillb a very quick pass on this and I note OPTIONS and PATCH are missing.

PATCH is one I would personally expect to see but more importantly, some (particularly browsers) cross-origin requests (esp POST/PUT/PATCH/DELETE) employ a preflight OPTION check - which would fail here.

Should we get these now before merging?

hardillb:
`PATCH` we can add, but the cors check IS an `OPTIONS` call, which is why it's not in the list

Steve-Mcl:
> IS an `OPTIONS` call, which is why it's not in the list

Yes! of course. Obvs :D  🧠 🌫️ 

I'll hang fire while you add PATCH 👍

hardillb:
@Steve-Mcl test now, I've done the update for the nr-launcher as well

frontend/src/pages/admin/Template/sections/Security.vue
```diff
@@ -138,6 +262,23 @@ export default {
...
+                return false
+            }
+
+            // template
+            if (!this.instance) {
+                return true
+            } else {
+                const launcherVersion = this.instance.meta?.versions?.launcher
+                if (launcherVersion) {
+                    return semver.gte(launcherVersion, '2.21.0')
```
hardillb:
```suggestion
                    return semver.gte(launcherVersion, '2.21.0', { includePrerelease: true })
```

hardillb:
NOT TO BE MERGED until 2.21.0 or we do a point release to include new Stacks.

test/unit/forge/db/controllers/ProjectTemplate_spec.js
```diff
@@ -296,6 +296,110 @@ describe('Project Template controller', function () {
...
+                            POST: true,
+                            PATCH: true,
+                            PUT: true,
+                            DELETE: false,
+                            HEAD: true
+                        }
+                    })
+                }).not.throw()
+            })
+            it('should allow origin foo', async function () {
```
Steve-Mcl:
```suggestion
            it('should not allow origin foo', async function () {
```

Steve-Mcl (approved):

PR #5889, opened 2025-08-08
HMR

cstns:
## Description

Webpack configuration used for development only. It introduces HMR, which is a big time saver for my use case, and provides somewhat improved performance.

The front-end build is now run by the webpack dev server. Assets are no longer served statically by Fastify.

### `devServer` Configuration Keys and Description
- **`port: 8080`**: Specifies the port number the development server runs on. Accessible at `http://localhost:8080`. A different port is needed to avoid conflicts with the backend process. The web app is still accessible on port 3000 because the assets are still written to disk in the same location.
- **`historyApiFallback: true`**  
  Ensures requests that would normally return 404 fall back to `index.html`, essential for SPAs with client-side routing.
- **`hot: true`**: Enables Hot Module Replacement (HMR), which updates modules in the browser without a full page refresh, improving the development experience.
- **`liveReload: false`**: Disables full page reload when files change, relying instead on HMR (`hot: true`) for updates.
- **`devMiddleware: { writeToDisk: true }`**: Configures the dev server to write the bundled files to disk. Useful for debugging or when processes require files on disk. As mentioned before, this allows Fastify to serve the front-end on port 3000 as well.
- **`proxy`**: Defines rules to forward specific requests to the appropriate backend or external services:
  - **`context: ['/api/v1', '/account/', '/storage', '/ee/billing']`**: Redirects API requests to the local backend running on `http://localhost:3000`.
  - **`context: ['/api']`**: Routes API requests to `https://registry.npmjs.com` and rewrites the path by removing `/api`. For example, `/api/packages` becomes `/packages`.
  - **`context: ['/api/**/projects/**/resources/stream']`**: Proxies WebSocket requests to `ws://localhost:3000` to support WebSocket connections with the backend.

The proxy list is not complete. The `/setup` endpoint does not work on port 8080 and must be accessed on port 3000. There may be other endpoints missing as well.

The `stats` override and `devServer.client.logging` settings hide some embarrassing HTML warnings.

I’m using this setup mainly for HMR during development, but it has also proved to be a better and more efficient alternative to the original setup.

## Related Issue(s)

None – this is just a playground.

## Checklist

  - [ ] Upgrade instructions
  - [ ] Configuration details
  - [ ] Concepts

## Labels

- [ ] Includes a DB migration? → add the `area:migration` label


hardillb:
@cstns I think we should close this as it's now way behind head and things have been better on the memory front for the normal dev env setup (I'm trying to get clean up the stale PRs on this repo)

cstns:
this was never meant to be released into the wild, just to serve as test bed for webpacks HMR capabilities but yeah, nobody bothered. Safe to close

PR #5896, opened 2025-08-08
Support for annual billing

### LINKED ISSUE #4318: Support for annual billing option
Part of https://github.com/FlowFuse/customer/issues/233

We want to provide the option for a user to pay annually, rather than the existing monthly model. This issue is for the technical work needed to achieve that.

### Current Model

 - A TeamType can have:
    - (A) a stripe product/price associated with it
    - (B) a stripe product/price associated with each instance type and device
 - Our Starter team has (A) - a monthly cost for the team to exist.
 - Our Team/Enterprise teams have (B) - no monthly cost for the team to exist, but cost per instance/device that exists (that varies based on the size of the instance)

### Desired Model

The desired model described in the linked issue is to have a minimum cost for Team/Enterprise - and for annual billing to be available in order to apply a discounted rate.

We have two options for implementing that:

1. Use the current model to add a stripe price/product to the Team/Enterprise tier teams. Then set the 'free' allocation of the available instance type to match what is included in that base price. This way the billing system won't do on-demand charging for the first X instances - only when they go above X.
   The problem with this approach is it doesn't allow the user to commit to more than the minimum number that the team price factors in. So whilst this may offer a quick solution, it doesn't get us closer to where we want to be.

2. The better model is, when setting up billing, let the user pick the desired team type, then be shown a slider for how many instances they want included - with the minimum applied. When they checkout, we then add the appropriate number of instances items to the subscription. The billing system can then ensure the item count on the subscription does not go below the defined minimum.

### Annual billing

Supporting annual billing option is fairly straight forward to do the basics:

1. The current billing interval is determined by the stripe price. So the starting point is to create yearly price objects for each stripe product.
4. The TeamType configuration (which holds tier-specific pricing for each instance type), needs extending to include the yearly price details (two fields to add: stripe price ID, and the price we show on the page) for the team.
5. Add toggle on Upgrade Team page (where the user picks Team Tier), for monthly/yearly pricing.
6. Add flag on api to indicate monthly/yearly billing choice

### Problems to resolve

If a user subscribes with Annual billing for 5 instances, what happens when they create their 6th instance.

 - we currently add prorata usage to their *next* invoice. On a monthly schedule, that's acceptable - they pay for what they have used over the last month.
 - For annual billing, that model doesn't work: user subscribes on Jan 1st for 5 instances. On Jan 2nd they create 30 instances - we don't charge them until their next invoice a year from now.
 - The alternative is we invoice immediately for any billing changes beyond their annual commitment. But then we are then back to raising an invoice for each billable change the user makes. That is something we were specifically asked to stop doing.

**We need to resolve the desired behaviour for on-demand usage that exceeds their initial subscription.**

knolleary:
Closes #4318 

This adds the ability for a TeamType to be configured with both monthly and annual stripe pricing options.

I have targeted the branch of #5875 to avoid merge conflicts.

## Commit Summary

 - https://github.com/FlowFuse/flowfuse/commit/cc05016bb9b032210c6bda627ea7f269779a17ec - updates Subscription model to store `interval` property and exposes on `Team` api object
 - https://github.com/FlowFuse/flowfuse/commit/233d3d3950bf9207e0f6ddaac8da3b201156681a - updates TeamTier Edit Dialog to allow entry of annual billing configuration
 - https://github.com/FlowFuse/flowfuse/commit/17d5edc8019af206cd7d2b1c430fabe86321f142 - backend work to ensure the annual billing prices are applied when `interval` is set to `year`.
 - https://github.com/FlowFuse/flowfuse/commit/5f84f514899b34a02f2de3275827ee2a11ba7a16 - adds a monthly/yearly toggle to the UI to allow the user to pick what they want.

## Details

In the team type UI, the options are presented as:

 - Team config
    image

 - Instance config
   image

It is important that annual pricing/descriptions are provided for everything within the team config (same idea as the existing monthly items).

----

There are then a number of scenarios where the user needs to be provided the choice of monthly/annual billing.

1. Creating their first team
2. Creating their second team
3. Changing the tier of their current team
4. Setting up billing post-trial

We appear to have three different UIs doing similar jobs across these scenarios.

### Creating their first team

Note - the code supports annual billing only being available for *some* team types. In the screenshot below, I have toggled to yearly billing, but I haven't configured Starter with any annual plan details so the select button is hidden.

image

### Creating their second team

This uses the original team select UI. As above, Starter doesn't have annual in my local setup, so gets disabled when the toggled to yearly.

image

### Changing tier

This uses the same look and feel as the 'Create second team' view, but is its own component.

image

### Setting up billing post trial

Don't have a screenshot to hand, but its reusing one of the above two UIs (not the shiny top UI).

---

## Backend

We record the `interval` type of the subscription on our `Subscription` object. This ensures we pickout the right stripe price items when updating the subscription as the user makes changes.

In `ee/lib/billing/index.js:updateTeamBillingCounts` we do the work to update the subscription as the use adds/removes instances.

If annual billing is enabled, we set `pending_invoice_item_interval` on the request to stripe as:
```
{
   interval: 'month',
   interval_count: 1
}
```

This means they will be invoiced within a month for any changes they make, but the charge will be pro-rata usage for the remaining year of their billing cycle.

### DIFF
--- a/frontend/src/pages/team/Tables/Table/TableExplorer/drawers/CreateTable.vue
+++ b/frontend/src/pages/team/Tables/Table/TableExplorer/drawers/CreateTable.vue
@@ -13,7 +13,7 @@
             <div class="section table-name">
                 <h3>Define name</h3>
                 <ff-text-input
-                    v-model="tableName"
+                    v-model="newTable.name"
                     placeholder="Your table's new name"
                     type="string"
                     :error="errors.name"
@@ -34,14 +34,14 @@
                     <!-- <span class="col-span-1 title -ml-2">Unsigned</span>-->
                 </div>
                 <ul class="columns">
-                    <li v-for="(column, $key) in columns" :key="$key">
-                        <table-column :column="column" @remove="removeColumn($key)" />
+                    <li v-for="(column, $key) in newTable.columns" :key="$key">
+                        <table-column :column="column" @remove="removeNewTableColumn($key)" />
                     </li>
                 </ul>
                 <div v-if="errors.columns" data-el="form-row-error" class="ml-4 text-red-400 text-xs text-center p-5">
                     {{ errors.columns }}
                 </div>
-                <ff-button type="button" kind="secondary" class="w-full" @click="onNewColumn">Add a new column</ff-button>
+                <ff-button type="button" kind="secondary" class="w-full" @click="addNewTableColumn">Add a new column</ff-button>
             </div>
         </div>
     </div>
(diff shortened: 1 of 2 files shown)


cstns:
## implemented fixes

a31621417de959f49025fe3da6f1c1df66ef94f8
- pass along billing interval params to non-trial teams to correctly display the annual billing amount while also passing it to the createTeam API to correctly select the annual price product for the specific product

58f832d9ab2493808817aaacb624058336dc5643
- reset team type selection when switching between monthly/yearly billing when changing team types

90c9921a96e2b8c328aadbe4328c0a329ba0fdbf  
- align ui between the three team creation/change pages
- fixed missing billing interval toggle and tile functionality for the create team page (when the user had an existing team)

e4dd1e9e0e90b9f9cbfd93de9cc8040a4799df9e
- fixed a bug/omission in which the teamTypeTile was using the old and deprecated way of contacting sales, replaced the tile selection action the new one in which we open the hs calendar for the user to book a meeting  
- required fixing it because this was the first step a user could take to create an enterprise team after creating an account

cstns:
## Test scenarios

- create a team after signing up with a new user with annual billing
	- see a31621417de959f49025fe3da6f1c1df66ef94f8
	- tested ok after the fix, subscription with yearly recurrence in stripe 

- update a team from monthly to annual billing (currently can only be done by also switching team types)
	- created new team with monthly billing
	- advanced stripe subscription time by 2 weeks
	- changed team to another team type but with annual billing
	- new invoice in stripe w $1404 charge ($1500 yearly -  unused time on TeamPlan which appears on the invoice as -$95.91)
inv-0

- update a team from annual to monthly billing (currently can only be done by also switching team types)
	- created a new team with annual billing
	- advanced stripe subscription time by 4 weeks
	- changed team to another team type but with monthly billing
	- stripe correctly shows price downgrade in logs with a new invoice, applied remaining balance of $1,384.92 to the account
inv-1
	- advanced stripe subscription time by 5 weeks
	- stripe deducted another month's worth from the applied balance, billed 0
inv-2
	- billing area correctly displays the credit banner balance with correct amount after two month's worth of $175 of  $1,034.92
inv-3

- create a team with annual billing when other teams present
	- see a31621417de959f49025fe3da6f1c1df66ef94f8
	- tested ok after the fix, subscription with yearly recurrence in stripe 

- create a trial with annual
	- subscription created only locally, picked up by the licenseCheck housekeeper task


- exceed instance limit & exceed device limit
	- additional subscriptions created
	- see https://github.com/FlowFuse/flowfuse/issues/5913
	- see https://github.com/FlowFuse/flowfuse/issues/5914

- for both devices and instances, after exceeded team limit & fast forward stripe by 2 weeks, delete instance
	- devices: No Stripe subscription created when exceeding team device allowance #5916
	- instance got created and billed anually
	- instances subscription created succesfully, after advancing time to bill additional device, the team plan and instance subscription seem to merge into one
inv-4

- create team, use annual billing fast forward stripe 2 weeks and delete team
	- subscriptions get cancelled
	- refund n/a

- create team, use annual billing fast forward stripe 2 weeks and delete a team that has exceeded devices and instances
	- required to manually delete instances and devices before deleting team
	- see no stripe subscription when exceeding device allowances #5916
	- additional instance subscription not deleted/prorated after instance numbers within limit

- downgrade to lesser type after fast forwarding 2 weeks of annual subscription, w and w/o instances/devices
	- prompted by warning stating i need to reduce remote instance numbers due to limit
		- returning to the billing > cchange team type page resulted in same error with same number of devices (efter deleting devices) hard refresh fixed it
		- subscription updated in stripe, charges are getting prorated

cstns:
Created the following follow up tasks:

- unify the create team selector/pages https://github.com/FlowFuse/flowfuse/issues/5909
- update stripe sdk from v8.222.0 to v18.4.0 https://github.com/FlowFuse/flowfuse/issues/5910
- allow switching between monthly/yearly billing https://github.com/FlowFuse/flowfuse/issues/5911
- E2E tests https://github.com/FlowFuse/flowfuse/issues/5912
- Annual billing for additional hosted/remote instances https://github.com/FlowFuse/flowfuse/issues/5913
- Billing page shows incomplete upcoming charges for teams with annual plan and monthly add-ons https://github.com/FlowFuse/flowfuse/issues/5914

## Bugs:
Extra instance subscriptions not cancelled, refunded, or prorated after deletion [#5917](https://github.com/FlowFuse/flowfuse/issues/5917)
No Stripe subscription created when exceeding team device allowance [#5916](https://github.com/FlowFuse/flowfuse/issues/5916)

cstns:
@joepavitt , tagged you for a quick re-review of the changes I made on the front end side

joepavitt:
@gstout52 FYI - see follow up tasks detailed [above](https://github.com/FlowFuse/flowfuse/pull/5896#issuecomment-3180158757)

joepavitt:
You have a conflict to resolve @cstns

joepavitt (approved):

joepavitt:
Awaiting tests to pass before final merge, but all looks good

joepavitt:
@cstns this is good to go now, right?

cstns:
yes, merged!

PR #5903, opened 2025-08-11
Allow Direct SSO login

### LINKED ISSUE #5606: Allow instant redirect to SSO provider if only 1 SSO provider configured
### Description

> At the moment, users have to log in twice: once in FlowFuse (entering their email address), and then they are redirected to our ADFS (Microsoft Cloud) environment to log in again with their email address, password, and 2FA.
> 
> This happens even though we’ve defined only one SSO provider in FlowFuse. In our view, the initial login step (especially since there is only one SSO provider) is redundant. We would prefer if users were redirected straight to our Microsoft Cloud environment for login. That way, they would only need to log in once to access FlowFuse.

As a FlowFuse Admin
I want to skip the FlowFuse login step before redirect to SSO provider
To speedup login

Enterprise Tier Only (EE)

hardillb:
fixes #5606

## Description

Adds the ability to directly jump to a given SAML SSO provider rather than entering email address and getting redirected.

image

Admin setting

image

Shows a button for any active SAML providers.

## Related Issue(s)

#5606

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/Login.vue
+++ b/frontend/src/pages/Login.vue
@@ -50,6 +50,17 @@
                         </GoogleLogin>
                         <span class="ff-error-inline" data-el="errors-googleSSO">{{ errors.googleSSO }}</span>
                     </template>
+                    <template v-if="directSSOEnabled">
+                        <hr class="mb-4">
+                        SSO Direct
+                        <ul>
+                            <li v-for="{name, id} in settings['platform:sso:direct:list']" :key="id">
+                                <ff-button kind="secondary" :data-action="`direct-sso-${id}`" @click="directSSO(id)">
+                                    <span>{{ name }}</span>
+                                </ff-button>
+                            </li>
+                        </ul>
+                    </template>
                 </div>
             </template>
             <template v-else>
@@ -118,6 +129,11 @@ export default {
         },
         googleSSOEnabled () {
             return this.settings['platform:sso:google'] && this.settings['platform:sso:google:clientId']
+        },
+        directSSOEnabled () {
+            return !!this.settings['platform:sso:direct:list'] &&
+                Array.isArray(this.settings['platform:sso:direct:list']) &&
+                this.settings['platform:sso:direct:list'].length >= 1
         }
     },
     watch: {
@@ -229,6 +245,9 @@ export default {
                 // Handle error response - not sure what this will look like yet
                 console.error(result)
             }
+        },
+        async directSSO (id) {
+            window.location = `/ee/sso/login?p=${id}${this.$route.query.r ? `&r=${this.$route.query.r}` : ''}`
         }
     }
 }
--- a/frontend/src/pages/admin/Settings/General.vue
+++ b/frontend/src/pages/admin/Settings/General.vue
@@ -189,6 +189,16 @@
             </FormRow>
         </template>
 
+        <template v-if="ssoEnabled">
+            <FormHeading>Direct SSO</FormHeading>
+            <FormRow v-model="input['platform:sso:direct']" type="checkbox" data-el="direct-sso">
+                Show buttons on Login page to jump directly to a SAML SSO provider
+                <template #description>
+                    Allows bypassing email matching for SAML SSO logins
+                </template>
+            </FormRow>
+        </template>
+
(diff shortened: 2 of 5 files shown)


joepavitt:
For anyone using this in a self-hosted environment, do they need to know it's "SSO Direct"? Also, if they're using it, I suspect it'll be favoured over social logins, and so, should appear above?

joepavitt (changes requested):
Feedback above

hardillb:
@joepavitt 

> For anyone using this in a self-hosted environment, do they need to know it's "SSO Direct"?

Not sure what you mean here? Do you mean the section in the Admin settings should call it "SSO Direct" or that we should change the label?

> Also, if they're using it, I suspect it'll be favoured over social logins, and so, should appear above?

I very much doubt that anybody using this will also be using social login.

I could just remove the "SSO Direct" label an pre-pend the SSO provider name with "Sign in with " so it matches the Google button?

joepavitt:
> I very much doubt that anybody using this will also be using social login.

> I could just remove the "SSO Direct" label an pre-pend the SSO provider name with "Sign in with " so it matches the Google button?

Yep, that's what I was alluding to. Let's go with that. Exposing the "SSO Direct" labelling felt too technical, users won't care about that.

frontend/src/pages/Login.vue
```diff
@@ -50,6 +50,17 @@
                         </GoogleLogin>
                         <span class="ff-error-inline" data-el="errors-googleSSO">{{ errors.googleSSO }}</span>
                     </template>
+                    <template v-if="directSSOEnabled">
+                        <hr class="mb-4">
+                        SSO Direct
```
hardillb:
```suggestion
```

frontend/src/pages/Login.vue
```diff
@@ -50,6 +50,17 @@
...
+                        <ul>
+                            <li v-for="{name, id} in settings['platform:sso:direct:list']" :key="id">
+                                <ff-button kind="secondary" :data-action="`direct-sso-${id}`" @click="directSSO(id)">
+                                    <span>{{ name }}</span>
```
hardillb:
```suggestion
                                    <span>SIGN IN WITH {{ name.toUpperCase() }}</span>
```

frontend/src/pages/admin/Settings/General.vue
```diff
@@ -189,6 +189,16 @@
             </FormRow>
         </template>
 
+        <template v-if="ssoEnabled">
+            <FormHeading>Direct SSO</FormHeading>
```
hardillb:
```suggestion
            <FormHeading>Direct SSO Login</FormHeading>
```

hardillb:
@joepavitt changes made

joepavitt:
Screenshot 2025-08-15 at 10 11 27

What documentation do we have here that we can link to? How does a user set this up, as it's not just as simple as ticking this box.

hardillb:
> What documentation do we have here that we can link to?

None yet, I can add a section to the SSO configuration docs (https://github.com/FlowFuse/flowfuse/pull/5903/commits/d28f67f79046d89d4044669b06097f40fd0a4f37)

> How does a user set this up, as it's not just as simple as ticking this box.

It is, but it's only going to show up if you have setup a SAML SSO provider, if you don't have any then you won't get any buttons.

see the test at ~130 on login.vue

joepavitt:
> None yet, I can add a section to the SSO configuration docs (https://github.com/FlowFuse/flowfuse/commit/d28f67f79046d89d4044669b06097f40fd0a4f37)

So how are customers setting this up now if we have no documentation? It is:

> You have setup a SAML SSO provider,

This step that I'm referring to for a link. We should point users to that documentation from the admin settings so that users can understand how to set it up if they need to.

hardillb:
@joepavitt https://flowfuse.com/docs/admin/sso/saml/

I've added a link to the Admin settings page

joepavitt:
Thanks Ben, I've just shifted it into the description so that it doesn't also toggle the option when clicking the link.

Issue #3666, opened 2024-04-01
SubFlow Version control


zackwasli:
### Description

The idea here is for a sub-flow to function like a snapshot and act as a single source of truth. 

If said sub-flow changes, then it could propagate to all instances that have said sub-flow.

This would be helpful in updating and maintaining similar instances that all reply upon the same sub-flow portion, but contain unique characteristics that can not be controlled for with snapshots and environment variables.

### Which customers would this be available to

Enterprise Tier Only (EE)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Crossover with https://github.com/FlowFuse/flowfuse/issues/3841

joepavitt:
And crossover with https://github.com/FlowFuse/flowfuse/issues/3230

knolleary:
Now that we have support for a team-wide npm repository, the proposal here is to take advantage of the fact subflows can be packaged and distributed as npm modules.

The high level workflow would be:

1. User develops subflow in Instance A.
2. When they want to share it with other instances, they package and publish to the team node repository
3. The module can then be installed into any other team instances via the palette manager
4. If there is an update to distribute, the subflow is updated in Instance A, then an updated module published to the team repo
5. The module can then be updated via the palette manager in other instances.

The technical guide for how the packaging works is available [here](https://nodered.org/docs/creating-nodes/subflow-modules).

Subflows already have a 'Module Properties' tab in their edit dialog that allows the user to enter much of the required info:

<img width="531" alt="Image" src="https://github.com/user-attachments/assets/f3c3b172-f45c-4cca-868e-cc2044fcaaf5" />

From a UX point of view, the missing part is then the ability to publish to the team repository.

### Proposal

 - Create a NR plugin that provides:
    - UI sidebar for managing subflow modules
      - List subflows, allow user to view/edit the module meta data
         - The existing form is free-form. We have strict requirements on module name when working with the team repo - need to validate and enforce in the sidebar UI
         - Have a 'publish' button to trigger npm publish on the backend

#### Open problem: npm credentials

Currently instances/devices are provisioned with individual npm read-only token that do not have permission to publish to the registry. But this work will require the instance to have access to a write token.

Write tokens are scoped to individual users; for RBAC and auditing.

One potential solution; create an api end point that generates a short-lived npm token for a user. This can be called by the plugin backend as part of the publish workflow. (This short-lived token should not replace any existing token the user has provisioned manually via the UI).

[...]

PR #5928, opened 2025-08-19
Subflow publish

hardillb:
part of #3666
## Related Issue(s)
Adds API endpoint to push a subflow module to the team npm registry

PUT `/api/v1/teams/:teamId/npm/subflow`

Also adds feature flag to Instance settings for nr-launcher to add to `settings.js`

#3666

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/ee/routes/catalogues/lib/subflow.js
```diff
@@ -0,0 +1,128 @@
...
+//         username: 'admin',
+//         password: 'secret'
+//     }
+// })
+// .then(n => {
+//     console.log(n.data)
+// })
+// .catch(err => {
+//     console.log(err)
+// })
```
Steve-Mcl: commented code

forge/ee/routes/catalogues/index.js
```diff
@@ -62,6 +64,88 @@ module.exports = async function (app) {
...
+     * @name /api/v1/teams/:teamId/npm/subflow
+     * @static
+     * @memberof forge.routes.api.team.npm
+     */
+    app.put('/npm/subflow', {
+        preHandler: async (request, reply) => {
+            if (request.session.ownerType !== 'project') {
+                reply.code(401).send({ code: 'unauthorized', error: 'unauthorized' })
+            }
+            const project = await app.db.models.Project.byId(request.session.ownerId)
```
Steve-Mcl, line 79: Are we not supporting devices?

forge/ee/routes/catalogues/index.js
```diff
@@ -62,6 +64,88 @@ module.exports = async function (app) {
...
+        const package = request.body.package
+        const subflowJSON = request.body.subflow
+        if (!package || !subflowJSON) {
+            reply.status(422).send({})
+            return
+        }
+        if (!package.name.startsWith(`@flowfuse-${request.team.hashid}/`)) {
+            reply.status(403).send({})
+            return
+        }
```
Steve-Mcl: should we not be doing a semver check on `package.version` and other sanity checks on `.license`, `.author`, `.description`, ...?

package.json
```diff
@@ -109,6 +109,7 @@
         "sequelize": "^6.33.0",
         "sqlite3": "^5.1.6",
         "stripe": "^8.222.0",
+        "tinytar": "^0.1.0",
```
Steve-Mcl: Ben, this lib is >9yo and has an engine check that is fixed at Node v4.x...

```
$ npm install tinytar
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'tinytar@0.1.0',
npm warn EBADENGINE   required: { node: '^4.0.0', npm: '^2.0.0' },
npm warn EBADENGINE   current: { node: 'v20.19.1', npm: '10.8.2' }
npm warn EBADENGINE }
```

We might want to consider using something more mainstream? The lib that seems to be "up there" (with over 40m DL per week) is `tar-stream` (https://www.npmjs.com/package/tar-stream)

hardillb: The tar file format hasn't changed in 50 years....

I need something that outputs a buffer not a stream, but I'll have a look later.

forge/ee/routes/catalogues/index.js
hardillb: Only on the version, the other can be empty

forge/ee/routes/catalogues/index.js
hardillb: No

package.json
Steve-Mcl: > The tar file format hasn't changed in 50 years....

😆😆😆

forge/ee/routes/catalogues/index.js
```diff
@@ -62,6 +65,92 @@ module.exports = async function (app) {
...
+        schema: {
+            summary: 'Allows Subflow packages to be stored in the Team Registry',
+            tags: ['NPM Package'],
+            body: {
+                type: 'object',
+                properties: {
+                    package: {
+                        type: 'object',
+                        additionalProperties: true
+                    },
```
Steve-Mcl: Since name and version are require props, we should let swagger take care of checking and handing this:


```suggestion
                    package: {
                        type: 'object',
                        properties: {
                            name: { type: 'string' },
                            version: { type: 'string' }
                        },
                        required: ['name', 'version'],
                        additionalProperties: true
                    },
```

forge/ee/routes/catalogues/index.js
```diff
@@ -62,6 +65,92 @@ module.exports = async function (app) {
...
+                }
+            }
+        }
+    }, async (request, reply) => {
+        const package = request.body.package
+        const subflowJSON = request.body.subflow
+        if (!package || !subflowJSON) {
+            reply.status(422).send({ error: 'missing_values', message: 'Missing values' })
+            return
+        }
```
Steve-Mcl:
```suggestion
```

This will never be hit because they are declared in schema

forge/ee/routes/catalogues/index.js
```diff
@@ -62,6 +65,92 @@ module.exports = async function (app) {
...
+        if (!package.name.startsWith(`@flowfuse-${request.team.hashid}/`)) {
+            reply.status(403).send({ error: 'not_authorized', message: 'Not Authorized' })
+            return
+        }
+        if (!package.version || !semver.valid(package.version)) {
+            reply.status(422).send({ error: 'bad_version', message: 'Invalid semver' })
+            return
+        }
```
Steve-Mcl:
```suggestion
        if (!semver.valid(package.version)) {
            reply.status(422).send({ error: 'bad_version', message: 'Invalid semver' })
            return
        }
```

Let swagger take care of missing props

Steve-Mcl (commented):
Some minor points here. I think it is better we allow the scema to take care of missing props.

This will also make them into `400` errors (which is probably best for missing props)

Apart from that, there are no tests to prevent regressions - I have wrote some and will PR them shortly.

forge/ee/routes/catalogues/index.js
hardillb: But a 422 error is more correct than a straight 400

forge/ee/routes/catalogues/index.js
hardillb: Same again 422 is the better status code

forge/ee/routes/catalogues/index.js
Steve-Mcl: Although I normally avoid WebDav status codes like 422, I wont argue over the use of it for general HTTP APIs.  However, while 422 is fine for "bad" values, 400 is the better call for "missing" props.  Its how we do the majority of our APIs (mark required fields as required in the schema and swagger/fastify do the necessary!)

That said, if there is a specific reason for differing from the norms and sending a 422 is necessary, we can proceed off-spec.  If so, then the tests in #5947 will need to be adjusted - let me know if you want me to pick that up

forge/ee/routes/catalogues/index.js
Steve-Mcl: Same again

> Although I normally avoid WebDav status codes like 422, I wont argue over the use of it for general HTTP APIs.  However, while 422 is fine for "bad" values, 400 is the better call for "missing" props.  Its how we do the majority of our APIs (mark required fields as required in the schema and swagger/fastify do the necessary!)

> That said, if there is a specific reason for differing from the norms and sending a 422 is necessary, we can proceed off-spec.  If so, then the tests in #5947 will need to be adjusted - let me know if you want me to pick that up

PR #5956, opened 2025-08-27
Add api o generate snapshot diff description

cstns:
## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -1419,14 +1419,16 @@ module.exports = async function (app) {
     }, async (request, reply) => {
         const options = {}
         let isTeamOnTrial
+
         const latestSnapshot = await request.project.getLatestSnapshot()
+        // we need to 'build' the snapshot to compare it
+        const currentSnapshot = await app.db.controllers.ProjectSnapshot.buildSnapshot(
+            request.project,
+            request.session.User,
+            options
+        )
+
         let previousState = {}
-
-        if (app.billing && request.project.Team.getSubscription) {
-            const subscription = await request.project.Team.getSubscription()
-            isTeamOnTrial = subscription ? subscription.isTrial() : null
-        }
-
         if (latestSnapshot) {
             const latestSnapshotJson = latestSnapshot.toJSON()
             previousState = {
@@ -1437,12 +1439,6 @@ module.exports = async function (app) {
             }
         }
 
-        const currentSnapshot = await app.db.controllers.ProjectSnapshot.buildSnapshot(
-            request.project,
-            request.session.User,
-            options
-        )
-
         const currentState = {
             name: currentSnapshot.name,
             description: currentSnapshot.description,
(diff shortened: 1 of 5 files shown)


cstns:
I may need to retrieve the last snapshot that's not auto-generated when sending requests to the llm because it can't figure out what's changed since the last snapshot is the current state

forge/routes/api/project.js
```diff
@@ -1415,6 +1415,96 @@ module.exports = async function (app) {
...
+        const currentSnapshot = await app.db.controllers.ProjectSnapshot.buildSnapshot(
+            request.project,
+            request.session.User,
+            options
+        )
+
+        const previousState = latestSnapshot
+            ? {
+                settings: latestSnapshot.toJSON().settings,
+                flows: latestSnapshot.toJSON().flows
```
Steve-Mcl: calling `latestSnapshot.toJSON()` multiple times is a bit inefficient TBH.

forge/routes/api/project.js
```diff
@@ -1438,6 +1528,86 @@ module.exports = async function (app) {
         return Object.entries(mergedKV).filter(e => !!e[0]).map(([k, v]) => { return { name: k, value: v } })
     }
 
+    function deepDiff (currentState, previousState) {
```
Steve-Mcl: Should `deepDiff` not be extracted to a reusable/unit-testable lib?

forge/routes/api/project.js
```diff
@@ -1415,6 +1415,96 @@ module.exports = async function (app) {
...
+                    $ref: 'APIError'
+                }
+            }
+        }
+
+    }, async (request, reply) => {
+        const options = {}
+        let isTeamOnTrial
+
+        const latestSnapshot = await request.project.getLatestSnapshot(true)
```
Steve-Mcl: What happens if there are no snapshots?
Should there be a test for "no latest snapshot"?

forge/routes/api/project.js
```diff
@@ -1415,6 +1415,96 @@ module.exports = async function (app) {
...
+
+        const { currentStateDiff, previousStateDiff } = deepDiff(currentState, previousState)
+
+        // redact env var values
+        if (currentStateDiff.settings?.env) {
+            Object.keys(currentStateDiff.settings.env).forEach(k => currentStateDiff.settings.env[k] === 'REDACTED')
+        }
+        if (previousStateDiff.settings?.env) {
+            Object.keys(previousStateDiff.settings.env).forEach(k => previousStateDiff.settings.env[k] === 'REDACTED')
+        }
```
Steve-Mcl, line 1501: If you set all env values to `REDACTED`, wont they match in the AI diff?

forge/routes/api/project.js
cstns: yes, I had the exact thought but did not now where exactly to extract it to and forgot about it along the way. Ill gladly extract it jsut point the way

forge/routes/api/project.js
cstns: that's a good catch, it might return undefined. Thinking of it, i can return a the latest snapshot if it exists or an empty object, functionality being preserved that way. The empty object will mock a empty snapshot

forge/routes/api/project.js
cstns: I remember adding a system prompt to take care of that but can't find it anymore. I added one now. It instructs the LLM that it's only given parts of the flows that changed not the full objects. Also added a new one reg redacted values

Steve-Mcl (approved):
looking good

Steve-Mcl:
Before we merge, there is one thing I should probably clarify:

This is for projects snapshots only right? I.e. you wont be adjusting the UI to use this feature for device snapshots (as the code is only touching project stuff)

cstns:
Yes, only projects at this time. The UI follow up is exposing the generate description button only for the project/instance create snapshot dialog.

Steve-Mcl:
> * Link to Changelog Entry PR, or note why one is not needed.

NOTE: I permitted this to merge as it is only backend part. I will ensure the front-end part has this checked 😉

Issue #5959, opened 2025-08-27
Add API to generate Snapshot diff description for remote instances


cstns:
### Description

similar implementation can be found for hosted instances: https://github.com/FlowFuse/flowfuse/issues/5843

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
@cstns is this now "In Review", and if so, it needs a reviewer assigned.

cstns:
still working on unit tests on it, will put it in review once done

Issue #5843, opened 2025-07-30
Add API to generate Snapshot diff description


joepavitt:
### Description

New API for FlowFuse (EE) which has the following input/output. Will then be used by https://github.com/FlowFuse/flowfuse/issues/5842

### Input:

- "Before" flow
- "After" flow

### Output

- Text-based description of the changes

### Epic/Story

https://github.com/FlowFuse/flowfuse/issues/5758

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
@Steve-Mcl Serban is going to start this work today, does it make sense (I think it does?) to add this as a `method` to the `flow-gen` flows running for NR Assistant?

Steve-Mcl:
It is certainty doable.

I would be tempted to iterate the flows and filter out unchanged nodes to minimise what is posted to avoid going over the context window (and speed it up as a by-product).  I would also likely breakdown the sections of the snapshot into constituent parts (flow, settings, modules etc). (modules, are an easy/basic logic check so would likely not send that to AI, but iterate it manually).

As for system prompts to accompany it, I can definitely offer some insight into what will work well. Including things like giving prop names a description to aid the AI.

As touched on in last meeting, there may not be any need to modify existing APIs for assistant (the current architecture _should_ accommodate new requests) but the flow-get routines will need a separate pathway/route.

PR #5966, opened 2025-08-29
Add api to generate remote instance snapshot description

cstns:
## Description

Expose a `/:deviceId/generate/snapshot-description` endpoint for remote instances

- similar implementation to https://github.com/FlowFuse/flowfuse/issues/5843
- Compares the device’s current state vs its latest snapshot, computes a structured diff, and invokes the internal assistant to produce a human-readable description
- Computes a structured diff for flows and settings, with sensitive values redacted
- Supports both instance-owned and application-owned devices
- Invokes the internal assistant with relevant context to produce the summary
- Includes permission checks; license/tier gating can be enabled later
- Refactors snapshot-building logic to separate “build” from “persist,” enabling non-persistent comparisons

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5959

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/api/application.js
+++ b/frontend/src/api/application.js
@@ -247,8 +247,10 @@ const updatePipeline = async (applicationId, pipeline) => {
  * @param {string} applicationId - The ID of application to get snapshots for
  * @param {string} cursor
  * @param {string} limit
+ * @param options
+ * @param query
  */
-const getSnapshots = async (applicationId, cursor, limit, options) => {
+const getSnapshots = async (applicationId, cursor, limit, options, query = null) => {
     const extraParams = {}
     if (options?.deviceId) {
         extraParams.deviceId = options.deviceId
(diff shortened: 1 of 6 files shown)


Steve-Mcl:
@cstns given the issue #5983 do you want to hold off on this?

cstns:
yes, fine with holding

Steve-Mcl (commented):
This does not work as expected.

I created a snapshot "from device" in the Developer Mode tab. This is visible in the Instance->snapshots.
Then i added a 2nd flow and created a 2nd snapshot - the AI said I had created 2 flows when actually I only added a 2nd flow.


#### No snapshots in dropdown
Likely the reason for above to the above:

Developer Mode -> Create Snapshot

Image

As you can see, there were snapshots taken
Device owner instance -> snapshots

Image

cstns:
I presume it's because of the legacy instance owned device snapshots and associated quirks. I used the same api endpoint as the snapshots tab uses to retrieve snapshots because it was also capable of searching through a devices snapshots and that one is bound to the application the device is assigned to. 

If the device is not assigned to an application I assume it's either sending a malformed request or receiving an empty response.

The fix here would be to use the device `/api/v1/devices/:deviceId/snapshots/` get endpoint when the device has no application assigned.

Give me a couple of moments

cstns:
this should be treated on the UI PR as that's where the functionality resides

cstns:
fixed in https://github.com/FlowFuse/flowfuse/pull/5970/commits/9524caced724844b92771ac655e8f6c571891a90. The question now becomes, should we toggle between the application/device snapshot endpoints or just stick to one?

Steve-Mcl (approved):
GTG

Steve-Mcl:
final checks  then merge

Issue #5939, opened 2025-08-20
Lefthand side bar missing when coming from shared link


Steve-Mcl:
### Current Behavior

![Image](https://github.com/user-attachments/assets/c2734fde-1c8c-4cb9-a99e-4b20a5eafb2a)

Link opens page without only top bar
After a few seconds, the page loads but no sidebar is present
A page refresh resolves this

### Expected Behavior

should open as swiftly as it does with a refresh

### Steps To Reproduce

visit an FFC device via `https://app.flowfuse.com/device/<device-id>/overview` hyperlink

### Environment

- FlowFuse version: 2.22.0
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

cstns:
The culprit seems to be a race condition caused by the fact that when opening a new page `teamMembership` is not yet loaded when the menu gets rendered and is no re-evaluated after it loads through the `/api/v1/teams/<team-id>/user` api call.

Once the api call is made it gets stored in local storage so a page refresh properly evaluates the prop and renders the menu.

cstns:
I created a follow-up task to fix any possible lingering reactivity issues related to permissions https://github.com/FlowFuse/flowfuse/issues/5974

PR #5973, opened 2025-09-01
Fix lefthand side bar missing when accessing from a shared link

cstns:
## Description

Fixes the left-hand sidebar not rendering when opening settings via shared/deep links.

- traced the root cause to a race condition where `teamMembership` isn’t loaded yet when the menu first renders, and the menu isn’t re-evaluated after the `/api/v1/teams/<team-id>/user` response arrives
- exposed the permissions functions as direct exports to be used in vuex modules
  - getters must be pure, sync, and only depend on state and other getters
  - vuex re-runs a getter only when it reads reactive state. Using the composable hid that dependency, so vuex couldn't track it
- because I refactored the permission composable and stored `teamMembership` into a prop, i also had to replace the permissions mixin with the composable to preserve reactivity

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5939

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/composables/Permissions.js
+++ b/frontend/src/composables/Permissions.js
@@ -1,19 +1,78 @@
+import { computed } from 'vue'
 import { useStore } from 'vuex'
 
 import { Permissions } from '../../../forge/lib/permissions.js'
 import { Roles } from '../utils/roles.js'
 
 /**
- * Fixes reactivity issue with the useStore utility but throws a warning while in dev mode
- * [Vue warn]: inject() can only be used inside setup() or functional components
- * Since teamMembership is defined outside the composable function and updated conditionally in usePermissions()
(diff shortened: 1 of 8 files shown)


joepavitt:
@cstns noticed that the underlying issue wasn't scheduled for `2.22`, so I've updated it. If this is ready for review, please assign it to someone.

cstns:
graciously assigned you as the reviewer

cstns:
we should also consider https://github.com/FlowFuse/flowfuse/issues/5974. It's a byproduct of this task and some tidying up needed for some loose ends with mixins. Did not assign it a milestone because it didn't go through planning, but I estimated the subtasks at a total of 1SP, it should be a quick win

PR #5979, opened 2025-09-02
API for assistant inline code completion

### LINKED ISSUE #5978: Add API support for inline completions
### Description

_No response_

### Epic/Story

_No response_

Steve-Mcl:
## Description

- Added settings endpoint to Assistant API
  - GET /api/v1/assistant/settings
  - This permits the assistant to retrieve settings for a specific device or project to understand if it can use inline code completions
- Added fim code completion service endpoint
  - POST /api/v1/assistant/fim/:nodeModule/:nodeType
  - This route is very similar to the generic /api/v1/assistant/:method route except for 2 differences:
    - It requires the nodeModule and nodeType to be specified in the URL - so we know exactly which node is requesting the completion and can apply logic specific to that node
    - It has a much higher rate limit (60 requests per 30 minute window) - because code completions are expected to be used more frequently
- Added tests for new fim service endpoint
- Added tests for new settings endpoint

### Tests added `test/unit/forge/routes/api/assistant_spec.js`
```
▼ Assistant API
  ▼ service enabled
    ▼ service tests
      ▼ fim code completion service
        ✔ anonymous cannot access
        ✔ random token cannot access
        ✔ user token can not access
        ✔ device token can access
        ✔ instance token can access
        ✔ fails when prompt is not supplied
        ✔ fails when transactionId is not supplied
        ✔ fails when transactionId is mismatched
        ✔ contains owner info in headers for an instance
        ✔ contains owner info in headers for a device
        ✔ should include tables hints in context and cache it for subsequent requests
        ✔ should fetch fresh tables context hints after cache expiration
        ✔ should fail for unlicensed requests
        ✔ should fail for non pro/enterprise tier requests
    ▼ settings endpoint
      ✔ should return settings for device
      ✔ should return settings for project
      ✔ should fail for unauthorized requests
      ✔ should have inlineCompletions enabled for pro tier team
      ✔ should have inlineCompletions enabled for enterprise tier team
      ✔ should have inlineCompletions disabled for starter team
```

## Related Issue(s)

closes #5978

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/routes/api/assistant.js
+++ b/forge/routes/api/assistant.js
@@ -157,6 +157,132 @@ module.exports = async function (app) {
         }
     })
     /**
+     * Endpoint for FIM (fill-in-the-middle) code completion requests
+     * For now, this is simply a relay to an external assistant service
+     * In the future, we may decide to bring that service inside the core or
+     * use an alternative means of accessing it.
+     * @name /api/v1/assistant/fim/:nodeModule/:nodeType
+     * @static
+     * @memberof forge.routes.api.assistant
+     */
+    app.post('/fim/:nodeModule/:nodeType', {
+        config: {
+            rateLimit: app.config.rate_limits
+                ? {
+                    hook: 'preHandler', // apply the rate as a preHandler so that session is available
+                    max: 60, // max requests per window. Since the assistant plugin debounces requests, this should minimise risk of overuse without impacting user experience
+                    timeWindow: 30000, // 30 seconds window
+                    keyGenerator: (request) => {
+                        return request.ownerId || request.ip
+                    }
+                }
+                : false
+        },
+        schema: {
+            hide: true, // dont show in swagger
+            params: {
+                nodeModule: { type: 'string' },
+                nodeType: { type: 'string' }
+            },
+            body: {
+                type: 'object',
+                properties: {
+                    // The prompt to send to the assistant (required)
+                    prompt: { type: 'string' },
+                    // A correlation id for the transaction (required)
+                    transactionId: { type: 'string' },
+                    // Additional context for the completion (optional)
+                    context: { type: 'object', additionalProperties: true }
+                },
+                required: ['prompt', 'transactionId']
+            },
+            response: {
+                200: {
+                    type: 'object',
+                    additionalProperties: true
+                },
+                '4xx': {
+                    $ref: 'APIError'
+                }
+            }
+        }
+    },
+    async (request, reply) => {
+        // Check license and tier - FIM is only available for certain tiers
+        const isLicensed = app.license?.active() || false
+        const licenseType = isLicensed ? (app.license.get('dev') ? 'DEV' : 'EE') : 'CE'
+        const tier = isLicensed ? app.license.get('tier') : null
+        const fimTiers = [
+            'teams', // AKA "pro" tier
+            'enterprise'
+        ]
+        if (!fimTiers.includes(tier) && licenseType !== 'DEV') {
+            return reply.code(403).send({ code: 'forbidden', error: 'Forbidden' })
+        }
+
+        const nodeModule = request.params.nodeModule
+        const nodeType = request.params.nodeType
+        const supported = [
+            { nodeModule: 'node-red', nodeName: 'function' },
+            { nodeModule: '@flowfuse/node-red-dashboard', nodeName: 'ui-template' },
+            { nodeModule: '@flowfuse/nr-tables-nodes', nodeName: 'tables-query' }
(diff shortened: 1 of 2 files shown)


forge/routes/api/deviceLive.js
```diff
@@ -296,14 +296,25 @@ module.exports = async function (app) {
...
             response.assistant.completions = { ...app.config.assistant.completions }
         }
+        const isLicensed = app.license.active() || false
+        const licenseType = isLicensed ? (app.license.get('dev') ? 'DEV' : 'EE') : 'CE'
+        const tier = isLicensed ? app.license.get('tier') : null
+        const completionsTiers = [
+            'starter',
+            'teams', // AKA "pro" tier
+            'enterprise'
+        ]
```
hardillb: We have never tested for license 'type' before.

Also why do we need to test for team type? it should just test if the correct flag is present in the team, not what we (or any other enterprise licensed user has chosen to name their team)

forge/routes/api/project.js
```diff
@@ -835,14 +835,25 @@ module.exports = async function (app) {
...
             settings.assistant.completions = { ...app.config.assistant.completions }
         }
+        const isLicensed = app.license.active() || false
+        const licenseType = isLicensed ? (app.license.get('dev') ? 'DEV' : 'EE') : 'CE'
+        const tier = isLicensed ? app.license.get('tier') : null
+        const completionsTiers = [
+            'starter',
+            'teams', // AKA "pro" tier
+            'enterprise'
+        ]
```
hardillb: This is also very brittle, what happens when we intoduce a new team type next week (given the churn in billing this is VERY likely)

Steve-Mcl: Glad you picked up on that. I was unsure how we could achieve this any other way so this MVP hardcodes it.

For reference, the issue request is that "pro" (teams tier) and above are permitted to use this.

IIRC, we have no means of understanding what tiers are available (or, to be more precise, what _price_ordering_ they are)

Any thoughts?

forge/routes/api/deviceLive.js
Steve-Mcl: >Also why do we need to test for team type? it should just test if the correct flag is present in the team, not what we (or any other enterprise licensed user has chosen to name their team)

Hmm, if I am reading this right, you saying that we should add a feature flag right?

forge/routes/api/project.js
hardillb: We do this with feature flags on the team type, no other way, so all the teams that need it need the flag setting

Steve-Mcl: follow up since I answered this PR comment first - if I am reading this right, you saying that we should add a feature flag right?

forge/routes/api/deviceLive.js
Steve-Mcl: Added feature flag in [Enable Inline Completions via feature flag](https://github.com/FlowFuse/flowfuse/pull/5979/commits/c553e2bdf6abd77d421bf5603b0044f3e041eeae)

forge/routes/api/project.js
Steve-Mcl: Added feature flag in [Enable Inline Completions via feature flag](https://github.com/FlowFuse/flowfuse/pull/5979/commits/c553e2bdf6abd77d421bf5603b0044f3e041eeae)

Steve-Mcl:
@hardillb - addressed concerns & reset the review request.

forge/ee/lib/index.js
```diff
@@ -53,4 +53,7 @@ module.exports = fp(async function (app, opts) {
 
     // Set the Editor Limits Feature Flag
     app.config.features.register('generatedSnapshotDescription', true, true)
+
+    // Set the assistant inline completions Feature Flag
+    app.config.features.register('assistantInlineCompletions', true, true)
```
hardillb, line 58: Just double checking, this is meant for Team/Pro level licesnes as well as Enterprise for self hosted (e.g. not in the `app.license.get('tier') === 'enterprise'` test

hardillb (commented):
Code looks OK, but haven't run it locally (I think @joepavitt did as that was what I needed to fix his environment for)

Let me know if I need to run it (and if so when does the token you generated for me expire)

forge/ee/lib/index.js
Steve-Mcl: Discussed with Ben. This is ok to go.

hardillb (approved):

Issue #5983, opened 2025-09-03
AI Snapshot Generator reporting massive changes when a small node move was made


Steve-Mcl:
### Current Behavior

When generating AI description on an instance with only "Auto Snapshots", it is reporting as though everything in the flows were newly created

<img width="1019" height="825" alt="Image" src="https://github.com/user-attachments/assets/fde19537-e45d-49fc-88be-41018bcb7950" />

### Expected Behavior

Should either take the very last snapshot and describe changes OR permit the user to chose which snapshot to compare against for AI comparison.

As it stands, the text is very wrong and very misleading - I did NOT make all those changes, I simply moved an inject node and deployed.

### Steps To Reproduce

- Visit https://app.flowfuse.com/instance/97650584-ccdf-47c3-8825-76893b653b4b/version-history/snapshots
- click "Create Snapshot"
- click the Generate with AI button

The only difference in those flows (as of writing) was an inject node was moved.

### Environment

- FlowFuse version: FFC Current
- Node.js version: NA
- npm version: NA
- Platform/OS: NA
- Browser: NA

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

cstns:
Given the nature of the problem and our discussions so far, a practical solution is to let the user choose which snapshot to compare the current state with.

The core issue is that we can't reliably determine the correct snapshot to run the AI model against. There are too many possible combinations: deployment/pipeline snapshots, auto snapshots, rollbacks, manual snapshots in different orders, plus the requirement that the same functionality must work on remote instances. This adds a layer of complexity we cannot handle automatically, which is why allowing the users to select what to run their current state against is the way to move forward.

I got the UI in a decent state that would support something along these lines, the following image is a work in progress on the matter:
<img width="3722" height="1544" alt="Image" src="https://github.com/user-attachments/assets/6089d571-0279-4b0e-a40c-3f224f4cf205" />

There are a few drawbacks in terms of the work required to implement this:
- we don’t store retrieved snapshots in Vuex or in a centralized way, and the `createSnapshot` dialog component has no access to them and allow users to search through them
- even if we did, the API endpoint we're using only returns the first 10 results, which isn’t enough for users to search for a specific snapshot and the returned content is far to broad and wouldn't make sense for the task at hand
- we need another endpoint that allows minimal searching through existing snapshots so users can look up a specific one
[...]

PR #5985, opened 2025-09-04
Select which snapshot to generate ai description against

cstns:
## Description

Allow for more accurate, context-aware ai-generated snapshot descriptions by letting users select the comparison baseline, including pipeline deploys or any specific snapshot, with a quick-search UI.

- added a new ff popover component which is a wrapper over [headless ui's popover](https://headlessui.com/v1/vue/popover) component
- extended the instances.generateSnapshotDescription api method to accept a target parameter for query searching
- extended the combobox to allow for remote option searching
- replaced single “Generate with AI” button with a popover offering:
  - Use latest manual snapshot
  - Use latest deploy snapshot
  - Search and pick a specific snapshot (combobox with remote search)

- added a new `target` parameter for the snapshot generation endpoint to ingest the target snapshot we would need to compare against
- added a new query `getLatestDeploySnapshot` method to query for the latest deployed snapshot through a pipeline

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5983

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/models/Project.js
+++ b/forge/db/models/Project.js
@@ -391,6 +391,18 @@ module.exports = {
                         limit: 1
                     })
                     return snapshots[0]
+                },
+                async getLatestDeploySnapshot () {
+                    const snapshots = await this.getProjectSnapshots({
+                        where: {
+                            name: {
+                                [Op.like]: 'Deploy Snapshot - %'
+                            }
+                        },
+                        order: [['createdAt', 'DESC']],
+                        limit: 1
+                    })
+                    return snapshots[0]
                 }
             },
             static: {
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -1474,15 +1474,15 @@ module.exports = async function (app) {
             targetSnapshot = (await request.project.getLatestDeploySnapshot()) ?? {}
             break
         default:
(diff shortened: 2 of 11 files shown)


cstns:
I may have been overly protective with the API endpoint and added 
```js
if (!targetSnapshot || Object.keys(targetSnapshot).length === 0) {
    return reply.code(404).send({
        code: 'not_found',
        error: 'Snapshot not found'
    })
}
 ```
 to avoid situations where the AI would make assumptions comparing the current state against an empty set of values

cstns:
Altered the endpoint to only return a 404 when a custom snapshot is not found. When using `Latest manual...` or `Latest Deployed...` the comparison will be made against an empty object, allowing generating descriptions when creating the first snapshot.

Steve-Mcl (approved):
LGTM. My only nitpick would be the drop-down caret is bold (as if the combo has content already) and it is clickable even though the search list has not yet been populated.

Since @joepavitt is tagged for review, I'll not merge.

joepavitt:
Awaiting for final test run before merging

Issue #5923, opened 2025-08-15
UI can completely lock up when it encounters a table with specific data


Steve-Mcl:
### Current Behavior

I have not pinned it down yet but I suspect it is due to accented chars, symbols or spaces in the table/field names or something else. 

Watch the below GIF. Once I click the 3rd table **"table 3 ok!"** the whole UI locks up and only a hard page refresh gets you past this.  

NOTE: I tagged this as frontend issue because the data is successfully retrieved and brought to the frontend (this is shown in the GIF after I click "table 3")

![Image](https://github.com/user-attachments/assets/3077b571-90f3-46bc-ae18-5a8a86c64905)

Also,  seeing **many hundreds** of vue warnings about missing props (related?)

### Expected Behavior

- UI should not lock up
- Missing prop warnings should not occur

### Steps To Reproduce

Not sure what the cause is yet though I suspect it is due to symbols/spaces/accented chars in table & field names

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Steve-Mcl:
DDL for that table:

```
CREATE TABLE public."table 3 ok!" (
	id serial4 NOT NULL,
	created_at timestamp DEFAULT now() NOT NULL,
	"field with spaces" text NULL,
	"fíèld wíth áççèntèd chárs" text NULL,
	CONSTRAINT "table 3 ok!_pkey" PRIMARY KEY (id)
);
```

some data:
```
INSERT INTO public."table 3 ok!" (created_at,"field with spaces","fíèld wíth áççèntèd chárs") VALUES
	 ('2025-07-24 14:44:42.172043','For our culinary adventure, you''ll need a robust selection of ingredients to ensure a symphony of flavors:

Two tablespoons of high-quality extra virgin olive oil: This forms the foundational fat for our dish.

One large yellow onion: Finely diced, to build aromatic layers.

Three cloves of garlic: Minced, for a pungent kick.

One pound of lean ground beef or turkey: Choose based on your preference; this will be browned and crumbled.

One 28-ounce can of crushed tomatoes: An essential liquid component.

One 15-ounce can of tomato sauce: Also crucial for the base, ideally unsalted to control sodium.

One tablespoon of granulated sugar: To introduce a touch of sweetness and depth.

Two teaspoons of dried oregano: A classic herb for Italian flavors.

[...]

PR #5986, opened 2025-09-04
Fix UI lock up when binding tables values with non utf encodings

cstns:
## Description

Fix UI lock-up caused by non-UTF8 encoded values bound to Vue components. Switched to hashing key values as a safer alternative to stripping non-UTF8 characters (hashing ensures deterministic output for identical inputs, regardless of encoding validity).
- separated datatable payload into original state and safe state, where keys are hashed to avoid binding issues
- refactored datatable component to bind only specific values to rows when a component bind property is provided, instead of binding the entire object

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5923

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/composables/String.js
+++ b/frontend/src/composables/String.js
@@ -51,3 +51,44 @@ export const generateUuid = (length = 6) => {
         byte.toString(36).padStart(2, '0')
     ).join('').substring(0, length)
 }
+
+/**
+ * Generates a hash string from a given input string.
+ * The hash value is computed using the FNV-1a hashing algorithm.
+ * The length of the returned hash can be configured.
+ *
+ * @param {string} str - The input string to be hashed.
+ * @param {number} [length=5] - The desired length of the resulting hash string. Defaults to 5 characters.
+ * @returns {string} - The computed hash string truncated to the specified length.
+ */
+export const hashString = (str, length = 5) => {
+    let h = 2166136261 >>> 0
+    for (let i = 0; i < str.length; i++) {
+        h ^= str.charCodeAt(i)
+        h = Math.imul(h, 16777619)
+    }
+    return (h >>> 0).toString(16).substring(0, length)
+}
+
+/**
+ * Generates a non-deterministic hashed string of a specified length.
+ *
+ * This function asynchronously computes a SHA-1 hash of the input string
+ * and converts the result into a hexadecimal representation. The output
+ * is truncated to the specified length, defaulting to 5 characters.
+ *
+ * Note: Due to the use of SHA-1, this function should not be used
+ * for cryptographic purposes or security-critical applications.
+ *
+ * @param {string} str - The input string to be hashed.
+ * @param {number} [length=5] - The desired length of the resulting hash string.
+ * @returns {Promise<string>} A promise that resolves to the non-deterministic hashed string.
+ */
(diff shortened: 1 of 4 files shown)


frontend/src/composables/String.js
```diff
@@ -51,3 +51,44 @@ export const generateUuid = (length = 6) => {
...
+ * @returns {string} - The computed hash string truncated to the specified length.
+ */
+export const hashString = (str, length = 5) => {
+    let h = 2166136261 >>> 0
+    for (let i = 0; i < str.length; i++) {
+        h ^= str.charCodeAt(i)
+        h = Math.imul(h, 16777619)
+    }
+    return (h >>> 0).toString(16).substring(0, length)
+}
```
Steve-Mcl: Nice fnv1a hash however I am a bit confused why the length limit to 5 chars and not leave it to the default 8 chars an FNV returns (effective hash space for 8 chars ~32bit, 5 chars = ~20 bits) so in effect increasing the chance of collision for no apparent reason (i.e. vue keys the keys dont need to be short - we could just keep the full 8 hex chars!?!)

frontend/src/composables/String.js
```diff
@@ -51,3 +51,44 @@ export const generateUuid = (length = 6) => {
...
+ * @returns {Promise<string>} A promise that resolves to the non-deterministic hashed string.
+ */
+export const nonDeterministicHashString = async (str, length = 5) => {
+    const data = new TextEncoder().encode(str)
+    const hashBuffer = await crypto.subtle.digest('SHA-1', data)
+    return Array.from(new Uint8Array(hashBuffer))
+        .map(b => b.toString(16).padStart(2, '0'))
+        .join('')
+        .substring(0, length)
+}
```
Steve-Mcl: I cannot see where this is used? Artefact?

frontend/src/store/modules/product/tables/index.js
```diff
@@ -141,13 +142,22 @@ const actions = {
...
     getTableData ({ commit }, { databaseId, tableName, teamId }) {
         return tablesApi.getTableData(teamId, databaseId, tableName)
             .then((data) => {
-                commit('setTableData', { databaseId, tableName, data })
+                const payload = {
+                    data,
+                    safe: data.map(row => Object.fromEntries(
+                        Object.entries(row).map(([key, value]) => [hashString(key), value])
+                    ))
+                }
```
Steve-Mcl, line 159: Not 100% certain why the whole table data is hashed? Cant the key just be hashed when used?

frontend/src/composables/String.js
cstns: the length part was a late night addition and did not think it through, will be reverting immediatly

frontend/src/composables/String.js
cstns: yes

frontend/src/store/modules/product/tables/index.js
cstns: only the key is getting hashed

Steve-Mcl: @cstns I wasnt clear. Why is every property name of every row being hashed? 
Perhaps a quick chat would clear this up?

Issue #5976, opened 2025-09-01
Replace permissions mixin with permissions composable across all Vue components


cstns:
### Description

Remove the permissions mixin from every component and use the permissions composable directly.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #5993, opened 2025-09-08
Replace the permissions mixin with it's composable counterpart

cstns:
## Description

This continues the migration away from the permissions mixin by aligning permission checks with the new composable API.

- replaced mixin-based permission checks with calls to the `usePermissions` composable
- ensured all permission-related logic is handled via composable functions instead of relying on implicit mixin state
- updated affected components to import and use explicit permission helpers

This keeps permission handling consistent across the codebase and avoids the reactivity issues caused by the mixin.
We can rely on unit and E2E tests to confirm that all permission checks behave correctly after the migration.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5976

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


cstns:
That's odd, I had multiple clean e2e runs locally. Investigating

frontend/src/composables/Permissions.js
```diff
@@ -88,7 +88,7 @@ export default function usePermissions () {
      *
      * @returns {boolean} True if the user has an admin role, otherwise false.
      */
-    const _isVisitingAdmin = () => isVisitingAdmin(teamMembership.value)
+    const _isVisitingAdmin = isVisitingAdmin(teamMembership.value)
```
knolleary: This doesn't look right to me; `_isVisitingAdmin` is exported as `isVisitingAdmin` - and is used as a function throughout the code base. This change means we now export the boolean value, not the function that the rest of the code expects.

cstns: This was a fix for a latent bug I introduced in https://github.com/FlowFuse/flowfuse/issues/5939. It did not manifest itself  because the fn was only used inside the composable so reactivity was preserved. 

looking at it with fresh eyes, i'd also need to wrap it in computed

cstns: Wrapping `isVisitingAdmin` in a computed ensures reactivity, and `toValue` lets consumers use it without unwrapping .value, while still preserving reactivity

knolleary: @cstns so is the result of `toValue()` a function that can be called? Because the code is still treating `isVisitingAdmin` as a function

cstns: good catch, it was a boolean, reverted to returning the computed value and the components that use the composable unwrap it automatically.

Computed returns a `ComputedRef` which integrates with vue's reactivity system when used in setup functions

PR #6017, opened 2025-09-12
Certified Nodes usage telemetry

hardillb:
part of https://github.com/FlowFuse/customer/issues/392

## Description

Gethers certified nodes count every 24hrs

~~Dual implementation, telemetry and~~ a standalone housekeeping task

~~need to pick one before merging~~

## Related Issue(s)

https://github.com/FlowFuse/customer/issues/392

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/housekeeper/tasks/certifiedNodes.js
+++ b/forge/housekeeper/tasks/certifiedNodes.js
@@ -38,32 +38,20 @@ async function collect (app) {
                 ]
             })
 
-            const instancePromise = new Promise((resolve, reject) => {
-                // uses getRuntimeSettings to include template merge
-                let instanceOffset = 0
-                const instanceInterval = setInterval(async () => {
-                    for (let i = 0; i < 20; i++) {
-                        const project = runningProjects[instanceOffset++]
-                        const settings = await app.db.controllers.Project.getRuntimeSettings(project)
-                        if (settings?.palette?.modules) {
-                            for (const mod of Object.keys(settings.palette.modules)) {
-                                if (mod.startsWith(SCOPE_CERTIFIED) || mod.startsWith(SCOPE_NODES)) {
-                                    if (payload.certifiedNodes[mod]) {
-                                        payload.certifiedNodes[mod]++
-                                    } else {
-                                        payload.certifiedNodes[mod] = 1
-                                    }
-                                }
+            for (const project of runningProjects) {
+                const settings = await app.db.controllers.Project.getRuntimeSettings(project)
+                if (settings.palette?.modules) {
+                    for (const mod of Object.keys(settings.palette.modules)) {
+                        if (mod.startsWith(SCOPE_CERTIFIED) || mod.startsWith(SCOPE_NODES)) {
+                            if (payload.certifiedNodes[mod]) {
+                                payload.certifiedNodes[mod]++
+                            } else {
+                                payload.certifiedNodes[mod] = 1
                             }
                         }
-                        if (instanceOffset === runningProjects.length) {
-                            clearInterval(instanceInterval)
-                            resolve()
-                            break
-                        }
                     }
-                }, 1000)
-            })
+                }
+            }
 
             const now = Date.now()
             const runningDevices = await app.db.models.Device.findAll({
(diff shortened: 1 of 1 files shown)


hardillb:
The consumer for this is all up and running now

Only outstanding question is should the scopes and upload URL be configurable?

hardillb:
@knolleary if you get a second

forge/housekeeper/tasks/certifiedNodes.js
```diff
@@ -0,0 +1,119 @@
...
+
+const ProjectSettings = require('../../db/models/ProjectSettings')
+const ProjectTemplate = require('../../db/models/ProjectTemplate')
+
+const { randomInt } = require('../utils')
+
+const SCOPE_CERTIFIED = '@flowfuse-certified-nodes/' // TODO change to certified nodes scope, maybe should be configurable
+const SCOPE_NODES = '@flowfuse-nodes/'
+const PING_TIME = `${randomInt(0, 59)} ${randomInt(0, 23)} * * *`
+const CERTIFIED_NODES_ENDPOINT = 'https://ff-certified-node.flowfuse.cloud/certified-nodes' // configurable?
```
knolleary:
```suggestion
const CERTIFIED_NODES_ENDPOINT = 'https://ff-certified-nodes.flowfuse.cloud/certified-nodes' // configurable?
```

forge/housekeeper/tasks/certifiedNodes.js
```diff
@@ -0,0 +1,119 @@
...
+    const payload = {
+        instanceId: app.settings.get('instanceId'),
+        certifiedNodes: {}
+    }
+    const snapshots = {}
+    // checking certified nodes enabled and active
+    const npmRegURLString = app.settings.get('platform:certifiedNodes:npmRegistryURL')
+    const token = app.settings.get('platform:certifiedNodes:token')
+    const catalogueString = app.settings.get('platform:certifiedNodes:catalogueURL')
+    if (app.config.features.enabled('certifiedNodes') && (npmRegURLString && token && catalogueString)) {
```
knolleary: Need to realign this with the changes made by https://github.com/FlowFuse/flowfuse/pull/6105

```suggestion

    const platformNPMEnabled = !!app.config.features.enabled('certifiedNodes', false) &&
                                   !!app.config.features.enabled('ffNodes', false) &&
                                   !!app.settings.get('platform:ff-npm-registry:token')     
    if (platformNPMEnabled) {
```

forge/housekeeper/tasks/certifiedNodes.js
```diff
@@ -0,0 +1,119 @@
...
+        }
+    } else {
+        // console.log('NOT ACTIVE')
+        // console.log(catalogueString, npmRegURLString, token)
+    }
+}
+
+module.exports = {
+    name: 'certifiedNodes',
+    startup: 15000,
```
knolleary:
```suggestion
```
Given the workload involved, we don't need to be reporting each time the app restarts

forge/housekeeper/tasks/certifiedNodes.js
```diff
@@ -0,0 +1,119 @@
...
+const { Op } = require('sequelize')
+
+const ProjectSettings = require('../../db/models/ProjectSettings')
+const ProjectTemplate = require('../../db/models/ProjectTemplate')
+
+const { randomInt } = require('../utils')
+
+const SCOPE_CERTIFIED = '@flowfuse-certified-nodes/' // TODO change to certified nodes scope, maybe should be configurable
+const SCOPE_NODES = '@flowfuse-nodes/'
+const PING_TIME = `${randomInt(0, 59)} ${randomInt(0, 23)} * * *`
```
knolleary:
```suggestion
const PING_TIME = `${randomInt(0, 59)} ${randomInt(0, 2)} * * *`
```
Suggest limiting to overnight due to the load this will put on the DB

forge/housekeeper/tasks/certifiedNodes.js
```diff
@@ -0,0 +1,119 @@
...
+
+            // all running devices seen in the last 24hrs
+            const now = Date.now()
+            const runningDevices = await app.db.models.Device.findAll({
+                where: {
+                    state: 'running',
+                    lastSeenAt: { [Op.gte]: new Date(now - 1000 * 60 * 60 * 24) }
+                }
+            })
+            for (const dev of runningDevices) {
```
knolleary, line 64: Need to review this in more detail - not sure how this'll handle devices in dev mode.

knolleary (changes requested):

forge/housekeeper/tasks/certifiedNodes.js
hardillb: We can't easily count devices in dev mode until they push a snapshot, which this should pick up. But if they add nodes and stay in devMode we will never know

hardillb:
@knolleary this now runs the queries in batches of 20 every second until complete to reduce load on db, we can tune the time period if 20/s is still too fast

forge/housekeeper/tasks/certifiedNodes.js
```diff
@@ -0,0 +1,145 @@
...
+                        model: ProjectSettings.model,
+                        where: { key: ProjectSettings.KEY_SETTINGS }
+                    }
+                ]
+            })
+
+            const instancePromise = new Promise((resolve, reject) => {
+                // uses getRuntimeSettings to include template merge
+                let instanceOffset = 0
+                const instanceInterval = setInterval(async () => {
```
knolleary: If, for any reason, it takes longer than 1s to process the whole batch of twenty, this will end up scheduling batches in a parallel - putting even more load on the DB than if it was just doing them all in a single batch.

Lets revert the batching change here and go back to a simple iteration over the list. Add some log statements at start/end of the task so we can check back on how long it takes to run on FFC. We can then decide if it needs batching etc.

knolleary (changes requested):
Sorry for bouncing this again. See me other comment - lets revert the batching change (for both instances and devices), add some logging and see how it does on FFC.

Issue #6028, opened 2025-09-17
Granular RBAC: Team UI


cstns:
### Description

Granular RBAC Team management user interface

<img width="2496" height="818" alt="Image" src="https://github.com/user-attachments/assets/73b4365a-128c-4ace-bac4-2c54f034353f" />

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
@cstns screenshots please

cstns:
closed by https://github.com/FlowFuse/flowfuse/pull/6034

PR #6021, opened 2025-09-15
RBAC UI to manage application scope roles from the application

cstns:
## Description

Backend:
- Fixed the `User` model to include a permissions attribute when fetching team/team-member relations
- fixed `teamMembers` API endpoint to check for the correct team permission & renamed `applicationRBAC` to `rbacApplication` to align with team naming and actual permission

Frontend:
- `changeTeamMemberRole` API call extended to support optional permissions parameter (in addition to role)
- added `EditApplicationPermissionsDialog` for editing user’s permissions in application scope
- added a `RoleRow` component integrated into `UserAccess` view to display user role + permissions
- split `ApplicationSettings`  into modular views

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/6027

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production
 - [ ] Link to Changelog Entry PR, or note why one is not needed.

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #6034, opened 2025-09-17
Add team management UI for applications RBAC

cstns:
## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/6028

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


joepavitt:
Merging back with the applications-ui branch as it needed some other fixes just pushed in https://github.com/FlowFuse/flowfuse/pull/6021

joepavitt:
A gap here whereby I'm logged in as a Team Owner, but that user is a Dashboard Only member of a single Application. The "1 x Override" shows, but then doesn't list, because the user doesn't have permissions to see the Application. It's an odd one that i can't see happening, but I've found it, so am reporting it.

joepavitt:
Screenshot 2025-09-24 at 11 33 42

Have moved the up/down/no-access graphic into a centralised component, then applied that in the Application-level view too for consistency.

frontend/src/components/tables/cells/UserRoleCell.vue
```diff
@@ -1,12 +1,20 @@
 <template>
-    <div>{{ formatRole(role) }}</div>
+    <div>{{ formattedRole }}</div>
 </template>
 <script>
+import { capitalize } from '../../../composables/String.js'
```
joepavitt, line 5: why not just use CSS?

frontend/src/pages/team/Members/General.vue
```diff
@@ -80,14 +112,79 @@ export default {
...
+                {
+                    label: '',
+                    key: 'overrides',
+                    sortable: false,
+                    class: ['w-40'],
+                    component: {
+                        // eslint-disable-next-line vue/one-component-per-file
+                        is: markRaw(defineComponent({
+                            // inheritAttrs: false,
+                            props: ['permissions', 'role'],
```
joepavitt: can you fix these lint warnings too please?

frontend/src/components/tables/cells/UserRoleCell.vue
cstns: Because my mind did not take me there at the moment of writing the unit tests and if i were to change them now would mean to break a lot of tests. (actual value lowercase when checking for value vs tests display value).

I'm a bit sorry that i did not think of the css approach sooner as that would have definitely been the way to go

PR #6035, opened 2025-09-18
fix: added missing loading state in AuditLogs

### LINKED ISSUE #5416: Some views missing a 'loading' state
### Current Behavior

Whilst demoing in Hannover on a slow network, I noticed some pages don't have a 'loading' placeholder - meaning the view shows an empty state, then suddenly populates with data. This can be misleading.

The main culprit was the various Audit Log views - but there may be others.

### Expected Behavior

Any view that loads data should have a loading placeholder to feedback to the user that things are happening

### Steps To Reproduce

1. enable chrome dev tools network throttling
2. access any audit log page

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

harshrajeevsingh:
## Description

Closes #5416 

I found that the `AuditLog` component had its own loading state, which wasn't synced with the `AuditLogBrowser` component. 

In `AuditLogBrowser`, I removed the loading variable and initialised `gettingEntries` =  `true` and passed it as a prop to `AuditLog`. Also, removed `gettingEntries` = `false` from `loadEntries()` method and instead added a _watcher_ for `logEntries` to handle gettingEntries state when the entries are updated. 

PS: BE HARSH WITH THE REVIEW & ALSO PINPOINT OTHER VIEWS WHICH I MIGHT HAVE MISSED. I'm new to this Repo, and this is my first time working with the Vue.js framework. 

Before: 

[Screencast from 2025-09-18 15-54-09.webm](https://github.com/user-attachments/assets/326a1355-5f49-4fd2-a054-13f91ea042db)

AFTER: 

[Screencast from 2025-09-18 13-26-17.webm](https://github.com/user-attachments/assets/6227f87d-681b-4710-83d0-8d4ecd080d40)

## Checklist

 - [X] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/admin/AuditLog.vue
+++ b/frontend/src/pages/admin/AuditLog.vue
@@ -7,7 +7,7 @@
                 </template>
             </ff-page-header>
         </template>
-        <AuditLogBrowser ref="AuditLog" :users="users" :logEntries="logEntries" logType="platform" @load-entries="loadEntries" />
+        <AuditLogBrowser ref="AuditLog" :users="users" :logEntries="logEntries" logType="platform" :loading="loading" @load-entries="loadEntries" />
     </ff-page>
 </template>
 
@@ -26,7 +26,8 @@ export default {
     data () {
         return {
             logEntries: [],
-            users: []
+            users: [],
+            loading: true
         }
     },
     computed: {
@@ -46,6 +47,8 @@ export default {
                 if (err.response?.status === 403) {
                     this.$router.push('/')
                 }
+            } finally {
+                this.loading = false
             }
         }
     }
--- a/frontend/src/pages/application/Activity.vue
+++ b/frontend/src/pages/application/Activity.vue
@@ -1,5 +1,5 @@
 <template>
-    <AuditLogBrowser ref="AuditLog" :users="users" :logEntries="logEntries" :associations="associations" :logType="logScope" @load-entries="loadEntries">
+    <AuditLogBrowser ref="AuditLog" :users="users" :logEntries="logEntries" :associations="associations" :logType="logScope" :loading="loading" @load-entries="loadEntries">
         <template #title>
             <SectionTopMenu hero="Audit Log" info="Recorded events that have taken place in within this application." />
         </template>
@@ -61,6 +61,7 @@ export default {
     data () {
         return {
             logEntries: [],
+            loading: true,
             associations: {}, // applications, instances, devices
             users: [],
             auditFilters: {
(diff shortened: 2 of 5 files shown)


cstns:
Thank you for your contribution!

I haven't started the review just yet but noticed an e2e test is failing:
image

If you don't have access to the GH action that's running the tests you can run them locally. Here's the [docs page](https://flowfuse.com/handbook/development/frontend/testing/#front-end-testing) to get you started.

harshrajeevsingh:
@cstns Done! :)

cstns:
Thank you! Will review it by EOD!

frontend/src/components/audit-log/AuditLogBrowser.vue
```diff
@@ -112,10 +112,8 @@ export default {
         users: function (users) {
             this.auditFilters.users = users
         },
-        logEntries: function (entries) {
-            if (entries) {
-                this.gettingEntries = false
-            }
+        logEntries: function () {
```
cstns: Adding a watcher on logEntries could be computationally expensive dependent on how many log entries are currently loaded in the component. Moreover, this watcher will trigger when _any_ non-deep changes occur to the array, from 0 logEntries to n or from n to 100 logEntries or other superficial ones.

frontend/src/components/audit-log/AuditLogBrowser.vue
```diff
@@ -68,8 +68,7 @@ export default {
     emits: ['load-entries'],
     data () {
         return {
-            loading: true,
-            gettingEntries: false,
+            gettingEntries: true,
```
cstns: by setting this property initially to true, you're just replicating the old behavior of setting loading to true when the component was created.

I'm in favor of removing duplicate state keys (`gettingEntries` and `loading`) which basically did the exact same thing but given that this project mainly uses loading in other components, I'd stick with that nomenclature

frontend/src/components/audit-log/AuditLogBrowser.vue
```diff
@@ -153,7 +154,6 @@ export default {
                 params.append('includeChildren', includeChildren)
             }
             this.$emit('load-entries', params)
-            this.gettingEntries = false
```
cstns, line 156: given that the calls are not async, this could have been left in. It would have spared the computation ow watching `logEntries`.

frontend/src/components/audit-log/AuditLogBrowser.vue
harshrajeevsingh: Thanks for such detailed feedback. :) 
I will remove this.

frontend/src/components/audit-log/AuditLogBrowser.vue
harshrajeevsingh: This will re-create the original flicker bug. 
While the $emit call is synchronous, the action it triggers in the parent component - the API call to fetch logs, is asynchronous.

I think the component that manages the asynchronous action must manage the loading state. Wdyt? 
If you have any other suggestions, plz let me know.

cstns: Correct. The loading state should be managed outside the component, with the component only responding to those external changes

harshrajeevsingh:
@cstns Lemme know if this looks good so that I can change other AuditLogs components too.

With these changes, the Loading overlay is shown on the first mount and whenever the EventScope changes. 

The loader on every filter change: 
I am not sure if it's required or not. I explored showing the loader for the eventType and username change, but since it's managed inside the `AuditLogBrowser` component, I found a race condition whenever the filter is changed quickly on a slow network. Another problem with implementing this could be the jarring "UI flicker" of the loading component on fast networks.

cstns:
I owe you a review. I haven’t forgotten, but I won’t be able to get to it today.

cstns:
This looks good. The changes are already a big improvement.

We don’t yet have a unified approach to loading states, so some flicker will remain until we rethink loaders more globally. In my view, for the `AuditLogBrowser`, the loader should only appear once on initial mount, not on every filter change, since that adds complexity and UI flicker as you noted.

Also, make sure to account for all places where `AuditLogBrowser` is used, including the admin section.

cstns (approved):
Looks good! 

Thanks for this contribution. Handling the audit logs loading state was long overdue, and this is as solid as it gets until we address application-wide loaders.

PR #6003, opened 2025-09-10
Team broker schema agent

hardillb:
closes #5951
## Description

Adds support for Team Broker to MQTT Schema capture.

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/5951

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production
 - [ ] Link to Changelog Entry PR, or note why one is not needed.

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #6041, opened 2025-09-19
Expose SAML SSO groups to Dashboard

### LINKED ISSUE #5902: SAML group assertions in _client.user object
### Description

Please add the possibility to see all SAML group assertions in the `_client.user` object in dashboard flows and not only the FlowFuse role.

This would help differentiate dashboard users in one team that have different permissions in those dashboards.

You already have a debug log for all groups:
https://github.com/FlowFuse/flowfuse/blob/main/forge/ee/lib/sso/index.js#L327

These could be added to a `_client.user.groups` array for example.

Enterprise Tier Only (EE)

hardillb:
fixes #5902

## Description

If enabled in the SSO provider settings this adds the list of groups to the FlowFuse User object on each SSO login.

This can then be retrieved by the nr-launcher middleware when using FF Authentication and exposed to the Dashboard

depends on https://github.com/FlowFuse/nr-launcher/pull/389

~~DO NOT MERGE BEFORE https://github.com/FlowFuse/flowfuse/pull/6003 as it contains db migration~~

## Related Issue(s)

#5902

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/ee/routes/sso/auth.js
```diff
@@ -100,6 +100,11 @@ module.exports = fp(async function (app, opts) {
                         return
                     }
                 }
+                if (providerOpts.exposeGroups) {
+                    // get SAML groups
+                    user.SSOGroups = app.sso.getUserGroups(samlUser, user, providerOpts)
+                    await user.save()
+                }
```
knolleary, line 110: Suggest if exposeGroups is not set, but the user object has groups, then clear the user group property. Avoid leaving stale info behind.

PR #6054, opened 2025-09-23
Account: Make it easier/clearer to delete account

### LINKED ISSUE #5935: Link to Account Settings in Team Settings area
### Description

A user who wants to delete their team often ends up in user settings. The text beside Delete Account says that they need to delete their team first. Instead of telling the user where to go, let's remove friction by linking to the Team Settings area.

Image

At the end of the text that says "Before you can delete your account, teams you own must be deleted or have at least 1 other owner", add another sentence: "Click here to go to Team Settings" and link to Team Settings.

### Epic/Story

_No response_

joepavitt:
## Description

- As discussed in the issue, we now list the teams that a user is part of, permit the user to delete the team from here, and then drive the `:disabled` state by whether they actually can delete their account. 

image

## Related Issue(s)

Closes #5935

### DIFF
--- a/frontend/src/pages/account/Settings.vue
+++ b/frontend/src/pages/account/Settings.vue
@@ -286,14 +286,25 @@ export default {
                 text: 'Are you sure you want to delete this team? This cannot be undone. All Instances and resources within this team will be removed.',
                 confirmLabel: 'Delete Team'
             }, async () => {
-                teamApi.deleteTeam(teamId).then(() => {
-                    alerts.emit('Team successfully deleted', 'confirmation')
-                    // refresh teams
-                    this.$store.dispatch('account/refreshTeams')
-                }).catch(err => {
-                    alerts.emit('Problem deleting team', 'warning')
-                    console.warn(err)
-                })
+                teamApi.deleteTeam(teamId)
+                    .then(() => {
+                        alerts.emit('Team successfully deleted', 'confirmation')
+                        // refresh teams
+                        return this.$store.dispatch('account/refreshTeams')
+                    }).then(() => {
+                        const activeTeam = this.$store.getters['account/team']
+                        // check if the active team is one deleted
+                        if (activeTeam?.id === teamId) {
+                            const teams = this.$store.getters['account/teams']
+                            if (teams.length > 0) {
+                                // get another team
+                                this.$store.dispatch('account/setTeam', teams[0].slug)
+                            }
+                        }
+                    }).catch(err => {
+                        alerts.emit('Problem deleting team', 'warning')
+                        console.warn(err)
+                    })
             })
         },
         selectTeam (team) {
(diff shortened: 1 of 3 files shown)


frontend/src/pages/account/Settings.vue
```diff
@@ -225,6 +271,33 @@ export default {
...
             })
+        },
+        deleteTeam (teamId) {
+            dialog.show({
+                header: 'Delete Team',
+                kind: 'danger',
+                text: 'Are you sure you want to delete this team?',
+                confirmLabel: 'Delete Team'
+            }, async () => {
+                teamApi.deleteTeam(teamId).then(() => {
```
cstns: the teams list is not refreshed after deletion. The teams data attr should be a computed prop

frontend/src/pages/account/Settings.vue
```diff
@@ -25,28 +25,43 @@
...
                 </div>
             </div>
-            <div class="min-w-fit flex-shrink-0">
-                <ff-button class="warning" kind="danger" data-action="delete-account" @click="deleteAccount">Delete Account</ff-button>
-            </div>
+        </div>
+        <div v-if="!canDeleteAccount" class=" max-w-2xl mt-4">
+            <h3>Teams</h3>
+            <ul class="space-y-2 border-t border-gray-200">
+                <li v-for="team in teams" :key="team.id" class="flex justify-between items-center border-b border-gray-200 h-11">
```
cstns: we should only iterate over teams that need additional actions, or mark the ones that need additional actions distinctly from the others

cstns: for example, if the other team has other owners add a checkmark with a title 'This team will not be deleted' or 'This team will pass on to the other owner'

frontend/src/pages/account/Settings.vue
```diff
@@ -63,8 +78,12 @@ export default {
                 defaultTeamName = team.name
             }
             return {
-                value: team.id,
-                label: team.name
+                id: team.id,
```
cstns: the team options should be a computed prop in order for it to react to changes of account/teams

frontend/src/pages/account/Settings.vue
joepavitt: I'm okay with how we have this for now

joepavitt:
@cstns noticing this got missed before the 2.22 cut - can you take a look when you have wiggle room again please

joepavitt:
> Currently, pressing delete removes the team immediately, even if it has another owner and an active instance. To avoid this confusion, teams with inheritors should simply not be listed.

Sensible

> A simple fix would be to automatically pre-select another team the user has access to when their active team is deleted.

Also sensible

cstns:
offboarding tests failed, re-running suite

PR #6062, opened 2025-09-25
Show a disabled "Add Instance" & "Create Instance" button for Members

### LINKED ISSUE #5425: Change UI for Member user accessing Device Agent locally
### Description

When a Member user is accessing the Device Agent locally, it is not possible to add an instance. This is by design, but confusing to the user. Surface the Add Instance button, but disable it, and create a tooltip that says, "Your role does not allow creating new instances. Contact a team admin to change your role."

### Epic/Story

_No response_

harshrajeevsingh:
## Description

Closes #5425 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/docs/device-agent/install/device-agent-installer.md
+++ b/docs/device-agent/install/device-agent-installer.md
@@ -135,7 +135,9 @@ To update the Device Agent package, use the `--update-agent` flag, optionally sp
 
 Specifying `--update-agent` without the `--agent-version` flag will update to the latest available version.
 
-## Managing
+## Troubleshooting
+
+### Managing the Device Agent service
 
 Services are named per-port, for example `flowfuse-device-agent-1880`. On macOS, the launchd label is `com.flowfuse.device-agent-1880`.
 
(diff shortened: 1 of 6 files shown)


frontend/src/ui-components/components/Button.vue
```diff
@@ -4,7 +4,7 @@
                  class="ff-btn transition-fade--color"
                  :target="target"
                  :class="computedClass"
-                 :to="to ?? '#'"
+                 :to="(!to || disabled) ? '' : to"
```
cstns, line 7: Why is this needed?

cstns (commented):
All in all looks good, I just have one question related to the ff-button change.

frontend/src/ui-components/components/Button.vue
harshrajeevsingh: The `router-link` is a wrapper around a standard `a` tag, and the HTML disabled attribute has no effect on `a` elements. So, the browser still allowed the click event to trigger the navigation specified by :to.
Therefore, I passed an empty string here( in disabled state). Although I am not sure if it should be `#` here instead of an empty string. I tried `#`, but I saw a slight flicker in UI on a click. Plz verify this UI behaviour on your side too.

cstns: That is a very nice catch! we're predominantly using the component as a button and never noticed it.

Issue #5745, opened 2025-07-07
Server-side changes for Granular RBAC


joepavitt:
### Description

Parent issue to capture the work required on the server-side as part of the #3093 Granular RBAC work.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Marking as "In Design" due to eng. design work required.

joepavitt:
@knolleary can you size and add breakdown tasks as appropriate for this please? You have a much better understanding of the scope of changes that will be required here.

gstout52:
Moved to 2.21 as I believe this is currently in progress.

knolleary:
My working notes on necessary changes. Unlikely to spin these out as discrete sub-issues/tasks as they all feed into each other.

## TeamType Flag

- [x] As this feature will be scoped to particular team types, a new feature flag will be added called `rbacApplication` (name may change prior to PR).

Will need to consider how that impacts the permission checking part.

## DB Changes

 - [x] The `TeamMembers` table is used to identify a users role in a given team. This table will be extended to add a `permissions` column as a JSON blob to contain the granular permission information for the user/team.

The JSON blob will have the structure:

```
{
   "applications": {
      <applicationId>: <role>
   }
}
```

By having 'application' as a top level key this leaves room for other granularities in the future if the need arises. Will need to decide if its the raw DB id or the hashid we store. There are pros/cons to both approaches. The key thing to remember will be to ensure it is returned as hashids over the external API as we don't expose raw db ids on the api.

## Permission Checking

### `needsPermission`

- [x] We guard routes using the `app.needsPermission(...)` pre-handler. This expects the incoming `request`  to already be decorated with `teamMembership`. It then compares `teamMembership.role` against the required role for the particular permission.

Where granular RBAC is required, we will need to decorate the request with either `applicationId` (the hashid) or `application` (the full model object) so we have the proper context of what resource the request is accessing.

Will need to consider how the TeamType feature flag comes into play here; whether we have knowledge of the teamtype already without doing additional DB lookups.

### `hasPermission`

- [x] This function is exposed as `app.hasPermission` and takes just the `TeamMembership` object and the name of the required permission.

This will need updating to be able to provide additional context so the permissions object can be properly inspected.

I think this be adding a third optional `context` object that will contain `applicationId` or `application` as we'd expect on the `request` object.


## Updating pre-handlers

[...]

PR #6076, opened 2025-09-29
Update bulk device api for granular rbac

knolleary:
Part of #5745 

Updates the `/api/v1/team/:teamId/devices/bulk` end points to apply granular rbac checks.

For delete
  - verifies none of the devices are currently assigned (directly, or via their instance) to a blocked application.

For update
  - verifies none of the devices are currently assigned (directly, or via their instance) to a blocked application.
 - verifies the target application/instance are not blocked

### DIFF
--- a/forge/db/controllers/Device.js
+++ b/forge/db/controllers/Device.js
@@ -243,8 +243,9 @@ module.exports = {
      * @param {*} team - User's team
      * @param {Array<string>} deviceIds - Array of device hashids
      * @param {*} user - User performing the deletion (required for audit logging)
+     * @param {*} teamMembership - Team membership of the user performing the deletion (required for rbac checks)
      */
-    bulkDelete: async function (app, team, deviceIds, user) {
+    bulkDelete: async function (app, team, deviceIds, user, teamMembership) {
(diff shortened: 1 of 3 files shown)


forge/db/controllers/Device.js
```diff
@@ -254,11 +255,34 @@ module.exports = {
...
+            include: {
+                model: app.db.models.Project,
+                attributes: ['id', 'ApplicationId']
+            }
+        })
         if (devices.length === 0) {
             throw new ControllerError('not_found', 'No devices found', 404)
         }
 
+        if (!user.admin && teamMembership && teamMembership.permissions?.applications) {
```
cstns, line 269: One concern: both methods check if RBAC is enabled by checking if teamMembership & teamMembership.permissions are passed, but disabling RBAC on a team doesn’t cause the check to fail since `TeamMembership.permissions` aren’t cleared.

Instead of adding another argument, we could verify the feature is enabled at both the team and platform level, and only pass `teamMembership` if it is. This avoids extra arguments and ensures checks reflect the current state.

cstns (changes requested):

forge/db/controllers/Device.js
knolleary: Added feature flag checks

Issue #5554, opened 2025-05-28
Scheduled instance restart


Esmewieringa:
### Description

we would like to implement a regular restart of our logic instance. 
We can do this with an API call of course, but it would be nice if this could be scheduled within Flowfuse.

This has been requested by:
https://app-eu1.hubspot.com/contacts/26586079/record/0-1/3995301

Greg edit July 9 2025:

As a FlowFuse user, I want to schedule updates to Node-RED as new versions become available, so that I always have the most up-to-date version and don't have to be bothered with maintenance restart requests from FlowFuse.

## Rationale

Users could benefit from having the option to approve of our automatically restarting their instances to apply updates. This helps users because it is one less thing they have to do, and because many users would likely choose to restart their instances if they were aware of the benefits that could follow them doing so. Relatedly, it allows us to have more control over the feature set that is provided to customers, ensuring that they have what we consider the best experience on offer.

### Which customers would this be available to

None

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

### Related:
https://github.com/FlowFuse/flowfuse/issues/4608

ZJvandeWeg:
Related issue: https://github.com/FlowFuse/CloudProject/issues/617

ppawlowski:
There are two issues related to this one, which preferably should be addressed in parallel:
* https://github.com/FlowFuse/nr-launcher/issues/386
* https://github.com/FlowFuse/driver-k8s/issues/233

hardillb:
How fine a grained control do we want?

Is hour intervals OK, e.g. uses can specify and hour window on a single day of the week. say between 03:00 and 04:00 on Tuesday mornings?

Also should restarts only happen if there is a new stack or happen regardless?

ZJvandeWeg:
> say between 03:00 and 04:00 on Tuesday mornings?

From customer feedback, that's alright. Disabling might be a nice Enterprise feature 🤔 

> Also should restarts only happen if there is a new stack or happen regardless?

What would be gained here, by restarting? If none, I think we should skip

hardillb:
> From customer feedback, that's alright. Disabling might be a nice Enterprise feature 🤔

This has to be opt in surely or at least off by default and enabled by Admin at a Team level (if we are saying that for e.g. Starter Teir on FFC is going to mandatory)

The opening comment says this is optional

> Users could benefit from having the **option** to approve of our automatically

hardillb:
OK, what exactly are we implementing here:

[...]

Issue #4608, opened 2024-10-03
Feature: automatic stack updates


knolleary:
### Description

On FFC we publish new stacks at least once a month with each release. Stacks enable new features and bring fixes/security updates.

Currently, we leave it to end users to decide when to update their instances. Evidence shows it isn't something most users pay attention to - with a lot of instances running older stacks.

The proposal here is to introduce a new (optional) feature where we will automatically apply stack updates at a schedule of the user's choosing. The user can pick a regular maintenance window for their instance - for eg '3am-4am PT' - when a restart of their instance will be most convenient. During that window, if there is a stack update available, the platform will apply it.

We need to consider:
 - schedule should allow both day-of-week and time-of-day configuration (inc timezone)
 - we should provide a sensible default; eg 1am-2am GMT - configured by Admin so self-hosted customers can set their own default that is sensitive to their situation.
 - Opt-in vs opt-out considerations; how to apply retroactively to existing instances.
 - we know small instances that haven't been restarted for a long time but have had lots of modules incrementally applied sometimes have issues reinstalling all the modules on restart. That is a risk here - especially if the window is weekend/overnight.

A side-effect benefit for us will be getting to a place where a decent proportion of the NR instances on the platform are being restarted with some level of predictability. This will make the maintenance task of migrating k8s nodes easier.

I believe this feature should be available to all users.

### Which customers would this be available to

Everyone - CE/Starter/Team/Enterprise

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

ZJvandeWeg:
Another consideration here is for users that read the docs - These are updated each release and so the user on an old stack expects them to have access to features that they don't. 

If we can update the stack the weekend after each stack update that makes it a better experience for them too.

hardillb:
Customer feedback:

> Yes, we have multiple instances on varying versions of Node-RED. We don't always notice when the update button appears.
> 
> Our instances run different functionalities for our customers, which are mainly used during office hours. Although some are used 24/7.
> 
> If we could manually or automatically choose a date (weekday) and time on a weekly or monthly basis to update to the latest version, that would be great. We could then choose a moment when it would have the least impact.

https://app-eu1.hubspot.com/help-desk/26586079/view/233410279/ticket/153791147254/thread/12202850553#live-chat

[...]

PR #6079, opened 2025-09-29
Scheduled maintenance for instances

hardillb:
part of #5554
part of #4608 

## Description

This wakes up once an hour (random offset in first 30mins) and checks for instances with matching hour and day of week. If they match and there is a new Stack version, it will upgrade and restart.

It also runs 45 seconds after start up to ensure the current hour isn't missed.

The UTC hour (0-23) and day (0-6 where 0 is Sunday) is used and the format in the ProjectSettings table is

```
`{"hour":1,"day":1}
```

It also includes the UI for instances to allow this to be changed and a Admin TeamType UI to for this on by default for a team and set a range of days/times it will pick at random from for new instances.

Screenshot From 2025-11-17 17-08-41

Screenshot From 2025-11-17 17-07-17

Still need to write docs for this and an announcement

## Related Issue(s)

#5554

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/views/TeamType.js
+++ b/forge/db/views/TeamType.js
@@ -31,7 +31,8 @@ module.exports = function (app) {
                     devices: { type: 'object', additionalProperties: true },
                     features: { type: 'object', additionalProperties: true },
                     instances: { type: 'object', additionalProperties: true },
-                    billing: { type: 'object', additionalProperties: true }
+                    billing: { type: 'object', additionalProperties: true },
+                    autoStackUpdate: { type: 'object', additionalProperties: true }
                 },
                 additionalProperties: true
             }
@@ -39,7 +40,8 @@ module.exports = function (app) {
     })
 
     function removeAdminOnlyProps (obj) {
-        const result = {}
+        // Handle both array and object properties - ensure the result is the right shape of thing
+        const result = Array.isArray(obj) ? [] : {}
         for (const [key, value] of Object.entries(obj)) {
             if (/^(price|product)Id$/.test(key)) {
                 continue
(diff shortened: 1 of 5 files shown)


hardillb:
Still needs a UI/UX defining

E.g. should there be a default that for Starter Team and editable for Pro/Enterprise?

(any default should not be a single day/time)

Also need to decide if there is a platform wide flag to turn this on/off as an admin

hardillb:
Pushing this back to draft as the multi-day component will will mean I need to rethink how the times are stored in the ProjectSettings table to be able to query for multiple matches

hardillb:
@cstns I've updated the API to support selecting multiple days, please have a look at updating the UI.

I have also added a test to the instance/Settings/index.js to cover if the feature is not enabled, but while talking about this in the feature-demo channel I think we will also need a Team level feature flag to enable being able to edit the setting (e.g. Starter team will be able to change the value but not disable it)

Will look at that in the morning

hardillb:
@cstns This looks good, just need to work out how we are going to do defaults (for Starter) and team level flags as who is allowed to turn it off vs just change the time (again for Starter)

cstns:
I’m guessing we want to avoid doing a mass insert for each currently running instance, while still letting the housekeeper work with real data, so that all instances don't upgrade themselves in the same slot over the weekend

knolleary:
I have pushed some UX tweaks.

- This feature should be available to all tiers. The 'enable' flag in the teamtype should only indicate whether a default schedule should be applied or not.
- If an instance does not have a schedule configured, but the teamtype does have a default one - apply that default as the initial config of the form rather than force the user to set all the values.
- Fixes the TeamType view in the api to render as an array not an object

knolleary:
Will wait for tests to pass following my set of changes - but then this can be approved and merged.

PR #6089, opened 2025-09-30
Home: "Add Instance" button for Hosted & Remote Instances

### LINKED ISSUE #6085: Add 'Create Instance' buttons on the home page
### Epic

_No response_

### Description

**As a FlowFuse user**

**I want to add instances from the homepage**

**So that I can avoid the friction step of having to navigate to the Hosted Instances or Remote Instances area and do it from there**

The new home page design, while more elegant than before, requires an extra step from users before they can create a new instance.

Image

The work here is to add a 'create instance' button on both the Hosted Instances overview area, and on the Remote Instances one.

A simple plus sign in the top-right corner of each panel that triggers the workflow for a new hosted instance or remote instance will suffice.

None

### Acceptance Criteria

- [ ] A small button is added to the Hosted Instances panel on the homepage that starts the 'create hosted instance' workflow
- [ ]  A small button is added to the Hosted Instances panel on the homepage that starts the 'create remote instance' workflow

joepavitt:
## Description

- "Add Instance" button for Hosted Instances
    - Directs to the "Create Instance" page when clicked
    - Add appropriate permissions check to the button too
- "Add Instance" button for Remote Instances
    - Opens "Add Remote Instance" dialog
    - Only shows when permissions are set 

## Related Issue(s)

Closes #6085

### DIFF
--- a/forge/db/migrations/20251006-01-fix-sso-group-type.js
+++ b/forge/db/migrations/20251006-01-fix-sso-group-type.js
@@ -1,0 +1,44 @@
+/**
+ * Change column type
+ */
+
+const { DataTypes } = require('sequelize')
+
+module.exports = {
+    up: async (context, Sequelize) => {
+        const dialect = context.sequelize.options.dialect
+        if (dialect === 'sqlite') {
+            // For SQLITE, we need to avoid triggering cascading deletes due to the way it does column changes
+            const sqlFind = "select sql from SQLITE_MASTER where name = 'Users' and type = 'table';"
+            const [results] = await context.sequelize.query(sqlFind)
+            if (results.length === 0) {
+                return // Nothing to do
+            }
+
+            // Check if the DDL is as expected already
+            const ddl = results[0].sql
+            const re = /(`SSOGroups`[^,]+?VARCHAR\(255\))/.exec(ddl)
+            if (!re || re.length < 2) {
+                return // Nothing to do
+            }
+
+            // create new column definition
+            const currentColDef = re[1]
+            const newColDef = currentColDef.replace('VARCHAR(255)', 'TEXT')
+
+            // update the table with the new column definition
+            await context.sequelize.query('pragma writable_schema=1;')
+            const sqlUpdate = `update SQLITE_MASTER set sql = replace(sql, '${currentColDef}', '${newColDef}') where name = 'Users' and type = 'table';`
+            context.sequelize.query(sqlUpdate)
+            await context.sequelize.query('pragma writable_schema=0;')
+        } else {
+            // For Postgres, we can use changeColumn directly.
+            await context.changeColumn('Users', 'SSOGroups', {
+                type: DataTypes.TEXT,
+                allowNull: true
+            })
+        }
+    },
+    down: async (useContext, Sequelize) => {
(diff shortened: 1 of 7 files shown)


joepavitt:
@hardillb FYI - assigning you as a reviewer, as I know Serban and nick are locked in RBAC, and Steve on ONNX atm

hardillb:
@joepavitt re-running backend tests as this is all frontend code so not sure why it's failing

hardillb:
@joepavitt adding a new device never shows the install instructions, just takes you straight to the device overview page, so no way to get the credentials without clicking on the "finish setup" button or going back to the device list and regenerating them (the finish setup also shows warning about regenerating credentials).


https://github.com/user-attachments/assets/1587290a-2650-4574-8de3-3903adace4fc

joepavitt:
Good point, I overlooked that step. Will address it.

gstout52:
@hardillb Are you able to pick this up?
cc @knolleary

hardillb:
@gstout52 no, not really

gstout52:
We'll need to find the right person for it. @knolleary Perhaps @cstns when he is back from OOO?

knolleary:
@cstns please pick this one up and help get over the line

cstns:
From what I can tell, this is a two-part problem:
1. We need to redirect to the newly created remote instance.
1. We need to open the remote instance credentials dialog.

The easiest short-term fix would be to only do step two: open the credentials dialog for that specific remote instance without performing a redirect.

The long-term and sustainable fix would be to decouple dialogs from components, similar to how the left/right drawers work. This would allow more complex operations, such as triggering a page redirect followed by a modal. Specifically, we should be able to open a modal from anywhere by passing a data payload and the modal component to load asynchronously. This would remove the need to include each modal in every page where it might be used (and rendered).

I’d estimate the long-term fix at 1 SP, plus 0.25 SP to implement the redirect and credentials modal for the current task, and 0.25 SP to apply the easy fix to meet the immediate requirements.

cstns:
I replaced the redirect callback after creating a device from the team dashboard with showing the credentials dialog. Let me know If you'd like me to follow up on my previous comment

hardillb:
@cstns is this ready for review again?

cstns:
yes, I replaced the redirect to the device page with opening the credentials dialog.

gstout52:
> yes, I replaced the redirect to the device page with opening the credentials dialog.

Thanks @cstns for laying out the options and pursuing the quicker win here.

hardillb:
@cstns this still isn't doing what I reported to joe when creating a new device.

It still doesn't open the device credentials dialog when creating a new device, it just opens the device's overview page

hardillb:
@cstns are you sure you pushed changes, the commit history just shows merges of main

cstns:
it was added in 6fc996e9aac389ae883db0a8c0d607a4ca735e60 ~5h ago, checking pre-staging once built to confirm

cstns:
confirming it works on pre-staging

hardillb:
The commit was not present in the tree when I pulled it to test or when I reviewed it, it just shows the 2 merge main commits 

image

PR #6105, opened 2025-10-07
Add FF exclusive nodes feature flag

knolleary:
## Description

Splits the cert-nodes feature flag into two - cert nodes and ff-exclusive nodes.

Updates the UI to simplify the configuration. Admins should only need to provide the registry token

image

I have renamed the settings we are using as certNodes is now a subset of the full thing. Only impact will be our staging env that'll need the token readding. The settings can be provided via the yaml file under:

-  `platform:ff-npm-registry:url`
- `platform:ff-npm-registry:token`
- `platform:ff-npm-registry:catalogue:certifiedNodes`
- `platform:ff-npm-registry:catalogue:ffNodes`

All except for `token` have the right defaults applied (albeit inline in the code rather than the defaults file... ).

### DIFF
--- a/forge/routes/api/project.js
+++ b/forge/routes/api/project.js
@@ -910,36 +910,46 @@ module.exports = async function (app) {
         }
 
         // Platform wide catalogue and npm registry
-        const platformNPMEnabled = !!app.config.features.enabled('certifiedNodes') && !!teamType.getFeatureProperty('certifiedNodes', false)
-        if (platformNPMEnabled) {
-            const npmRegURLString = app.settings.get('platform:certifiedNodes:npmRegistryURL')
-            const token = app.settings.get('platform:certifiedNodes:token')
-            const catalogueString = app.settings.get('platform:certifiedNodes:catalogueURL')
-            if (npmRegURLString && token && catalogueString) {
-                const npmRegURL = new URL(npmRegURLString)
-                const catalogue = new URL(catalogueString)
-                if (!settings.settings?.palette) {
-                    settings.settings.palette = {}
+        const platformNPMEnabled = !!app.config.features.enabled('certifiedNodes', false) &&
+                                   !!app.config.features.enabled('ffNodes', false) &&
+                                   !!app.settings.get('platform:ff-npm-registry:token')
+
+        const certifiedNodesEnabledForTeam = teamType.getFeatureProperty('certifiedNodes', false)
+        const ffNodesEnabledForTeam = teamType.getFeatureProperty('ffNodes', false)
+        if (platformNPMEnabled && (certifiedNodesEnabledForTeam || ffNodesEnabledForTeam)) {
+            try {
+                const npmRegURL = new URL(app.settings.get('platform:ff-npm-registry:url') || 'https://registry.flowfuse.com/')
+                const token = app.settings.get('platform:ff-npm-registry:token')
+
+                const certNodesCatalogue = app.settings.get('platform:ff-npm-registry:catalogue:certifiedNodes') || 'https://ff-certified-nodes.flowfuse.cloud/catalogue.json'
+                const ffNodesCatalogue = app.settings.get('platform:ff-npm-registry:catalogue:ffNodes') || 'https://ff-certified-nodes.flowfuse.cloud/ff-catalogue.json'
+
+                // Handle FF Exclusive Nodes
+
+                if (certNodesCatalogue || ffNodesCatalogue) {
+                    // At least one is configured - so initialise the settings
+                    settings.settings.palette = settings.settings.palette || {}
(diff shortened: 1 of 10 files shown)


knolleary:
Fixes pushed for the tests.

Turns out devices expect `palette.catalogues` and projects expect `palette.catalogue`.

Updated tests to check the settings are properly updated depending on the feature flags - more so on the device side.

hardillb:
@knolleary just checking

>     - platform:ff-npm-registry:url
>     - platform:ff-npm-registry:token
>     - platform:ff-npm-registry:catalogue:certifiedNodes
>     - platform:ff-npm-registry:catalogue:ffNodes

maps to this in the `flowforge.yml` file

```
platform:
  ff-npm-registry:
    url: https://registry.flowfuse.com
    token: fdfafdsafsad
    catalogue:
      certifiedNodes: https://ff-certified.nodes.flowfuse.cloud/catalogue.json
      ffNodes: https://ff-certified-nodes.flowfuse.cloud/ff/catalogue.json
```

hardillb:
Actually I don't see anything that would load these default settings from the `flowforge.yml` file or am I missing something?

frontend/src/pages/admin/TeamTypes/dialogs/TeamTypeEditDialog.vue
```diff
@@ -137,6 +137,7 @@
                     <FormRow v-model="input.properties.features.instanceResources" type="checkbox">Instance Resources</FormRow>
                     <FormRow v-model="input.properties.features.tables" type="checkbox">Tables</FormRow>
                     <FormRow v-model="input.properties.features.certifiedNodes" type="checkbox">Certified Nodes</FormRow>
+                    <FormRow v-model="input.properties.features.ffNodes" type="checkbox">FlowFuse Exclusive Nodes</FormRow>
```
hardillb, line 140: Needs the spacer uncommenting at the end of the list of items.

I can't suggest a change because its not in the changed section.

forge/settings/defaults.js
```diff
@@ -71,9 +71,9 @@ module.exports = {
...
-    // Certified Nodes
-    'platform:certifiedNodes:npmRegistryURL': null, // NPM registry URL for certified nodes
-    'platform:certifiedNodes:token': null, // Token for certified nodes
-    'platform:certifiedNodes:catalogueURL': null // Catalogue URL for certified nodes
-
+    // FlowFuse npm registry
+    'platform:ff-npm-registry:url': null,
+    'platform:ff-npm-registry:token': null,
+    'platform:ff-npm-registry:catalogue:certifiedNodes': null,
+    'platform:ff-npm-registry:catalogue:ffNodes': null
```
hardillb: I can't see how to set these from the `flowforge.yml` file so I can test them locally

knolleary: I thought settings would get automaticall populated via the yml file - been a while since I've touched that bit. Will fix up.

frontend/src/pages/admin/TeamTypes/dialogs/TeamTypeEditDialog.vue
knolleary: Fixed

knolleary:
@hardillb I completely overlooked config/settings being separate entities.

1. have removed all but the `token` property from settings - as that's the only one to dynamically change
2. the registry url and catalog urls can now be overridden via the config file as:
```
ff-npm-registry:
   url: http://example.com
   catalogue:
      certifiedNodes: http:/example.com
      ffNodes: http:/example.com
```

Note that `token` cannot be set via config file - must be done via settings in the admin panel. This avoids having two sources of truth for the value.

forge/routes/api/project.js
```diff
@@ -910,36 +910,45 @@ module.exports = async function (app) {
...
-                        `@flowfuse-nodes:registry=${npmRegURL.toString()}\n` +
-                        `//${npmRegURL.host}:_auth="${token}"\n`
+                if (certifiedNodesEnabledForTeam && certNodesCatalogue) {
+                    updateSettingsForCatalogue('@flowfuse-certified-nodes', certNodesCatalogue)
                 }
+                if (ffNodesEnabledForTeam && ffNodesCatalogue) {
+                    updateSettingsForCatalogue('@flowfuse-nodes', ffNodesCatalogue)
+                }
+            } catch (err) {
+                app.log.error('Failed to configure platform npm registry for device', err)
```
hardillb:
```suggestion
                app.log.error('Failed to configure platform npm registry for project', err)
```

hardillb (approved):
Just a copy/paste in the error message `device` vs `project`

forge/routes/api/project.js
knolleary: copy pasta

Issue #3642, opened 2024-03-26
Scalability: move ephemeral state to redis


knolleary:
Part of #2782 

Whilst the work under #2782 has removed most in-memory state handling, there are some pieces that are not suitable for managing via the postgres database.

 - API Rate-limiting - state is currently held in memory. The module we use supports connecting to redis out of the box (or a redis-api compatible alternative)
 - Instance/Device inflight state - when an instance/device is transitioning between states, we store some state in memory and some in the database. Need to review what, if anything, would be more suited to holding separately.

### Tasks - Phase One
- [x] Identify what ephemeral state we still have
- [x] Create 'cache' component in forge app to centralise via API
- [x] Update forge code to use cache component in place of local in-memory state

### Tasks - Phase two
- [x] Identify what technology to use (redis, or an alternative given recent licensing changes)
- [x] Add to our helm charts
- [ ] Setup authentication mechanism
- [ ] Update rate-limiting to use redis if configured
- [ ] Review instance/device lifecycle to see if any other state needs moving in

hardillb:
Or some suitable fork/alternative (though current new license only impacts SaaS offerings of Redis iirc) e.g.

- dragonflydb
- keydb
- memcache

hardillb:
First quick pass on existing code

### Forge App
- `forge/settings/index.js` Platform settings are loaded at startup and only updated if the instance changes them. They are never re-read from the db, the object could be shared or reads could pull from the database.
- `forge/db/controllers/Projects.js` includes inflightProjectState (object), latestProjectState (object), inflightDeploys (Set)
- ~~`forge/comms/devices.js` keeps a list of inflight commands, and device logs websocket clients, would need to ensure that session affinity is used here and heartbeats are in shared cache~~


### Drivers
- driver-localfs is not intended to be used with multiple instances (keeps used ports and file handles for logs in memory)
- driver-docker not intended to be used with multiple instances, does keep an internal in memory list of project states (this._projects) used to shortcut methods for suspened instances
- driver-k8s has the same internal project state cache (this._projects), mainly used to shortcut methods to return instantly if not found. would need to move to shared cache

I need to dig a little harder to make sure I've not missed any.

@cstns is all the ephemeral state you added recently just in the Projects controller?

cstns:
Yes, implemented similarly to `inflightProjectState`. It should be safe to move to a cache store

hardillb:
I'm looking at the `forge/settings` usecase

[...]

PR #6115, opened 2025-10-08
Shared Cache implementation

hardillb:
part of #3642 
## Description

This adds a new caches component to the `app` object. It can be backed by in memory objects or a remote Redis/Valkey instance to allow sharing between multiple instances of the Forge App.

- `app.caches.initCache()` used to set backend up
- `app.caches.getCache(name)` used to create/retrieve a cache instance with a given name

- `cache.get(key)` gets a value for a key
- `cache.set(key, value)` sets a value for key
- `cache.del(key)` removes a key
- `cache.keys()` array of all keys
- `cache.all()` object with all key value pairs in the cache instance

Needs tests

## Related Issue(s)

#3642

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/caches/index.js
+++ b/forge/caches/index.js
@@ -1,0 +1,37 @@
+/**
+ * The a pluggable object cache
+ * 
+ *
+ * @namespace cache
+ * @memberof forge
+ */
+
+
+/**
+ * @typedef {Object} forge.Status
+ * @property {string} status
+ */
+const fp = require('fastify-plugin')
+
+const CACHE_DRIVERS = {
+    memory: './memory-cache.js',
+    redis: './redis-cache.js'
+}
+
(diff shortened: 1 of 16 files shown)


hardillb:
Test coverage looks strange as it contains files I've not touched.

Also still getting the odd UI RBAC test failures that again should be unrelated.

Tests are all running with the memory cache, not redis/valkey, but we should be able to enable valkey in the test suite

hardillb:
To test run 

```
docker run --rm -p 6379:6379 --name valkey valkey/valkey
```

And add to end of `flowfuse.local.yaml`

```
ache:
  driver: 'redis'
  options:
    url: redis://localhost:6379
```

knolleary (approved):

PR #6142, opened 2025-10-16
Add caching to TeamBroker topic logging

### LINKED ISSUE #6141: Scaling: improve caching of MQTTTopicSchemas to reduce db load
### Description

Part of our monitor of the TeamBroker uses the authentication routes used by EMQX to identify topics being used by a team. This triggers an upsert on the DB to add the topic if not already present in the `MQTTopicSchema` table.

EMQX already does some caching to avoid auth checks on every request, but we don't currently do any caching in the forge app and instead rely on the upsert command to figure out if the topic should be added or not.

The task here is to introduce some caching in the forge app to avoid unnecessary DB work.

`forge/ee/lib/teamBroker/index.js:addUserTopic` should cache the most recent 500(?) topics/teamids seen to avoid duplicate db upserts.

As part of its init, it can grab the 500(?) most recently updated entries to prepopulate the cache.

### Epic/Story

_No response_

hardillb:
fixes #6141

## Description

This adds a 5000 entry LRU cache with a 30min expiry to buffer updating topic entries for the MQTTTopicSchema table from the TeamBroker authorization check.

## Related Issue(s)

#6141

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/ee/lib/billing/trialTask.js
+++ b/forge/ee/lib/billing/trialTask.js
@@ -33,6 +33,7 @@ module.exports.init = function (app) {
                         await sendTrialEmail(subscription.Team, 'TrialTeamSuspended', {
                             teamSettingsURL: `${app.config.base_url}/team/${subscription.Team.slug}/billing`
                         })
+                        await sendTrialSuspendedNotification(subscription.Team, 'TrialTeamEnded')
                     }
 
                     // We have dealt with this team
@@ -110,6 +111,16 @@ module.exports.init = function (app) {
                     }
                 )
             }
+        }
+    }
+
+    async function sendTrialSuspendedNotification (team) {
+        const owners = await team.getOwners()
+        for (const user of owners) {
+            const { id, TeamTypeId, ...teamPayload } = team.toJSON()
+            app.notifications.send(user, 'team-trial-suspended', {
+                team: teamPayload
+            })
         }
     }
 
--- a/frontend/src/components/notifications/Generic.vue
+++ b/frontend/src/components/notifications/Generic.vue
@@ -20,7 +20,7 @@
 </template>
 
 <script>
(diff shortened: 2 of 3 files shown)


forge/ee/lib/teamBroker/index.js
```diff
@@ -39,15 +41,38 @@ module.exports.init = function (app) {
...
+    try {
+        const topics = await app.db.models.MQTTTopicSchema.findAll({
+            where: {
+                BrokerCredentialsId: app.settings.get('team:broker:creds')
+            },
+            order: [['updatedAt', 'DESC']],
+            limit: 5000
         })
+        for (const topic of topics) {
+            topicCache.set(`${topic.teamId}#${topic.topic}`, true)
```
knolleary:
```suggestion
            topicCache.set(`${topic.TeamId}#${topic.topic}`, true)
```

forge/ee/lib/teamBroker/index.js
```diff
@@ -39,15 +41,38 @@ module.exports.init = function (app) {
...
+
+    const topicCache = new LRUCache({
+        max: 5000,
+        ttl: 1000 * 60 * 30 // 30 min grace period
+    })
+    try {
+        const topics = await app.db.models.MQTTTopicSchema.findAll({
+            where: {
+                BrokerCredentialsId: app.settings.get('team:broker:creds')
+            },
```
knolleary, line 67:
```suggestion
            },
            attributes: ['TeamId', 'topic'],
```

Only include the properties we care about - no need to retrieve any schema info

knolleary (changes requested):
`teamId` -> `TeamId` fix

knolleary:
Unrelated UI test failures - merging as-is

Issue #6162, opened 2025-10-22
Scalablity: Housekeeper task review


hardillb:
### Description

Re-assessment of House keeper tasks (https://github.com/FlowFuse/flowfuse/issues/2793)

List all tasks with when/how often they run and what problems would come from having 2 instances try and run them.

### Epic/Story

#2782

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

hardillb:
Tasks:

- `blueprintImport.js` runs 5mins after startup and then at a random min, in a random hour at a random day in the week. Doesn't matter if it runs more regularly, first pass inspection suggest 2 instances at the same time should be OK
- `deviceUnusedReminder.js` runs between 7 & 8 am (utc or local?) every day sends email about unused devices, given it runs in a really short window, best to stop it sending same email twice
- `expireInvites.js` runs random min in random hour every day, only removes invites that have expired, not a problem to run multiple times a day
- `expireTokens.js` runs random min in random hour every day, only removes expired Session, AccessTokens and OAuthSessions (should check if others should be added to this list). Safe to run multiple times a day
- `inviteReminder.js` runs between 6 & 7 am (utc or local?) every day, sends email about outstanding team invites, best to only run once to prevent multiple identical emails
- `licenseCheck.js` runs at 2am every day, checks license is about to expire or has expired. It sends emails so should only run once
- `licenseOverage.js` runs midnight Sunday, only adds to audit log, but should only run once
- `telemetryMetrics.js` calls home with useage metrics 10seconds after startup and random min & random hour every day, shouldn't matter if runs multiple times

not yet merged

- `certifiedNodes.js` runs 2.5mins after startup and random min, random hour every day, reports usage of certified nodes, backend does an upsert for each day, so should be fine to run multiple times

knolleary:
Next step - spin out tasks for updates to housekeeper tasks to work in scaled up environment

hardillb:
Closing as we added leader vote to core task runner, so tasks only run by leader

PR #6239, opened 2025-11-04
First attempt at leadership vote for housekeeper

hardillb:
## Description

Uses MQTT to vote on a which forge app instance will run housekeeper tasks in a horizontally scaled environment.

It generates a random number between 0-99 at startup and if it's number is the lowest it runs the housekeeper tasks.

It publishes the number every 15 seconds to ensure that the cache in each instance is up to date as cron tasks have a maximum fidelity of 1 minute so this should ensure it's always correct when an instance starts a task.

LWS and a death message are published when a instance shuts down.

topic used is `ff/v1/platform/leader`

## Related Issue(s)

#6162

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/comms/index.js
+++ b/forge/comms/index.js
@@ -73,12 +73,7 @@ module.exports = fp(async function (app, _opts) {
         app.addHook('onClose', async (_) => {
             app.log.info('Comms shutdown')
             await deviceCommsHandler.stopLogWatcher()
-            const promise = new Promise((resolve, reject) => {
-                client.publish('ff/v1/platform/leader', JSON.stringify({ id: client.platformId, vote: -1 }), function () {
-                    resolve()
-                })
-            })
-            await promise
+            client.publish('ff/v1/platform/leader', JSON.stringify({ id: client.platformId, vote: -1 }))
             await client.disconnect()
         })
     } else {


hardillb:
I extended the startup time for the Blueprint task to make sure it only runs after comms broker has connected, may want to push it a little further than 60 seconds to be 100% sure.

forge/comms/commsClient.js
```diff
@@ -102,6 +106,10 @@ class CommsClient extends EventEmitter {
                     if (jsonPayload.srcId !== this.platformId) {
                         this.app.settings.refresh(jsonPayload.key)
                     }
+                } else if (ownerType === ' leader') {
```
knolleary:
```suggestion
                } else if (ownerType === 'leader') {
```

forge/housekeeper/index.js
```diff
@@ -20,6 +20,12 @@ const fp = require('fastify-plugin')
 module.exports = fp(async function (app, _opts) {
     const tasks = {}
     const delayedStartupTasks = []
+    const localLeaderVote = (Math.random() * 100)
```
knolleary:
```suggestion
    const localLeaderVote = Math.round(Math.random() * 100000)
```
Ensure its an integer and increase the range; 1/100000 chance of a duplicate feels better than 1/100 for the sake of a couple bytes.

forge/comms/index.js
```diff
@@ -64,6 +73,7 @@ module.exports = fp(async function (app, _opts) {
         app.addHook('onClose', async (_) => {
             app.log.info('Comms shutdown')
             await deviceCommsHandler.stopLogWatcher()
+            client.publish('ff/v1/platform/leader', JSON.stringify({ id: client.platformId, vote: -1 }))
             await client.disconnect()
```
knolleary: Small concern here the publish returns async and then triggers the disconnect; can that lead to the publish being dropped? Can we await the publish? Need to verify the API as I don't know if that's supported.

knolleary (changes requested):

forge/comms/index.js
hardillb: publish isn't async, but does have a callback option, so could move the disconnect to in the callback

hardillb: OK, that didn't work, it hung up all the tests.

Need to think of something else, may be a `setTimeout()` with a 1 second wait?

hardillb: I tried with a promise resolved from the callback, but it's a QOS 0 publish so I don't think there is a callback and it's instant.

Also the disconnect doesn't need to be awaited as there is no async calls inside it.

knolleary (approved):

PR #6247, opened 2025-11-05
Add ability to override features for teamtype

### LINKED ISSUE #6222: Enable TeamType overrides
### Description

It is possible for an FFC admin to override instance number, broker client count, and user allocation that is set at the TeamType level. This issue is to expand the overrides to allow for changes in the feature set allowed by the TeamType. For example, I want to be able to give a Pro customer access to Tables, or SSO. This task is to enable that.

### Epic/Story

_No response_

hardillb:
fixes #6222

## Description

Adds ability to override
- Tables
- Instance Resources
- Bill of Materials
- NPM Packages
- Git Integration

## Related Issue(s)

#6222

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,25 @@
+#### 2.24.5: Release
+
+ - build(deps): bump aws-actions/configure-aws-credentials (#6354)
+ - build(deps): bump @modelcontextprotocol/sdk from 1.23.0 to 1.24.0 (#6359) @app/dependabot
+ - docs(instances): add detailed instance states list (#6365) @cstns
+ - Allow initial Admin to modify default stack (#6326) @hardillb
+ - build(deps): bump actions/checkout from 6.0.0 to 6.0.1 (#6376) @app/dependabot
+ - build(deps): bump actions/setup-node from 6.0.0 to 6.1.0 (#6377) @app/dependabot
+ - Intercept kebab menu item click to close the menu (#6372) @cstns
+ - Update group entry display for non-assignable devices and permissions (#6371) @cstns
+ - Add Seprate RBAC Document (#6370) @sumitshinde-84
+ - Force plain links to open in new tabs (#6334) @cstns
+ - Align device snapshots ux (#6358) @cstns
+ - build(deps): bump jws (#6368) @app/dependabot
+ - Devices full height pages and scrolling behavior (#6324) @cstns
+ - Update FlowFuse Concept Section (#6366) @sumitshinde-84
+ - Show broker client limit notification and disable client creation when limit is reached (#6363) @cstns
+ - Add note that Suspend/Delete team removes billing Subscription (#6362) @hardillb
+ - build(deps): bump validator from 13.15.20 to 13.15.23 (#6360) @app/dependabot
+ - build(deps): bump express (#6355) @app/dependabot
+ - build(deps): bump nodemailer from 7.0.10 to 7.0.11 (#6356) @app/dependabot
+
 #### 2.24.4: Release
 
  - Ensure TeamType passed to instanceCount (#6351) @hardillb
--- a/docs/user/ff-tables.md
+++ b/docs/user/ff-tables.md
@@ -4,7 +4,7 @@ navTitle: FlowFuse Tables
 
 # FlowFuse Tables
 
-From FlowFuse v2.20.0 Teams (Pro & Enterprise teams on FlowFuse Cloud) can create a relational database to use to store data.
+From FlowFuse v2.20.0 Teams (Enterprise teams only) can create a relational database to use to store data.
 
 You can create a database by selecting the Tables entry in the left hand menu
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@flowfuse/flowfuse",
-    "version": "2.24.4",
+    "version": "2.24.5",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@flowfuse/flowfuse",
-            "version": "2.24.4",
+            "version": "2.24.5",
             "license": "SEE LICENSE IN ./LICENSE",
             "dependencies": {
                 "@aws-sdk/client-sesv2": "^3.916.0",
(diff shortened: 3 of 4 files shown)


hardillb:
image

Needs test

hardillb:
Picks up defaults from the current team type

knolleary:
For each feature, we need to check the permission checks in the backend to ensure they operate against the Team object and not doing Team->TeamType->Check - as that will fail to apply the team override.

For example, the npm catalogue does the check via the TeamType here: https://github.com/FlowFuse/flowfuse/blob/8bed051fa0dd2b509868bcc475d990874c05c398/forge/ee/routes/catalogues/index.js#L18

hardillb:
@knolleary Realised, this will store overrides for everything on first edit so they match the current TeamType.

Because it pulls through from the TeamType if it doesn't match a Team Property

```
            this.editableLimits.features.tables = this.getTeamProperty('features_tables')
            this.editableLimits.features.instanceResources = this.getTeamProperty('features_instanceResources')
            this.editableLimits.features.bom = this.getTeamProperty('features_bom')
            this.editableLimits.features.npm = this.getTeamProperty('features_npm')
            this.editableLimits.features.gitIntegration = this.getTeamProperty('features_gitIntegration')
```

So if a team say moves from Pro to Enterprise later it would be pinned at Pro features.

EDIT: nevermind, fixed

knolleary:
The PR is adding `getFeatureOverride` to the Team model and then relies on all checks to look at both the teamType and the team to get the appropriate value.

```
teamType.getFeatureProperty('deviceAutoSnapshot', false) || device.Team.getFeatureOverride('deviceAutoSnapshot')
```

It would be cleaner, and easier to maintain, to add `getFeatureProperty` to the Team model and centralise the logic that checks for the presence of a local override, otherwise defer to `this.teamType.getFeatureProperty(...)`.

Then all the callers just need to ask the team if the feature is set without having to worry about checking for overrides.

hardillb:
> It would be cleaner, and easier to maintain, to add getFeatureProperty to the Team model and centralise the logic that checks for the presence of a local override, otherwise defer to this.teamType.getFeatureProperty(...).

The problem with that is i makes `team.getFeatureProperty()` async as it needs to be confident the team object has TeamType populated. While this is being done in all the current cases because we are extracting the TeamType object first it makes later additions harder because we can't assume TeamType is always populated.

hardillb:
@knolleary ready for review again

forge/ee/lib/alerts/index.js
```diff
@@ -42,7 +42,7 @@ module.exports = {
                     } else if (emailAlerts?.resource?.memory && event === 'resource.memory') {
                         template = 'InstanceResourceMemoryExceeded'
                     }
-                    if (!template || !teamType.getFeatureProperty('emailAlerts', false)) {
+                    if (!template || (!teamType.getFeatureProperty('emailAlerts', false) || project.Team.properties?.feature?.emailAlerts)) {
```
knolleary: Think the logic here needs looking at again.

Just checked the implementation of `app.db.models.Project.byId` - it will include the `Team` model, but `properties` is not an included property, so it won't be set here.

forge/routes/api/project.js
```diff
@@ -838,9 +838,9 @@ module.exports = async function (app) {
         settings.forgeURL = app.config.base_url
         settings.fileStore = app.config.fileStore ? { ...app.config.fileStore } : null
 
-        const teamType = await request.project.Team.getTeamType()
-
-        const assistantInlineCompletionsFeatureEnabled = !!(app.config.features.enabled('assistantInlineCompletions') && teamType.getFeatureProperty('assistantInlineCompletions', false))
+        await request.project.Team.ensureTeamTypeExists()
+        const team = await request.project.Team
```
knolleary:
```suggestion
        const team = request.project.Team
```

forge/routes/auth/oauth.js
```diff
@@ -499,9 +499,9 @@ module.exports = async function (app) {
             let response
             if (request.headers['ff-quota']) {
                 const project = await app.db.models.Project.byId(request.session.ownerId)
-                const teamType = await project.Team.getTeamType()
-                const fileStorageLimit = teamType.getFeatureProperty('fileStorageLimit', 100)
-                const contextLimit = teamType.getFeatureProperty('contextLimit', 1)
+                await project.Team.ensureTeamTypeExists()
```
knolleary, line 512: `app.db.models.Project.byId` doesn't include the `properties` attribute of the `Team` object it loads. Need to update that to do so.

hardillb:
@knolleary tests are passing again, should be good for a second look

forge/routes/api/team.js
```diff
@@ -98,6 +108,7 @@ module.exports = async function (app) {
         result.instanceCountByType = await team.instanceCountByType()
 
         await appendBillingDetails(result, team, request)
+        await updateTeamFeatures(result, team)
```
knolleary, line 114: Been debugging an odd behaviour with this PR and I *think* it's due to this line.

This is changing the `type` object in the response to represent the team-type feature flags with the team overrides applied. Whereas, I think, this should be the underlying team type flags - and the overrides are returned in the top-level properties object for the team.

Still digging into it.

knolleary: Ok -so there are still lots of frontend feature flag checks that are examining the `team.type.properties` object directly, rather than use the featureCheck composable that would allow us to apply overrides centrally.

Going to look at how quick it would be to switch over.

knolleary: It is not a quick switch over. There is a lot of inconsistent use of the `featureCheck` function and it doesn't cover all feature flags we have.

So I can see why the choice was made to modify the type property. It isn't the ideal solution, but it may be what we have to run with.

frontend/src/pages/team/Settings/TeamAdminTools.vue
```diff
@@ -381,6 +420,15 @@ export default {
                 delete properties.devices.combinedFreeType
             }
 
+            Object.keys(this.editableLimits.features).forEach(feature => {
+                // only store the delta from the TeamType
```
knolleary, line 446: This isn't working as intended - because the API is modifying `team.type.properties.features` to include the overrides, this check is not comparing against the underlying type properties.

Working on a fix.

knolleary:
I've pushed a UI rework for the edit dialog:

image


Including an indicator for any applied overrides:

image


As noted in my inline comments, I think there is a big tidy up needed in how feature flags are checked in the frontend. We have lots of places access `team.type.properties` directly - which has meant this PR is mutating that object to apply the overrides. It would be much clearer for `team.type` to be unchanged, and for the frontend to provide a central place to do feature checks. That central place does exist for some flags, but not used consistently. I will raise an issue to address that, then we can revisit the change made by this PR.

PR #6275, opened 2025-11-07
Temp patch to fix broken rollup package

hardillb:
## Description

temp override of rollup version while we wait for broken package to be fixed.

https://github.com/rollup/rollup/issues/6168


Steve-Mcl:
FYI, this is now resolved with a release of 4.53.1

PR #6311, opened 2025-11-19
fix(housekeeper): add optional chaining for broker availability check

dimitrieh:
## Description

The housekeeper leadership voting feature introduced in PR #6239 assumes the MQTT broker is always configured. This causes a crash when running FlowFuse without broker configuration (which is valid for local dev).

Add optional chaining to safely handle cases where app.comms.platform is undefined, consistent with how other parts of the codebase handle missing broker configuration (e.g., forge/settings/index.js).

## Related Issue(s)

No issue as of yet. But the change was necessary due to #6239

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/housekeeper/index.js
+++ b/forge/housekeeper/index.js
@@ -24,7 +24,7 @@ module.exports = fp(async function (app, _opts) {
     const leaderVotes = {}
     // vote every 15 seconds
     const voteInterval = setInterval(() => {
-        app.comms.platform.housekeeper.vote(localLeaderVote)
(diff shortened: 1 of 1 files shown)


dimitrieh:
@hardillb can you see if this makes sense according to you for dev usecases?

hardillb:
@dimitrieh running with a broker locally is probably worth it as a lot of the EE features now depend on it, but this should allow housekeeper tasks to run without.

dimitrieh:
@hardillb 👍 . merging in that case

Issue #6281, opened 2025-11-10
Hyperlinks in chat widget should always open in external tab


Steve-Mcl:
### Description

In FlowFuse Expert, when a chat type answer contains a hyperlink, it opens over the top of the flowfuse application (including immersive mode editor). All links should open to a target (or _blank)

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

cstns:
pausing work in favor of surfacing flows into the expert

Steve-Mcl:
@cstns look at the dompurify options I added to the website PR. Dom purify strips target attribute without it.

PR #6334, opened 2025-11-25
Force plain links to open in new tabs

cstns:
## Description

forces pain links received in responses to always open in new tabs and also adds utm parameters.

- parametrized the sanitize method in the String composable to accept target and append query params
- because dompurify only has a global parametrization option to configure attr handling I opted with adding/removing a `afterSanitizeAttributes` hook to keep the global instance clean

image

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/6281

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/composables/String.js
```diff
@@ -84,6 +84,43 @@ export const slugify = (str) => {
...
+
+    // add a per-call temporary hook
+    //  - dom purify can only be configured globally
+    //  - this hook will be applied removed after sanitization
+    DOMPurify.addHook('afterSanitizeAttributes', hook)
+
+    const result = DOMPurify.sanitize(str)
+
+    // remove all hooks for this stage (including the one we just added)
+    DOMPurify.removeHook('afterSanitizeAttributes')
```
Steve-Mcl: I think this should be in a finally block to ensure it is removed

WDYT?

cstns: sanitize is a synchronous operation and we'd be introducing race conditions between subsequent calls because hooks are a shared mutation

Steve-Mcl: I meant wrap the code in try..finally to ENSURE `removeHook` is called - even in the event of a throw

Steve-Mcl (approved):

PR #6339, opened 2025-11-26
Allow default TeamType to enabled all feats default

### LINKED ISSUE #4866: Add 'auto-enable all feature flags' to TeamType configuration
### Description

We manage features via feature flags. Each TeamType is configured via a set of checkboxes to enable the features available to that particular TeamType.

On FFC, our Enterprise team type is expected to have all features available; so we have to ensure we tick the box as part of rolling out a new feature. This can be helpful as it allows us to deploy code to production and then selectively enable it - useful when there are other things that need to be in place in a co-ordinated fashion.

For our self-hosted customers, this can mean they don't automatically see new features when they upgrade if they don't know to tick the box under their TeamType config.

I'm proposing we add a 'enable all feature flags' option to the TeamType config. If ticked, then any team-level feature flag check will come back 'enabled'. Whilst we may elect not to use that on FFC, for self-hosted customers who will typically only have a single TeamType, it will reduce the admin task of accessing new features after an upgrade.

As features are also flagged via whether a license is applied or not, we could probably make this setting on-by-default for the default team type that gets created when initialising the platform.

### Tasks

Listing out the required changes as tasks to tick off. Holding off raising sub-issues as I think they are all just what needs doing

 - Add 'Enable all available features' checkbox to team type dialog. When ticked, hide the feature list.
 - Update feature flag checks logic to do: 'is all-flag enabled? if not, check specific flag'. In the backend, this is centralised in `TeamType.getFeatureProperty`. Frontend should have similar centralised logic, just not sure off the top of my head where that is.
 - Update the platform setup wizard (that creates a default TeamType) to set this flag by default.
 - Verify feature flag checks apply as expected
     - eg non-enterprise licensed install should not start showing EE features as available (because the platform flag will be false regardless of the teamtype)
     -

hardillb:
fixes #4866

## Description

Adds a check box to the TeamType config that enables all features.

image

It disables the individual feature checkboxes when ticked.

This defaults to enabled for the default team that is create on first startup.

Need to decide if we want a migration to enable this for existing installs if there is only 1 team type.

This PR is built on top of #6247 and targets it (so it should be merged first).

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/ee/routes/catalogues/index.js
+++ b/forge/ee/routes/catalogues/index.js
@@ -40,26 +40,30 @@ module.exports = async function (app) {
         }
     }, async (request, reply) => {
         try {
-            const packageList = await axios.get(`${app.config.npmRegistry?.url}/-/all`, {
-                // If we can swap this for a teams creds or token then the
-                // filtering will all be done in the npm repo
-                auth: {
-                    username: app.config.npmRegistry.admin.username,
-                    password: app.config.npmRegistry.admin.password
-                }
-            })
+            if (app.config.npmRegistry?.url) {
+                const packageList = await axios.get(`${app.config.npmRegistry?.url}/-/all`, {
+                    // If we can swap this for a teams creds or token then the
+                    // filtering will all be done in the npm repo
+                    auth: {
+                        username: app.config.npmRegistry.admin.username,
+                        password: app.config.npmRegistry.admin.password
+                    }
+                })
 
-            const packages = {}
-            for (const package in packageList.data) {
-                if (package === '_updated') {
-                    continue
+                const packages = {}
+                for (const package in packageList.data) {
+                    if (package === '_updated') {
+                        continue
+                    }
+                    if (package.startsWith(`@flowfuse-${request.params.teamId}/`)) {
+                        packages[package] = packageList.data[package]
+                    }
                 }
-                if (package.startsWith(`@flowfuse-${request.params.teamId}/`)) {
-                    packages[package] = packageList.data[package]
-                }
+
+                reply.send(packages)
+            } else {
+                reply.status(412).send({ error: 'npm_not_configured', message: 'NPM not configured' })
             }
-
-            reply.send(packages)
         } catch (err) {
             reply.status(500).send({ error: 'unknown_error', message: err.toString() })
         }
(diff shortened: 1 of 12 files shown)


hardillb:
Still need to decide on a migration to make this retro active?

frontend/src/pages/admin/TeamTypes/dialogs/TeamTypeEditDialog.vue
```diff
@@ -167,8 +167,11 @@
...
 
                 <FormHeading>Features</FormHeading>
+                <div>
+                    <FormRow v-model="input.properties.enableAllFeatures" type="checkbox">Enable all features</FormRow>
+                </div>
                 <div class="grid gap-3 grid-cols-2">
-                    <FormRow v-for="(feature, index) in featureList" :key="index" v-model="input.properties.features[feature]" type="checkbox">{{ featureNames[feature] }}</FormRow>
+                    <FormRow v-for="(feature, index) in featureList" :key="index" v-model="input.properties.features[feature]" :disabled="input.properties.enableAllFeatures" type="checkbox">{{ featureNames[feature] }}</FormRow>
                     <!-- to make the grid work nicely, only needed if there is an odd number of checkbox features above-->
                     <span v-if="featureList.length % 2 === 1" />
```
knolleary:
```suggestion
                    <template v-if="!input.properties.enableAllFeatures">
                        <FormRow v-for="(feature, index) in featureList" :key="index" v-model="input.properties.features[feature]" :disabled="input.properties.enableAllFeatures" type="checkbox">{{ featureNames[feature] }}</FormRow>
                        <!-- to make the grid work nicely, only needed if there is an odd number of checkbox features above-->
                        <span v-if="featureList.length % 2 === 1" />
                    </template>
```

Suggest hiding the full list if enabled-all is ticked. For self-hosted users with the option enabled, it makes declutters the form.

forge/ee/routes/catalogues/index.js
knolleary, line 1: Not sure what this is fixing in relation to the enable-all option?

knolleary (changes requested):

forge/ee/routes/catalogues/index.js
hardillb: It wraps the request in a test to see if the feature is configured, but yeah looks like it slipped in by accident with this. Possibly discovered while updating the tests.

hardillb: I'll revert and add as seperate PR

knolleary (approved):

PR #6408, opened 2025-12-12
Enable schedule instance restart

### LINKED ISSUE #6232: Schedule instance restart without NR update
### Description

Follow-on task. Once work to allow scheduling *updates* is complete, this task is to enable the scheduling of an instance restart whether or not there is an update available.

### Epic/Story

_No response_

hardillb:
fixes #6232

## Description

Allows scheduling to restart Node-RED and not apply updates

image

Default is to do stack update, but can be changed to just restart on same schedule.

## Related Issue(s)

#6232

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/instance/Settings/Maintenance.vue
+++ b/frontend/src/pages/instance/Settings/Maintenance.vue
@@ -10,7 +10,7 @@
                 </template>
             </FormRow>
             <FormRow v-model="scheduledUpgrade.restart" :disabled="!scheduledUpgrade.enabled" type="checkbox">
-                Restart Even if no update available
+                Restart even if no update available
                 <template #description>
                     This will trigger a Node-RED restart even if no update is available.
                 </template>


knolleary:
The options appear to be mutually exclusive; does this mean a user can't have both regular restarts *and* automatic upgrades? It feels like it should be possible to enable both?

hardillb:
@knolleary this was a deliberate choice, let me see if I can find it in the original isssue

hardillb:
@knolleary I can't find it, but I'm sure it was discussed, I think with regard to the original upgrade PR as that is when the backend was done for just restarting.

I can look at changing it if needed?

knolleary:
@hardillb it feels like it should be possible to enable both.

```
 - [ ] Apply updates when available
 - [ ] Restart Node-RED even if no update is available
```

hardillb:
@knolleary pushed updates

frontend/src/pages/instance/Settings/Maintenance.vue
```diff
@@ -2,17 +2,20 @@
...
-                    <p>
-                        Select the day of the week and the hour during which the automatic upgrade will occur if available. The
-                        upgrade will start within the selected hour.
-                    </p>
-                    <p class="my-3"><span class="font-bold">Note:</span> All times are stated in UTC.</p>
+                    Select the days of the week and the hour during which the automatic upgrade will occur if available. The upgrade will start within the selected hour.
                 </template>
             </FormRow>
+            <FormRow v-model="scheduledUpgrade.restart" :disabled="!scheduledUpgrade.enabled" type="checkbox">
+                Restart Even if no update available
```
knolleary:
```suggestion
                Restart even if no update available
```

cstns (approved):

PR #6441, opened 2025-12-21
Fix `TokenCreated` dialog for null tokens

GogoVega:
## Description

By default, `token` is null, which triggers an error when the default template is mounted due to the evaluation of the expression `{{ token.token }}`.

`showToken` is declared in `setup` and uses `this` which is theoretically not possible.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/pages/account/Security/dialogs/TokenCreated.vue
```diff
@@ -1,7 +1,7 @@
 <template>
     <ff-dialog ref="dialog" data-el="add-token-confirmation" header="Token Created">
         <template #default>
-            <p>Your token is <code>{{ token.token }}</code></p>
+            <p>Your token is <code>{{ token?.token }}</code></p>
```
AllanOricil, line 4: Component is called TokenCreated because it means there should exist a token already when the component is created. If you add "?." It means, semantically, that this component could possibly display "Your token is ", which isn't right. A proper fix is to ensure that the whole TokenCreated is rendered only if token is set. I believe the error occurs because TokenCreated is created and mounted when Token isn't even available, which triggers 1 rendering cycle and the evaluation of the {{ token.token }} expression. This is a waste of rendering cycle. Could you check if it is possible to create TokenCreated only when token is available?

AllanOricil: @gogovega

The issue is that TokenCreated is created and mounted to the dom when it isn't even needed, wasting a rendering cycle for nothing


https://github.com/FlowFuse/flowfuse/blob/main/frontend%2Fsrc%2Fpages%2Faccount%2FSecurity%2FTokens.vue#L28

It would be more efficient to just create and mount the component with a v-if in this case, since it is a very small component Users wouldn't notice the time spent rendering and mounting a small template to the dom. You can define a new state variable called "showCreatedToken", and use it in v-if. When the following method is called, you could set showCreatedToken to true, then call showToken. If calling showToken in the same tick doesn't work, put it on the nextTick callback.

https://github.com/FlowFuse/flowfuse/blob/main/frontend%2Fsrc%2Fpages%2Faccount%2FSecurity%2FTokens.vue#L81-L84

Another solution is to control the rendering of the component inside its own template. In other words add v-if inside of it instead of in the parent, then trigger the rendering when calling showToken from the parent. The trade-off is that vue will again spend time mounting something that isn't used yet.

cstns: Agreed, we don't  have the best state management when it comes to dialogs, mainly because we're using a mix of vue and the built in html5 dialog api.

Issue #6484, opened 2026-01-09
Allow admin-created users to get a trial-mode personal team


knolleary:
### Description

The admin UI for creating a new user (rather than have them self-register) includes an option to create them a personal team.

This creates a team for the user however:

- it assumes there's a teamtype called 'starter' and uses that
- it doesn't create applicaiton/instance
- it doesn't apply trial settings

We have a requirement to be able to administratively register a user on their behalf and put them into a trial team.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #6485, opened 2026-01-09
Add option to disable provisioning new users via google login

knolleary:
Part of #6484 

Currently, the google social login flow will always automatically create a new user if they are logging in for the first time.

If we disable self-service sign-up of the app, we also need to prevent the social login from provisioning new users.

This adds a new admin option for the social login as to whether new users should be created when attempting to login for the first time.

### DIFF
--- a/forge/ee/routes/sso/social/google.js
+++ b/forge/ee/routes/sso/social/google.js
@@ -61,7 +61,7 @@ module.exports = fp(async function (app, opts) {
                 reply.send({
                     url: '/'
                 })
-            } else {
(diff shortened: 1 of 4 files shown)


cstns:
Should we "link" this platform option to the existing ` Allow new users to register on the login screen ` option and offer the possibility of enablement only if the first option is also enabled?

I assume this is mostly intended for FFC so not sure how much we want to complicate things or if a change request to enable is required.

knolleary:
Lets not complicate it for now. We're the only ones using these options.

PR #6487, opened 2026-01-09
Fix button width calculation and update route names in VersionHistory page

cstns:
## Problem
CleanShot 2026-01-09 at 15 26 18

## Description

Fixes the `ToggleButtonGroup` width to make room for the indicator right border:
Before:
image
After:
image

The `ToggleButtonGroup` was not identifying the correct `selectedIndex` because it was receiving a relative path as a prop from the `VersionHistory` page when running in immersive mode. We relied on the `router-link` component to handle the active class until recently and it didn't pose a problem.

Switching to named paths in the `VersionHistory` when setting the `pageToggle` props on the  `ToggleButtonGroup` fixes the problem.

## Related Issue(s)

N/A

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/elements/ToggleButtonGroup.vue
+++ b/frontend/src/components/elements/ToggleButtonGroup.vue
@@ -76,7 +76,7 @@ export default {
             if (this.selectedIndex < 0) return { opacity: 0 }
             return {
                 transform: `translateX(${this.selectedIndex * 100}%)`,
-                width: `${100 / this.buttons.length}%`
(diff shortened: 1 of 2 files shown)


cstns:
that's happening because the router link doesn't match the current route on the `Version History` tab because of my fix, thinking of an alternative

cstns:
covered by https://github.com/FlowFuse/flowfuse/pull/6658

PR #6679, opened 2026-02-10
Allow snapshot to be restored to a remote instance in dev mode

### LINKED ISSUE #6654: Add snapshot restore option to app-assigned remote instances
### Description

Part of #6496 - full details in there

Add restore option to the snapshot list for remote instances.

I *think* this should only be available to app-assigned instances as instance-assigned instances are in fleet mode.

Tasks (not sure they need to be further split out... I could be wrong)

 - Frontend: add option to context menu
 - Backend: ensure api allows this scenario

This will have a dependency on an updated device agent to support this action; see the other child tasks of #6496 for details. The UI should *only* offer this action if the device agent is up to date. If it is not up to date, the option should be disabled - ideally with a tooltip of why its disabled (not sure if we can do tooltips on context menu items...)

### Epic/Story

_No response_

knolleary:
Closes #6654 

This adds a `forceUpdate` flag to the status update sent to a device when changing the snapshot version manually (ie, not via a pipeline).

When running a device with https://github.com/FlowFuse/device-agent/pull/570 applied, this will allow the snapshot to be applied even if in developer mode.

In the UI:

 - Updates the `canDeploy` logic that determines if the snapshot restore button is enabled so that if the device is in dev mode, but is running 3.8.0 (as yet unreleased), then the button should be enabled
 - Adds the ability to set a tooltop on the drawer action buttons - and does so for the restore button if it will be disabled to help the user know what's up
 - Relaxes the spacing of the Snapshot info table as it was very bunched up.

### DIFF
--- a/.github/workflows/branch-deploy.yaml
+++ b/.github/workflows/branch-deploy.yaml
@@ -89,7 +89,7 @@ jobs:
       needs.validate-user.outputs.is_org_member == 'true' &&
       github.event_name == 'workflow_dispatch' &&
       inputs.driver_k8s_branch != 'main'
-    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.47.0'
+    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.48.0'
     with:
       package_name: driver-k8s
       publish_package: true
@@ -106,7 +106,7 @@ jobs:
       needs.validate-user.outputs.is_org_member == 'true' &&
       github.event_name == 'workflow_dispatch' &&
       inputs.nr_project_nodes_branch != 'main'
-    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.47.0'
+    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.48.0'
     with:
       package_name: nr-project-nodes
       publish_package: true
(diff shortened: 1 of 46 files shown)


knolleary:
> Restore snapshot 1 ✅
> Restore snapshot 2 ❌
> After refreshing the FF app, it is permitted to be deployed.

I *think* this is a timing issue. You are deploying snapshot 1, then fairly quickly asking to restore snapshot 2 - however the device is still updating to snapshot1 and has not yet reported back that it's running snapshot1 - so the platform thinks its active snapshot is still snapshot 2.

I'll try to reproduce, but with a longer delay between the restores to see if that theory is correct. If so, I think we can live with it.

Steve-Mcl:
> I _think_ this is a timing issue.

I dont think so unfortunately.

I took that gif about 20 mins ago then left it running.

It still happens.

chrome_PQjk2PvJtY


I had to refresh FF page then I could deploy `Snapshot 1`


---

EDIT

it is hard to tell from the GIF (since its a loop) but whats happening is:
1. I try to send Snapshot 1 - get "Cannot" toast
2. I go to the editor to see what is running - it is Snapshot 2
3. I go back to FF and refresh
4. I can now deploy Snapshot 1

knolleary:
- Fixed the refresh of device details so the frontend knows the snapshot has changed.
 - Updated the dialog if in dev mode to include the warning about overwriting changes.

image


@Steve-Mcl ready for another review

Issue #6683, opened 2026-02-10
Re-enable the immersive mode for remote instances


cstns:
### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #6684, opened 2026-02-10
Reenable immersive mode for remote instances

cstns:
## Description

created a new editor link button to replicate styling of the hosted instances editor link.
styling

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/6683

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/team/Home/components/RecentlyModifiedDevices.vue
+++ b/frontend/src/pages/team/Home/components/RecentlyModifiedDevices.vue
@@ -139,6 +139,7 @@ export default {
             display: flex;
             width: 100%;
             height: fit-content;
+            min-height: 55px;
             justify-content: space-between;
             align-items: center;
             gap: 10px;
--- a/frontend/src/pages/team/Home/components/RecentlyModifiedInstances.vue
+++ b/frontend/src/pages/team/Home/components/RecentlyModifiedInstances.vue
@@ -110,6 +110,7 @@ export default {
             display: flex;
             width: 100%;
             height: fit-content;
+            min-height: 55px;
             justify-content: space-between;
             align-items: center;
             gap: 10px;


Steve-Mcl:
Cool it fixes alignment 👊

However, the font style is still mismatched 😢

image

<br>

It also fixes button spacing 👊

### before
image


### this PR
image

Steve-Mcl:
My only other feedback would be:

The Hubspot ticked stated: 

> Previously when opening a node-red editor from the flowfuse platform, the instance was opened in a new tab. I see that this functionality changed to the expert pane to be more integrated. This is verry inconvenient when working on multiple projects side by side. There is also no link available to right-click

Since we need the "Open Editor" to be a `button` element (due to needing to check or initiate the tunnel connection) we cannot simply provide a hyperlink.  Therefore I suggest, for those not aware of `ctrl-click` or `middle-button-click` we should add a 2nd item in the list e.g:
image

Happy for that to be a follow up tho.

Steve-Mcl:
Last point

Since we are no longer redirecting back to FF app upon disconnections - is there any reason NOT to include the restart dropdown?

cstns:
I'd leave them to a follow-up, they're not hard to implement

Steve-Mcl:
> I'd leave them to a follow-up, they're not hard to implement

That is fine however I just want to spell it out (seeing as we had a complaint about this in the 1st iteration) for a person who doesnt know ctrl+click or middle-mouse-click but DOES want separate tabs, we lose the thing we are trying to achieve - "engagement in the immersive mode".  

For those not aware of ctrl+click/middle-mouse-click they only see 2 options. 

| action | result | immersive |
| -- | -- | -- |
| click the button | opens in same tab | yes |
| click the dropdown then the Direct URL | opens in new tab | no |

PR #6762, opened 2026-02-25
Add Device Agent on Kubernetes documentation

### LINKED ISSUE #6578: Kubernetes support for Device Agent
### Description

In a recent kickoff and architecture discussion
📼 https://fathom.video/calls/537276487

This PR requests clarity on the following:
1. Recommended way to run the Device Agent on Kubernetes today
	•	e.g. supported / expected deployment model (high-level)
2. Helm chart or reference manifests
	•	Do we provide an official Helm chart?
	•	If not, do we endorse reference Kubernetes manifests?
3. Provisioning token handling (Kubernetes-native)
	•	Is token injection via environment variables supported?
	•	Should device.yml be mounted as a Secret?
	•	Any patterns we explicitly recommend or discourage?

### Epic/Story

_No response_

hardillb:
closes #6578

## Description

Describes 2 deployment options
- pre-configured device.yml
- provisioning token

This is bare minimum docs, might need some more "why" language at the top...

## Related Issue(s)

#6578 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/.github/workflows/branch-deploy.yaml
+++ b/.github/workflows/branch-deploy.yaml
@@ -97,7 +97,7 @@ jobs:
       needs.validate-user.outputs.is_org_member == 'true' &&
       github.event_name == 'workflow_dispatch' &&
       inputs.driver_k8s_branch != 'main'
-    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.49.0'
+    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.51.0'
     with:
       package_name: driver-k8s
       publish_package: true
@@ -114,7 +114,7 @@ jobs:
       needs.validate-user.outputs.is_org_member == 'true' &&
       github.event_name == 'workflow_dispatch' &&
       inputs.nr_project_nodes_branch != 'main'
-    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.49.0'
+    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.51.0'
     with:
       package_name: nr-project-nodes
       publish_package: true
@@ -131,7 +131,7 @@ jobs:
       needs.validate-user.outputs.is_org_member == 'true' &&
       github.event_name == 'workflow_dispatch' &&
       inputs.nr_file_nodes_branch != 'main'
-    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.49.0'
+    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.51.0'
     with:
       package_name: nr-file-nodes
       publish_package: true
@@ -148,7 +148,7 @@ jobs:
       needs.validate-user.outputs.is_org_member == 'true' &&
       github.event_name == 'workflow_dispatch' &&
       inputs.nr_assistant_branch != 'main'
-    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.49.0'
+    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.51.0'
     with:
       package_name: nr-assistant
       publish_package: true
@@ -165,7 +165,7 @@ jobs:
       needs.validate-user.outputs.is_org_member == 'true' &&
       github.event_name == 'workflow_dispatch' &&
       inputs.nr_tables_nodes_branch != 'main'
-    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.49.0'
+    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.51.0'
     with:
       package_name: nr-tables-nodes
       publish_package: true
@@ -182,7 +182,7 @@ jobs:
       needs.validate-user.outputs.is_org_member == 'true' &&
       github.event_name == 'workflow_dispatch' &&
       inputs.nr_subflow_export_branch != 'main'
-    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.49.0'
+    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.51.0'
     with:
       package_name: nr-subflow-export
       publish_package: true
@@ -199,7 +199,7 @@ jobs:
       needs.validate-user.outputs.is_org_member == 'true' &&
       github.event_name == 'workflow_dispatch' &&
       inputs.nr_mqtt_nodes_branch != 'main'
-    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.49.0'
+    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.51.0'
     with:
       package_name: nr-mqtt-nodes
       publish_package: true
@@ -229,7 +229,7 @@ jobs:
       needs.publish_nr_tables_nodes.result == 'success' || 
       needs.publish_nr_mqtt_nodes.result == 'success' || 
       needs.publish_nr_subflow_export.result == 'success'
-    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.49.0'
+    uses: 'flowfuse/github-actions-workflows/.github/workflows/publish_node_package.yml@v0.51.0'
     with:
       package_name: flowfuse-nr-launcher
       publish_package: true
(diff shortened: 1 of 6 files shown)


ppawlowski:
I thought about placing manifests in the device agent repo, in the `examples` directory, to make it easier to reference in some kind of scripts/pipelines/etc. It will also be easier for us to automatically scan them and keep image tags up to date. WDYT?

hardillb:
I might just change them to `:latest` :stuck_out_tongue: 

But yeah, they should probably end up in there as well

ppawlowski:
@hardillb please check my changes. Merge the PR anytime.

docs/device-agent/install/kubernetes.md
```diff
@@ -0,0 +1,174 @@
...
+  description: Run the FlowFuse Device Agent in a Kubernetes cluster
+  tags:
+     - device agent
+     - kubernetes
+     - installation
+---
+
+# Kubernetes Install
+
+Any deployment on Kubernetes is going to be specific to the environment and requirements of the solution. The following are examples that can be adopted to fit specific needs.
```
allthedoll: I think we can give more direction here, like


```suggestion
Any deployment on Kubernetes is going to be specific to the environment and requirements of the solution. The following examples show two common patterns for running the FlowFuse Device Agent on Kubernetes:

- Fixed configuration using a static `device.yml`
- Automatic provisioning using a FlowFuse Provisioning Token

Choose the approach that matches how you manage device lifecycle and credentials.
```

docs/device-agent/install/kubernetes.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Fixed Configuration
+
+If you have an existing `device.yml` file containing a set of Device Agent credentials.
+
+```bash
+kubectl create secret generic device-one-secret --from-file=device.yml=./device.yml
+```
+
+The following manifest will create a Deployment and Service for a device using the supplied Secret as it's credentials
```
allthedoll:
```suggestion
The following manifest will create a Deployment and Service for a device using the supplied Secret as its credentials
```

ppawlowski:
To clarify: the `navOrder` value change in the `README.md` file fixes ordering in the higher level navigation in the sidebar.

Current:
Zrzut ekranu 2026-02-25 o 13 36 49

After the order change:
Zrzut ekranu 2026-02-25 o 13 37 26

docs/device-agent/install/kubernetes.md
```diff
@@ -0,0 +1,174 @@
...
+    matchLabels:
+      app: device-one
+  template:
+    metadata:
+      labels:
+        app: device-one
+    spec:
+      containers:
+      - name: device-one
+        image: flowfuse/device-agent:latest
```
allthedoll, line 64: Consider replacing `latest` with a version placeholder (e.g. <version>) or adding a note recommending pinning to a specific version in production. Using latest can make deployments non-reproducible.

allthedoll:
I am still reviewing 😇

docs/device-agent/install/kubernetes.md
```diff
@@ -0,0 +1,174 @@
...
+      volumes:
+      - name: config
+        secret:
+          secretName: device-one-secret
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: device-one-service
+spec:
```
allthedoll, line 88: Just to clarify, since `type` is not specified this will default to `ClusterIP`, which is internal only, right? Do we want to mention in the text how users would expose this externally using an Ingress, LoadBalancer, or NodePort? or is internal only the intended default here?

docs/device-agent/install/kubernetes.md
```diff
@@ -0,0 +1,174 @@
...
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: device-one
+  labels:
+    app: device-one
+spec:
+  replicas: 1
```
allthedoll: Since this deployment uses a single pvc and provisioning token, should we explicitly note that this example assumes one device agent per deployment and should not be scaled beyond one replica?

docs/device-agent/install/kubernetes.md
```diff
@@ -0,0 +1,174 @@
...
+    matchLabels:
+      app: device-one
+  template:
+    metadata:
+      labels:
+        app: device-one
+    spec:
+      initContainers:
+      - name: config-copy
+        image: busybox:latest
```
allthedoll, line 133: similar to the main container image, should we avoid latest here and either pin to a specific busybox version or add a note recommending version pinning for production?

docs/device-agent/install/kubernetes.md
```diff
@@ -0,0 +1,174 @@
...
+        command:
```
allthedoll, line 134: since this `initContainer` only copies the file if it does not already exist, should we add a short note explaining that the provisioning configuration is only applied on first start and that subsequent restarts preserve the existing `device.yml `on the pvc?

docs/device-agent/install/kubernetes.md
```diff
@@ -0,0 +1,174 @@
...
+          requests:
+            cpu: 500m
+            memory: 128Mi
+      volumes:
+      - name: initial-config
+        secret:
+          secretName: device-provisioning-secret
+      - name: config
+        persistentVolumeClaim:
+          claimName: device-one-pvc
```
allthedoll, line 166: since this pvc is per device, do we want to call out that each device agent instance should have its own unique pvc (and deployment name) so storage is not shared across devices?

docs/device-agent/install/kubernetes.md
```diff
@@ -0,0 +1,174 @@
...
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: device-one-pvc
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
```
allthedoll, line 190: this assumes the cluster has a default `StorageClass`. Should we add a brief note that users may need to set `storageClassName` depending on their Kubernetes setup, for example `storageClassName: <your-storage-class>`?

Also, should we mention what gets stored here (the updated `device.yml` and installed nodes) so users can size the pvc appropriately?

docs/device-agent/install/overview.md
```diff
@@ -14,7 +14,7 @@ meta:
 
 ## Choose your install path
 
-Most users should use the Device Agent Installer. Power users can choose Manual ( using `npm`) or Docker deployments.
+Most users should use the Device Agent Installer. Power users can choose Manual ( using `npm`), Docker, or Kubernetes deployments.
```
allthedoll: since we are expanding install options, should we strengthen the recommendation slightly here to make the installer the clearly preferred path for most users? ex: calling it out as “Recommended” or “Quick Start"


```suggestion
**Recommended for most users:** Use the Device Agent Installer (Quick Start), the fastest way to deploy with minimal configuration. Power users can choose Manual (using `npm`), Docker, or Kubernetes deployments.
```

allthedoll (changes requested):

docs/device-agent/install/kubernetes.md
hardillb: That's why it's named `device-one-pvc`

docs/device-agent/install/kubernetes.md
hardillb: No, because if they want to expose it they should really add an Ingress object

docs/device-agent/install/kubernetes.md
hardillb: No, this was changed from a specific version, we want to force users to run the latest device agent.

docs/device-agent/install/kubernetes.md
hardillb: We could add a note that horizontal scaling requires using a StatefulSet

docs/device-agent/install/kubernetes.md
hardillb: Nah, it's so trivial a start up script that latest is fine

docs/device-agent/install/kubernetes.md
```diff
@@ -0,0 +1,179 @@
...
+navOrder: 4
+meta:
+  description: Run the FlowFuse Device Agent in a Kubernetes cluster
+  tags:
+     - device agent
+     - kubernetes
+     - installation
+---
+
+# Kubernetes Install
```
allthedoll, line 13:
```suggestion
# Kubernetes Install

## When to Use Each Option

Running the Device Agent in Kubernetes is appropriate when devices are containerized or managed as part of a Kubernetes-based edge or infrastructure platform.

Choose your deployment pattern based on how you manage device identity:

- **Fixed Configuration**  
  Use when the device already exists in FlowFuse and you have a `device.yml` with its credentials. One deployment maps to one device identity.

- **Automatic Provisioning**  
  Use when devices should register themselves at startup using a Provisioning Token. Each instance requires writable persistent storage.
```

allthedoll (approved):
I added a hot suggested why/when if you care for it @hardillb. Otherwise it's looking good.

PR #6780, opened 2026-03-02
Update Instances AutoStackUpdate when TeamType Changes

### LINKED ISSUE #6349: Apply default scheduled maintenance to all instances in a team tier
### Description

The work done via #6079 allows an admin to set a default schedule for maintenance restarts for instances in a given team tier.

When a new instance is created, it gets a slot from the default schedule applied to its settings. Existing instances however do not get the option applied.

The goal of this task is to apply the default to all instances in the tier.

**My proposed solution is:**

A new task runs once a day, looks for any instances belonging to a TeamType that has maintenance defaults but the instance doesn't have their own setting enabled. It can then assign a slot to the instances. This has the benefit of allowing the default to be a bigger time window and the instances get distributed within it.

We will need to refine the exact criteria for enabling the flag; it should only apply if the TeamType config disallows disabling the schedule.

This will catch-up the backlog, but also handle cases where a team downgrades to Starter following a Pro trial and will ensure its instances get the maintenance schedule applied.

The downside of this is that *if* the default schedule is changed or turned off etc, it won't apply that change to the existing instances.

### Epic/Story

_No response_

hardillb:
fixes #6349

## Description

Adds a daily (at random time) job that checks for Teams with a TeamType that includes mandatory AutoStackUpdates. It then checks all Instances in those teams to see if they have a restart time set and if not adds one.

## Related Issue(s)

#6349

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/docs/install/introduction.md
+++ b/docs/install/introduction.md
@@ -10,6 +10,10 @@ meta:
      - kubernetes
      - trial license
      - deployment models
+templateEngineOverride: njk,md    
+installationServiceHubspot:
+  formId: "22edc659-d098-4767-aeb1-6480daae41ad"
+  targetId: "hs-form-installation-service"
 ---
 
 # Installing FlowFuse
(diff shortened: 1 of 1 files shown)


hardillb:
Also included a fix for the original code when creating a new instance

forge/ee/lib/autoUpdateStacks/tasks/enforce-team-rules.js
```diff
@@ -0,0 +1,59 @@
...
+                                            }
+                                        }
+
+                                        if (!found) {
+                                            const days = autoStackUpdate.days
+                                            const hours = autoStackUpdate.hours
+                                            // generate random day and hour in ranges
+                                            const day = days[Math.floor(days.length * Math.random())]
+                                            const hour = hours[Math.floor(hours.length * Math.random())]
+                                            await instance.updateSetting(`${KEY_STACK_UPGRADE_HOUR}_${day}`, { hour })
```
knolleary, line 49:
```suggestion
                                            app.log.info(`AutoStackUpgrade: applying update schedule ${day}/${hour} to instance ${instance.id} in team ${team.hashid}`)
                                            await instance.updateSetting(`${KEY_STACK_UPGRADE_HOUR}_${day}`, { hour })
```

Adding by inspection; please verify I've got the inserts right

knolleary (changes requested):
Added some logging - but it needs testing to ensure I've got the inserts right.

Issue #6804, opened 2026-03-04
Pinia Task 1 - ux-dialog


n-lark:
## Task 1 — `ux-dialog` (PR 2)

**Gate:** Phase 0 merged
**Vuex module:** `frontend/src/store/modules/ux/dialog/index.js`
**New file:** `frontend/src/stores/ux-dialog.js`
**No persistence. No cross-store dependencies.**

### 1.1 — Create the Pinia store

```javascript
// frontend/src/stores/ux-dialog.js
import { defineStore } from 'pinia'

export const useUxDialogStore = defineStore('ux-dialog', {
    state: () => ({
        dialog: {
            boxClass: null,
            cancelLabel: null,
            canBeCanceled: true,
            confirmLabel: null,
            contentClass: '',
            disablePrimary: false,
            header: null,
            html: null,
            is: null,
            kind: null,
            onCancel: null,
            onConfirm: null,
            subHeader: null,
            text: null,
            textLines: null,
            notices: []
        }
    }),
    actions: {
        clearDialog (cancelled = false) {
            if (cancelled) {
                this.dialog.onCancel?.()
            }
            this.$reset()  // Pinia re-invokes the state factory; no initialDialogState helper needed
        },
        showDialogHandlers ({ payload, onConfirm, onCancel }) {
            this.clearDialog(false)

            if (typeof payload === 'string') {
                this.dialog.content = payload
            } else {
                this.dialog.header = payload.header
                this.dialog.text = payload.text
                this.dialog.textLines = payload.text?.split('\n')
                this.dialog.html = payload.html
                this.dialog.is = payload.is ?? undefined
                this.dialog.confirmLabel = payload.confirmLabel
                this.dialog.cancelLabel = payload.cancelLabel
                this.dialog.kind = payload.kind
                this.dialog.disablePrimary = payload.disablePrimary
                this.dialog.notices = payload.notices
                if (Object.prototype.hasOwnProperty.call(payload, 'canBeCanceled')) {
                    this.dialog.canBeCanceled = payload.canBeCanceled
                }
                this.dialog.boxClass = payload.boxClass
            }
            this.dialog.onConfirm = onConfirm
            this.dialog.onCancel = onCancel
        },
        setDisablePrimary (value) {
            this.dialog.disablePrimary = value
        },
        setDialogDevices (devices) {
            this.dialog.is.payload.devices = devices
        }
    }
})
```

[...]

Issue #6812, opened 2026-03-05
Standardise Import Aliases for Components, Stores, and Composables


n-lark:
# Standardise Import Aliases for Components, Stores, and Composables

> **Before picking this up:** The ESLint resolver configuration and initial alias groundwork was completed as part of **#6804 — Pinia Task 1: ux-dialog**. Review that PR first to understand what has already been set up and avoid duplicating work.

## Summary

The FlowFuse frontend currently uses a mix of deep relative import paths (e.g. `../../components/FormHeading.vue`) across components, stores, and composables. The `@/` alias — pointing to `frontend/src` — is already configured in both Webpack and Vitest, but is not being used in source files. This issue proposes standardising all frontend imports to use the `@/` alias, completing the ESLint resolver configuration to support it, and applying the pattern consistently to components, stores, and composables.

---

## What is an Import Alias?

An **import alias** is a shorthand path token that a bundler (Webpack, Vite) or test runner (Vitest) resolves to a real directory at build/test time. Rather than counting parent directory hops, developers write a consistent, intention-revealing prefix.

**Without alias:**
```js
import FormHeading from '../../components/FormHeading.vue'
import { useNavigationHelper } from '../../../composables/NavigationHelper.js'
import { useAccountStore } from '../../../stores/account.js'
```

**With `@/` alias:**
```js
import FormHeading from '@/components/FormHeading.vue'
import { useNavigationHelper } from '@/composables/NavigationHelper.js'
import { useAccountStore } from '@/stores/account.js'
```

Both resolve to the same file. The alias version is unambiguous regardless of where the importing file lives in the tree.

---

### Alias is already configured — but unused in source

`config/webpack.config.js` (lines 170–176):
```js
resolve: {
    alias: {
        vue: 'vue/dist/vue.esm-bundler.js',
        '@': path.resolve('frontend/src')
    }
}
```

`config/vitest.config.ts` (lines 26–30):
```ts
resolve: {
    alias: {
        '@': fileURLToPath(new URL('../frontend/src', import.meta.url))
    }
}
```

Despite this, every `.vue` and `.js` file in `frontend/src/` uses relative paths. Example from `frontend/src/pages/instance/Overview.vue`:
```js
import InstanceApi from '../../api/instances.js'
import FormHeading from '../../components/FormHeading.vue'
import { useNavigationHelper } from '../../composables/NavigationHelper.js'
```

### Alias targets (once adopted)

| Alias | Resolves to |
|---|---|
[...]

PR #6813, opened 2026-03-05
[6804] Pinia Task 1 - ux-dialog

n-lark:
Resolves https://github.com/FlowFuse/flowfuse/issues/6804

## Description

- See full details [here](https://github.com/FlowFuse/flowfuse/issues/6804#issue-4024271237). 
- See manual testing plan [here](https://github.com/FlowFuse/flowfuse/issues/6804#issuecomment-4007448161). 

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/6804

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


package-lock.json
```diff
@@ -9884,13 +9885,17 @@
             }
         },
         "node_modules/array-buffer-byte-length": {
-            "version": "1.0.0",
-            "resolved": "https://registry.npmjs.org/array-buffer-byte-length/-/array-buffer-byte-length-1.0.0.tgz",
-            "integrity": "sha512-LPuwb2P+NrQw3XhxGc36+XSvuBPopovXYTR9Ew++Du9Yb/bx5AzBfrIsBoj0EZUifjQU+sHL21sseZ3jerWO/A==",
+            "version": "1.0.2",
```
n-lark: This change set is sus - figure it out

n-lark: Ended up killing install since it caused issues with the CI - see details under Standardise Import Aliases for Components, Stores, and Composables #6812

frontend/src/components/DevicesBrowser.vue
```diff
@@ -459,8 +461,8 @@ export default {
     computed: {
         ...mapState('account', ['team', 'teamMembership']),
         ...mapState('ux/tours', ['tours']),
-        ...mapState('ux/dialog', ['dialog']),
         ...mapGetters('account', ['featuresCheck']),
+        dialog () { return useUxDialogStore().dialog },
```
cstns: This logic should be moved to a setup method. Since the component includes mutators that alter state, the computed properties should remain read-only. Specifically, the checkedDevices watcher is currently acting as a mutator and should be refactored.

the mutator is the checkedDevices watcher
```javascript
        checkedDevices (devices) {
            if (this.dialog?.is?.payload?.devices) {
                this.dialog.is.payload.devices = devices
            }
        }
```


fix: 
```javascript
setup () {
    const dialogStore =  useUxDialogStore()
     return {
          dialog: dialogStore.dialog
     }
}
```

frontend/src/mixins/Dialog.js
```diff
@@ -1,20 +1,17 @@
-import { mapActions, mapState } from 'vuex'
-
 import dialog from '../services/dialog.js'
 
+import { useUxDialogStore } from '@/stores/ux-dialog.js'
+
 export default {
     computed: {
-        ...mapState('ux/dialog', ['dialog'])
+        dialog () { return useUxDialogStore().dialog }
```
cstns: this computed prop is not used within this component, it can be removed

cstns: just now realizing this is a mixin.

the dialog should be exposed as a reactive key in a setup method, parts of it are being bound to components in `frontend/src/layouts/Platform.vue:43` (there may be others) and computed props cannot/should not be mutated.

`frontend/src/layouts/Platform.vue:52` dialog.notices seem to be bound as well.

cstns: Dwelling on it, exposing computed props from mixins is not a good pattern to begin with, i know it was my doing.. We also veered in a direction where we try to steer away from mixins in favor of composables. 

not asking to remove the mixin altogether, but since this mixin is used in only two places, we should at least get rid of the computed prop and expose the dialog store key directly there via the setup method

frontend/src/components/dialogs/device-group-management/AddDeviceToGroupDialog.vue
```diff
@@ -119,7 +121,7 @@ export default {
         }
     },
     computed: {
-        ...mapState('ux/dialog', ['dialog']),
+        dialog () { return useUxDialogStore().dialog },
```
cstns: I'm not finding any usage of this computed prop

frontend/src/stores/ux-dialog.js
```diff
@@ -0,0 +1,61 @@
+import { defineStore } from 'pinia'
+
+const initialDialogState = () => ({
+    boxClass: null,
```
cstns: Does it still make sense to store state like this? Storing it this way in vuex made it possible to expose it to the store twice, once as the actual state and 2ndly as the initialState key which was used when resetting the store

test/unit/frontend/stores/ux-dialog.spec.js
```diff
@@ -0,0 +1,57 @@
+// frontend/src/tests/stores/ux-dialog.spec.js
+import { createPinia, setActivePinia } from 'pinia'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { useUxDialogStore } from '@/stores/ux-dialog.js'
```
cstns, line 4: this should reside in `test/unit/frontend/stores` to match our current testing scaffolding

jsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+    "compilerOptions": {
```
cstns: Since path resolution for @/* is scoped to the frontend, it might be cleaner to handle this within the frontend/jsconfig.json. 

My IDE (and VS Code) isn't picking up the aliased imports currently, but adding it there should fix the resolution for everyone

frontend/src/stores/ux-dialog.js
n-lark: Hey so from my investigation you do not need to store the initial state separately as you did in Vuex. 

Pinia automatically resets the state by re-invoking the state function you define in your store. So, you can simply define your state as a function, and Pinia will handle resetting to the initial state internally when you call $reset. There’s no need to keep a separate initialState property in the store.

I went ahead and updated the store to use $reset: 

```
export const useUxDialogStore = defineStore('ux-dialog', {
    state: () => ({
        dialog: {
            boxClass: null,
            cancelLabel: null,
            canBeCanceled: true,
            confirmLabel: null,
            contentClass: '',
            disablePrimary: false,
            header: null,
            html: null,
            is: null,
            kind: null,
            onCancel: null,
            onConfirm: null,
            subHeader: null,
            text: null,
            textLines: null,
            notices: []
        }
    }),
    actions: {
        clearDialog (cancelled = false) {
            if (cancelled) {
                this.dialog.onCancel?.()
            }
            this.$reset()
        },
```

frontend/src/components/DevicesBrowser.vue
n-lark: Alright so this was an interesting one. We have some considerations to make here: 

## Using `setup()`

The problem with defining the values in `setup()` is that when we access the values in the template we need to hit like `dialog.value.header` since its stored in a ref. 

Quick note on storeToRefs: 

> You use storeToRefs with Pinia to ensure that each property you access from the store (like dialog) is a ref that stays reactive when destructured.
> 
> Without storeToRefs, if you do:
> 
> ```
> const { dialog } = useUxDialogStore()
> ```
> 
> dialog is just a plain value, not a ref—so if dialog changes in the store, your component won’t reactively update.
> 
> With storeToRefs:
> 
> ```
> const { dialog } = storeToRefs(useUxDialogStore())
> ```
> 
> dialog is a ref, so it stays in sync with the store and updates your UI reactively.

So the pattern would be like the below. The issue I see with it is its gonna be annoying to refactor all the templates to hit .value, however this is a best practice so I am willing to suffer. 

```
    setup () {
        const { dialog } = storeToRefs(useUxDialogStore())

        return { dialog }
    },

// Then in template 

 <ff-dialog
                ref="dialog" data-el="platform-dialog"
                :header="dialog.value.header"
```

## Using computed 

Alternatively, we can expose the store property via a computed property in the Options API. This allows us to access dialog directly in the template, without needing to use .value, since computed properties are automatically unwrapped in Options API templates.

```
computed: {
    dialog () { return useUxDialogStore().dialog }
}

// Template 

<ff-dialog
    ref="dialog" data-el="platform-dialog"
    :header="dialog.header"
```

### Pros:

- No need to use .value in the template, just use dialog.header, dialog.kind, etc.
- Simpler migration for legacy Options API components.
- Less template refactoring required.

### Cons:

- Less idiomatic for modern Vue 3 + Pinia, which encourages Composition API and composables.
- Computed properties should not be mutated, but store state (like dialog) is often updated elsewhere.
- If/when we migrate to <script setup> or full Composition API, this pattern will need to be refactored.

n-lark: So I am of the opinion that we should do `setup()` for sure. I just wanted sign off before I go and refactor all the templates to use the `.value` syntax. 

As of now, I left everything in computed just FYI.

test/unit/frontend/stores/ux-dialog.spec.js
n-lark: Moved! As a consequence of this we had some issues resolving import aliases and I went ahead and did an override. I felt it was a fair trade off since its only targeting that specific folder and the other routes were installing alias resolvers unless I am missing a simpler solution. Lmk what you think. 

```
   // .eslintrc
   // Ignore unresolved @ imports and Node import errors only for test stores files, has issues with aliases
        {
            "files": [
                "test/unit/frontend/stores/**/*"
            ],
            "rules": {
                "import/no-unresolved": [2, { "ignore": ["^@/"] }],
                "n/no-missing-import": "off"
            }
        },
```

jsconfig.json
n-lark: Moved and deleted the root one, thanks!

frontend/src/components/dialogs/device-group-management/AddDeviceToGroupDialog.vue
n-lark: Removed ty

frontend/src/mixins/Dialog.js
n-lark: So for this, I went ahead and removed the mixin and exposed the mixin logic for methods and mounted within each file. See the comment above this one around computed vs setup.

frontend/src/components/DevicesBrowser.vue
cstns: I’m rethinking the `setup` approach; altering our templates just to accommodate refs feels counterproductive and will set us up for failure. 

To be honest, this is my first time pairing Pinia with Options API components, so we're exploring this territory together. After some research, I found that Pinia provides helpers specifically for the Options API—`mapState`, `mapWritableState`, and `mapActions` so the workflow is actually quite similar to Vuex.

here's an example: 
```javascript
export default {
  computed: {
    ...mapState(useStore, ['count']),          // Access via this.count
    ...mapWritableState(useStore, ['name'])    // Access via this.name (supports assignment)
  },
  methods: {
    ...mapActions(useStore, ['increment'])     // Access via this.increment()
  }
}
```

We should use `mapState` for read-only properties, whereas `mapWritableState` enables [read-write state](https://vuejs.org/guide/essentials/computed.html#writable-computed) within the computed object. This is perfect for v-model integration. Meanwhile, `mapActions` functions almost exactly like it did in Vuex.

Since this mirrors the patterns we already know, I think this is our best path forward.

n-lark: Great suggestion thank you! All files in this PR have been updated to use Pinia's Options API helpers (`mapState`, `mapWritableState`, `mapActions`). The setup() + storeToRefs approach has been removed entirely.

### On `setDialogDevices`

The watcher in `DevicesBrowser.vue` needed to update `dialog.is.payload.devices` when the checked devices selection changes. At first glance this looks like a candidate for `mapWritableState`, but neither helper is the right tool here. 

 - `mapWritableState` only enables reassigning the top-level dialog reference (this.dialog = newObject)
 - `mapState` allows mutating nested properties on the returned object without replacing the reference. 
 
Both would technically permit `this.dialog.is.payload.devices = devices`, but doing so from a component watcher bypasses the store's action layer entirely, which the Pinia docs don't explicitly sanction for nested mutations.

The correct approach per Pinia's design is a dedicated store action:

```
// ux-dialog.js
setDialogDevices (devices) {
    this.dialog.is.payload.devices = devices
}

// DevicesBrowser.vue
methods: {
    ...mapActions(useUxDialogStore, ['setDialogDevices'])
},
watch: {
    checkedDevices (devices) {
        if (this.dialog?.is?.payload?.devices) {
            this.setDialogDevices(devices)
        }
    }
}
```

This keeps all state mutations inside the store where they belong, makes the change traceable in Pinia devtools, and is consistent with how `setDisablePrimary` is already handled in this store.

I'll go ahead and apply these patterns across the other PRs & update the docs on the parent issue, great catch!

frontend/src/stores/ux-dialog.js
cstns: awesome, i love it when we remove chunks of code!

cstns:
please address the merge conflicts

PR #6855, opened 2026-03-11
Add Docker dev configuration for EMQX, Postgres, and supporting services

cstns:
## Description

A standardized FlowFuse development setup designed for rapid onboarding.

This project provides a Dockerized environment that enables the full FlowFuse feature set locally. It's an attempt to bridge the gap for new hires by pre-configuring essential services like EMQX, Postgres, and SMTP, ensuring everyone has access to a complete development suite.

### Configuration for the flowforge.local.yml:
```yml
db:
  logging: false
  ## The database type: sqlite|postgres
  type: postgres
  host: 'localhost'
  database: 'flowfuse'
  user: 'ff'
  password: 'super-secret'
  ssl: false
````

```yml
broker:
  ## Internal url used by the platform to connect to the broker
  url: mqtt://localhost:1883
```

```yml
email:
  enabled: true
  debug: true
  smtp:
    host: localhost
    port: 1020
    secure: false
 ```

### Build a local sandbox docker image
Create a Dockerfile on your os
```Dockerfile
FROM debian:bookworm-slim

# Avoid interactive prompts during build
ENV DEBIAN_FRONTEND=noninteractive

# Combine commands to keep the image size down
RUN apt-get update && apt-get install -y \
    sudo \
    curl \
    procps \
    iputils-ping \
    netcat-openbsd \
    iproute2 \
    vim \
    && rm -rf /var/lib/apt/lists/*

# Set the default command
CMD ["bash"]
```
NB:  you can add as many utility libraries as you need to your docker image

build the image locally and name it `sandbox` by running:
```bash
docker build -t sandbox .
```

### Aliasing and using throwaway containers
Define the following aliases in your CLI

```bash
sbox() { 
  # spins up a throwaway container in which you can install the device agent
  docker run -it --rm \
    --dns 172.16.1.1 \
    --dns 8.8.8.8 \
    sandbox
}

sbox_h() { 
  # same as sbox but mounts the container's network to the host os
  docker run -it --rm \
    --dns 172.16.1.1 \
    --dns 8.8.8.8 \
    --net host \
    sandbox
}

sbox_nr() {
  # spins up a throwaway Node-RED instance
  docker run -it --rm \
    --user root \
    --entrypoint /bin/bash \
    --dns 172.16.1.1 \
    --dns 8.8.8.8 \
    --net host \
    nodered/node-red \
    -lc "apk update && apk add --no-cache sudo curl procps iputils netcat-openbsd iproute2 vim bash && exec bash"
}
```

### Usage

With the current proposed setup, only the `sbox_h` alias will work for devices because both emqx and device agent need to be running on the same host for discovery. That means that the device agent will clash with other processes running on your host on the same ports.

Once you exit a container, all data will be discarded.

## Related Issue(s)

N/A

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
There already is a data and emqx at the dev-env level, why are we duplicating them?

https://github.com/flowFuse/dev-env

cstns:
It's not perfect but it's a starting point.

Dev's can spin up local docker containers and run throwaway remote instances in them and get their job done.

Remote instances installed on other hardware that the local host will not work.

cstns:
> There already is a data and emqx at the dev-env level, why are we duplicating them?
> 
> https://github.com/flowFuse/dev-env

I was not aware of it

cstns:
can we run it locally and use it when connecting devices?

I used the emqx config you gave me a while back for this

cstns:
it seems they are the same files

cstns:
It was a nice try but the concept doesn't work on macs because, why would it?

Docker for mac only supports bridge networking mode so `flowfuse` <-> `emqx` <-> `device agent` discovery doesn't work. 

I'll poke around with it over the weekend to attempt a workaround, marking it as draft in the meantime.

hardillb:
@cstns Can we close this? Anything like this should probably go in the dev-env project rather than here?

hardillb:
@cstns bump on closing this and/or moving it to https://github.com/FlowFuse/dev-env

Just trying to clean up stale PRs ahead of release week

Issue #6822, opened 2026-03-06
Pinia Task 5 - context


n-lark:
## Task 5 — `context` (PR 6)

**Gate:** Phase 0 merged
**Vuex module:** `frontend/src/store/modules/context/index.js`
**New file:** `frontend/src/stores/context.js`
**No persistence. No external dependencies for state/actions.**
**Note on `expert` getter:** This getter reads from `account`, `product.assistant`, and `product.expert` — all still on Vuex at this point. Use the bridge pattern to read from Vuex during migration; update the getter in PR 12 when `product-expert` ships.

### 5.1 — Create the Pinia store

State and actions are simple — four properties, four actions. The `expert` getter is the only complex part. Port it with bridge reads for now:

```javascript
// frontend/src/stores/context.js
import { defineStore } from 'pinia'
import store from '../store/index.js' // bridge — remove after all deps migrated

export const useContextStore = defineStore('context', {
    state: () => ({
        route: null,
        instance: null,
        device: null
    }),
    getters: {
        // Bridge version — reads from Vuex for account/assistant/expert data.
        // Update in PR 12 (product-expert) to use Pinia store imports directly.
        expert (state) {
            const rootState = store.state
            const rootGetters = store.getters

            if (!state.route) {
                return {
                    assistantVersion: rootState.product.assistant.version,
                    assistantFeatures: rootState.product.assistant.assistantFeatures,
                    palette: null,
                    debugLog: null,
                    userId: rootState.account?.user?.id || null,
                    teamId: rootState.account?.team?.id || null,
                    teamSlug: rootState.account?.team?.slug || null,
                    instanceId: null,
                    deviceId: null,
                    applicationId: null,
                    isTrialAccount: rootGetters['account/isTrialAccount'] || false,
                    nodeRedVersion: rootState.product.assistant.nodeRedVersion,
                    pageName: null,
                    rawRoute: {},
                    selectedNodes: null,
                    scope: 'ff-app'
                }
            }

            // ... paste full existing getter logic from context/index.js
            // replacing rootState/rootGetters references with store.state / store.getters
        }
    },
    actions: {
        updateRoute (route) { this.route = route },
        setInstance (instance) { this.instance = instance },
        setDevice (device) { this.device = device },
        clearInstance () { this.instance = null }
    }
})
```

### 5.2 — Find and update all consumers

```bash
[...]

PR #6863, opened 2026-03-12
[6822] Pinia Task 5 - context

n-lark:
## Description

See [test plan](https://github.com/FlowFuse/flowfuse/issues/6822#issuecomment-4068291954) and [details](https://github.com/FlowFuse/flowfuse/issues/6822#issue-4034632064).  

⚠️ @cstns just wanted to highlight this observation from the test plan: 

> Note: When you navigate from one instance Expert drawer to another, the messages from the previous conversation persist. This does appear to be the functionality on main, just wanted to note it here.

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/6822

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


cstns:
please address the merge conflicts

frontend/src/stores/context.js
```diff
@@ -0,0 +1,123 @@
...
+        route: null,
+        instance: null,
+        device: null
+    }),
+    getters: {
+        // Bridge version — reads from Vuex for account/assistant/expert data.
+        // Update in PR 12 (product-expert) to use Pinia store imports directly.
+        expert (state) {
+            // Use require() to avoid circular module dependencies at load time.
+            // Falls back to safe defaults when the Vuex store is unavailable (e.g. isolated unit tests).
```
cstns, line 16: a different approach should be taken here, we should not be altering functionality to suit testing but the other way round

frontend/src/stores/_account_bridge.js
```diff
@@ -0,0 +1,25 @@
+// Temporary bridge — reads account data from Vuex during migration.
+// Delete this file after the account stores are migrated to Pinia.
+
+export function useAccountBridge () {
```
n-lark, line 4: I ended up with two of these files `_account-bridge.js` - need to fix

cstns: I only see one file in the diff?

cstns (changes requested):
Looking good!

I don't like the try/catch approach in the stores module though, would like to explore alternatives

frontend/src/stores/_account_bridge.js
n-lark: Oh sorry I fixed and didn't resolve my own comment

frontend/src/stores/context.js
n-lark: So we do remove this within [Task 11 ](https://github.com/FlowFuse/flowfuse/pull/6886/changes#diff-9b8bd48a8b6f1a702564f8bbc45b9e0731f1ad1881de4d42d7b7bbcb26be0e4bR14) it just becomes:

```
      const account = useAccountBridge()
```

This is a temporary workaround since we are split between Vuex and Pinia so I figured it would be okay for now. Lmk your thoughts.

PR #6870, opened 2026-03-14
Ensure getProjectsToUpgrade Matches all Project Settings

hardillb:
## Description

When looking at why only a few projects where updated, noticed that the query only matches settings that also include a restart flag.

This now tests for just bare hour markers.


hardillb:
I have tested this locally.

There is also a possible update to the task that checks for team type changes to add the `restart: false` to the injected setting, but that can wait till Monday

Issue #6826, opened 2026-03-06
Pinia Task 9 - product-expert-ff-agent


n-lark:
## Task 9 — `product-expert-ff-agent` (PR 10)

**Gate:** Phase 0 merged
**Vuex module:** `frontend/src/store/modules/product/expert/ff-agent/index.js`
**New file:** `frontend/src/stores/product-expert-ff-agent.js`
**Persistence:** `context`, `sessionId` → localStorage (cleared on logout)
**No external dependencies — this is a leaf store.**

### 9.1 — Create the Pinia store

```javascript
// frontend/src/stores/product-expert-ff-agent.js
import { defineStore } from 'pinia'
import { markRaw } from 'vue'

export const useProductExpertFfAgentStore = defineStore('product-expert-ff-agent', {
    state: () => ({
        context: null,
        sessionId: null,
        messages: [],
        sessionStartTime: null,
        sessionWarningShown: false,
        sessionExpiredShown: false,
        sessionCheckTimer: null
    }),
    actions: {
        reset () {
            if (this.sessionCheckTimer) clearInterval(this.sessionCheckTimer)
            Object.assign(this, {
                context: null, sessionId: null, messages: [],
                sessionStartTime: null, sessionWarningShown: false,
                sessionExpiredShown: false, sessionCheckTimer: null
            })
        },
        setSessionCheckTimer (timer) {
            this.sessionCheckTimer = markRaw(timer)
        }
    },
    persist: {
        pick: ['context', 'sessionId'],
        storage: localStorage
    }
})
```

**Note on `sessionCheckTimer`:** This is a `setInterval` reference. Use `markRaw()` to prevent Vue from proxying it. The `setSessionCheckTimer` action handles this. The parent `product-expert` store is the only caller.

### 9.2 — No component consumers

This store is only accessed by the parent `product-expert` store. No component updates needed.

### 9.3 — Do not delete the Vuex module yet

The parent `product/expert` Vuex module still references this as a nested module. Delete both in PR 12 when `product-expert` ships.

**Logout bridge:** uncomment `useProductExpertFfAgentStore().$reset()` in the Vuex logout action (Task 0.7).

### 9.4 — Write store tests

```javascript
// frontend/src/tests/stores/product-expert-ff-agent.spec.js
import { setActivePinia, createPinia } from 'pinia'
import { describe, it, expect, beforeEach, vi } from 'vitest'
[...]

PR #6882, opened 2026-03-16
[6826] Pinia Task 9 - product-expert-ff-agent

n-lark:
## Description

See details [here](https://github.com/FlowFuse/flowfuse/issues/6826#issue-4034637416). 

⚠️ This is cut from [`6825-pinia-task-8-product-assistant`](https://github.com/FlowFuse/flowfuse/tree/6825-pinia-task-8-product-assistant) to daisy chain. 

No UI test plan required for this task.

> The `product-expert-ff-agent` Pinia store has no direct component consumers, all access goes through the parent product/expert Vuex module, which remains unchanged. There is nothing new visible in the UI as a result of this PR.
> 
> The only user-facing behavior to verify is:
> 
> - Login: no console errors on page load
> - Logout: `useProductExpertFfAgentStore().$reset()` fires cleanly; no console errors
> 
> All store logic (state initialization, `reset()`, `setSessionCheckTimer()`) is covered by the 18 unit tests in `test/unit/frontend/stores/product-expert-ff-agent.spec.js`.

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/6826

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


cstns:
This is not right, both the `ff-agent` and `operator` stores are used intensively within the `expert` store dynamically, wherever `[state.agentMode]` is used

cstns:
we should also rename the store and agent key while we're at it. We settled on `insights` with product after a couple of variations and we should also switch to it to avoid disconnects moving forward

n-lark:
Hey @cstns 

> This is not right, both the ff-agent and operator stores are used intensively within the expert store dynamically, wherever [state.agentMode] is used

As discussed in slack, we are waiting to convert the vuex store to pinia so we don't do double the work here. This is just a file setup PR. 

> we should also rename the store and agent key while we're at it. We settled on insights with product after a couple of variations and we should also switch to it to avoid disconnects moving forward

Updated the name as below. For the agent key - I'll update that in the [Task 11 PR](https://github.com/FlowFuse/flowfuse/pull/6886/changes#r2975941640). 

```
const useProductExpertInsightsAgentStore = defineStore('product-expert-insights-agent', {
```

cstns:
yes, my bad, I was expecting everything to be done in one go

PR #6839, opened 2026-03-10
Expert/scalability

cstns:
## Description

This task refactors the Expert panel to make it more modular, scalable, and easier to extend. The existing monolithic implementation has been broken down into smaller focused Vue components for message rendering, mode switching, banners, context selection, and answer presentation.

The PR also introduces a more structured rendering model for AI responses, including dedicated components for:

- human, AI, and system messages,
- answer sections such as guides, resources, flows, packages, issues, and suggestions,
- tool/MCP call rendering,
- streaming-aware content presentation.

In addition to the architectural cleanup, the PR consolidates chat input behavior so it interacts more directly with the Expert store, improves loading-state handling, and simplifies scroll management

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/6597

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production
 - [ ] Link to Changelog Entry PR, or note why one is not needed.

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #6894, opened 2026-03-19
Fix expert messages reactivity when switching between agent modes

cstns:
## Description

The message component was using array indices as key props. When switching between different agent modes, the underlying data changed but the indices remained the same (0,1,2...). This caused the framework to incorrectly reuse DOM elements, leading to rendering artifacts and stale state.

I updated the message rendering logic to use each message's unique uuid as the key. This ensures that the UI correctly disposes of old message components and mounts new ones when the agent context changes.

## Related Issue(s)

introduced by https://github.com/FlowFuse/flowfuse/pull/6839

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/expert/components/ExpertMessages.vue
+++ b/frontend/src/components/expert/components/ExpertMessages.vue
@@ -1,7 +1,7 @@
 <template>
     <div ref="messagesWrapper" class="messages-wrapper">
         <ul class="flex flex-col gap-3">
-            <li v-for="(message, index) in messages" :key="index" class="flex flex-col gap-3">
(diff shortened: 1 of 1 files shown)


dimitrieh:
Reported at https://flowfuse.slack.com/archives/C09QS57R8AD/p1773911852313069?thread_ts=1773865648.690409&cid=C09QS57R8AD

Steve-Mcl (approved):
Makes sense but I cannot test in pre-staging so hanging back on merging.

However, I accept it is a small / somewhat inconsequential PR so please feel free to merge.

Issue #6829, opened 2026-03-06
Pinia Task 12 - account-auth


n-lark:
## Task 12 — `account-auth` (PR 13)

**Gate:** All other PRs merged.
**State:** `user`, `pending`, `loginInflight`, `loginError`, `redirectUrlAfterLogin`, `offline`
**Actions:** `login`, `checkIfAuthenticated`, `setUser`, `setOffline`, `setRedirectUrl` (plus internal setters)
**Consumers:** ~25 files (user identity and auth state)

> **Note:** `logout` stays in the Vuex `account` module until Task 15 — it needs `dispatch('$resetState')` to reset remaining Vuex state that hasn't moved to Pinia yet. It moves to Pinia in teardown.

### 12.1 — Create `account-auth.js`

Create the store with its state shape and actions. Port the auth-specific logic out of Vuex:
- `login` — sets `user`, clears `loginInflight`
- `checkIfAuthenticated` — calls `userApi.getUser()`, sets user
- `setUser`, `setOffline`, `setRedirectUrl` — simple setters; consumers call Pinia directly (no Vuex proxy action)

### 12.2 — Update Vuex `checkState` to delegate auth parts to Pinia

Remove the auth commits from `checkState` and replace with Pinia calls:

```js
// Before
commit('login', user)
commit('clearPending')

// After
useAccountAuthStore().login(user)
useAccountAuthStore().clearPending()
```

Remove the auth-related mutations from the Vuex module (`login`, `setUser`, `clearPending`, `setPending`, `setLoginInflight`, `loginFailed`, `setOffline`, `setRedirectUrl`, `sessionExpired`). They are no longer needed.

### 12.3 — Update Vuex `logout` to call Pinia

```js
// Vuex logout after Task 12
async logout({ dispatch }) {
    return userApi.logout()
        .then(() => {
            useAccountAuthStore().$reset()   // auth state cleared via Pinia
            dispatch('$resetState', null, { root: true })  // remaining Vuex state
            // existing Pinia $reset() calls for already-migrated stores...
        })
}
```

### 12.4 — Add persistence

```js
persist: {
    pick: ['redirectUrlAfterLogin'],
    storage: localStorage
}
```

`user`, `pending` etc. are not persisted — always re-fetched on load.

### 12.5 — Update `_account_bridge.js`

The bridge currently reads `user` from Vuex. Update it to read from `useAccountAuthStore()`:

```js
// _account_bridge.js after Task 12
[...]

PR #6945, opened 2026-03-23
[6829] Pinia Task 12 - account-auth 

n-lark:
## Description

See [details](https://github.com/FlowFuse/flowfuse/issues/6829#issue-4034642248) & [test plan](https://github.com/FlowFuse/flowfuse/issues/6829#issuecomment-4114078894)

**Todos**

- [x] Test plan 

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/6829

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/stores/account-auth.js
```diff
@@ -0,0 +1,56 @@
+import { defineStore } from 'pinia'
+
+import userApi from '../api/user.js'
+
+export const useAccountAuthStore = defineStore('account-auth', {
+    state: () => ({
+        user: null,
+        pending: true,
```
cstns: The pending key currently toggles the application's main loading overlay, so I don't think it belongs in the account store.

Since we’re already slicing the account store, this is the perfect time to extract this into a dedicated store. Given it’s purely UI/UX logic, it should be namespaced accordingly. I’m leaning toward ux-loading, but I'm open to other names.

The current key names also don’t quite reflect their purpose. In preparation for upcoming page and component-level loaders, I suggest a more scalable structure within a new `ux-loading` store:
     - `global` (renamed from pending)
     - `network` (renamed from offline)
     - `page` (upcoming)
     - `components` (upcoming)

These aren’t set in stone, but grouping them this way or similar fashion feels more intuitive for the long term.

TL;DR: Let's extract the pending state and ancillary loading functionality into a distinct store under the ux namespace and fight me on the naming

frontend/src/stores/account-auth.js
```diff
@@ -0,0 +1,56 @@
...
+        loginInflight: false,
+        loginError: null,
+        redirectUrlAfterLogin: null,
+        offline: null
```
cstns: see the above comment about the `pending` key, this should also be moved to a distinct store

frontend/src/stores/account-auth.js
```diff
@@ -0,0 +1,56 @@
...
+
+import userApi from '../api/user.js'
+
+export const useAccountAuthStore = defineStore('account-auth', {
+    state: () => ({
+        user: null,
+        pending: true,
+        loginInflight: false,
+        loginError: null,
+        redirectUrlAfterLogin: null,
```
cstns: This should also be moved but not in a dedicated loading store. I feel this still relates to UX/UI so maybe the default ux store would be a good place for it

n-lark: As discussed in slack, leaving as is since all login logic should remain together.

frontend/src/stores/account-auth.js
n-lark: Separated out into new `ux-loading` store with the following name changes: 

- `pending` -> `appLoader`
- `offline` -> `offline` - kept this as is since network is confusing when set to true, implies we are online.

frontend/src/store/modules/account/index.js
```diff
@@ -499,19 +440,19 @@ const actions = {
         const teams = await teamApi.getTeams()
         state.commit('setTeams', teams.teams)
     },
-    async login ({ state, dispatch, commit, getters }, credentials) {
+    async login ({ dispatch }, credentials) {
```
cstns, line 443: Something's off here. It looks like the Vuex mutations were moved into Pinia actions, but the main login/logout methods are still residing in the Vuex account store.

n-lark: Hi since `logout` stays in the Vuex account module until Task 15, it needs `dispatch('$resetState')` to reset remaining Vuex state that hasn't moved to Pinia yet. We fully move this over in [Task 15 here](https://github.com/FlowFuse/flowfuse/pull/6987/changes#diff-f35d85e7b5efff377a1b3e30a56e95742f6e0b73ba8c7715f06b0e70e080af28R58).

cstns:
ok, that's what was unclear to me. GTG then!

Issue #6943, opened 2026-03-23
Pinia Task 14 - account-settings


n-lark:
## Task 14 — `account-settings` (PR 15)

**Gate:** Task 13 merged (`featuresCheck` imports `team` from `account-team`).
**State:** `settings`, `features`
**Getters:** `featuresCheck` (31 properties — imports `team` from `useAccountTeamStore()`)
**Actions:** `refreshSettings`
**Consumers:** ~50 files (almost all via `featuresCheck`)

### 14.2 — Port `featuresCheck` getter

This is the most-consumed getter in the codebase — 31 computed properties. Port it line-by-line. No shortcuts. `team` is now imported from `useAccountTeamStore()` inside the getter body:

```js
featuresCheck () {
    const { team } = useAccountTeamStore()   // cross-store import inside getter
    const { settings, features } = this
    // ... same logic as today
}
```

### 14.3 — Update Vuex `checkState` to delegate settings to Pinia

```js
// Before
commit('setSettings', settings)

// After
useAccountSettingsStore().setSettings(settings)
```

Remove `setSettings` mutation from Vuex.

### 14.4 — Update Vuex `logout`

```js
useAccountAuthStore().$reset()
useAccountTeamStore().$reset()
useAccountSettingsStore().$reset()    // ← add — skipReset: ['settings', 'features'] preserves them
dispatch('$resetState', null, { root: true })
```

### 14.5 — Add persistence

```js
persist: {
    pick: ['settings', 'features'],
    storage: localStorage
},
skipReset: ['settings', 'features']   // clearOnLogout: false equivalent
```

### 14.6 — Update `_account_bridge.js`

```js
import { useAccountSettingsStore } from './account-settings.js'

export function useAccountBridge () {
    return {
        user: useAccountAuthStore().user,
        team: useAccountTeamStore().team,
        featuresCheck: useAccountSettingsStore().featuresCheck   // fully on Pinia now
    }
}
```

At this point the bridge no longer reads from Vuex at all. It can be simplified to a thin re-export if needed.

### 14.7 — Delete `mixins/Features.js` and update its 16 consumers

`mixins/Features.js` is a thin Vuex wrapper with a `// todo` comment saying it should be replaced. Delete it and update each consumer:

1. Remove the mixin import and `mixins: [Features]` entry
2. Add `...mapState(useAccountSettingsStore, ['featuresCheck'])` to computed
[...]

PR #6979, opened 2026-03-26
[6943] Pinia Task 14 - account-settings 

n-lark:
## Description

See details [here](https://github.com/FlowFuse/flowfuse/issues/6943#issue-4123498348) & [test plan here](https://github.com/FlowFuse/flowfuse/issues/6943#issuecomment-4137750679). 

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/6943

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/.github/workflows/branch-deploy.yaml
+++ b/.github/workflows/branch-deploy.yaml
@@ -309,7 +309,7 @@ jobs:
       
       - name: Login to AWS ECR
         id: login-ecr
-        uses: aws-actions/amazon-ecr-login@c962da2960ed15f492addc26fffa274485265950 # v2.0.2
+        uses: aws-actions/amazon-ecr-login@f2e9fc6c2b355c1890b65e6f6f0e2ac3e6e22f78 # v2.1.2
         with:
           mask-password: true
 
@@ -448,7 +448,7 @@ jobs:
         
       - name: Login to AWS ECR
         id: login-ecr
-        uses: aws-actions/amazon-ecr-login@c962da2960ed15f492addc26fffa274485265950 # v2.0.2
+        uses: aws-actions/amazon-ecr-login@f2e9fc6c2b355c1890b65e6f6f0e2ac3e6e22f78 # v2.1.2
         with:
           mask-password: true
 
@@ -470,7 +470,7 @@ jobs:
           aws eks update-kubeconfig --region eu-west-1 --name ${{ secrets.EKS_CLUSTER_NAME }}
 
       - name: Install 1Password CLI
-        uses: 1password/install-cli-action@9a0c9dd934086b7ab1d90115d455bda1c53c2bdb # v2.0.2
+        uses: 1password/install-cli-action@8d006a0d0a4fd505af7f7ce589e7f768385ff5e4 # v3.0.0
         with:
           version: 2.25.0
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -51,7 +51,7 @@ jobs:
     - name: Run forge system tests
       run: npm run cover:system
     - name: Upload coverage reports to Codecov
-      uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # v5.5.2
+      uses: codecov/codecov-action@57e3a136b779b570ffcdbf80b3bdc90e7fab3de2 # v6.0.0
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         flags: backend
--- a/docs/user/expert/index.md
+++ b/docs/user/expert/index.md
@@ -6,7 +6,7 @@ navTitle: FlowFuse Expert
 
 FlowFuse Expert is the AI built into FlowFuse and the Node-RED editor. It is not a generic AI assistant bolted onto the side of your workflow, it understands your flows, your installed nodes, your live data, and your environment in real time.
 
-**FlowFuse Expert is automatically installed and available in all hosted and remote instances running within or connected to FlowFuse**, no manual installation or configuration required.
+**FlowFuse Expert is automatically installed and available in all hosted and remote instances running within or connected to FlowFuse**, no manual installation or configuration required. For self-hosted Enterprise customers, FlowFuse Expert can be enabled on request. [Contact us](https://flowfuse.com/contact-us/) to get it set up on your infrastructure.
 
 ## What FlowFuse Expert Can Do
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -10579,13 +10579,21 @@
             "dev": true
         },
         "node_modules/axios": {
-            "version": "1.13.5",
-            "resolved": "https://registry.npmjs.org/axios/-/axios-1.13.5.tgz",
-            "integrity": "sha512-cz4ur7Vb0xS4/KUN0tPWe44eqxrIu31me+fbang3ijiNscE129POzipJJA6zniq2C/Z6sJCjMimjS8Lc/GAs8Q==",
+            "version": "1.15.0",
+            "resolved": "https://registry.npmjs.org/axios/-/axios-1.15.0.tgz",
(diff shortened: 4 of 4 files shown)


frontend/src/pages/instance/Settings/Security.vue
```diff
@@ -180,10 +178,10 @@ export default {
             }
         }
     },
-    mounted () {
-        this.checkAccess()
+    async mounted () {
+        await this.checkAccess()
         this.getSettings()
-        if (this.isHTTPBearerTokensFeatureEnabled()) {
+        if (this.featuresCheck.isHTTPBearerTokensFeatureEnabled) {
```
n-lark, line 184: `isHTTPBearerTokensFeatureEnabled` was a computed property called with `()`, throwing a TypeError that silently prevented `getTokens()` from ever running.

Fixing the boolean call exposed two issues:

1. Permission race — `checkAccess()` wasn't returning a value so `mounted()` was accidentally bailing out before `getTokens()` could fire. Once fixed, `getTokens()` ran for the first time, including for viewers/members without permission, causing an unhandled 403 that failed Cypress. Fixed by making `checkAccess()` return `false` so `mounted()` bails immediately for unauthorized users. The same permission check was added to the `httpNodeAuth_type` watcher which bypassed `checkAccess()`.

2. Application owner vs team member mismatch — a user can be an application-level owner (passing the client-side `hasPermission('project:edit', { application })` check) but only a team-level viewer/member. The server-side tokens endpoint checks team-level membership and returns 403 in this case. This mismatch pre-existed our changes but was hidden by the TypeError. The `try/catch` handles it gracefully,  the token table stays empty, which is the correct UI state for this permission combination.

The `try/catch` isn't ideal here so we should probably create another ticket to address - the proper fix is to align the client-side permission check for `getTokens()` with the team-level check the server uses. Lmk if we are cool with me raising an issue for this.

frontend/src/stores/account-settings.js
```diff
@@ -0,0 +1,179 @@
...
+            const { team, isTrialAccount } = useAccountTeamStore()
+            const isNotAdmin = user && !user.admin
+            return isNotAdmin &&
+                state.features.billing &&
+                !team?.billing?.unmanaged &&
+                (!isTrialAccount || team?.billing?.trialEnded) &&
+                !team?.type?.properties?.billing?.disabled &&
+                !team?.billing?.active
+        },
+        canCreateTeam (state) {
```
n-lark, line 26: Move this into account-team store maybe - fight with Serban.

frontend/src/components/banners/LicenseBanner.vue
```diff
@@ -18,9 +18,9 @@
 import { ChevronRightIcon, ExclamationCircleIcon } from '@heroicons/vue/outline'
```
n-lark, line 18: Will need to fix issues with deconstructing getters/state due to reactivity issues. This comment is just a reminder.

frontend/src/stores/account-settings.js
```diff
@@ -0,0 +1,181 @@
...
+                // Mqtt Broker
+                isMqttBrokerFeatureEnabledForPlatform: !!state.features?.teamBroker,
+                isMqttBrokerFeatureEnabledForTeam: !!team?.type?.properties?.features?.teamBroker || team?.type?.properties?.enableAllFeatures,
+
+                isExternalMqttBrokerFeatureEnabledForPlatform: !!state.features?.externalBroker,
+
+                // DevOps Pipelines
+                devOpsPipelinesFeatureEnabledForPlatform: !!state.features?.['devops-pipelines'],
+
+                isGitIntegrationFeatureEnabledForPlatform: !!state.features?.gitIntegration,
```
n-lark, line 120: Nick is doing some refactoring here - keep an eye out for merge conflicts or let him know if you merge first.

cstns: more likely to tackle this next week, thanks for pointing it out!

n-lark: this is done

frontend/src/pages/instance/Settings/Security.vue
cstns: I'm not sure if I follow. In this specific component, `isHTTPBearerTokensFeatureEnabled` is accessed via featuresMixin (which is currently being sunset) where it's a method, not called through the account store.

frontend/src/stores/account-settings.js
cstns: seems sensible! not sure if i'm rubbing off on you or are you just messing with me

frontend/src/stores/account-settings.js
cstns: am i missing the refactoring bits?

frontend/src/composables/Components/multi-step-forms/instance/InstanceFormHelper.js
```diff
@@ -20,8 +15,8 @@ export function useInstanceFormHelper () {
     }
 
     const decorateInstanceTypes = (instanceTypes) => {
-        const { team: teamRef } = storeToRefs(useContextStore())
-        const team = teamRef.value
+        const team = useContextStore().team
+        const features = useAccountSettingsStore().features
```
cstns, line 19: why are we ditching storeToRefs in this case?

frontend/src/mixins/Features.js
```diff
@@ -1,91 +0,0 @@
-import { mapGetters, mapState } from 'vuex'
-
-export default {
```
cstns, line 3: thank you for finally closing this chapter

frontend/src/pages/account/Settings.vue
```diff
@@ -299,13 +299,13 @@ export default {
                         // refresh teams
                         return useAccountStore().refreshTeams()
                     }).then(() => {
-                        const { teams } = storeToRefs(useAccountStore())
+                        const teams = useAccountStore().teams
```
cstns, line 302: why ditch storeToRefs?

frontend/src/pages/instance/Settings/Security.vue
cstns: nevermind.we should open a ticket for it indeed

frontend/src/composables/Components/multi-step-forms/instance/InstanceFormHelper.js
cstns: thinking of it, it makes sense, this is a method/action so we don't really care about reactivity just the state of the prop at the moment of execution

cstns: thinking of it, it makes sense, this is a method/action so we don't really care about reactivity just the state of the prop at the moment of execution

frontend/src/pages/account/Settings.vue
cstns: nvm, same reasoning as the previous comment

frontend/src/stores/account-settings.js
n-lark: I resolved it in a merge conflict - I just had to make sure I ported over Nick's changes from Fix gitIntegration feature flag check to include all-feature override #7050

frontend/src/stores/account-settings.js
n-lark: Discussed in slack, leaving in account.

frontend/src/composables/Components/multi-step-forms/instance/InstanceFormHelper.js
n-lark: Yahh and this pattern is cleaner to me.

frontend/src/pages/instance/Settings/Security.vue
n-lark: Made this: https://github.com/FlowFuse/flowfuse/issues/7082

Issue #6944, opened 2026-03-23
Pinia Task 15 - teardown


n-lark:
## Task 15 — Teardown (PR 16)

**Gate:** Tasks 12, 13, 14 merged. Vuex account module is now a near-empty shell — all state has been moved to Pinia, `checkState` delegates everything to Pinia stores.

### 15.1 — Move `checkState` fully to `account-auth`

The Vuex `checkState` action is now mostly delegation calls to Pinia. Move the whole thing into `account-auth.js` as the `checkState` action. Remove it from Vuex.

### 15.2 — Finalise `logout` in `account-auth`

Replace the Vuex `logout` with the full Pinia version. With the `skipReset` plugin, call `$reset()` uniformly on every store:

```javascript
// account-auth.js — final logout
async logout() {
    await userApi.logout()
    useAccountAuthStore().$reset()
    useAccountTeamStore().$reset()
    useAccountSettingsStore().$reset()  // skipReset: ['settings', 'features'] preserves them
    useContextStore().$reset()
    useUxDialogStore().$reset()
    useUxToursStore().$reset()
    useUxNavigationStore().$reset()
    useUxDrawersStore().$reset()
    useProductTablesStore().$reset()
    useProductBrokersStore().$reset()
    useProductAssistantStore().$reset()
    useProductExpertInsightsAgentStore().$reset()
    useProductExpertOperatorAgentStore().$reset()
    useProductExpertStore().$reset()
}
```

### 15.3 — Clean up `api/client.js`

During Task 11, `api/client.js` was changed to lazily `require('../store/index.js')` to break a circular dependency chain:

```
product-expert.js → api/expert.js → api/client.js → store/index.js → account/index.js → routes.js → team/index.vue → product-expert.js
```

With Vuex gone, the cycle is broken permanently. Replace the lazy require with direct Pinia store calls:

```js
// Before (workaround from Task 11):
const store = require('../store/index.js').default
store.dispatch('account/setOffline', true)
store.dispatch('account/logout')

// After:
import { useAccountAuthStore } from '@/stores/account-auth.js'
useAccountAuthStore().setOffline(true)
useAccountAuthStore().logout()
```

### 15.4 — Update route guards

`routes.js` uses `useStore()` from Vuex to read user state in `ensureAdmin` and `ensurePermission`. Replace with Pinia and swap the Vuex store watcher for `watch()` on Pinia reactive state.

### 15.5 — Update `App.vue`

[...]

PR #6987, opened 2026-03-27
[6944] Pinia Task 15 - teardown

n-lark:
## Description

See [details](https://github.com/FlowFuse/flowfuse/issues/6944#issue-4123502061) & [test plan](https://github.com/FlowFuse/flowfuse/issues/6944#issuecomment-4144643687). 

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/6944

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/api/client.js
+++ b/frontend/src/api/client.js
@@ -2,6 +2,9 @@ import axios from 'axios'
 
 import Alerts from '../services/alerts.js'
 
+import { useAccountAuthStore } from '@/stores/account-auth.js'
+import { useUxLoadingStore } from '@/stores/ux-loading.js'
+
 const client = axios.create({
     headers: {
         'Content-Type': 'application/json'
@@ -18,9 +21,6 @@ client.interceptors.response.use(function (response) {
         return Promise.reject(error)
     }
 
-    // Dynamic import breaks the circular dep: client.js → account-auth.js → api/* → client.js
-    const { useAccountAuthStore } = await import('../stores/account-auth.js')
-    const { useUxLoadingStore } = await import('../stores/ux-loading.js')
     if (error.code === 'ERR_NETWORK') {
         // Backend failed to respond
         useUxLoadingStore().setOffline(true)


test/e2e/frontend/cypress/tests-ee/landing/deploy-blueprint.js
```diff
@@ -15,10 +15,8 @@ function checkIfLandedOnLoginPage () {
     cy.intercept('GET', '/api/*/user/').as('getUser')
     cy.wait('@getUser')
 
-    cy.window()
-        .then((win) => {
-            expect(win.location.href).to.match(/^[a-zA-Z]+:\/\/localhost:\d+\/?$/)
-        })
+    // Assert that the login form is visible
+    cy.get('[data-form="login"]').should('be.visible')
```
n-lark, line 19: With Pinia, `checkState()` is fully async, it awaits `getSettings()` and `getUser()` before processing the 401 and pushing to the login route. 

The old `cy.wait('@getUser')` + immediate `win.location.href `assertion resolved as soon as the XHR completed, but the URL hadn't changed yet because Pinia's async chain (store update → router.push) hadn't finished. `cy.get('[data-form="login"]').should('be.visible')` retries automatically, so it correctly waits for the DOM to settle after the redirect.

frontend/src/api/client.js
```diff
@@ -12,24 +12,21 @@ const client = axios.create({
 // Common error handler
```
n-lark, line 15: Will need to fix issues with deconstructing getters/state due to reactivity issues. We will need to sweep the entire codebase for this as well for all the other stores already implemented either in this PR or in another task. 

This comment is just a reminder.

n-lark: this is done skrrrt

frontend/src/stores/product-brokers.js
```diff
@@ -15,7 +15,7 @@ export const useProductBrokersStore = defineStore('product-brokers', {
         hasFfUnsClients: (state) => state.UNS.clients.length > 0,
         hasBrokers: (state) => state.UNS.brokers.length > 0,
         brokerExpandedTopics: (state) => (brokerId) => {
-            const { team } = useAccountBridge()
+            const team = useContextStore().team
```
cstns, line 18: because it's used in a getter, i feel that team should be stored with pinia's storeToRefs helper function to preserve reactivity

frontend/src/stores/ux-navigation.js
```diff
@@ -25,16 +26,19 @@ export const useUxNavigationStore = defineStore('ux-navigation', {
...
-                features: accountFeatures,
-                teamMembership,
-                featuresCheck: features,
-                requiresBilling,
-                isTrialAccountExpired
-            } = useAccountBridge()
+            const contextStore = useContextStore()
+            const team = contextStore.team
+            const teamMembership = contextStore.teamMembership
+            const isTrialAccountExpired = contextStore.isTrialAccountExpired
```
cstns, line 33: I feel like all of these should use pinia's helper to preserve reactivity because they're used in a getter or am i mistaken?

frontend/src/api/client.js
```diff
@@ -12,24 +12,21 @@ const client = axios.create({
...
+    // Dynamic import breaks the circular dep: client.js → account-auth.js → api/* → client.js
+    const { useAccountAuthStore } = await import('../stores/account-auth.js')
+    const { useUxLoadingStore } = await import('../stores/ux-loading.js')
     if (error.code === 'ERR_NETWORK') {
         // Backend failed to respond
         useUxLoadingStore().setOffline(true)
     } else if (error.response && error.response.status === 401 && !useUxLoadingStore().appLoader && !useAccountAuthStore().loginInflight) {
         // 401 when !pending && !loginInflight means the session has expired
-        store.dispatch('account/logout')
+        useAccountAuthStore().logout()
```
cstns, line 29: I know we weren't waiting before, but should we wait now?

frontend/src/stores/account-auth.js
```diff
@@ -36,6 +54,159 @@ export const useAccountAuthStore = defineStore('account-auth', {
         async checkIfAuthenticated () {
             const user = await userApi.getUser()
             this.user = user
+        },
+        async checkState (redirectUrlAfterLogin) {
+            // Lazy require to break circular: account-auth.js → routes.js → Home.vue → ... → account-auth.js
+            const router = require('../routes.js').default
```
cstns, line 61: why not use the useRouter helper?

frontend/src/stores/account-auth.js
```diff
@@ -36,6 +54,159 @@ export const useAccountAuthStore = defineStore('account-auth', {
...
+                    useAccountStore().$reset()
+                    useUxLoadingStore().$reset()
+                    useAccountSettingsStore().$reset()
+                    useUxDialogStore().$reset()
+                    useUxToursStore().$reset()
+                    useUxNavigationStore().$reset()
+                    useUxDrawersStore().$reset()
+                    useUxStore().$reset()
+                    // Lazy require to break circular: account-auth.js → context.js → account-auth.js
+                    require('@/stores/context.js').useContextStore().$reset()
```
cstns: not encountering any circular dependency with a top level import, am i missing anything?

frontend/src/stores/context.js
```diff
@@ -67,34 +67,34 @@ export const useContextStore = defineStore('context', {
             const { matched, redirectedFrom, ...rawRoute } = state.route ?? {}
             let selectedNodes = null
 
-            if (scope === 'immersive' && assistant.selectedNodes.length > 0) {
+            if (scope === 'immersive' && assistantStore.selectedNodes.length > 0) {
                 // Lazy require to avoid circular dependency:
-                // context.js → product-expert.js → product-assistant.js → context.js
+                // context.js → product-expert.js → product-assistantStore.js → context.js
```
cstns: I'm definitely missing something, i can't for the life of me replicate any circular dependency behavior

frontend/src/stores/product-brokers.js
n-lark: So `storeToRefs` is needed when you deconstruct a store to avoid snapshotting the value (e.g., `const { team } = useContextStore()` loses reactivity). 

Here we're reading `.team` directly off the reactive store object, so Vue's reactivity system tracks the access properly whenever this runs inside a reactive context.

frontend/src/stores/ux-navigation.js
n-lark: Same as above, these read properties directly off the reactive store object inside a `getter` (`computed()`), so Vue tracks the access. `storeToRefs` would only be needed if we were deconstructing the store.

frontend/src/api/client.js
n-lark: I don't think so, it seems like the previous behavior was fire-and-forget intentionally. `logout()` ends with `window.location = '/'` in its `.finally()` block, which triggers a full page reload regardless. Do you see another reason why we should here?

frontend/src/stores/account-auth.js
n-lark: `useRouter()` only works inside a component's `setup()` context. It relies on Vue's `inject()` under the hood. Pinia actions run outside that context, so it would return undefined.

frontend/src/stores/account-auth.js
n-lark: So I promise that when I worked on this it was an issue - but now I cannot replicate either. I'm sure someone changed something and I'm totally not losing my mind. Removed.

frontend/src/stores/context.js
n-lark: Removed.

frontend/src/api/client.js
```diff
@@ -12,24 +12,21 @@ const client = axios.create({
...
     }
 
-    // This is an error response from our own API (or failure to reach it)
-    // Lazy require to avoid circular dependency:
-    // api/client.js → stores/account-auth.js → api/user.js → api/client.js
-    const { useAccountAuthStore } = require('../stores/account-auth.js')
-    const { useUxLoadingStore } = require('../stores/ux-loading.js')
-    const store = require('../store/index.js').default
+    // Dynamic import breaks the circular dep: client.js → account-auth.js → api/* → client.js
+    const { useAccountAuthStore } = await import('../stores/account-auth.js')
```
cstns: I was sure I mentioned these imports too yesterday, must not have pressed send.. These do not pose a problem on my end. I'm taking the liberty to extract them as top level imports for the sake of brevity and get this in asap

PR #6994, opened 2026-03-30
Show error if user roles are managed by SSO

### LINKED ISSUE #6958: SSO enabled users' roles unchangeable in teams
### Current Behavior

When a SSO enabled user is in a team with a specific role it is not possible to change the role. There is no UI feedback, but the reqest returns with a 400. It is also not possible to remove the user, nor to re-invite them.

This makes user administration on team level practically unusable, e.g. promoting a member to owner when the current owner leaves.

The only way we found to fix this is to delete the user as flowfuse admin and re-create it with a new login, but that is not reasonable for a user that has multiple teams and needs to be re-added everywhere.

Image

_Modification fails_

Image

_Deletion fails_

Image

_re-invite fails_

### Expected Behavior

- If there is an issue with the underlying request the UI should show it and not fail silently
- There should be a possibilty to change roles inside a team for SSO users

### Steps To Reproduce

- Add a team member with SSO to the team in any role
- Open dev tools
- Try to change the role or remove the user from the team

### Environment

- FlowFuse version: 2.28.0
- Node.js version: 20.20.1
- npm version: 10.8.2
- Platform/OS: Windows
- Browser: Firefox 148.0.2 (64-Bit)

hardillb:
fixes #6958

## Description

Currently if you try to change the role of a Team member who's role is managed by a SSO group it silently fails.

This adds an alert toast to explain why

Screenshot From 2026-03-30 16-27-38

## Related Issue(s)

#6958

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/views/User.js
+++ b/forge/db/views/User.js
@@ -100,7 +100,7 @@ module.exports = function (app) {
 
     async function ssoManaged (users, team) {
         const list = {}
-        for (u of users) {
+        for (const u of users) {
             list[u.hashid] = await app.sso.isUserMembershipManaged(u, team)
         }
         return list
--- a/frontend/src/components/blueprints/BlueprintTile.vue
+++ b/frontend/src/components/blueprints/BlueprintTile.vue
@@ -17,7 +17,7 @@
         <div class="ff-blueprint-tile--info">
             <label>{{ blueprint.name }}</label>
             <p v-if="blueprint.description" :title="blueprint.description">
-                <ff-markdown-viewer :content="blueprint.description"/>
+                <ff-markdown-viewer :content="blueprint.description" />
             </p>
         </div>
         <div class="ff-blueprint-tile--actions justify-between">
--- a/frontend/src/pages/team/Members/General.vue
+++ b/frontend/src/pages/team/Members/General.vue
@@ -19,16 +19,20 @@
             </template>
             <template v-if="canEditUser" #context-menu="{row}">
                 <ff-kebab-item
-                    v-if="(hasPermission('team:user:change-role') && !requiresBilling) || isAdminUser"
+                    v-if="((hasPermission('team:user:change-role') && !requiresBilling) || isAdminUser) && !ssoManaged({row})"
                     data-action="member-change-role"
                     label="Change Role" @click="changeRoleDialog(row)"
                 />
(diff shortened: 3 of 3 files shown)


hardillb:
reverting to draft as I've not actually read @knolleary's comments on the issue about doing it properly (which explains why he up'd it to 2 hours)

hardillb:
@knolleary I was trying to knock this out quickly without looking too closely at your earlier suggested route.


> This will require:
> 
>    - add a flag to the User object to indicate whether its team memberships are SSO managed. We already have the SSOGroups property that is only set if the option is enabled to expose membership to dashboard logins. Can we reuse that somehow whilst retaining the exposeGroups option?
>    - add the flag on the api schema so the frontend receives it
>    - update UX for changing role to prevent changes, with appropriate hint on why its disabled
> 

I've now gone back to have another look and there is a small niggle with your approach. `SSOGroups` contains the raw group names, not the team names e.g. on my test system

```
{
  "email":"ben@example.com",
   "email_verified":true,
   "sso_enabled":true,
   "mfa_enabled":false,
   "SSOGroups":[
    "FlowFuse",
    "ff-development-owner",
    "minio",
    "test_ff-development-owner_test"
  ],
  "id":"r3VG59G85J",
  "username":"ben-example_come",
  "name":"ben",
  "avatar":"http://localhost:3000/avatar/YmVu",
  "admin":false,
  "createdAt":"2024-07-22T13:32:03.589Z",
  "suspended":false
}
```

So to get SSO managed teams we would need to filter this list based on the SSO prefix/suffix settings to get just the team names (and remove the extra groups)

Any thoughts?

hardillb:
I see 2 approaches here:

1. Provide function to parse existing groups based on SSO config prefix/suffix lengths
2. Add the SSO Teams list to the User object

1 means we don't need to add anything else to the user object, but does mean always populating the SSOGroups array for a SSO user

2. adds more to the User object, but is probably the cleanest

hardillb:
OK, have worked out how to add a `ssoManaged` flag to a teams membership list.

Will need some help getting the UI to play ball with that flag now.

hardillb:
image

vs

image

PR #7003, opened 2026-03-31
Bump sqlite3 to 6.0.1 to pick up tar fixes

hardillb:
## Description

Upgrade to fix tar dependencies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -68,7 +68,7 @@
                 "random-words": "~2.0.0",
                 "semver": "~7.6.0",
                 "sequelize": "^6.33.0",
-                "sqlite3": "^5.1.6",
+                "sqlite3": "^6.0.1",
                 "stripe": "^8.222.0",
                 "tar-stream": "^3.1.7",
                 "uuid": "^9.0.0",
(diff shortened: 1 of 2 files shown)


knolleary:
There is an open issue against this version of sqlite3 that it includes a build of sqlite that has since been withdrawn due to 'backward compatibility' issues.

https://github.com/TryGhost/node-sqlite3/issues/1859

We need to due diligence on what those issues are and if they are relevant to us.

hardillb:
I've found the following in the sqlite forum

Why 3.52.0 Was Withdrawn (https://sqlite.org/forum/forumpost/1996d7e81ea968274ad2340210024d516223a7fb4dd9387aaa2493f1c982467b)

> 
> Version 3.52/3.53 contains enhancements to the internal text→float conversion routines used by SQLite. These enhancements cause the results to be more accurate in some corner cases. But "more accurate" means "different", and this lead us to discover the concept of [stale expression indexes](https://sqlite.org/draft/staleexpridx.html). The stale expression index problem has always existed. However, it had never before come to the attention of the developers; it is not something we had previously ever thought about.
> 
> Stale expression indexes are not often an issue, it seems. There were improvements to the accuracy of text→float conversions going from version 3.41 to 3.42 on non-x86/x64 platforms and nobody noticed. But someone did notice the change going from 3.51 to 3.52, and so we withdrew 3.52 in order to investigate the problem and devise a solution.
> 
> The solution is three-fold:
> 
>    - Stale expression indexes are now "[self-healing](http://sqlite.org/draft/staleexpridx.html#selfheal)".
> 
>    - There is a new "[REINDEX EXPRESSIONS](https://sqlite.org/draft/lang_reindex.html)" SQL statement that rebuilds only expression indexes, which is guaranteed to thoroughly cleanse a database of any stale expression indexes.
> 
>    - The "PRAGMA integrity_check" command has been enhanced to detect stale expression indexes and report the problem as such, rather than proclaiming that the database is corrupt.

If I've understood that all properly, we should be fine as I don't think we do any test -> float queries

hardillb:
I'm also not convinced we will get a 6.0.2 of sqlite3, the big THIS is DEPRECATED banner is flying on sqlite3 and there has been no comment on that issue having been open for 3 weeks

PR #7004, opened 2026-03-31
Expert/mqtt FrontEnd scaffolding

### LINKED ISSUE #6962: Scaffolding to support MQTT
Parent: #6923

Create an MQTT service that handles all MQTT client lifecycles.

cstns:
## Description

This PR scaffolds frontend MQTT support and restructures service startup/cleanup into a centralized orchestrator, while updating dependent code and adding tests.

- Replaces the old service.factory with a new services.orchestrator that boots and tears down frontend services more cleanly.
- Adds a new, substantial mqtt.service.js that manages MQTT connections, publishing, subscribing, lifecycle cleanup, payload serialization, and validation.
- Renames the messaging service class/file to post-message.service.js and updates consumers accordingly.
- Updates app bootstrap (main.js) to use the new orchestrator and initialize services through it.
- Adjusts a drawer helper and product assistant store to reference the new service setup.
- Adds @types/mqtt to dev dependencies.
- Includes extensive unit tests for the new MQTT service and related service wiring.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/6962

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/services/bootstrap.service.ts
```diff
@@ -1,40 +1,39 @@
...
 import { useAccountSettingsStore } from '@/stores/account-settings.js'
 import { useContextStore } from '@/stores/context.js'
 
+export interface BootstrapServiceI extends AppService {
+    init(): Promise<void>
+    destroy(): Promise<void>
+    setupReadyPromise(): void
+    waitForAppMount(): Promise<void>
+    waitForRouterReady(): Promise<void>
+    checkUser(): Promise<unknown>
```
n-lark, line 15: Lets not use `<unknown>` here please - could probably be `Promise<void>` from looking at the implementation.

frontend/src/services/service.contract.ts
```diff
@@ -0,0 +1,52 @@
...
+        name,
+        app,
+        router,
+        services
+    }) {
+        this.$name = name
+        this.$app = app
+        this.$router = router
+        this.$services = services
+    }
```
n-lark, line 43: `{ name: string } & CreateServiceOptions` - this is an implicit any.

frontend/src/services/service.contract.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Lightweight base class for shared lifecycle surface.
+ */
+export abstract class BaseService {
+    protected $name: string
+
+    protected $app: App = null
+
+    protected $router: Router = null
+
+    protected $services: ServiceInstances = null
```
n-lark, line 31:
```
protected $app: Maybe<App> = null          
protected $router: Maybe<Router> = null    
protected $services: Maybe<ServiceInstances> = null
```

These could utilize the `Maybe<T>` - that should probably be an exported shared utility type. 

```
export type Maybe<T> = T | null | undefined
```

frontend/src/services/bootstrap.service.ts
```diff
@@ -1,40 +1,39 @@
...
-        app,
-        router,
-        services = {}
-    }) {
-        this.$app = app
-        this.$router = router
-        this.$services = services
+class BootstrapService extends BaseService implements BootstrapServiceI {
+    protected isReady: boolean = false
+    protected readyPromise: Promise<never> | null = null
```
n-lark, line 27: `Promise<void>`?

frontend/src/services/bootstrap.service.ts
```diff
@@ -116,15 +117,7 @@ class BootstrapService {
 
 let BootstrapServiceInstance = null
```
n-lark, line 118: `Maybe<BootstrapService>`

frontend/src/services/mqtt.service.ts
```diff
@@ -0,0 +1,991 @@
...
+
+    bindManagedListeners (managed: ManagedMqttClient, client: MqttClient) {
+        const register = (eventName: keyof MqttClientEventCallbacks, handler: (...args: any[]) => void) => {
+            const wrapped = (...args: any[]) => {
+                if (managed.destroyed || this.$destroyed || managed.client !== client) return
+                handler(...args)
+            }
+            client.on(eventName, wrapped)
+            managed.listeners.add(() => client.off(eventName, wrapped))
+        }
```
n-lark, line 557: Two `any`s here - could we benefit from a generic in this? 

```
bindManagedListeners (managed: ManagedMqttClient, client: MqttClient) {
    const register = <K extends keyof MqttClientEventCallbacks>(
        eventName: K,
        handler: MqttClientEventCallbacks[K]
    ) => {
        const wrapped = ((...args: Parameters<MqttClientEventCallbacks[K]>) => {
            if (managed.destroyed || this.$destroyed || managed.client !== client) return
            ;(handler as Function)(...args)
        }) as MqttClientEventCallbacks[K]

        client.on(eventName, wrapped)
        managed.listeners.add(() => client.off(eventName, wrapped))
    }

    register('connect', async () => { /* ... */ })
    register('close', () => { /* ... */ })
    register('error', (error: Error) => { /* ... */ })
    register('message', (topic: string, message: Buffer, packet: IPublishPacket) => { /* ... */ })
}
```

frontend/src/services/mqtt.service.ts
```diff
@@ -0,0 +1,991 @@
...
+    onConnect?: (client: MqttClient) => void
+    onClose?: (client: MqttClient) => void
+    onOffline?: (client: MqttClient) => void
+    onError?: (error: Error, client: MqttClient | null) => void
+    onMessage?: (topic: string, message: Buffer, packet: IPublishPacket, client: MqttClient) => void
+}
+
+export interface MqttPublishRequest {
+    topic: string
+    payload: unknown
```
n-lark, line 48: Could do a union here: 


`type MqttPayload = string | BinaryPayload | Record<string, unknown> | unknown[]`

frontend/src/services/bootstrap.service.ts
Steve-Mcl:
```suggestion
    checkUser(): Promise<void>
```

frontend/src/services/post-message.service.ts
```diff
@@ -18,43 +20,67 @@ const ACTIONS_FLOWFUSE_EXPERT = {
...
+        payload: unknown
+    }
+)
+
+export interface PostMessageServiceI extends AppService {
+    init(): void
+    setupMessageHandlers(): void
+    destroy(): Promise<void>
+    handleFlowFuseExpertMessage(event: MessageEvent<any>): Promise<void>
+    handleAssistantMessage(event: MessageEvent<any>): Promise<void>
```
n-lark, line 53: `MessageEvent<PostMessagePayload>` for both `event: MessageEvent<any>`

frontend/src/services/post-message.service.ts
```diff
@@ -18,43 +20,67 @@ const ACTIONS_FLOWFUSE_EXPERT = {
...
+    sendReadyMessage(): void
+    setExpertContext(payload: unknown): void
```
n-lark, line 55: `{data: object, sessionId: string}` is what its called with - can you use that?

frontend/src/services/post-message.service.ts
```diff
@@ -18,43 +20,67 @@ const ACTIONS_FLOWFUSE_EXPERT = {
...
-     * @type {import('vue').App} - Vue app instance
-     */
-    $app
-
-    /**
-     * @type {import('vue-router').Router} - Vue router instance
-     */
-    $router
+class PostMessageService extends BaseService implements PostMessageServiceI {
+    protected $onMessage: { (event: any): Promise<void>; (this: Window, ev: MessageEvent<any>): any; (this: Window, ev: MessageEvent<any>): any }
```
n-lark, line 78: `Anys` - I think this can just be:

```
protected $onMessage: ((event: MessageEvent<PostMessagePayload>) => Promise<void>) | null = null
```

frontend/src/services/post-message.service.ts
```diff
@@ -91,13 +117,22 @@ class MessagingService {
...
+    }
+
+    async destroy () {
+        if (this.$onMessage) {
+            window.removeEventListener('message', this.$onMessage)
+            this.$onMessage = null
+        }
     }
 
     async handleFlowFuseExpertMessage (event) {
```
n-lark, line 132: `event: MessageEvent<PostMessagePayload>`

frontend/src/services/post-message.service.ts
n-lark, line 142: `event: MessageEvent<PostMessagePayload>`

n-lark, line 247: `Maybe<PostMessageService>`

frontend/src/services/service.orchestrator.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+class ServiceOrchestrator {
+    $serviceInstances: ServiceInstances = Object.fromEntries(SERVICE_REGISTRY.map(service => [service.key, null])) as ServiceInstances
+
+    /**
+     * @type {import('vue').App} - Vue app instance
+     */
+    $app: App = null
+
+    $router: Router = null
```
n-lark, line 17:
```
  $app: Maybe<App> = null

   $router: Maybe<Router> = null
```

frontend/src/services/service.orchestrator.ts
```diff
@@ -0,0 +1,108 @@
...
+                import.meta.hot.dispose(async () => {
+                    await this.dispose()
+                })
+            }
+        }
+    }
+}
+
+// Create a singleton factory instance
+let orchestratorInstance = null
```
n-lark, line 92: `Maybe<ServiceOrchestrator>`

frontend/src/services/service.orchestrator.ts
```diff
@@ -0,0 +1,108 @@
...
+
+        await this.$serviceInstances.bootstrap.init()
+
+        app.provide('$services', this.$serviceInstances)
+
+        return this.$serviceInstances
+    }
+
+    async dispose () {
+        for (const service of Object.keys(this.$serviceInstances)) {
```
n-lark, line 57: So `Object.keys` loses the type inference and returns `Array<string>` - can just cast here. Will yell at you in strict mode. 

```
for (const service of Object.keys(this.$serviceInstances) as Array<keyof ServiceInstances>) {
    await this.$serviceInstances[service]?.destroy?.()
    //  is now BootstrapServiceI | PostMessageServiceI | MqttServiceI | null
    this.$serviceInstances[service] = null
}

```

frontend/src/services/service.registry.ts
```diff
@@ -0,0 +1,15 @@
...
+    bootstrap: BootstrapServiceI | null
+    postMessage: PostMessageServiceI | null
+    mqtt: MqttServiceI | null
+}
+
+export default [
+    { key: 'bootstrap', create: createBootstrapService, requiredLifecycle: ['init', 'destroy'] },
+    { key: 'postMessage', create: createMessagingService, requiredLifecycle: ['destroy'] },
+    { key: 'mqtt', create: createMqttService, requiredLifecycle: ['destroy'] }
+]
```
n-lark, line 15: You do get an implicit type here but if you want the key to not be `string` could do this: 

```
export default [
    { key: 'bootstrap' as const, create: createBootstrapService, requiredLifecycle: ['init', 'destroy'] as const },
    { key: 'postMessage' as const, create: createMessagingService, requiredLifecycle: ['destroy'] as const },
    { key: 'mqtt' as const, create: createMqttService, requiredLifecycle: ['destroy'] as const }
]
```

Without this, key is string and can't safely index `ServiceInstances`:

```
for (const serviceDefinition of SERVICE_REGISTRY) {
    this.$serviceInstances[serviceDefinition.key] = serviceDefinition.create({ ... })
    //  .key is a string so cannot index errrr
}
```

n-lark (changes requested):
Hey! This is just the typescript stuff I found I will also poke this with a stick - great work so excited to see .ts files in the codebase! 

Update - I poked with a stick and didn't find anything unusual.

frontend/src/components/immersive-editor/EditorDrawer.vue
```diff
@@ -119,7 +119,7 @@ import { storeToRefs } from 'pinia'
 import { computed, nextTick, onMounted, onUnmounted, ref, watch } from 'vue'
 import { useRoute } from 'vue-router'
 
-import { getServiceFactory } from '../../services/service.factory.js'
+import getServiceOrchestrator from '../../services/service.orchestrator'
```
Steve-Mcl, line 122: shouldnt we be explicit & include the file.ext

(I forget reasons why we needed to do this, but for now, just consistency)


---

Edit, this is a ~~compiled~~ transpiled thing isnt it :( Now I recall why i dislike TS!

frontend/src/types/services/bootstrap.types.ts
```diff
@@ -0,0 +1,13 @@
+import type { AppService } from '@/types'
+
+export interface BootstrapServiceI extends AppService {
```
Steve-Mcl, line 3: not keen on `xxxxI` but wont block

frontend/src/types/services/mqtt.types.ts
```diff
@@ -0,0 +1,109 @@
...
+    onMessage?: (topic: string, message: Buffer, packet: IPublishPacket, client: MqttClient) => void
+}
+
+export interface MqttPublishRequest {
+    topic: string
+    payload: MqttPayload
+    qos?: 0 | 1 | 2
+    retain?: boolean
+    onError?: ((error: Error) => void) | null
+    correlationData?: string | BinaryPayload | null
```
Steve-Mcl: I think this should be opinionated. Under the hood it any well be a uint8array (or Buffer) but the content is always a string.
```suggestion
    correlationData?: string | null
```

frontend/src/types/services/mqtt.types.ts
```diff
@@ -0,0 +1,109 @@
...
+    userProperties?: Record<string, string | string[]> | null
```
Steve-Mcl, line 53: Why is string array an option here?

Steve-Mcl (commented):
really close - some superficial feedback

frontend/src/types/services/mqtt.types.ts
Steve-Mcl: Obviously applying this ^ would mean changes to surrounding code, but perhaps worth it for a clean implementation?

frontend/src/types/services/bootstrap.types.ts
cstns: I'm not either, we should make a decision on how to move forward with them. we'll let it slip for now to get this in and circle back to it

cc @n-lark

frontend/src/types/services/mqtt.types.ts
cstns: done

frontend/src/types/services/mqtt.types.ts
cstns: v5 user properties can contain multiple values for the same property name

Steve-Mcl: yes, that is true, but never looed into how they (MQTT JS lib) parses records with same name - now i know!

frontend/src/types/services/bootstrap.types.ts
n-lark: `IBootstrapService` is what I would prefer here.

frontend/src/services/mqtt.service.ts
```diff
@@ -0,0 +1,911 @@
...
+            try {
+                return await this.connectManagedClient(managed, false)
+            } catch (error) {
+                this.$clients.delete(key)
+                throw error
+            }
+        })
+    }
+
+    async publishMessage (key: string, options?: MqttPublishRequest): Promise<void> {
```
n-lark, line 165: Nit - we should make this not optional for `options?` just a mismatch with the fn signature below in the types file:   

```
publishMessage(key: string, options: MqttPublishRequest): Promise<void>
```

Then we can drop the `options!`

frontend/src/services/service.orchestrator.ts
```diff
@@ -0,0 +1,111 @@
...
+ */
+class ServiceOrchestrator {
+    $serviceInstances: ServiceInstances = Object.fromEntries(SERVICE_REGISTRY.map(service => [service.key, null])) as ServiceInstances
+
+    /**
+     * @type {import('vue').App} - Vue app instance
+     */
+    $app: Maybe<App> = null
+
+    $router: Maybe<Router> = null
```
n-lark, line 20: So `$app`, `$router`, and `$serviceInstances.bootstrap` are now correctly typed as nullable (`Maybe<App>`, `BootstrapServiceI | null`), but the access sites here treat them as non-null, see lines 45, 52, 54, 79-80. Under strict null checks these are compile errors so we should guard with `if (!this.thing) throw or whatever`

Steve-Mcl (approved):

frontend/src/services/service.orchestrator.ts
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Create all services
+     * @returns {Promise<ServiceInstances>} Promise that resolves when all services are booted
+     */
+    async bootServices (): Promise<ServiceInstances> {
+        const app = this.$app
+        const router = this.$router
+
+        for (const serviceDefinition of SERVICE_REGISTRY) {
+            ;(this.$serviceInstances as Record<string, unknown>)[serviceDefinition.key] = serviceDefinition.create({
```
n-lark, line 45: Dislike the unknown here, could we do like: 

```
 type ServiceDefinition<K extends keyof ServiceInstances> = {
     key: K
     create: (opts: CreateServiceOptions) => NonNullable<ServiceInstances[K]>
     requiredLifecycle: ReadonlyArray<'init' | 'destroy'>
 }
 ``` 

Then `serviceDefinition.key` stays as `'bootstrap' | 'postMessage' | 'mqtt'` and indexes `ServiceInstances` directly without the cast.

File: frontend/src/services/post-message.service.ts [hunk 23]
```diff
@@ -91,20 +103,30 @@ class MessagingService {
...
     }
 
-    async handleFlowFuseExpertMessage (event) {
-        switch (event.data.action) {
+    async handleFlowFuseExpertMessage (event: MessageEvent<PostMessagePayload>) {
+        const data = event.data as PostMessageBase & { action: string, payload?: { data: object, sessionId: string } }
+        switch (data.action) {
         case sourceActions[DATA_SOURCE_FLOWFUSE_WEBSITE].SET_CONTEXT:
-            await this.setExpertContext(event.data.payload)
+            this.setExpertContext(data.payload)
```
n-lark, line 122: Two things:

1. `PostMessagePayload` is already a discriminated union and setting action as string widens it, `SetContextAction`
2. `data.payload` is optional after the cast but `setExpertContext` requires it. 

```
if (event.data?.action === ACTIONS_FLOWFUSE_EXPERT.SET_CONTEXT) {
    const payload = event.data.payload as { data: object, sessionId: string } | undefined
    if (payload) this.setExpertContext(payload)
}
```

Or tighten `payload: unknown` in `PostMessagePayload` to the real shape so no cast is needed.

n-lark (approved):
Nothing here is major up to you - great work!

cstns:
we can circle back when we have a clear way on how to name intarfaces too, see https://github.com/FlowFuse/flowfuse/pull/7004#discussion_r3167463750

PR #7033, opened 2026-04-07
feat: visual flow diff viewer for snapshot comparison

### LINKED ISSUE #7025: feat: visual flow diff viewer for snapshot comparison
## Summary

Replaces the basic snapshot comparison UI with a full visual diff viewer that combines the \`@flowfuse/flow-renderer\` canvas with a property-level sidebar diff panel.

## What's new

**Canvas + navigation**
- Flow canvas is always visible; a collapsible property diff sidebar appears on the right when a comparison is active
- Prev / Next navigation bar to step through each changed/added/deleted node one at a time
- The canvas highlights and scrolls to the current node automatically, adjusting the opacity slider to show the relevant snapshot layer (added → 90%, deleted → 10%)
- Sidebar is resizable via a drag handle

**Property diff sidebar**
- **Compact row** for structural props (`name`, `type`, `wires`, `position`, `z`, `disabled`): shows old → new inline
- **Git-style diff** for all other props (multiline strings like function code, JSON, templates): collapsible with `+N / -N` line counts, collapsed unchanged sections, click to expand
- Wires resolved to node names, tabs resolved to tab labels, `disabled` shown as `enabled`/`disabled`

**Correctness fixes vs previous implementation**
- Comparison order was reversed in the old code (`[newer, older]`); flipped to `[older, newer]` so added/deleted is correctly oriented
- Flow renderer CSS scope was derived from Tailwind utility classes (e.g. `flex-1`), leaking `svg { width: 100%; height: 100% }` globally and making all page icons huge — fixed with an explicit `ff-flow-compare-view` scope
- Renderer `highlight()` was called without `layerNo`, relying on internal diffType detection that broke for tab entries (`diffType: 'tab'` → default `layerNo = 0` → always 10%) — now passes explicit layerNo from our own diff classification

andypalmi:
Closes #7025

## Summary

- Replaces the basic snapshot comparison UI with a full visual diff viewer combining the `@flowfuse/flow-renderer` canvas with a property-level sidebar diff panel
- New `SnapshotDiffChangePanel` component handles per-property diffs (compact inline mode for structural props, git-style line diff for everything else)

## Changes

**Canvas + navigation**
- Prev / Next nav bar to step through each changed/added/deleted node — buttons are grouped side-by-side with the counter between them
- ← → arrow key shortcuts for keyboard navigation
- Canvas highlights and scrolls to the current node, adjusting the opacity slider (added → 90%, deleted → 10%)
- Resizable sidebar via drag handle

**Property diff sidebar**
- Compact inline row for structural props (`name`, `type`, `wires`, `position`, `z`, `disabled`)
- Git-style collapsible diff for multiline props (function code, templates, JSON)
- Wires resolved to node names, tabs to tab labels, `disabled` shown as `enabled`/`disabled`

**Bug fixes**
- Comparison order was reversed (`[newer, older]`) — flipped to `[older, newer]`
- Renderer CSS leaked `svg { width: 100%; height: 100% }` globally via Tailwind class scope — fixed with dedicated `ff-flow-compare-view` scope
- `highlight()` called without `layerNo` caused tab entries to always show 10% — now passes explicit layerNo from our own diff classification
- Slider animation dead zone: after landing at 10 or 90, navigating in the opposite direction left the slider stuck — fixed by skipping the animation when already at target
- Tab highlight in canvas did not navigate to newly added tabs — fixed by proxying to a child node's highlight which triggers the tab click as a side effect
- Nav bar showed raw UUID for nodes without a name/label (e.g. inject) — now falls back to node type
- Slider input had a white background rectangle — removed with `background: transparent`
- Slider opacity animation was a slow JS stepping loop (up to 900ms) — replaced with a direct value set + CSS `transition: opacity 250ms` on SVG layers
- Spurious `→ Flow 1` tab-move badge on added/deleted nodes — suppressed for non-`changed` diffType
- `name` property showed as `—` for added nodes — skipped from propChanges since it's already in the nav header

## Test plan

- [ ] Open snapshot compare dialog on an instance with multiple snapshots
- [ ] Select an older snapshot to compare against — canvas renders, nav bar appears
- [ ] Step through Prev/Next — canvas highlights correct node, sidebar shows property changes
- [ ] Prev and Next buttons are adjacent; ← → arrow keys also navigate
- [ ] Added nodes show slider at ~90% (new snapshot layer visible); deleted at ~10%
- [ ] Navigate between added and deleted nodes — slider switches direction smoothly each time
- [ ] New tab in newer snapshot: navigating to it scrolls canvas to show the tab
- [ ] Multiline property (e.g. function node `func`) shows git-style diff, collapsible
- [ ] Drag the sidebar resize handle — width changes within bounds
- [ ] Page icons (header, version history) are not enlarged after opening the dialog
- [ ] Slider input has no white background rectangle

### DIFF
--- a/frontend/src/components/dialogs/AssetCompareDialog.vue
+++ b/frontend/src/components/dialogs/AssetCompareDialog.vue
@@ -147,7 +147,7 @@ export default {
             loading: false,
             nodeMap: {},
             currentGroupIndex: 0,
-            sidebarWidth: 570,
+            sidebarWidth: 380,
             resizing: false,
             resizeStartX: 0,
             resizeStartWidth: 0


Steve-Mcl:
Quick look on pre-staging - one thing I did note (it may have always been this way) but the compare dropdown includes the currently selected snapshot (and so if selected by the user, there are (obviously) zero differences). IMO, the selected snapshot should be excluded from the "compare to" drop-down.

Happy for this to be a follow up issue!

image

andypalmi:
Hi @Steve-Mcl thank you for testing it out.

I noticed that too and I am thinking we could include it in this pull request as it seems like an easy fix!

I will work on it tomorrow!

andypalmi:
@Steve-Mcl I managed to exclude the currently selected snapshot from the target list. Thank you for catching that! I also caught the issue of positions being undefined when only one coordinate changes instead of both. Now the unchanged coordinate is shown instead of `undefined`

frontend/src/components/dialogs/SnapshotDiffChangePanel.vue
```diff
@@ -0,0 +1,249 @@
...
+                    <template v-for="(line, i) in lines" :key="i">
+                        <!-- Collapsed unchanged section -->
+                        <div
+                            v-if="line.type === 'collapsed'"
+                            class="flex leading-5 text-blue-600 bg-blue-50 cursor-pointer hover:bg-blue-100 select-none border-y border-blue-100"
+                            @click="expandSection(i)"
+                        >
+                            <span class="line-num border-r border-blue-200 text-blue-400" />
+                            <span class="line-num border-r border-blue-200 text-blue-400" />
+                            <span class="px-3 flex-1 text-center">&#8597; {{ line.count }} unchanged lines</span>
```
n-lark: Nitpick but this will render "1 unchanged lines", can pluralize with: 

`${line.count} unchanged line${line.count === 1 ? '' : 's'}`

frontend/src/components/dialogs/AssetCompareDialog.vue
```diff
@@ -99,55 +153,309 @@ export default {
...
-            const flowRenderer = new FlowRenderer()
-            flowRenderer.renderFlows(this.flow, {
-                container: this.$refs.compareViewer
-            })
-        },
-        async renderComparison (snapshotId) {
+        async renderComparison () {
             this.cleanup()
+            this.hasCompared = false
             const compareSnapshot = await SnapshotsApi.getFullSnapshot(this.compareSnapshot)
```
n-lark: Do we need some kind of loading state for the UI here? Can add later nbd.

frontend/src/components/dialogs/SnapshotDiffChangePanel.vue
andypalmi: Done!

frontend/src/components/dialogs/AssetCompareDialog.vue
andypalmi: Done!

Issue #7032, opened 2026-04-07
Immersive editor: open snapshot detail as modal rather than right drawer


dimitrieh:
### Current Behavior

In the standalone instance view, opening a snapshot slides in a right-side drawer with the details. In the immersive editor this drawer func is missing at this moment, making the feature not available.

### Expected Behavior

In the immersive editor, opening a snapshot opens the detail view in a **modal** rather than a drawer. Standalone instance view keeps its existing drawer, unchanged.

### Scope

- Immersive editor only (remote instance editor)
- Covers every place snapshots open in a drawer today (Version History timeline, Snapshots list, etc.)
- Modal contains the same content as the drawer

### Why now

Unblocks customers who need snapshot review and edit inside the immersive editor. Ships alongside the Developer Mode tab restoration to give a complete create-and-review snapshot path without leaving the editor. A more refined drawer-interaction model is part of the broader immersive editor refactor (#6267); this is the in-between step.

### Considerations

- Standard modal close (Esc, backdrop, close button), no impact on the main immersive drawer beneath
- Sized for the content the drawer normally holds

n-lark:
# Snapshot Detail Modal – Test Plan (#7032)

<details><summary> ✅  All tests passed</summary>

---

## Hosted Instance – Immersive Editor

- [x] Open the immersive editor for a hosted instance → navigate to **Version History → Snapshots**
- [x] Click a snapshot row → modal opens (not a right drawer)
- [x] Modal shows snapshot name in header, flow viewer, description, author, date
- [x] **Edit/Save/Discard** buttons appear and function correctly
- [x] **Restore** button appears and triggers restore confirmation
- [x] Clicking the X button closes the modal
- [x] After **deleting** a snapshot, modal closes and list refreshes
- [x] After **saving** edits, modal stays open and snapshot name updates in header

## Hosted Instance – Standalone View (no regression)

- [x] Open the standalone instance view → **Version History → Snapshots**
- [x] Click a snapshot row → right drawer opens (unchanged behaviour)
- [x] Version History **Snapshots / Timeline toggle** is visible and navigates correctly

---

## Remote Instance – Immersive Editor

- [x] Open the immersive editor for a remote instance → navigate to **Version History**
- [x] **Snapshots / Timeline toggle** is visible in the drawer
- [x] Navigate to **Snapshots** tab → click a snapshot row → modal opens
- [x] Modal shows correct snapshot details
[...]

PR #7037, opened 2026-04-07
[7032] Immersive editor: open snapshot detail as modal rather than right drawer

n-lark:
## Description

See [details](https://github.com/FlowFuse/flowfuse/issues/7032#issue-4218402229) & [test plan](https://github.com/FlowFuse/flowfuse/issues/7032#issuecomment-4201018326). 

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7032

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/ee/lib/sso/emailTemplates/SSOCertsExpiring.js
+++ b/forge/ee/lib/sso/emailTemplates/SSOCertsExpiring.js
@@ -1,0 +1,32 @@
+// Sent when a SSO Cert will expire in under 2 weeks
+
+// Inserts:
+// - name: SSO profile name
+// - data: Date of cert Expiry
+
+module.exports = {
+    subject: 'FlowFuse SSO Certificate expiring soon',
+    text:
+`Hello {{{user.name}}},
+
+The certificate for the SSO Profile "{{name}}" will expire
+at {{{date}}}.
+
+Please talk to your SSO administrator about issuing a new
+certificate for this provider.
+
+Your friendly FlowFuse Team
+`,
+    html:
+`
+<p>Hello {{{user.name}}},</p>
+
+<p>The certificate for the SSO Profile "{{name}}" will expire
+at {{{date}}}.</p
+
+<p>Please talk to your SSO administrator about issuing a new
+certificate for this provider.</p>
+
+<p>Your friendly FlowFuse Team</p>
+`
+}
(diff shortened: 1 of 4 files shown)


frontend/src/pages/device/VersionHistory/Snapshots/index.vue
```diff
@@ -1,5 +1,11 @@
 <template>
     <div id="device-snapshots" class="flex-1 flex flex-col overflow-auto">
+        <SnapshotDetailsModal
```
cstns, line 3: we're sunsetting this method of using modals in favor of the global dynamic one in which we can inject the component directly. 

this is good for now as it wil be replace when we migrate to the new embedded immersive mode

frontend/src/pages/device/VersionHistory/index.vue
```diff
@@ -70,6 +70,7 @@
     </router-view>
 
     <SnapshotCreateDialog
+        v-if="device"
```
cstns, line 73: why are these needed?

frontend/src/components/dialogs/SnapshotDetailsModal.vue
```diff
@@ -0,0 +1,174 @@
...
+
+<style scoped lang="scss">
+// ff-dialog-box has overflow: auto on the whole box — override so only the
+// body scrolls and the header stays pinned.
+:deep(.snapshot-details-modal) {
+    overflow: hidden;
+}
+
+:deep(.ff-dialog-content) {
+    padding: 0;
```
cstns: this does not work and the modal should do without padding as it's already quite narrow

cstns (changes requested):
for the sake of brevety, please only adress the padding issue, we'll sort out the rest in a later revision

frontend/src/pages/device/VersionHistory/index.vue
n-lark: These guard against a pre-existing timing issue where `DeviceEditor` passes `device=null` to `VersionHistory` while the async load completes. Both have required Object props that fail validation when null is passed, causing Vue warnings on every page load.

frontend/src/components/dialogs/SnapshotDetailsModal.vue
n-lark: Removed

PR #7067, opened 2026-04-10
[7065] Typescript: Phase 2 — Domain Types

### LINKED ISSUE #7065: Typescript: Phase 2 — Domain Types
## Phase 2 — Domain Types 

> No backend changes required. `@fastify/swagger` already generates a live OpenAPI spec — `scripts/dump-openapi.js` boots the app, dumps the spec, and exits without starting the HTTP server. Requires the same env config and DB as a normal dev run.

- [ ] Add `scripts/dump-openapi.js`
- [ ] Install `openapi-typescript` as a dev dep; add `generate:types` script to `package.json`:
  ```
  node scripts/dump-openapi.js && openapi-typescript openapi.json -o frontend/src/types/generated.ts
  ```
- [ ] Run `npm run generate:types`
- [ ] Review `frontend/src/types/generated.ts` — confirm `Team`, `Device`, `Application`, `Instance`, `User`, `Snapshot` are present and accurate
- [ ] Create `frontend/src/types/index.ts` to re-export generated types + hand-write any shapes not in the OpenAPI spec (component-internal state, enum constants like `Roles`, `TeamType`)
- [ ] Goal: `import type { Team } from '@/types'` works as a single import surface

**No manually maintained domain type files.** Generated types regenerate when the backend schema changes.

n-lark:
## Description

See https://github.com/FlowFuse/flowfuse/issues/7065

See [comment here](https://github.com/FlowFuse/flowfuse/issues/7065#issuecomment-4226327296) around some issues with optional properties

> **TLDR**: Generated types mark every field as optional because the backend schemas don't specify which fields are required. Workaround: manually override the worst offenders in index.ts. Real fix: add required arrays to forge/db/views/*.js and regenerate.

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7065

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/types/generated.ts
```diff
@@ -0,0 +1,10671 @@
+/**
+ * This file was auto-generated by openapi-typescript.
+ * Do not make direct changes to the file.
+ */
+
+export interface paths {
+    "/api/v1/settings/": {
```
cstns, line 7: I like the fact that this paves the way of strictly typing our api clients, maybe even re-interpret the way we use them

frontend/src/types/index.ts
```diff
@@ -0,0 +1,46 @@
...
+ *   npm run generate:types
+ *
+ * Hand-written types below cover shapes that are not in the OpenAPI spec:
+ * component-internal state, enum constants, and frontend-only unions.
+ */
+
+// Re-export all auto-generated OpenAPI types.
+// Run `npm run generate:types` to produce this file.
+export * from './generated.js'
+
```
cstns, line 15: should we also move frontend/src/components.d.ts within this directory to have all our types together? writing the comment here for lack of a better place

frontend/src/types/generated.ts
```diff
@@ -0,0 +1,10671 @@
...
+        trace?: never;
+    };
+    "/api/v1/teams/{teamId}/broker/clients": {
+        parameters: {
+            query?: never;
+            header?: never;
+            path?: never;
+            cookie?: never;
+        };
+        /** List MQTT clients for the
```
cstns, line 9915: I think we should be re-exporting schemas as top level name so we can use them like 
```ts
import type { FlowBlueprint } from '@/types'
```

cstns, line 9913: why is this needed?

cstns, line 9974: I'm only writing this once but it applies to all schemas that inherit `[key: string]: unknown;`.

Why are making the type so lax?

cstns, line 10749: why are defs and operation types needed?

cstns, line 10664: why are the responses, parameters, requestBodies, headers and pathItems keys needed?

scripts/dump-openapi.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Boots the FlowFuse app, dumps the live OpenAPI spec to openapi.json, then exits.
+ * Does NOT start the HTTP server.
+ *
+ * Requires the same env config and database as a normal dev run.
+ * Usage: node scripts/dump-openapi.js
+ */
+
+const fs = require('fs')
```
cstns, line 17: we need to find a way, maybe through ci to have this run/checked if it's in sync with the backend. I'm not to be trusted to remember to keep them in sync and I think we should expect anyone to.

cstns (changes requested):
see questions/issues raised below.

It's a good starting point for type defs but we're building on shaky ground.

I'm very keen on getting this in btw!

frontend/src/types/index.ts
n-lark: `components.d.ts` is a Vue module augmentation (`declare module 'vue'`) picked up implicitly by Volar, not something you import. Its location at `src/` follows standard Vue convention. `types/` is for importable domain types, so mixing them doesn't seem ideal.

frontend/src/types/generated.ts
n-lark: `generated.ts` is auto-generated by [`openapi-typescript`](https://openapi-ts.dev/) from the live OpenAPI spec. We don't edit this file, it's the output of `npm run generate:types`. The full structure (`paths`, `components`, `$defs`, `operations`, etc.) maps 1:1 to the OpenAPI 3.x spec. The tool generates all of it whether or not we use every section.

That said, I looked into the CLI options and added `--root-types --root-types-no-schema-prefix` to the `generate:types script`. After regenerating, this:

- Exports schema types at the root level, so `import type { Team } from '@/types'` works directly without reaching into `components["schemas"]`
- The `$defs`, `operations`, and empty `responses/parameters/pathItems` keys are still in the file (part of the full OpenAPI spec structure) but consumers never need to touch them

n-lark: See https://github.com/FlowFuse/flowfuse/pull/7067#discussion_r3088908334

n-lark: See https://github.com/FlowFuse/flowfuse/pull/7067#discussion_r3088908334

In addition - This also comes from the backend spec, not the tool. Schemas without explicit `additionalProperties: false` produce this. Tightening requires backend-side changes which we could include when we make the required array changes.

n-lark: See https://github.com/FlowFuse/flowfuse/pull/7067#discussion_r3088908334

n-lark: See https://github.com/FlowFuse/flowfuse/pull/7067#discussion_r3088908334

scripts/dump-openapi.js
n-lark: Agreed, we shouldn't rely on remembering to run `generate:types`. A CI step that runs `npm run generate:types` and then checks for a dirty diff (`git diff --exit-code frontend/src/types/generated.ts`) would catch any drift. If the generated output doesn't match what's committed, the check fails. Is this something I should bug Piotr for?

cstns: yes, I'll create a ticket in that regard

frontend/src/types/generated.ts
cstns: awesomeness

cstns:
@n-lark, @ppawlowski was gracious to lend a hand and setup the ci step to check for TS spec drift we can work off that

scripts/dump-openapi.js
```diff
@@ -0,0 +1,68 @@
...
+        )
+    }
+
+    return JSON.parse(specStr)
+}
+
+;(async () => {
+    let server
+    try {
+        server = await forge()
```
cstns: Something that I didn't pick up yesterday is that the `npm run generate:types` command fails silently locally because this initialization bit is expecting a config file based on `process.env.NODE_ENV`

scripts/dump-openapi.js
```diff
@@ -0,0 +1,68 @@
...
+        const spec = normaliseSchemaNames(server.swagger())
+        const outPath = path.resolve(__dirname, '..', 'openapi.json')
+        fs.writeFileSync(outPath, JSON.stringify(spec, null, 2))
+    } finally {
```
cstns, line 89: we should also catch errors and output some sort of information when something fails, it took me a minute to figure out what was happening

cstns (changes requested):
As an additional thing, the ci action is throwing an error due to diff's being detected in the `/frontend/src/types/generated.ts` file. For some reason this is not happening locally.

You can check the new ci action logs for more details

scripts/dump-openapi.js
n-lark: Fixed! Added a catch block that logs the error and exits with code 1 instead of swallowing it in the finally. You'll now see `Failed to generate OpenAPI spec: <error message>` on failure.

scripts/dump-openapi.js
n-lark: Fixed! `dump-openapi.js` now defaults `NODE_ENV` to `'development'` if unset, so the config resolves to the repo root the same as a normal dev run. Won't override if someone sets it explicitly.

n-lark:
Hey @cstns @ppawlowski - First Piotr thank you for setting up the test 💪 

Second, the test is failing and I think this is why, please advice on what you all think is best. 

# CI Drift Failure — EE Schema Problem

## What's happening

`dump-openapi.js` boots the full app via `forge()` to grab the OpenAPI spec. During boot, `forge/ee/index.js` checks `app.license.active()` — without a valid EE license, the entire EE module is skipped. This means Pipeline, FlowBlueprint, MCP, DatabaseCredentials, and DatabaseTable schemas are never registered and don't appear in the generated types.

The committed `generated.ts` was produced with an EE license active, so it has all the types. When CI (or a dev without an EE license) regenerates, those types are missing — drift detected.

## Solution

`forge()` accepts a programmatic config object (same pattern used by the test suite). Updated `dump-openapi.js` to boot with:

- An **enterprise-tier dev license** (from `test/unit/forge/ee/routes/mcp/index_spec.js`) — unlocks all EE routes including MCP
- `tables: { enabled: true }` — unlocks DatabaseCredentials/DatabaseTable schemas
- In-memory SQLite + stub driver — no external DB or config file needed, works identically in CI and locally

scripts/dump-openapi.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    return JSON.parse(specStr)
+}
+
+;(async () => {
+    let server
+    try {
+        // Dev-only license so EE schemas are included in the spec.
+        const DEV_LICENSE = 'eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJGbG93Rm9yZ2UgSW5jLiIsInN1YiI6IkZsb3dGb3JnZSBJbmMuIERldmVsb3BtZW50IiwibmJmIjoxNjYyNDIyNDAwLCJleHAiOjc5ODY5MDIzOTksIm5vdGUiOiJEZXZlbG9wbWVudC1tb2RlIE9ubHkuIE5vdCBmb3IgcHJvZHVjdGlvbiIsInVzZXJzIjoxNTAsInRlYW1zIjo1MCwicHJvamVjdHMiOjUwLCJkZXZpY2VzIjo1MCwiZGV2Ijp0cnVlLCJpYXQiOjE2NjI0ODI5ODd9.e8Jeppq4aURwWYz-rEpnXs9RY2Y7HF7LJ6rMtMZWdw2Xls6-iyaiKV1TyzQw5sUBAhdUSZxgtiFH5e_cNJgrUg'
```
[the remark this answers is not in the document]
n-lark: This is fine and we already use this in `test_environment_ee.js`

scripts/dump-openapi.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    return JSON.parse(specStr)
+}
+
+;(async () => {
+    let server
+    try {
+        // Enterprise-tier dev license so all EE schemas (including MCP, tables) are included.
+        const DEV_LICENSE = 'eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6ImZkNDFmNmRjLTBmM2QtNGFmNy1hNzk0LWIyNWFhNGJmYTliZCIsInZlciI6IjIwMjQtMDMtMDQiLCJpc3MiOiJGbG93Rm9yZ2UgSW5jLiIsInN1YiI6IkZsb3dGdXNlIERldmVsb3BtZW50IiwibmJmIjoxNzMwNjc4NDAwLCJleHAiOjIwNzc3NDcyMDAsIm5vdGUiOiJEZXZlbG9wbWVudC1tb2RlIE9ubHkuIE5vdCBmb3IgcHJvZHVjdGlvbiIsInVzZXJzIjoxMCwidGVhbXMiOjEwLCJpbnN0YW5jZXMiOjEwLCJtcXR0Q2xpZW50cyI6NiwidGllciI6ImVudGVycHJpc2UiLCJkZXYiOnRydWUsImlhdCI6MTczMDcyMTEyNH0.02KMRf5kogkpH3HXHVSGprUm0QQFLn21-3QIORhxFgRE9N5DIE8YnTH_f8W_21T6TlYbDUmf4PtWyj120HTM2w'
```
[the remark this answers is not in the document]
cstns: the dev license is already exposed and used in multiple places throughout the application and is a public license

cstns (approved):
awesome

PR #7098, opened 2026-04-17
[6267] Allow users to interact with the Instance drawer/FlowFuse expert in the immersive experience without it overlapping one another

### LINKED ISSUE #6267: Allow users to interact with the Instance drawer/FlowFuse expert in the immersive experience without it overlapping one another.
## Description

Add pin mode functionality to make the drawer part of the UI rather than a slide over.

## Context

From PR #6234 description, mentioned to @gstout52.

## Acceptance Criteria

- [ ] Implement pin mode for drawer

n-lark:
## Description

See details & test plan [here.](https://github.com/FlowFuse/flowfuse/issues/6267#issuecomment-4282098950) 

embedded

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/6267

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/layouts/ImmersiveEditor.vue
```diff
@@ -0,0 +1,100 @@
+<template>
```
cstns, line 1: we should remove the plain layout file, it's only use was for the old immersive mode

frontend/src/App.vue
```diff
@@ -37,6 +37,11 @@
                     <router-view />
                 </ff-layout-plain>
             </template>
+            <template v-else-if="pageLayout === 'immersive'">
```
cstns, line 35: we can remove the plain layout case

frontend/src/composables/DrawerHelper.js
```diff
@@ -1,120 +0,0 @@
...
-                        toggleDrawer()
-                    }
-                    isInitialTease.value = false
-                    teaseCloseTimeout = null
-                }, 2000)
-            }
-        }
-    }
-
-    function runInitialTease () {
```
cstns, line 89: very happy to get rid of this

frontend/src/pages/device/Editor/index.vue
```diff
@@ -236,6 +175,10 @@ export default {
                 }
             ]
         },
+        homeRoute () {
```
cstns: i'm not sure why we need to define this here. It's only passed along in the EditorDrawer component which already receives the device on the entity prop.

frontend/src/layouts/ImmersiveEditor.vue
cstns: writing this here because it's the closest i can comment to the file

frontend/src/layouts/ImmersiveEditor.vue
```diff
@@ -0,0 +1,103 @@
...
+    <div class="ff-layout--immersive" :class="{ 'ff-layout--immersive--fullscreen': editorImmersiveDrawer.fullscreen }">
+        <PageHeader v-show="!editorImmersiveDrawer.fullscreen" />
+        <div class="ff-layout--immersive--wrapper">
+            <slot />
+        </div>
+        <ff-dialog
```
cstns: This looks very similar to the dialog from the platform layout. Would it warrant extraction into a reusable component given it's the placeholder for the global dialog which we're intending to become the standard dialog

frontend/src/layouts/ImmersiveEditor.vue
```diff
@@ -0,0 +1,103 @@
...
+            <div v-else class="space-y-2" v-html="dialog.html" />
+            <div v-if="dialog.notices?.length" class="notices flex flex-col gap-3 mt-5">
+                <hr class="mb-5">
+                <template v-for="notice in dialog.notices" :key="notice">
+                    <component :is="notice" v-if="typeof notice === 'object'" />
+                    <notice-banner v-else :text="notice" />
+                </template>
+            </div>
+        </ff-dialog>
+        <TransitionGroup class="ff-notifications" name="notifications-list" tag="div">
```
cstns: The same for the notification toasts. For this instance it will become particularly useful when we decide to make improvements to the toasts (they're not.. let's say optimal at the moment). 

Having them in the same basket across layouts would make our lives easier in the future

cstns: also, modularity

frontend/src/layouts/ImmersiveEditor.vue
```diff
@@ -0,0 +1,103 @@
...
+            <ff-notification-toast
+                v-for="(a, $index) in alertsReversed" :key="a.timestamp"
+                :type="a.type" :message="a.message" data-el="notification-alert"
+                :countdown="a.countdown || 3000" @close="clearAlert($index)"
+            />
+        </TransitionGroup>
+    </div>
+</template>
+
+<script setup>
```
cstns, line 12: love the use of the composition api

cstns: why not lang="ts"?

cstns: not asking to change, just curious

frontend/src/stores/product-expert.js
```diff
@@ -96,6 +96,22 @@ export const useProductExpertStore = defineStore('product-expert', {
             const featuresCheck = useAccountSettingsStore().featuresCheck
             if (featuresCheck.isExpertAssistantFeatureEnabled === false) return
 
+            // In immersive editor context, navigate to the Expert tab instead of opening RightDrawer
+            const router = require('@/routes.js').default
```
cstns: why are we importing this inline? from the top of my head i can't find a reason why it would be a circular dependency

frontend/src/stores/product-expert.js
```diff
@@ -96,6 +96,22 @@ export const useProductExpertStore = defineStore('product-expert', {
...
+            const currentRoute = router.currentRoute.value
+            const isImmersiveEditor = currentRoute.name?.startsWith('instance-editor-') ||
```
cstns: just a thought here, wouldn't it make sense to extract this into the context store? we already have all the metadata of the route loaded under the route key, and it seems to be a pattern we keep repeating throughout the code.

This is exactly the kind of thing I envisioned the context store for

frontend/src/stores/ux-drawers.js
```diff
@@ -15,6 +15,15 @@ import { useUxStore } from './ux.js'
 
 export const useUxDrawersStore = defineStore('ux-drawers', {
     state: () => ({
+        editorImmersiveDrawer: {
+            active: false,
```
cstns, line 19: the other drawers use state to determine if it's open or not, we should use the same notation or update the others for consistency

cstns: also, we should normalize the drawers. they (should) all pretty much do the same thing (accept a component, resize, be able to bind things to them). 

the right drawer got a bit more attention and is more dynamic than the left one, we should take care of that.

I'll create a follow-up task for it

cstns: actually.. we should not add a new type of drawer just for the use case of the immersive editor.

both left/right drawers should be able to have a view stack. when switching the immersive drawer from left to right or vice versa, the component alongside it's configuration & stack should swap parents from the left drawer to the right drawer.

this way we'll have a cohesive drawer management and open up future use cases with other features not only the immersive drawer.

let's talk about it when you get in

cstns (changes requested):
awesome start, and a massive improvement on usability & aesthetics!

I have a couple of things i'd like done slighly differently and a couple of questions as well.

NB. I also found some inconsistencies in guardrails when using the app without a license attached. I don't think they were introduced with the current changes though but i'll create follow ups for them

frontend/src/stores/product-expert.js
n-lark: So two reasons: 

1. Top-level import router from '@/routes.js' would form a real cycle (Platform → RightDrawer → product-expert.js, plus routes.js → team/routes.js → team/index.vue → product-expert.js). It does crash on my end on startup with a top level import. 
2. `useRouter()` doesn't work inside Pinia actions because that hook needs the component-injection context.

frontend/src/layouts/ImmersiveEditor.vue
n-lark: So the main benefit would be getting types from the Pinia stores, and since they're not converted yet, I'd be hand-authoring types that drift the moment the stores get their TS pass. Better to flip it once the stores are typed imo.

frontend/src/pages/device/Editor/index.vue
n-lark: Good call killed

frontend/src/stores/product-expert.js
n-lark: Moved!

frontend/src/stores/ux-drawers.js
n-lark: So I tried this approach at first, then shifted away since I'd need to refactor the layout of the app and that felt out of scope. 

The immersive drawer in pinned mode is a flex sibling of the iframe so the iframe physically resizes when the drawer opens. RightDrawer goes relative when pinned, so the positioning is fine, but it's mounted in Platform.vue, not in the editor's flex wrapper, so the iframe never becomes its sibling. 

To merge them, RightDrawer would need to grow tabs nav, slot-driven actions, configurable side, fullscreen, view stack, and persistence (or hide those features if we don't want them) most of what immersive has today. 

I am willing to do this work if we think its worth the effort - I just wanted to keep the scope tight to start. Happy to talk through it.

n-lark: > the other drawers use state to determine if it's open or not, we should use the same notation or update the others for consistency

Did update this.

frontend/src/layouts/ImmersiveEditor.vue
n-lark: Killed

frontend/src/App.vue
n-lark: Killed

frontend/src/layouts/ImmersiveEditor.vue
n-lark: done

frontend/src/layouts/ImmersiveEditor.vue
n-lark: done

frontend/src/stores/product-expert.js
cstns: i know that `useRouter()` doesn't work and it saddens me. Inline imports/require also sadden me but i'll live with it

frontend/src/mixins/Alerts.js
```diff
@@ -1,30 +0,0 @@
-import alerts from '../services/alerts.js'
-
-export default {
```
cstns, line 3: nice

cstns (approved):
It has a green light for me in principle. I found a lot of ux niggles that may/may not have been introduced by this change, I'll leave it up to you on what/when to tackle them, for the time being i added them as sub task follow ups to the parent issue.

On a second note, i'm still not 100% happy with the drawer situation but I can live with it. I admit it's more of a personal/ocd note. I'll raise a follow up on that as well at one point

frontend/src/stores/ux-drawers.js
```diff
@@ -15,6 +15,15 @@ import { useUxStore } from './ux.js'
...
+            state: true,
+            side: 'left',
```
n-lark: default to right

cstns (approved):

PR #7119, opened 2026-04-21
Allow Forcing all users to always use SSO

### LINKED ISSUE #6319: Allow instant redirect to SSO provider if only 1 SSO provider configured
Hi,

This is a request to re-open: Allow instant redirect to SSO provider if only 1 SSO provider configured #5606 
Reason: solution does not meet request.

Original request was to remove the login page if only one SS) provider is configured\enabled. Objective was to remove the login page entirely (again, on certain conditions) as to have fewer login screens i.e. a shorter login process.

The solution that was built places a button on the login screen, redirecting to a ID provider. This solution unfortunately does not remove the login page from the login process and does not shorten the login process. 
As the email field is still visible, it may even lead to unnecessary confusion for users (although both will lead to the ID provider and ultimately have the same effect).

Thanks for considering this request.
DenW

hardillb:
fixes #6319
## Description

Allows configuration of a single SSO provider that all users will be redirected to

image

Users trying to access `/admin/*` routes are not redirected to the SSO provider allowing admins to be able to login even when SSO is not working.

## Related Issue(s)

#6319

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
Possible keycloak option for the logout url

https://suedbroecker.net/2021/05/18/simply-logout-from-keycloak/

hardillb:
One thing to note, this does load the login page and then redirect so it will flash up for a moment, not sure how to prevent that.

It does the redirect in the Vue `mounted`

frontend/src/pages/admin/Settings/General.vue
```diff
@@ -320,12 +343,14 @@ export default {
...
                 termsAndConditions: null,
-                offboardingUrl: null
+                offboardingUrl: null,
+                ssoOnlyProvider: null
             },
             teamTypes: [],
             instanceTypes: [],
             teamTypesOptions: [],
             platformStatsTokenGenerating: false,
+            ssoProvidersOptions: []
```
hardillb:
```suggestion
            ssoProvidersOptions: [],
```

frontend/src/pages/admin/Settings/General.vue
```diff
@@ -256,6 +256,25 @@
...
+            <FormRow v-if="input['platform:sso:only']" v-model="input['platform:sso:only:provider']" :error="errors.ssoOnlyProvider" :options="ssoProvidersOptions" data-el="single-sso-provider">
+                Which active SAML SSO provider to use for all logins
+            </FormRow>
+            <FormRow v-if="input['platform:sso:only']" v-model="input['platform:sso:only:logoutURL']" type="text" data-el="single-sso-url">
+                URL to redirect to on logout
+                <template #description>
+                    Prevents redirect loops automatically logging user back in from SSO provider
+                </template>
+            </FormRow>
+        </template>
```
knolleary, line 278:
```suggestion
        <template v-if="ssoEnabled">
            <FormHeading>Automatic SSO Redirect</FormHeading>
            <FormRow v-model="input['platform:sso:only']" type="checkbox" data-el="single-sso">
                Automatically redirect all logins to a single SAML SSO provider
                <template #description>
                    Users will be automatically redirected to the SAML SSO provider without waiting for confirmation on the login page.
                    <br>
                    Admin users can still access the login page by going to /admin/
                </template>
            </FormRow>
            <FormRow v-if="input['platform:sso:only']" v-model="input['platform:sso:only:provider']" :error="errors.ssoOnlyProvider" :options="ssoProvidersOptions" containerClass="max-w-sm ml-9" data-el="single-sso-provider">
                Which active SAML SSO provider to use for all logins
            </FormRow>
            <FormRow v-if="input['platform:sso:only']" v-model="input['platform:sso:only:logoutURL']" containerClass="max-w-sm ml-9" type="text" data-el="single-sso-url">
                URL to redirect to on logout
                <template #description>
                    Prevents redirect loops automatically logging user back in from SSO provider
                </template>
            </FormRow>
        </template>
```

Slight tweak to the language used, and indent the child settings

knolleary (changes requested):
Works for me. We *may* get a bit of pushback on the fact it flashes the forge UI up even for a small moment - but we can iterate on that if needs be.

One suggested change around the language used on the settings dialog.

hardillb:
> Works for me. We may get a bit of pushback on the fact it flashes the forge UI up even for a small moment - but we can iterate on that if needs be.

Yeah, I made a comment somewhere (probably slack) about that. I couldn't see a way round it as it needs to be done from the UI with the current model I think.

knolleary (approved):

PR #7132, opened 2026-04-24
Move expert to ee

hardillb:
fixes FlowFuse/engineering#6

## Description

Moves the Expert code to where it is supposed to be

## Related Issue(s)

FlowFuse/engineering#6


forge/ee/lib/index.js
cstns: should we move these in the enterprise tier block as well

cstns (changes requested):
looks good to my eyes. On concern about also moving the expert features under the ee block, may not be mandatory though 

tested on a non-licensed instance of ff, the expert is disabled (did not check immersive mode but that should follow along with the broader feature gate )

forge/ee/lib/index.js
hardillb: I only moved the stuff that was outside EE, if that should be a Enterprise only feature then yes it should be moved inside the 

```
if (app.license.get('tier') === 'enterprise') {
...
}
```

block

cstns: both features rely on the general expert to be enabled so it doesn't make sense to have them outside of the ee block

hardillb: Just to be clear, there are 4 assistant/expert flags outside the ee block.

(The first of which is not labelled properly.)

```
// Set the Editor Limits Feature Flag
    app.config.features.register('generatedSnapshotDescription', true, true)

    // Set the assistant inline completions Feature Flag
    app.config.features.register('assistantInlineCompletions', true, true)

    // Set the expert assistant Feature Flag
    app.config.features.register('expertAssistant', app.config?.expert?.enabled ?? false, true)

    // temporary until FF Expert Insights can be enabled on Self Hosted EE instance
    const isInsightsEnabled = app.config?.expert?.enabled && app.config?.expert?.insights?.enabled
    app.config.features.register('expertInsights', isInsightsEnabled ?? false, false)
```

Should all these be moved?

cstns: `app.config.features.register('expertAssistant', app.config?.expert?.enabled ?? false, true)` and `    app.config.features.register('expertAssistant', app.config?.expert?.enabled ?? false, true)` most definitely do, because they won't work without the expert feature enabled.

The `assistantInlineCompletions` seem to work without the expert being enabled, @Steve-Mcl could clarify it.

I'm not sure about `generatedSnapshotDescription`. In my mind it shouldn't work without the expert being enabled because that's what generates the description based on the diff. On the other hand, it seems to be disabled by default on the main branch after i disabled my dev license because the `/api/v1/settings` api call returns an empty features object so that practically removes the possibility of generating diffs using the expert from the UI it so the UI behaves as I expected it. Long story short, I believe it's worth moving it only for cohesion, not for any practical reasons

cstns: @dimitrie hypothetical for you:

A user installs FlowFuse on-prem without a license, so they're on the open-source version. They spin up a Node-RED instance and install the Assistant plugin on it. Who does that plugin call to check for a valid license or login: the on-prem installation, or our cloud platform?

cstns (approved):

Steve-Mcl (approved):
Looks good. 


Tested on pre-staging.
Image

PR #7174, opened 2026-04-29
[7151 + 7156] Follow Ups: Console errors when navigating hosted instances in immersive mode (open-source, no license) + Performance tab missing from hosted instance immersive mode

### LINKED ISSUE #7151: Follow Up: Console errors when navigating hosted instances in immersive mode (open-source, no license)
### Description

When running the open-source version without a license loaded, navigating hosted instances in immersive mode produces errors in the browser console. Investigate the source of these errors and resolve them so the open-source experience is clean when no license is present.

Image

remote instances should also be checked

### Epic/Story

_No response_

### LINKED ISSUE #7156: Follow Up: Performance tab missing from hosted instance immersive mode
### Description

The Performance tab is available on hosted instances in non-immersive mode but is missing from the immersive mode view. Add the Performance tab to immersive mode to achieve parity with the non-immersive experience.

immersive:
Image

non-immersive:
Image

### Epic/Story

_No response_

n-lark:
## Description

Fixes #7151 & #7156

<details><summary> ✅ All tests passed </summary>

- [x] **7151 clean console (regression case):** open a hosted instance in immersive mode → DevTools → Console + Network tabs. No `projects/undefined/...` 404s on initial paint, **and** no `projects/<real-uuid>/autoUpdateStack` 404 either. Console is clean.
- [x] **7151 happy path (no-regression):** open a hosted instance with a real instance loaded. Logs and update-schedule still fetch correctly once `loadInstance()` resolves; "Scheduled Maintenance" badge shows "Disabled" if no schedule, "Enabled" if one is set.
- [x] **7151 Maintenance settings — no schedule (no-regression):** open Settings → Maintenance on an instance with no auto-update schedule. Page loads with team-type defaults applied (or empty defaults if team type has none). No console errors.
- [x] **7151 Maintenance settings — schedule set (no-regression):** with an auto-update schedule configured, open Settings → Maintenance. Form populates correctly with the saved days/hours.
- [x] **7156 immersive Performance tab:** open a hosted instance → immersive editor. **Performance** appears in the nav between Node-RED Logs and Settings. Clicking it lands on `instance-editor-performance` and renders the Performance page.
- [x] **7156 non-immersive (no-regression):** non-immersive instance Performance page still works at `instance-performance` (route untouched).

</details>

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7156
Resolves https://github.com/FlowFuse/flowfuse/issues/7151

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/pages/instance/Settings/Maintenance.vue
```diff
@@ -322,11 +325,10 @@ export default {
             } else {
                 return instanceApi.removeUpdateSchedule(this.project.id)
                     .then(() => {
-                        this.scheduledUpgrade.initialValue = null
-                        this.scheduledUpgrade.selectedWeekdays = null
+                        this.scheduledUpgrade.initialValue = { enabled: false, days: [], restart: false }
+                        this.scheduledUpgrade.selectedWeekdays = []
```
n-lark, line 329: **Drive-by fix:** `Maintenance.vue` `saveSettings` remove-schedule path crashed with `TypeError: Cannot set properties of null` (set `initialValue = null` then `initialValue.enabled = false`). Replaced with the empty-state shape the form expects.

frontend/src/pages/instance/Settings/Maintenance.vue
```diff
@@ -150,6 +150,14 @@ export default {
...
     },
     watch: {
+        'project.id': {
+            immediate: true,
+            handler (id) {
+                if (id) {
+                    this.getUpdateSchedule()
+                }
+            }
+        },
```
n-lark, line 160: Avoids a `projects/undefined/autoUpdateStack` 404. The `Instance.js` mixin pre-fills `instance: {}`, so mounted was firing with `this.project.id === undefined`. Watcher gates on id. Mirrors the `Overview.vue` fix in this PR.

cstns: nice, i prefer this approach of reacting to data as opposed to doing stuff on the mounted hook for these types of scenarios

forge/ee/routes/autoUpdateStacks/index.js
```diff
@@ -70,11 +70,7 @@ module.exports = async function (app) {
                 list.push({ ...conf, day: i })
             }
         }
-        if (list.length > 0) {
-            reply.send(list)
-            return
-        }
-        reply.code(404).send({ code: 'not_found', error: 'Not Found' })
+        reply.send(list)
```
cstns, line 73: This might have returned a 404 because it's under the ee namespace. When running the os version of FF the ee routes are not bound to fastify so they return a 404. The 404 response might have been a method of aligning the payload between os/ee.

frontend/src/pages/instance/Settings/Maintenance.vue
```diff
@@ -150,6 +150,14 @@ export default {
         }
     },
     watch: {
+        'project.id': {
+            immediate: true,
+            handler (id) {
+                if (id) {
+                    this.getUpdateSchedule()
```
cstns, line 157: we should be calling this endpoint only if we have the feature enabled (see my comment on the autoUpdateStack api endpoint)

frontend/src/pages/instance/Overview.vue
```diff
@@ -273,12 +273,11 @@ export default {
         },
         getUpdateSchedule: async function (instanceId) {
             try {
-                await InstanceApi.getUpdateSchedule(instanceId)
-                this.autoStackUpgrade = true
-                return
+                const list = await InstanceApi.getUpdateSchedule(instanceId)
```
cstns, line 278: we should only be calling the get update schedule if we're on an ee version of the platform (see the api endpoint comment)

cstns (changes requested):

forge/ee/routes/autoUpdateStacks/index.js
n-lark: So for EE, I noticed it was still returning a 404 for an empty list `[]`, which didn't seem like the right behavior. I've gated the calls behind `features.autoStackUpdate` so OS never hits the endpoint and EE gets a clean 200, but I think empty schedules should return cleanly rather than 404.

cstns: I agree with the clean 200. It seems to be sufficient in my mind to gate the API call from getting executed by the FE when running the app in ee. If anyone insists on calling the api without a license the 404 response is warranted.

frontend/src/pages/instance/Settings/Maintenance.vue
n-lark: added

frontend/src/pages/instance/Overview.vue
n-lark: added

cstns (approved):

Issue #7168, opened 2026-04-29
Create EXPERT_COMMS_BETA_ENABLED feature flag for frontend


n-lark:
# EXPERT_COMMS_BETA_ENABLED feature flag added

`EXPERT_COMMS_BETA_ENABLED` is a per-team PostHog flag. Flip it on for specific teams in the
PostHog UI — no deploy needed.

PostHog flag: [EXPERT_COMMS_BETA_ENABLED](https://eu.posthog.com/project/2209/feature_flags/178633)

## Use it in a component

```js
import { mapState } from 'pinia'
import { useAccountSettingsStore } from '@/stores/account-settings.js'

export default {
    computed: {
        ...mapState(useAccountSettingsStore, ['isExpertCommsBetaEnabled'])
    }
}
```

```html
<div v-if="isExpertCommsBetaEnabled">…beta UI…</div>
```

The getter updates reactively as PostHog re-evaluates flags.

## Enable it for a team

1. Open the [flag in PostHog](https://eu.posthog.com/project/2209/feature_flags/178633).
2. Under **Match by**, pick **Group → `team`**.
3. **Add filter**: `team-id` `equals` `<team-hashid>` (from `GET /api/v1/user/teams` → `id` field).
4. Save. Have the user reload FlowFuse — the flag flips on within a few seconds.

<img width="956" height="394" alt="Image" src="https://github.com/user-attachments/assets/cda65822-e0f8-4aca-9be9-a2daff5a6d8a" />

## Local testing

To test locally, add to `flowforge.local.yml` and restart the forge backend:

```yaml
telemetry:
  enabled: true
  frontend:
    posthog:
      apiurl: https://eu.posthog.com
      apikey: <phc_…project-token>
      capture_pageview: false
```

To get `apikey`:  Posthog -> Settings -> General -> Project token

> **Disable ad blockers on `localhost`** (uBlock, Brave Shields, Privacy Badger all block
> PostHog). If you see `ERR_BLOCKED_BY_CLIENT` in the console, that's why. Easiest fix:
> test in a clean Incognito window with extensions off.

Verify in the console while logged into the target team:

```js
window.posthog.getGroups()                                  // → { team: "<id>" }
window.posthog.isFeatureEnabled('EXPERT_COMMS_BETA_ENABLED') // → true
```

PR #7176, opened 2026-04-30
Refactor/account settings feature checks

cstns:
## Description

 The featuresCheck getter in account-settings.js has ~130 lines of repetitive boilerplate. Nearly every feature follows the same 3-property pattern and there are several naming inconsistencies that have drifted in over time. The goal is to replace the boilerplate with a data-driven config table + builder function, and fix all naming inconsistencies across the app since we're building on top of it.
 
 Something the we should be considering is removing the built in functionality that surrounded the now defunct free team tier. For now i just hardcoded the complex eligibility check and disabled e2e tests that were covering it.

## Related Issue(s)

part of https://github.com/FlowFuse/flowfuse/issues/7168

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


test/e2e/frontend/cypress/tests-ee/free-tier/free-tier.spec.js
```diff
@@ -1,4 +1,4 @@
-describe('FlowFuse EE - Free Tier', () => {
+describe.skip('FlowFuse EE - Free Tier', () => {
```
n-lark, line 1: You sure you want this? Nbd if you do just making sure.

frontend/src/stores/ux-navigation.js
```diff
@@ -264,7 +264,7 @@ export const useUxNavigationStore = defineStore('ux-navigation', {
                                 tag: 'team-pipelines',
                                 icon: PipelinesIcon,
                                 disabled: requiresBilling,
-                                featureUnavailable: !features.devOpsPipelinesFeatureEnabled,
+                                featureUnavailable: !features.isDevOpsPipelinesFeatureEnabled,
```
n-lark, line 267: In `ux-navigation.spec.js` we have `devOpsPipelinesFeatureEnabled: true` on lines 22 & 56. This is now stale due to the rename, so we need to update them to match the new key.

n-lark (changes requested):
Looks good

test/e2e/frontend/cypress/tests-ee/free-tier/free-tier.spec.js
cstns: yes, i mentioned it in the description

> Something the we should be considering is removing the built in functionality that surrounded the now defunct free team tier. For now i just hardcoded the complex eligibility check and disabled e2e tests that were covering it.

At one point we offered a free team tier where users were allowed only devices but moved away from it. This also raises the question if we should also clean up the UI with the leftovers that we didn't clean after discontinuing the free plan (there are quite a few, and I have a working stash but now's not the time for it)

frontend/src/stores/ux-navigation.js
cstns: good catch

cstns: not being used though... I'll update it nonetheless

test/e2e/frontend/cypress/tests-ee/free-tier/free-tier.spec.js
n-lark: All good ty

n-lark (approved):

PR #7184, opened 2026-05-03
Add automated EMQX bridge setup for Expert broker

### LINKED ISSUE #7181: Automated EMQX Bridge Setup for Expert over pubsub
### Description

Admin routines for applying the bridge settings to the FF App Instance Broker that bridges to the new Central Expert Broker (See https://github.com/FlowFuse/engineering/issues/38)

This will pick up values from https://github.com/FlowFuse/helm/issues/889#issuecomment-4359368291

This is the implementation step.

Code that can tear down and reapply the EMQX connection, rules and actions with scheduled tasks to init and sync licence

- [ ] init: sync license in emqx bridge
- [ ] weekly: license check and update

### Epic/Story

_No response_

Steve-Mcl:
## Description

TLDR
Adds setup and maintenance for automated configuration of FF App team broker to new central Expert Broker 

DETAILS
This PR introduces provisioning for the EMQX bridge between the FlowFuse App Instance Broker and the Expert Broker. It adds resource templates, orchestrator and helper functions, and scheduled tasks to ensure the bridge is always correctly configured and synchronized, both at startup and on a weekly basis. Additionally, it extends the license API to expose the raw license string (for configuring the bridge), and adds corresponding unit tests.

**EMQX Bridge Provisioning and Management**

* Added a new module `setup.js` that provides high-level orchestration (`syncBridge`), low-level resource management (`addBridge`, `removeBridge`), and health-check (`validateBridge`) functions for the EMQX bridge, including dependency-aware resource creation and teardown, configuration validation, and error handling.
* Defined resource templates for the EMQX connector, actions, sources, and rules in `templates.js`, supporting the full bridge lifecycle and MQTT v5 features required for expert chat and inflight requests.

**Automated Bridge Synchronization**

* Registered a startup task to perform a full bridge resynchronization at application startup, ensuring any changes to templates or license details are applied immediately.
* Added a weekly scheduled task that checks and repairs the bridge if any resources are missing or if license details have changed, minimizing disruption to active sessions.

**License API**

* Updated the license API to expose a `raw()` method, returning the raw license string for use in bridge configuration.
* Added unit tests to verify the new `raw()` method and license status reporting in both licensed and unlicensed scenarios. [[1]](diffhunk://#diff-41a098a85f3883b8488c44d92888823d2495030c65466cca8e828f530d52d446R149-L154) [[2]](diffhunk://#diff-41a098a85f3883b8488c44d92888823d2495030c65466cca8e828f530d52d446R228-R235)

## Related Issue(s)

closes #7181

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/ee/lib/expert/emxq-bridge/setup.js
```diff
@@ -0,0 +1,329 @@
...
+    client = client || makeClient(cfg)
+
+    // Clone the connector template so per-call config doesn't mutate module-level
+    // state — matters for test isolation and any future concurrent callers.
+    const connectorPayload = {
+        ...connector,
+        server: cfg.expertBrokerServerAddress,
+        username: cfg.licenceId,
+        password: cfg.licenceJwt
+    }
```
hardillb, line 317: We are going to need to add support for marking this as a MQTT over TLS connection.

I think it's safe to have it default to TLS on as the prod FFC Expert broker will be TLS only


```suggestion
    const connectorPayload = {
        ...connector,
        server: cfg.expertBrokerServerAddress,
        username: cfg.licenceId,
        password: cfg.licenceJwt,
        ssl: {
          enable: true
      }
    }
```

hardillb: https://docs.emqx.com/en/enterprise/v5.8/admin/api-docs.html#tag/Connectors

forge/ee/lib/expert/tasks/startup.js
```diff
@@ -0,0 +1,12 @@
+const { syncBridge } = require('../emxq-bridge/setup.js')
+module.exports = {
+    name: 'EmqxExpertBridgeSetup',
+    startup: true,
+    schedule: '',
```
hardillb: I would suggest changing `startup` from `true` to a value so it becomes a offset from startup.

E.g. 180 * 1000 for 3mins

Also check on empty `schedule`, I'd just remove it rather than have it empty

hardillb (changes requested):
Need to sort out the TLS stuff, give me a shout if needed.

And just a schedule bit

forge/ee/lib/expert/tasks/startup.js
Steve-Mcl: without `schedule`, `reportTask` will show `undefined` - I think it might be better to leave it as empty string?

> I would suggest changing startup from true to a value so it becomes a offset from startup.

Can we meet half way - make it 90sec after startup (to avoid the very small window of trying expert immediately after launching?

```suggestion
    startup: 90 * 1000, // Run 1.5min after startup
    schedule: '', // no schedule
```

@hardillb ^

hardillb: The 3 mins was just an example, I just like to ensure everything has time to start up cleanly.

hardillb (approved):

PR #7210, opened 2026-05-06
Convert device logs and performance to use MqttService

### LINKED ISSUE #6963: follow-up: standardise existing mqtt clients (remote instances logs & performance)
Parent: #6923

Follow-up from MQTT scaffolding.

cstns:
## Description

- Refactored DeviceLog.vue and DevicePerformance.vue to use the centralized MqttService via the service orchestrator instead of managing raw MQTT connections directly
- Removes direct dependency on the mqtt npm module from device components, gaining automatic reconnection with exponential backoff, subscription replay, and credential refresh
  on reconnect

## how to test

- Navigate to a device's Logs tab — verify live logs stream in and the connection status indicator works
- Navigate to a device's Performance tab — verify CPU and Memory charts populate with data
- Simulate a broker disconnect (e.g. toggle network) — verify reconnection happens automatically and data resumes
- Navigate away from the logs/performance pages — verify no console errors from stale MQTT clients

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/6963

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/device/components/DeviceLog.vue
+++ b/frontend/src/pages/device/components/DeviceLog.vue
@@ -60,8 +60,15 @@ export default {
             return `ff/v1/${this.device.team.id}/d/${this.device.id}/logs`
         }
     },
-    mounted () {
-        this.connectMQTT()
+    watch: {
+        device: {
+            immediate: true,
+            handler (device) {
+                if (device && device.id) {
+                    this.connectMQTT()
+                }
+            }
+        }
     },
     unmounted () {
         this.disconnectMQTT()
@@ -72,7 +79,7 @@ export default {
         },
         async connectMQTT () {
             const mqttService = this.getMqttService()
-            window.qwe = mqttService
+
             await mqttService.createClient(this.mqttConnectionKey, {
                 getCredentials: () => deviceApi.getDeviceLogCreds(this.device.id),
                 onConnect: async (_connack, client) => {
--- a/frontend/src/pages/device/components/DeviceLog.vue
+++ b/frontend/src/pages/device/components/DeviceLog.vue
@@ -27,7 +27,9 @@
 <script>
 import deviceApi from '../../../api/devices.js'
 
-let mqtt
+import getServicesOrchestrator from '@/services/service.orchestrator'
+
+const HEARTBEAT_INTERVAL = 10000
 
 export default {
     name: 'DeviceLogView',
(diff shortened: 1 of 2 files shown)


frontend/src/pages/device/components/DeviceLog.vue
```diff
@@ -43,103 +45,96 @@ export default {
...
+            return `ff/v1/${this.device.team.id}/d/${this.device.id}/logs`
         }
     },
-    async mounted () {
-    // need to subscribe to log stream
-        const { default: mqttImp } = await import('mqtt')
-        mqtt = mqttImp
+    mounted () {
         this.connectMQTT()
     },
```
n-lark, line 72: So while testing I got these errors: 

```
DeviceLog.vue:57 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'id')
    at Proxy.mqttConnectionKey (DeviceLog.vue:57:47)
    at Proxy.connectMQTT (DeviceLog.vue:76:49)
    at Proxy.mounted (DeviceLog.vue:64:14)
```

&& 

```
installHook.js:1 [Vue warn]: Invalid prop: type check failed for prop "device". Expected Object, got Null  
  at <DeviceLogView device=null onConnected=fn onDisconnected=fn > 
  at <DeviceLogs device=null instance=undefined onDeviceUpdated=fn<bound refreshDevice>  ... > 
```

I think this is due to `DeviceLog` getting mounted with device still null.

Possible options: 

1. Bail early if device isn't ready in `mounted()`, then watch for it.
2. Guard in the parent Logs.vue: `v-if="device?.id"`

frontend/src/pages/device/components/DeviceLog.vue
```diff
@@ -43,103 +45,96 @@ export default {
...
-                // if no broker to connect to, we see this event
-                this.loading = false
-                this.$emit('disconnected')
-            })
+        getMqttService () {
+            return getServicesOrchestrator().$serviceInstances.mqtt
+        },
+        async connectMQTT () {
+            const mqttService = this.getMqttService()
+            window.qwe = mqttService
```
n-lark: Do you want this line or was it from debugging?

n-lark (changes requested):
Results of test plan - I'll keep poking but wanted to get these out. 

- [x] Navigate to a device's Logs tab — verify live logs stream in and the connection status indicator works
- [x] Navigate to a device's Performance tab — verify CPU and Memory charts populate with data
- [ ] ❌ Simulate a broker disconnect (e.g. toggle network) — verify reconnection happens automatically and data resumes

For this one - I toggled off network and then reconnected, never got any additional logs after the reconnection. 

- [x] Navigate away from the logs/performance pages — verify no console errors from stale MQTT clients

frontend/src/pages/device/components/DeviceLog.vue
cstns: damn, nice catch

frontend/src/pages/device/components/DeviceLog.vue
cstns: the watcher pattern seems sensible here

cstns:
marking as draft for the time being, there are a couple of loose ends with re connection that i don't have time to cover

cstns:
Reconnection on the performance and logs pages behaves the same as on main, this task is only about aligning the MQTT-over-WS usage.

@hardillb I might be misremembering, but I think both device tabs had a known limitation. If so, this probably isn't the right time to tackle it.

hardillb:
The logs tab only sends it's history buffer to the first client that connects, so reconnecting will not resend the history, but it should send new messages if the device is generating them  (e.g. a inject set to fire once a second and a debug to console node). This also impacts if multiple people/tabs open the logs page.

Would need to double check the performance page, but I think it's the same.

hardillb (approved):
Looks to work when I test it locally with a device performance and logs

n-lark (approved):

PR #7308, opened 2026-05-19
[5637] Upgrade hericons to v 2.1.5

### LINKED ISSUE #5637: Upgrade hericons to v 2.1.5
### Epic

_No response_

### Description

We're missing out on a lot of new and modern heroicons due to our current package version

https://heroicons.com/

None

### Acceptance Criteria

upgrade the package

n-lark:
## Description

See details [here](https://github.com/FlowFuse/flowfuse/issues/5637#issuecomment-4490855840). 

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/5637

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/components/blueprints/BlueprintTile.vue
```diff
@@ -61,20 +61,28 @@
...
 import FormRow from '../FormRow.vue'
 import AssetDetailDialog from '../dialogs/AssetDetailDialog.vue'
 
 import { useContextStore } from '@/stores/context.js'
 
+let outlineIconsPromise = null
+function loadOutlineIcons () {
+    outlineIconsPromise ??= import('@heroicons/vue/24/outline')
+    return outlineIconsPromise
+}
```
n-lark, line 84: So, the old code lazy-loaded each icon. Heroicons v2's exports field broke that, and `…/esm/${name}.js` didn't work either.

Both tiles (this and `BlueprintTileSmall.vue`) now lazy-load the whole outline namespace as one dynamic import, webpack splits it into its own chunk, so the ~20 KB loads on demand instead of inflating `main.js`. `loadOutlineIcons` is at module scope so all tiles share one fetch. Tried a few different things here lmk if there is a better solution.

test/unit/frontend/components/nav-item.spec.js
```diff
@@ -23,7 +23,7 @@ describe('NavItem', () => {
         const wrapper = mount(NavItem, {
             props: {
                 label: 'nav item',
-                icon: CogIcon
+                icon: Cog8ToothIcon
```
cstns, line 26: I am deeply opposed to this type of testing..

cstns (approved):
Code-wise I have no objections, and I'm going to defer my OCD about small bits and bobs unrelated to this PR to a later date.

The only thing that stood out: we might want to upgrade to the actual latest version of Heroicons. It doesn't really do much for us beyond making the next version bump smaller, so I'm not going to block on it.

I did a very thorough pass of the UI, and OCD aside, there's nothing I could pick out as wrong (the swapped icons that are missing in the new version are also sensible choices).

Good job!

test/unit/frontend/components/nav-item.spec.js
n-lark: All I did here was update the icon name.

cstns: I know, it wasn't a poke at what you did just how these unit tests were done (not to my liking)

PR #7327, opened 2026-05-21
[7320] Live team updates over MQTT/WS

### LINKED ISSUE #7320: Live team updates over MQTT/WS
# Live team updates over MQTT/WS

## Problem

When one user changes team state (billing mode, team type, settings) or another user's role/membership, other connected members don't see the change until they reload. The UI silently goes stale.

## Proposal

Add a per-team MQTT-over-WebSocket channel between the browser and forge so team-state and membership changes are pushed live. Two new topics:

- `ff/v1/<team>/team/updated` — broadcast; subscribers refetch the team.
- `ff/v1/<team>/u/<user>/membership` — targeted; the named user refetches their membership.

MQTT carries the signal, the existing REST endpoints still do the fetch. Same pattern as `comms.platform.settings.sync`.

## Broker requirement

The platform's MQTT broker is optional — instances without `broker.url` configured run without it today, and features that depend on it (project nodes, device actions, remote device editing) degrade gracefully. This work follows that same pattern: where a broker is configured, members get live updates; where it isn't, the UI behaves exactly as it does today (refresh-on-navigate). No new install requirement.

## Acceptance criteria

- [x] Two members viewing the same team see each other's team-state changes (billing mode, type, settings) within ~1s, no reload.
- [x] A role change is reflected in the affected user's UI without a reload; other members are unaffected.
- [x] A removed user's session sees the membership event and redirects cleanly.
- [x] On instances without a broker configured, behavior is unchanged from today.

n-lark:
## Description

See [test plan](https://github.com/FlowFuse/flowfuse/issues/7320#issuecomment-4519636147) and [details](https://github.com/FlowFuse/flowfuse/issues/7320#issue-4489170993). 

## Related Issue(s)

Resolves #7320 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/services/team-channel.service.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    protected _onMqttDisconnect (): void {
+        // no-op — reconnect handled by mqtt.service
+    }
+
+    protected _onMqttError (err: Error): undefined {
+        // no-op — terminal failures handled by connect()'s catch; transients retried by mqtt.service
+        return err && undefined
+    }
```
cstns: why not inline these on the create client method? (with comments ofc)

frontend/src/services/team-channel.service.ts
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        const membershipMatch = MEMBERSHIP_TOPIC_REGEX.exec(topic)
+        if (membershipMatch) {
+            this._handleMembership(payload)
+            return
+        }
+        if (topic.endsWith('/team/updated')) {
+            this._handleTeamUpdated()
+        }
```
cstns: it's doing what it needs to but thinking ahead, i'd like to avoid future use cases to be added as individual if cases

frontend/src/services/team-channel.service.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    protected _handleMembership (payload: { reason?: string }): void {
```
cstns: `_handleMembership` and `_handleTeamUpdated` feel like domain-specific logic that probably fits better on the context store rather than the service. The service could be a plain orchestrator: match the topic, dispatch to the right store method, and let Pinia handle the interpretation (what `reason: 'removed'` means, what to refresh, etc.).

Same thinking for the `window.location.assign('/')` bit. Navigation side effects might be worth delegating to the vue router instead of poking at `window` directly.

frontend/src/services/team-channel.service.ts
cstns: a small map of topic pattern -> store method would age better

frontend/src/services/team-channel.service.ts
```diff
@@ -0,0 +1,187 @@
...
+            name: 'teamChannel',
+            app,
+            router,
+            services
+        })
+    }
+
+    // Minted per page-load so duplicated tabs (which clone sessionStorage in
+    // most browsers) still get distinct credentials and don't kick each other.
+    getSessionId (): string {
```
cstns: great call! we need something similar to easily identify active browser sessions moving forward (we got away with using the chat session id for the expert) with the upcoming platform automation feature to name one

cstns: placement wise, I think it should reside inside a pinia store though

frontend/src/services/team-channel.service.ts
```diff
@@ -0,0 +1,187 @@
...
+import type { CreateServiceOptions } from '@/types/services/service.types'
+import type { TeamChannelServiceI, TeamRef } from '@/types/services/team-channel.types'
+
+const MEMBERSHIP_TOPIC_REGEX = /^ff\/v1\/[^/]+\/u\/([^/]+)\/membership$/
+
+function connectionKey (teamId: string): string {
+    return `team:${teamId}`
+}
+
+class TeamChannelService extends BaseService implements TeamChannelServiceI {
```
cstns: Conceptually this is screaming gateway/subscriber to me not service, concepts that we don't currently have on the frontend but might consider adding. case in point: this is the first step in subscribing frontend entitites to live updates but we'll be expanding soon enough to instances, devices, maybe apps etc so this approach won't scale well..

These two statements align with me: team-channel.subscriber.ts: "I subscribe to backend events for this domain" and team-channel.gateway.ts: "I'm the boundary between MQTT and Pinia".

so the question now becomes, who will own these subscribers/gateways and who will manage their lifecycle?

cstns: thinking as I type, we could repurpose the service.orchestrator into a plain orchestrator or app.orchestrator and introduce these new concepts there alongside services

cstns: we could also go down the path of  a new subscriber/gateway registry and initialize it as a parallel set of entities (structural separation)

cstns: we need to have a chat about this, I love it!

forge/comms/aclManager.js
```diff
@@ -360,6 +390,16 @@ module.exports = function (app) {
                 { topic: /^ff\/v1\/([^/]+)\/d\/([^/]+)\/resources\/heartbeat$/, verify: 'checkDeviceIsAssigned' }
             ]
         },
+        // browser-side team channel (per-tab, per-team)
+        teamFrontend: {
+            sub: [
+                // - ff/v1/<team>/team/updated
+                { topic: /^ff\/v1\/([^/]+)\/team\/updated$/, verify: 'checkUserIsTeamMember' },
```
cstns: team should use the t shorthand to align with the existing pattern

cstns: I'm only writing this up once but the same should apply on all topic patterns introduced

forge/comms/aclManager.js
```diff
@@ -412,6 +453,8 @@ module.exports = function (app) {
                 aclList = ACLS.project[aclType]
             } else if (/^device:/.test(username)) {
                 aclList = ACLS.device[aclType]
+            } else if (/^team-frontend:/.test(username)) {
+                aclList = ACLS.teamFrontend[aclType]
             } else if (/^frontend:/.test(username)) {
                 aclList = ACLS.frontend[aclType]
```
cstns, line 459: why not use the `frontend:` namespace? this won't scale well when we introduce new sync-able entities (instances/devices/etc)

forge/routes/api/team.js
```diff
@@ -945,6 +947,63 @@ module.exports = async function (app) {
         }
     })
 
+    /**
+     * Issue MQTT/WS credentials for the team-level browser channel.
+     * @name /api/v1/teams/:teamId/comms-credentials
+     */
+    app.post('/:teamId/comms-credentials', {
```
cstns, line 954: just typing out loud here, should we consider consolidating all credential generation requests into it's own controller/thing? the reason i'm asking is that this is the same pattern used across multiple controller entities in the application. I just used the same pattern when I needed it (I'm also guilty of taking the same approach) when implementing the expert credential routes. what do you think?

cstns: we could do it as a cleanup/consolidation task not asking for a change here

cstns (commented):
Very good approach, works as expected and you caught all the gotcha's that came with it!  

My concerns are about how this is going to scale (conceptual, not technical) moving forward and how to mitigate it since it'll be doing some really heavy lifting in the near future with the addition of instance/device/etc live status tracking

frontend/src/services/team-channel.service.ts
cstns: also related: https://github.com/FlowFuse/flowfuse/issues/7419

frontend/src/services/team-channel.service.ts
n-lark: done good call out

forge/routes/api/team.js
n-lark: Solid point - made follow up here: Consolidate broker credential generation #7441

frontend/src/services/team-channel.service.ts
n-lark: added!

forge/comms/aclManager.js
n-lark: Switched `team/updated` → `t/updated`. The membership topic already used the `u` shorthand so it was already aligned. Those were the only two new ones.

frontend/src/services/team-channel.service.ts
n-lark: Moved to the `account-auth` store!

forge/comms/aclManager.js
n-lark: Rename to `fe:team` - we want to nest based on `fe:instance`, `fe:device`, `fe:team`. We will eventually change frontend but not right now.

n-lark: Mkay updated to `fe-team` to match existing conventions - the `frontend` one will eventually need to be moved to `fe-device` but out of scope for this work.

n-lark:
# Frontend subscribers & transport

Live team/membership updates (MQTT-over-WebSocket) used to be a "service". They
now live in a dedicated **subscriber** behind a generic **transport**, booted by
the app orchestrator alongside services. Behaviour is unchanged — this makes the
live-update plumbing reusable for future entities (instances, devices, apps).

## Layers

```
                    AppOrchestrator
            bootServices()   bootSubscribers()
                  │                  │
          services/             subscribers/
      (mqtt, bootstrap)     TeamChannelSubscriber
                                     │ depends on
                                     ▼
                               Transport (interface)
                               MqttTransport (impl) ─▶ Broker (EMQX)

  in:  Broker ─▶ Subscriber (match topic) ─▶ Pinia store (interpret + act)
```

- **Transport** — thin, swappable layer over the channel (today MQTT). Only
  `connect`/`subscribe`/`disconnect`, so subscribers are read-only.
- **Subscriber** — owns a connection's lifecycle, routes each topic to a store
  action. No domain logic.
- **Store** — interprets the message (what a `reason` means, what to refresh).

frontend/src/services/team-channel.service.ts
n-lark: See https://github.com/FlowFuse/flowfuse/pull/7327#issuecomment-4722131760 - refactor we discussed has been applied.

frontend/src/services/team-channel.service.ts
n-lark: See https://github.com/FlowFuse/flowfuse/pull/7327#issuecomment-4722131760 - refactor we discussed has been applied. 

> Same thinking for the `window.location.assign('/')` bit. Navigation side effects might be worth delegating to the vue router instead of poking at window directly.

Specifically around this, I kept the window assignment since a `router.push` to `Home` just bounces back to the current team (`Home.vue` redirects there, and the team-list still holds the removed team), so a hard reload is needed to re-bootstrap clean. Did move the decision out of the subscriber into the store action though.

frontend/src/stores/account-auth.js
```diff
@@ -193,7 +201,10 @@ export const useAccountAuthStore = defineStore('account-auth', {
             if (useAccountSettingsStore().settings['platform:sso:only']) {
                 logoutURL = useAccountSettingsStore().settings['platform:sso:only:logoutURL'] || '/'
             }
-            return userApi.logout()
+            const teamChannel = getAppOrchestrator().$subscriberInstances.teamChannel
+            const disconnect = teamChannel ? teamChannel.disconnect().catch(() => {}) : Promise.resolve()
```
cstns, line 205: While this works for now but in the future we may want to emit a logout event and all subscribers terminate their connection. The alternative is to keep piling individual subscribers in here

cstns (approved):
very nice work! one thing that's bodering me, i left a comment on the logout cleanup but we can tackle that when we add more subscribers

PR #7353, opened 2026-05-27
[7257] Add proper pagination to Hosted and Remote Instances pages + N+1 fix for Hosted Instances

### LINKED ISSUE #7257: Add proper pagination to Hosted and Remote Instances pages + N+1 fix for Hosted Instances
### Description

Hosted Instances currently has no pagination, and Remote Instances uses a "Load More" button. Both should be replaced with standard pagination for consistency and usability.

Requirements:

- Add bottom table navigation (< 1 2 3 >) on both Hosted and Remote Instances pages
- Include a page-size selector with a reasonable default (10–25 entries per page)
- Search should be handled on the backend when paginated. Remote Instances already does this, so extend the same functionality to Hosted Instances

### Epic/Story

_No response_

n-lark:
## Description

See https://github.com/FlowFuse/flowfuse/issues/7257#issuecomment-4558189618

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7257

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/db/models/Project.js
```diff
@@ -558,8 +558,12 @@ module.exports = {
                     includeSettings = false,
                     includeMeta = false,
                     limit = null,
+                    offset = null,
```
cstns: Should we switch to a generic `pagination` object to house these params, similar to the device model? This method's arguments are getting out of hand really fast.

maybe set it up as an object with default values?

forge/db/models/Project.js
```diff
@@ -558,8 +558,12 @@ module.exports = {
...
+                    sort = null,
+                    dir = 'asc',
                     orderByMostRecentFlows = false,
-                    excludeApplications = null
+                    excludeApplications = null,
+                    withTotal = false
```
cstns: with total could be derived off if we're sending in the pagination object or not

forge/routes/api/team.js
```diff
@@ -365,32 +365,40 @@ module.exports = async function (app) {
...
+        schema: {
+            query: {
+                allOf: [
+                    { $ref: 'PaginationParams' },
+                    {
+                        type: 'object',
+                        properties: {
+                            sort: {
+                                type: 'string',
+                                enum: ['name', 'createdAt', 'updatedAt', 'application.name', 'flowLastUpdatedAt']
```
cstns, line 377: pagination params already include sort as an option, was this only so we can type the endpoint more tightly?

forge/routes/api/team.js
```diff
@@ -406,7 +414,9 @@ module.exports = async function (app) {
             }
         }
 
-        const projects = await app.db.models.Project.byTeam(request.params.teamId, options)
+        const queryResult = await app.db.models.Project.byTeam(request.params.teamId, options)
```
cstns, line 411: this is where we could send the `pagination` param in

forge/routes/api/team.js
```diff
@@ -365,32 +365,40 @@ module.exports = async function (app) {
...
     }, async (request, reply) => {
         const includeMeta = request.query.includeMeta
+        const paginated = typeof request.query.page === 'number'
+        const limit = paginated ? (request.query.limit || 25) : (request.query.limit ?? null)
+        const offset = paginated ? (request.query.page - 1) * limit : null
+
         const options = {
             includeSettings: true,
-            limit: request.query.limit,
+            limit,
```
cstns: we could extract the pagination options similarly to how we do it for devices?
`const paginationOptions = app.db.controllers.Device.getDevicePaginationOptions(request)`

forge/routes/api/teamDevices.js
```diff
@@ -95,7 +95,6 @@ module.exports = async function (app) {
         }
         const devices = await app.db.models.Device.getAll(paginationOptions, where, options)
         devices.devices = devices.devices.map(d => app.db.views.Device.device(d, { statusOnly: paginationOptions.statusOnly }))
-        devices.count = devices.devices.length
         reply.send(devices)
```
cstns, line 98: I'm trying to figure out why this was in place to begin with..

frontend/src/api/team.js
```diff
@@ -197,18 +197,35 @@ const getTeamInstancesList = async (teamId) => {
 
 const getInstances = async (teamId, {
     limit = 20,
+    page = null,
```
cstns: we could define a pagination object and just use that on the frontend to make our lives a bit easier

cstns: I would also extract these parameters as a separate argument

frontend/src/api/team.js
```diff
@@ -197,18 +197,35 @@ const getTeamInstancesList = async (teamId) => {
...
+    if (query) params.append('query', query)
+    if (sort) params.append('sort', sort)
+    if (dir) params.append('dir', dir)
     if (includeMeta) params.append('includeMeta', includeMeta.toString())
     if (orderByMostRecentFlows) params.append('orderByMostRecentFlows', orderByMostRecentFlows.toString())
 
-    return await client.get(`/api/v1/teams/${teamId}/projects?${params.toString()}`)
-        .then(res => res.data)
+    const res = await client.get(`/api/v1/teams/${teamId}/projects?${params.toString()}`)
+    res.data.projects = res.data.projects.map(r => {
```
cstns, line 222: what changed that we need to do this now, and we didn't before? nothing stood out for me

frontend/src/components/DevicesBrowser.vue
```diff
@@ -20,13 +20,14 @@
                 :rows="devicesWithStatuses"
                 :show-search="true"
                 search-placeholder="Search Remote Instances"
-                :show-load-more="moreThanOnePage"
+                :pagination="paginationProps"
                 :check-key="row => row.id"
                 :show-row-checkboxes="hasPermission('team:device:bulk-edit', applicationContext)"
                 @rows-checked="checkedDevices = $event"
-                @load-more="loadMoreDevices"
                 @update:search="updateSearch"
```
cstns, line 27: I would be extremely happy if we could get rid of the get mor functionality on the datatable. maybe a follow up if it's not used anywhere else

frontend/src/ui-components/stylesheets/ff-components.scss
```diff
@@ -899,6 +899,18 @@ li.ff-kebab-item {
       justify-content: flex-end;
     }
   }
+
+  // Flatten last-row corners so the footer (pagination) carries the bottom radius.
+  &--has-footer .ff-data-table--data tbody .ff-data-table--row:last-child {
+    .ff-data-table--cell {
+      &:first-child {
```
cstns, line 907: shouldn't be both rules applied to the last-child?

frontend/src/ui-components/components/data-table/Pagination.vue
```diff
@@ -0,0 +1,109 @@
...
+                data-action="page-next"
+                @click="goTo(props.page + 1)"
+            >
+                <template #icon><ChevronRightIcon /></template>
+            </ff-button>
+        </nav>
+    </div>
+</template>
+
+<script setup>
```
cstns: why not zoidberg/typescript?

frontend/src/components/DevicesBrowser.vue
```diff
@@ -482,16 +486,14 @@ export default {
...
                     component: {
                         is: markRaw(DeviceAssignedToLink)
                     }
                 })
                 if (this.featuresCheck.isDeviceGroupsFeatureEnabled) {
                     columns.push({
                         label: 'Group',
-                        key: 'deviceGroup.name',
-                        sortable: !this.moreThanOnePage
+                        key: 'deviceGroup.name'
```
cstns: why disable sorting for device group name?

frontend/src/components/DevicesBrowser.vue
```diff
@@ -502,23 +504,14 @@ export default {
             } else if (this.displayingApplication && this.featuresCheck.isDeviceGroupsFeatureEnabled) {
                 columns.push({
                     label: 'Group',
-                    key: 'deviceGroup.name',
-                    sortable: !this.moreThanOnePage
+                    key: 'deviceGroup.name'
```
cstns: why disable sorting for device group name?

frontend/src/components/DevicesBrowser.vue
```diff
@@ -502,23 +504,14 @@ export default {
...
-            const devicesToDisplay = new Set(this.filter?.devices)
-
-            return Array.from(this.devices.values()).filter((device) => {
-                if (!this.filter || this.unfilteredHasMoreThanOnePage) {
-                    return true
-                }
-
-                return devicesToDisplay.has(device.id)
-            })
+            return Array.from(this.devices.values())
```
cstns: do we even need filtered devices now?

frontend/src/components/DevicesBrowser.vue
```diff
@@ -544,7 +537,18 @@ export default {
             )
         },
         moreThanOnePage () {
-            return !!this.nextCursor
+            return this.totalRows > this.pageSize
+        },
+        paginationProps () {
+            // Hide only when no page size could ever split the result — keeps the size selector reachable otherwise.
+            if (this.totalRows <= 10) {
```
cstns: this is somewhat leaking context imo. 10 only makes sense within the pagination component because it's a default there. this type of decision (whether to display or not the pagination component) should be made within the pagination component not outside of it. If we'l ever want to change the defaults we'll have to do a sweep of the app to see where and how the paginator is used

cstns: we could to this.totalRows <= this.pageSize in the pagination component. slightly more dynamic and more encapsulated

frontend/src/components/DevicesBrowser.vue
```diff
@@ -898,27 +901,17 @@ export default {
...
             }
 
-            // Actually fetch the data
-            const data = await this.fetchData(nextCursor, null, extraParams)
-
-            if (resetPage) {
-                this.devices = new Map()
-            }
+            // Actually fetch the data — null cursor, pageSize as limit, extra params carry page/sort/filters/query.
+            const data = await this.fetchData(null, this.pageSize, extraParams)
```
cstns: I think we may be able to remove the nextCursor argument from the underlying method, a quick search didn't return any uses of the method with this argument

frontend/src/mixins/DeviceActions.js
```diff
@@ -67,6 +67,9 @@ export default {
                         Alerts.emit('Successfully deleted the device', 'confirmation')
                         this.deleteLocalCopyOfDevice(device)
                         this.$emit('delete-device')
+                        if (typeof this.loadDevices === 'function') {
```
cstns, line 70: this has me thinking on why we need to do this if we're already deleting the local copy of the device (which should in turn remove it from the ui)

frontend/src/pages/team/Instances.vue
```diff
@@ -211,20 +217,29 @@ export default {
     data () {
         return {
             loading: false,
+            // Gates the full-page spinner to first paint so refetches don't unmount the search input.
+            hasLoadedOnce: false,
```
cstns: this is hurting my brain. I'm not going to block it, but we need to understand why we need to resort to these sorts of things and address the underlying issue

frontend/src/pages/team/Instances.vue
```diff
@@ -211,20 +217,29 @@ export default {
...
                     class: ['w-44'],
-                    sortable: true,
                     component: {
                         is: markRaw(InstanceStatusBadge),
                         map: {
                             instanceId: 'id',
                             pendingStateChange: 'pendingStateChange',
                             optimisticStateChange: 'optimisticStateChange',
-                            status: 'meta.state'
+                            status: 'status'
```
cstns, line 240: I may be wrong, but status and meta.status are not the same

frontend/src/pages/team/Instances.vue
```diff
@@ -211,20 +217,29 @@ export default {
     data () {
         return {
             loading: false,
+            // Gates the full-page spinner to first paint so refetches don't unmount the search input.
+            hasLoadedOnce: false,
             instancesMap: new Map(),
+            page: 1,
```
cstns, line 221: this is the 2nd time i've seen these params (granted without the default sort) but i'd say we could have a default object that we could instantiate where needed with these props and their defaults

frontend/src/pages/team/Instances.vue
cstns: I think i'm getting what you're doing. you're now pre-loading every instance status  by default, so we won't need to re-query all instances iteratively/in a cascade. I can't place exactly where you're doing that just yet

cstns: nvm, found it, you're including `includeMeta: true` in the initial fetch data request

cstns: the problem with that is that the `byTeam` method returns the db entries of the statuses (which do not include the transitory states instances can be in, only the stable ones - see https://flowfuse.com/docs/user/instance-states/#flowfuse-node-red-instance-states). the old getInstanceMeta was calling `'/:instanceId/status'` which in turn was getting the actual live state of the instance by calling `await instance.liveState({ omitStorageFlows: true })` which retrieved it's data twofold, corroborating from the db and from the cache. the `byTeam` endpoint is not doing that

cstns: the result is that the db instance status can be running (which it well may be), but its actual state could be (installing,pending etc)

cstns: This little bit needs a rethink. My thinking is that the best way about it would be to make the `byTeam` method use the project's actual `liveState`. I'm sure there were reasons why this was not done this way in the first place, but now maybe that we have proper caching in place for statuses, we might get away with it.

@hardillb I'd appreciate your input on the above proposition

frontend/src/pages/team/Instances.vue
```diff
@@ -265,64 +280,90 @@ export default {
         },
         instancesAvailable () {
             return this.featuresCheck?.isHostedInstancesEnabledForTeam
+        },
+        paginationProps () {
+            if (this.dashboardRoleOnly) return null
+            // Hide only when no page size could ever split the result — keeps the size selector reachable otherwise.
+            if (this.totalRows <= 10) return null
```
cstns: again, I feel this should be handled within the component and is a leak

cstns (changes requested):

frontend/src/pages/team/Instances.vue
n-lark: For the `liveState` thing, isn't that what this does in `Project.js`? It runs through the same `liveState` method in the view layer the old endpointed used.

```
if (includeMeta) {
    const liveState = await project.liveState()
    result.meta = liveState?.meta
    result.flowLastUpdatedAt = liveState?.flowLastUpdatedAt
}
```

So its like `byTeam` -> `Project.js` `liveState` in view -> we get back the correct `pending`, `installing`, `starting` state etc.

I also tested this for starting  in the UI and it returns the pending state so it is working. 

Screenshot 2026-06-03 at 8 20 06 AM

--- 

> I may be wrong, but status and meta.status are not the same

They hold the same value on hosted-instance rows, we normalize `meta.state` onto `r.status` in `api/team.js:223-224`. The reason we map the column to `status` instead of `meta.state` is that the dashboard-role endpoint (`/dashboard-instances`) returns top-level `status` and no `meta` object, so the old `meta.state` mapping gave dashboard users `undefined` and a forever spinner. Reading `status` works for both response shapes.

frontend/src/ui-components/components/data-table/Pagination.vue
n-lark: Converted and added the eslint rule for the parser.

frontend/src/components/DevicesBrowser.vue
n-lark: Nope killed.

frontend/src/components/DevicesBrowser.vue
n-lark: Dropped param from signature + both callers good catch.

frontend/src/components/DevicesBrowser.vue
n-lark: Fixed

frontend/src/pages/team/Instances.vue
n-lark: Fixed

forge/db/models/Project.js
n-lark: Yup, gone. `withTotal` now derives from `pagination.page != null` inside `byTeam`. Callers no longer pass it explicitly.

forge/db/models/Project.js
n-lark: Done. `byTeam` now takes a pagination object alongside the other opts. Defaults are inline so callers can pass a partial object or skip it entirely.

forge/routes/api/team.js
n-lark: Done. New `app.db.controllers.Project.getProjectPaginationOptions(request)` mirrors the device pattern. `?limit=` alone still gets the legacy array.

forge/routes/api/team.js
n-lark: Yup refactored.

frontend/src/api/team.js
n-lark: Done. `getInstances(teamId, { pagination, includeMeta, orderByMostRecentFlows })` now. Both callers (Instances page + RecentlyModified home widget) pass `pagination: { page, limit, ... }`

frontend/src/pages/team/Instances.vue
n-lark: Done.

frontend/src/components/DevicesBrowser.vue
n-lark: Restored. It was only client-side sortable on single-page before, now backend-supported.

frontend/src/pages/team/Instances.vue
n-lark: Yah issue was the table would remount so it felt glitchy flashy. So I went down the path of a skeleton loader first but ended up just gating the load on `loading && !instancesMap.size` so we don't flashy glitch on page change.

forge/routes/api/team.js
n-lark: Yeah `PaginationParams.sort` is just a string, so unknown values would silently fall through to the default order. The enum makes the route 400 instead.

forge/routes/api/teamDevices.js
n-lark: Yahhh it was masking a pre-existing bug. `Device.getAll`'s count query wasn't applying filters, so `meta.total` came back unfiltered.

frontend/src/api/team.js
n-lark: So when I tested with a dashboard-only user I got a forever spinner. The old code path called `getInstanceMeta()` per-row after the initial fetch, which set `meta.state` on every row including dashboard ones. 

The N+1 fix removed that fan-out, so dashboard rows (whose endpoint doesn't return meta) lost their `meta.state`. Swapped the column mapping from `meta.state` to `status` and added this normalization to copy `meta.state` onto `status` so both shapes now read the same field.

frontend/src/components/DevicesBrowser.vue
n-lark: Did a quick grep, `<ff-data-table>` Load More is still used by a few other consumers so not an easy murder.

frontend/src/ui-components/stylesheets/ff-components.scss
n-lark: `:first-child` and `:last-child` target the first and last cell of the last row (left and right corners). Same row, different cells.

frontend/src/mixins/DeviceActions.js
n-lark: `deleteLocalCopyOfDevice` only removes the row from the in-memory map. Refetching keeps `totalRows` (server-side count) honest and pulls in a backfill row from the next page so we don't show "29 of 30" with a missing row. Skipping the refetch leaves the footer drifting and rows missing on small pages.

n-lark:
Hey @cstns all feedback should be addressed - thanks!

frontend/src/ui-components/stylesheets/ff-components.scss
cstns: I must have missed that the classes applied to the td's

frontend/src/components/DevicesBrowser.vue
cstns: unfortunate

cstns:
Really happy with the code overall Please excuse my pickiness! 

Everything worked as expected when I tested locally. The only thing left is the linting rule causing the gh actions to fail; once that's fixed, this is good to go.

cstns (approved):

cstns:
And a follow-up I noticed on the performance page https://github.com/FlowFuse/flowfuse/issues/7419

PR #7386, opened 2026-06-05
[7385] Expert Tools row not available on device  

### LINKED ISSUE #7385: Expert Tools row not available on device
### Current Behavior

When using expert on a remote instance, the expert context tools row is not shown

Image

### Expected Behavior

Expert should have a tools row

### Steps To Reproduce

Open the editor on an application remote instance & open the FF Expert

### Environment

- FlowFuse version: Latest/Live FFC
- Node.js version: NA
- npm version: NA
- Platform/OS: Win
- Browser: Chrome

n-lark:
## Description

The Expert tools row didn't render on remote instance editors. The device editor never called `setIsImmersive(true)` so the context selector was gated out, fixed by restoring that call in `mounted()`.

<s>Also hid the `node-selection` chip on devices: `selectedNodes` is only sent when `scope === 'immersive'`, which is instance-only in `context.js`, so on devices the chip showed a node count that was never actually sent. </s>

> [!IMPORTANT]
> Expert is broken in my dev env(I probably have an outdated flow for it), so no end-to-end check, but confirmed palette/debugLog payloads are sent on devices. 

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7385

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/expert/components/context-selection/index.vue
+++ b/frontend/src/components/expert/components/context-selection/index.vue
@@ -4,7 +4,7 @@
         <div class="chips-container" @wheel="horizontalScrolling">
             <context-chip v-for="(context, index) in selectedContextFiltered" :key="index" :contextItem="context" />
             <debug-chip v-if="hasDebugLogsSelected && !isInsightsAgent" />
-            <selection-chip v-if="hasUserSelection && !isInsightsAgent" />
+            <selection-chip v-if="hasUserSelection && !isInsightsAgent && !isDevice" />
         </div>
     </div>
 </template>
@@ -18,6 +18,7 @@ import SelectionChip from '../chips/SelectionChip.vue'
 
 import ContextSelectorButton from './ContextSelectorButton.vue'
 
+import { useContextStore } from '@/stores/context.js'
 import { useProductAssistantStore } from '@/stores/product-assistant.js'
 import { useProductExpertStore } from '@/stores/product-expert.js'
 
@@ -32,6 +33,7 @@ export default {
     computed: {
         ...mapState(useProductAssistantStore, ['getSelectedContext', 'hasDebugLogsSelected', 'hasUserSelection']),
         ...mapState(useProductExpertStore, ['isInsightsAgent']),
+        ...mapState(useContextStore, { isDevice: store => store.editorEntityType === 'device' }),
         selectedContext () {
             // for insights mode, return empty array
             if (this.isInsightsAgent) {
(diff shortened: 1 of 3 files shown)


frontend/src/components/expert/components/context-selection/index.vue
```diff
@@ -4,7 +4,7 @@
         <div class="chips-container" @wheel="horizontalScrolling">
             <context-chip v-for="(context, index) in selectedContextFiltered" :key="index" :contextItem="context" />
             <debug-chip v-if="hasDebugLogsSelected && !isInsightsAgent" />
-            <selection-chip v-if="hasUserSelection && !isInsightsAgent" />
+            <selection-chip v-if="hasUserSelection && !isInsightsAgent && !isDevice" />
```
n-lark: Hey @Steve-Mcl I just wanna confirm this is the intended behavior: 

> Hide the `node-selection` chip on devices: `selectedNodes` is only sent when `scope === 'immersive'`, which is instance-only in `context.js`, so on devices the chip showed a node count that was never actually sent.

Steve-Mcl: If I understand the change made here you would be hiding the "selected flow" chip if it is a device (or rather only show it when Selection is made AND not in insigts mode AND not a device.

Then no, hosted and remote should work the same (selection of flow is valid for both)

So I suspect this is not required.

```suggestion
            <selection-chip v-if="hasUserSelection && !isInsightsAgent" />
```


@cstns ?

Steve-Mcl: Updated in [Ensure Devices hows selection chip](https://github.com/FlowFuse/flowfuse/pull/7386/commits/c87410a4bf2f692bb0d3b8a89e44645a1ed033b1)

Steve-Mcl:
@n-lark I pulled this and tested. I ended up pushing 2 commits to get this over the line:
* [Ensure Devices shows selection chip](https://github.com/FlowFuse/flowfuse/pull/7386/commits/c87410a4bf2f692bb0d3b8a89e44645a1ed033b1)
* [Update scope determination logic for immersive editors to include device](https://github.com/FlowFuse/flowfuse/pull/7386/commits/67441b61868a0a2d853e2e48dc95114a48e1de1d)

I am happy to approve but as a contributor I think a 3rd party should probably buy off!

@cstns - tagging you as 3rd part reviewer.

Steve-Mcl (approved):
working with remote instances locally.

n-lark:
Awesome thanks sm @Steve-Mcl!

Steve-Mcl:
With Serban OOO until Wednesday, I think we should make an executive decision to get this into Tuesdays point release

WDYT @n-lark

n-lark:
Agreed - send it 🚀

PR #7398, opened 2026-06-09
fix: await DOM flush before rendering snapshot comparison

andypalmi:
## Summary
- Snapshot comparison rendering was broken: ports appeared at the top-left corner of nodes and wires were misaligned
- Root cause: `v-show="!loading"` hides the container with `display: none`. After setting `loading = false`, Vue hasn't flushed the DOM update yet, so `flowRenderer.compare()` runs while the SVG is still inside a `display: none` ancestor — `getBBox()` returns zeros for all elements
- Fix: `await this.$nextTick()` after `loading = false` to ensure the DOM is updated before rendering

## Test plan
- [x] Open a snapshot comparison dialog with two snapshots that have differing nodes
- [x] Verify nodes render with correct port positions (input on left, output on right)
- [x] Verify wires connect between ports, not from node top-left corners

### DIFF
--- a/frontend/src/components/dialogs/AssetCompareDialog.vue
+++ b/frontend/src/components/dialogs/AssetCompareDialog.vue
@@ -292,6 +292,10 @@ export default {
             } finally {
                 this.loading = false
             }
+            // v-show="!loading" uses display:none — wait for Vue to flush
+            // the DOM update so the container is visible before rendering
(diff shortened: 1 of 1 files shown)


andypalmi:
### Before

Screenshot 2026-06-09 at 14 11 14

### After

Screenshot 2026-06-09 at 14 10 40

@hardillb Finally found the culprit

hardillb (approved):

PR #7399, opened 2026-06-09
fix: snapshots list dropdown not flexible on resize

### LINKED ISSUE #7397: UI: snapshots list is not flexible
### Current Behavior

 When window is resized, the snapshots list does not resize like all the other components. see screenshot

Image
Image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

andypalmi:
Closes #7397

## Summary
- Removed firstChild width expansion in `TeleportedMenuMixin` that created a feedback loop preventing the dropdown from shrinking on window resize
- Use actual rendered menu width (instead of trigger width) for overflow repositioning, so kebab menus and other small-trigger menus still correct their position correctly
- Removed stale `w-full` class from `ListBox` options container (overridden by inline style and incorrect on `position: fixed` elements)

## Test plan
Tested on:
- [x] Snapshots list in the asset compare dialog — dropdown resizes correctly with the window
- [x] Kebab menu on the hosted instances table — context menu appears at the correct position and does not break existing functionality

### DIFF
--- a/frontend/src/mixins/TeleportedMenuMixin.js
+++ b/frontend/src/mixins/TeleportedMenuMixin.js
@@ -67,19 +67,8 @@ export default {
             // Use viewport-relative coordinates (no window.scrollX/Y)
             let triggerTop = triggerRect.bottom + this.optionsOffsetTop
             let triggerLeft = triggerRect.left
-            let width = triggerRect.width
+            const width = triggerRect.width
             const transform = ''
-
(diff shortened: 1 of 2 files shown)


andypalmi:
List now resizes when window is resized

Screenshot 2026-06-09 at 14 58 11

n-lark (approved):

PR #7409, opened 2026-06-09
[7408] Typescript: API Modules - Utils

### LINKED ISSUE #7408: Typescript: API Modules - Utils
# Utils (subtask)

Convert `frontend/src/utils/` to TypeScript.

- 16 util files `.js` → `.ts` with types.
- `roles.ts` now re-exports the canonical `Roles`/`RoleNames`/`TeamRoles` from `@/types`.
- Added webpack `resolve.extensionAlias` (`.js` → `.ts`) so existing `.js` imports resolve to the new `.ts` files — zero importer changes.

Verified green: eslint, `vue-tsc`, vitest, webpack build.

n-lark:
## Description

See https://github.com/FlowFuse/flowfuse/issues/7408#issue-4624838630

## Related Issue(s)

Resolves #7408

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


cstns:
the utils directory should also be repurposed to composables as that's the vue way of tackling helpers

n-lark:
> the utils directory should also be repurposed to composables as that's the vue way of tackling helpers

@cstns I do think that is out of scope for this work since we are just converting utils to typescript. Would you like me to make a separate ticket for that?

cstns:
Yes please, that's what I meant. Excuse my bluntness

n-lark:
Follow up here: Repurpose src/utils into composables #7437

frontend/src/utils/InstanceStateMutator.ts
```diff
@@ -1,13 +1,21 @@
...
+
+type MutableInstance = Instance & {
+    optimisticStateChange?: boolean
+    pendingStateChange?: boolean
+}
+
 export class InstanceStateMutator {
-    constructor (instance) {
+    instance: MutableInstance
+    prevState?: unknown
```
andypalmi: Is there a reason `prevState` is `unknown` here while `DeviceStateMutator` types it as `string`? Since the two files are nearly identical, might be nice to align them - but happy to leave it if the difference is intentional.

frontend/src/utils/daysSince.ts
```diff
@@ -1,6 +1,8 @@
 import elapsedTime from './elapsedTime.js'
 
-export default function (dateString, shortenExtendedPeriods = false) {
+import { Maybe } from '@/types/common/types.js'
```
andypalmi: Small thing - would it be worth dropping the `.js` here to match the other imports (`@/types/common/types`)? Or is the extension intentional?

frontend/src/utils/elapsedTime.ts
```diff
@@ -10,34 +12,34 @@ const periodSeconds = {
...
 
-function dateDiff (toStringDateOrNumber, fromStringDateOrNumber = new Date()) {
-    let to = toStringDateOrNumber
+function dateDiff (toStringDateOrNumber: DateInput, fromStringDateOrNumber: DateInput = new Date()): Record<string, number> {
+    let to: DateInput = toStringDateOrNumber
     if (typeof to === 'string' || typeof to === 'number') {
-        if (!isNaN(to)) {
-            to = Number.parseInt(to)
+        if (!isNaN(to as number)) {
+            to = Number.parseInt(to as string)
```
andypalmi: Would `Number.parseInt(String(to))` read a bit cleaner here, so we don't cast a `number` as `string`? Totally fine to keep as-is if you prefer - same applies to the `from` branch below.

frontend/src/utils/snapshot.ts
```diff
@@ -84,26 +86,26 @@ function isSnapshot (snapshot) {
     // must haves
     checkProperty(snapshot, 'name', 'string', true)
     checkProperty(snapshot, 'flows', 'object', true)
-    checkProperty(snapshot.flows, 'flows', 'array', true)
+    checkProperty(snapshot.flows as SnapshotLike, 'flows', 'array', true)
```
andypalmi: Just thinking out loud - since this function is what validates the structure, do the `as SnapshotLike` casts on `snapshot.flows`/`settings` feel slightly circular? Not blocking at all given the note about avoiding a JSON schema, just curious on the intent.

andypalmi:
This looks good, and I like that the `.js` to `.ts` switch didn't require changing any of the files that import these utils. I ran the frontend tests locally and all 517 pass, and the converted files type-check cleanly.

One thing I noticed is more about the TypeScript migration overall than this PR specifically: nothing in CI actually type-checks our `.ts` files yet. The `lint` script only looks at `.js` and `.vue` files, and there's no `tsc`/`vue-tsc` script in `package.json`. The build (webpack) compiles the `.ts` files but doesn't fail on type errors, so the types we're adding aren't really being checked anywhere. For example, there's already a type error in `MqttExpertTopicHelper.ts` that no check is catching today.

Would it be worth adding a `vue-tsc --noEmit` step as part of this migration work, even if we don't make it blocking at first? Otherwise we're adding types without anything verifying them. I'm happy to open a separate issue for it so this PR can stay focused.

frontend/src/utils/elapsedTime.ts
n-lark: Solid fixed

frontend/src/utils/daysSince.ts
n-lark: fixed

frontend/src/utils/InstanceStateMutator.ts
n-lark: Totally fair fixed

frontend/src/utils/snapshot.ts
n-lark: Addressed, the per-call-site casts are gone. `checkProperty` now takes `unknown` and casts once internally.

That one cast is just to allow indexing (`object[key]` isn't allowed on `unknown`); the `typeof`/`Array.isArray` checks still do the real validation. A proper `isRecord()` type guard would drop the cast entirely, but it'd also reject truthy non-objects which a behavior change, so I left it out to keep this a pure TS conversion.

n-lark:
Hey @andypalmi solid point - I created a follow up here for that Typescript: API Modules > Enforce TypeScript checks in CI https://github.com/FlowFuse/flowfuse/issues/7552

Rest of feedback has also been addressed thanks!

andypalmi (approved):

andypalmi (approved):
Looks good! Happy to approve

PR #7420, opened 2026-06-11
Add hint to SSO SAML authentication query parameters

### LINKED ISSUE #7350: SSO: include original URL as query param to SSO request
### Description

As a follow on from https://github.com/FlowFuse/flowfuse/issues/6319, we have a user with a single SSO configuration pointed at keycloak, which then further redirects the user to their own SSO endpoint.

However, that redirect requires the user to click through and select their endpoint to get redirected.

If the original URL the user was trying to access (ie, the URL of the Node-RED instance) was to be included as a query param to the original redirect to keycloak, they can then dynamically route the user to the right endpoint.

The request for us is to provide an option to include the url in the request.

1. We will need to confirm that would work with keycloak
2. Is there a specific query parameter to use?
3. Assuming this will need to be an option on the SSO Config as to whether to include the url in the request

This will likely require more discussion with the customer to ensure any proposed solution will work on their side.

Other (See comments)

hardillb:
fixes #7350 

## Description

Adds a flag to the SAML SSO configuration that will trigger the hostname of the initial entry point URL to be sent to the SAML provider as a query string argument.

image

First pass only supports KeyCloak's `kc_idp_hint` query parameter, but there may be scope to expand to other SAML providers or to make it more generic.

Also would be useful to look at if we can send the email address when in interactive mode

## Related Issue(s)

#7350 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/pages/Login.vue
+++ b/frontend/src/pages/Login.vue
@@ -259,7 +259,8 @@ export default {
             }
         },
         async directSSO (id) {
-            window.location = `/ee/sso/login?p=${id}${this.$route.query.r ? `&r=${this.$route.query.r}` : ''}`
+            const matched = this.redirectUrlAfterLogin.match(/^\/account\/request\/([a-zA-Z0-9\-_]+)(\/editor)?/)
+            window.location = `/ee/sso/login?p=${id}${this.$route.query.r ? `&r=${this.$route.query.r}` : ''}${matched[1] ? `&t=${matched[1]}` : ''}`
         }
     }
 }
--- a/frontend/src/pages/Login.vue
+++ b/frontend/src/pages/Login.vue
@@ -259,8 +259,8 @@ export default {
             }
         },
         async directSSO (id) {
-            const matched = this.redirectUrlAfterLogin.match(/^\/account\/request\/([a-zA-Z0-9\-_]+)(\/editor)?/)
-            window.location = `/ee/sso/login?p=${id}${this.$route.query.r ? `&r=${this.$route.query.r}` : ''}${matched[1] ? `&t=${matched[1]}` : ''}`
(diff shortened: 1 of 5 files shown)


frontend/src/pages/Login.vue
```diff
@@ -259,7 +259,8 @@ export default {
             }
         },
         async directSSO (id) {
-            window.location = `/ee/sso/login?p=${id}${this.$route.query.r ? `&r=${this.$route.query.r}` : ''}`
+            const matched = this.redirectUrlAfterLogin.match(/^\/account\/request\/([a-zA-Z0-9\-_]+)(\/editor)?/)
+            window.location = `/ee/sso/login?p=${id}${this.$route.query.r ? `&r=${this.$route.query.r}` : ''}${matched[1] ? `&t=${matched[1]}` : ''}`
```
knolleary:
```suggestion
            const matched = this.redirectUrlAfterLogin.match(/^\/account\/request\/([a-zA-Z0-9\-_]+)(\/editor)?$/)
            window.location = `/ee/sso/login?p=${id}${this.$route.query.r ? `&r=${this.$route.query.r}` : ''}${matched?.[1] ? `&t=${matched[1]}` : ''}`
```

1. Ensure the regex matches the whole url
2. `matched` will be undefined if the regex doesn't match, so need to guard trying to access `[1]` otherwise this throws an error and prevents the auto-redirect from happening.

knolleary (changes requested):
Tested locally and hit a bug.

knolleary (approved):

knolleary:
Have verified the hint query parameter is included when redirecting to the IdP. This was with Entra as I don't have local keycloak setup for testing - but chrome dev tools confirmed the query param was present and correct as expected.

Issue #7522, opened 2026-06-15
Theme sync - Node-RED side


n-lark:
# Theme sync — Node-RED side

Lets an embedding parent (FlowFuse) drive the editor's light/dark preference. Generic and FlowFuse-agnostic; lives in `@node-red/editor-client` (`userSettings.js`). Requires NR 5+ (the `view-dark-theme` / `applyThemeClass` machinery).

## What this side does
1. On boot, when embedded (`window.parent !== window`): attach a `message` listener and post `{type:'request-theme'}` to the parent.
2. On `{type:'set-theme', payload:{theme}}`: apply via `setViewDarkTheme(theme)` — validates `theme ∈ auto|light|dark`, then `setSelected('view-dark-theme', theme)`, which persists to localStorage and applies through the existing `applyThemeClass`.

The request is posted *after* the listener attaches, so there's no boot race. Standalone NR (not embedded) skips it entirely.

## Contract
- **Out:** `{ type: 'request-theme' }` → parent (`'*'`).
- **In:** `{ type: 'set-theme', payload: { theme: 'auto' | 'light' | 'dark' } }`.

## Notes
- No origin check on the inbound message, matching the existing embedded-message handlers (impact is cosmetic theme only).
- Pairs with the FlowFuse-side change; absent that, NR requests, gets no reply, and uses its own stored/default theme (no-op).

PR #7523, opened 2026-06-15
[7521] Sync FlowFuse theme preference into the embedded Node-RED editor

### LINKED ISSUE #7380: Sync FlowFuse dark theme preference with Node-RED 5+
# Sync FlowFuse dark theme with Node-RED 5+

Propagates the FlowFuse light/dark preference into the embedded NR 5+ editor. One-way (FF → NR); FlowFuse is the source of truth. 

## Flow

Cross-origin iframe, so it uses `postMessage` with a handshake (avoids the boot race):

1. NR editor boots, attaches a listener, posts `{type:'request-theme'}` to the parent.
2. FlowFuse replies `{type:'set-theme', payload:{theme}}` (`auto`/`light`/`dark`; `system`→`auto`).
3. NR applies + persists via `setViewDarkTheme` → existing `view-dark-theme` / `applyThemeClass`.

FlowFuse also re-sends on its own theme change (live cross-tab case).

## Notes

- Works on the FlowFuse `forge` theme because it declares `schemes: ['light','dark']` — `hasCustomTheme()` is a deliberate carveout, not a blocker.
- NR 4 / missing either half → silent no-op (gate + unknown messages ignored).
- FlowFuse wins on (re)load: an in-editor theme change is overridden next open.
- Known limitation: brief flash of the prior theme on open (postMessage lands after first paint).

## Test plan

- [x] **Hosted instances (NR 5+): verified end-to-end.** Theme applies on editor open via the handshake and live on FlowFuse theme change; persists across reloads; the `forge` theme switches light/dark.
- [ ] ⚠️  **Remote (device) instances: not verified end-to-end.** The local device-agent harness reinstalls NR on restart (clobbering the test build) and proxies the editor through an extra frame, which blocked local testing. The code path is identical to Hosted (same service + handshake) plus a Remote-specific origin fix (`device.editor.url` is a full proxied URL, so it compares origins), so it should behave the same, worth confirming on a real NR 5+ device before release.

### LINKED ISSUE #7521: Theme sync - FlowFuse side
# Theme sync — FlowFuse side

Sends the user's FlowFuse light/dark preference to the embedded Node-RED 5+ editor iframe. One-way (FF → NR); FlowFuse is the source of truth.

## What this side does
- `post-message.service.ts` — tracks the single open editor iframe (`registerEditorTarget` / `unregisterEditorTarget`) and pushes the theme to it (`broadcastTheme`), mapping `system`→`auto`.
- `HostedInstanceEditorWrapper.vue` / `RemoteInstanceEditorWrapper.vue` — register the iframe on the NR `load` and `request-theme` messages, and re-send when `themeMode` (the theme store) changes. Both gated to NR 5+ via `isInstanceOnNR5Plus`.
- Remote compares the message origin against the *origin* of `device.editor.url` (it's a full proxied URL, not a bare origin).

## Contract
- **In:** `{ type: 'request-theme' }` from the editor → reply with the current theme.
- **Out:** `{ type: 'set-theme', payload: { theme: 'auto' | 'light' | 'dark' } }` → editor iframe.

## Notes
- Gated to NR 5+; older editors and missing parts are silent no-ops.
- FlowFuse re-asserts its preference on every editor (re)load, so a user's in-editor theme change is overridden next open.
- Pairs with the Node-RED-side change; absent that, `set-theme`/`request-theme` are ignored by the editor (no-op).

n-lark:
## Description

See https://github.com/FlowFuse/flowfuse/issues/7380#issue-4590752633

Depends on Node-RED work: #7522 (PR here: https://github.com/node-red/node-red/pull/5819)

⚠️ **This is not fully tested on a remote device** - its the same path as hosted instances I just couldn't figure out how to get a device to point to my local node-RED like I could with hosted instances. If anyone has any tips lmk. 

## Related Issue(s)

Resolves #7521
Resolves #7380

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


cstns (approved):

cstns:
code changes look good to me, please look into the failing tests

frontend/src/components/immersive-editor/HostedInstanceEditorWrapper.vue
```diff
@@ -112,6 +115,11 @@ export default {
             }
         }
     },
+    watch: {
```
cstns: ^

Issue #407, opened 2022-03-22
User can create templates


sammachin:
### Epic

#141

### Description

**As a:** user (should this be user or owner level?)

**I want to:** create a template

**So that:** I can share it with my team

### Acceptance Criteria

- [ ] Template is based on a system template (for policy)
- [ ] User can edit fields
- [ ] Template appears as available to the team
- ...

knolleary:
Not a full design, but some high-level considerations.

Today only admins can create (and edit) Templates and they are visible/usable to all users on the platform.

For a User to be able to create a Template, we need to introduce the concept of Template owner. That needs to consider whether a Template can be owned by a Team, or an individual User (or eventually both...).

That also needs to keep in mind that a user may want to share a Team with other Teams - so this isn't a simple 'ownership' question. Restricting Templates to within the same team would make a lot of this easier for MVP

A user template *must* derive from a system level template to pick up its policy. The user cannot modify the policy for any locked setting. They *may* be allowed to lock a currently unlocked setting - tighten security, not reduce it.

If the template is owned by a Team, then the Team view needs a Template section where the users can list the templates owned by the team and manage them - edit/delete/duplicate actions etc.

The Template section of the team view includes a button to 'create new template' - which opens the template create screens in the same style as the admin user workflow.

PR #7556, opened 2026-06-18
feat(expert): grouped clarifying questions UI + render robustness (#407)

andypalmi:
## Summary

Frontend for the Expert "ask before acting" clarifying-questions UX (FlowFuse/product#407), plus a couple of robustness fixes in the answer rendering.

**Grouped clarifying questions:** the Expert can ask 1-4 questions in a single turn, each rendered as its own option group (single- or multi-select), and all answers are collected before the turn is submitted. Answered questions can be edited and resubmitted; the card is disabled once a newer message arrives so a stale turn can't be answered. The options are built on the existing form components (`ff-radio-group` for single-select, `ff-checkbox` for multi-select) so they look and behave like the rest of the app.

**Cadence control:** a follow-up-questions setting in the composer menu (all at once vs one at a time), shipped to the agent via the expert context.

**Composer auto-size:** the chat input grows with its content (typed text, or an edited question loaded back in) using CSS `field-sizing: content` capped by a max-height, and only pins to a fixed height once the user drag-resizes it.

**Resource card robustness:** guarded the optional `streamable` chains in `StandardResourceCard` and `FlowResourceCard` that could throw on a null/malformed value and blank the whole message. They now degrade to "nothing to show" for that field instead, and the streaming list still advances.

## Testing

`npm run lint`, `npm run build` and `npm run test:unit:frontend` (570 tests) all pass.

Manual: grouped questions render and collect answers; a single missing value renders as a single question; cadence setting persists; composer grows with loaded/edited content.

Relates to FlowFuse/product#407


frontend/src/components/expert/components/messages/components/ErrorBoundary.vue
```diff
@@ -0,0 +1,39 @@
+<template>
```
n-lark: Hey so I'd love a second opinion from @cstns, but this looks like the only `ErrorBoundary` in the frontend, everywhere else just leans on the global `app.config.errorHandler` in `main.js`. Introducing a new local-catch pattern feels a bit sus, especially since at least one of the things it's catching (`FlowResourceCard`'s `flow.metadata.streamable.flows`, L43) is something we could fix directly. 

Boundaries make sense as a last-resort net for genuinely unexpected failures, but here it reads more like masking a fixable bug than handling it. Can we guard the actual throw and keep the boundary (if at all) as a backstop, rather than the primary fix? Want to make sure we're not setting a "wrap it in a boundary" precedent when the rest of the app doesn't work that way.

frontend/src/components/expert/components/messages/components/resources/QuestionsList.vue
```diff
@@ -0,0 +1,259 @@
+<template>
```
n-lark, line 1: Hey on styling - the unselected options read as disabled and don't lift off the background, so they don't look clickable. 

Is there a reason why we can't reuse our existing form components here? `ui-components/form/` already has `RadioGroup` (options with label + description + disabled, supports a grid layout) and `Checkbox`. That could solve the styling issues & would make it consistent with the rest of the codebase. 

Image

frontend/src/components/expert/components/ExpertChatInput.vue
```diff
@@ -181,6 +239,38 @@ export default {
...
 
             this.inputText = ''
+            // Collapse the composer back to its default height if we had grown it
+            // (180 matches the CSS min-height of .ff-expert-input).
+            if (this.composerAutoGrown) {
+                this.setHeight(180)
+                this.composerAutoGrown = false
+            }
+        },
+        growComposerToContent () {
```
n-lark: `growComposerToContent` does direct DOM manipulation which is generally sus for Vue. Could we let the `textarea` auto-size in CSS instead with like `field-sizing: content` with a `max-height`? That drops the manual measuring, the reflows, and the `composerAutoGrown` flag.

frontend/src/components/expert/components/messages/components/resources/QuestionsList.vue
cstns: yes, reuse as many existing components

frontend/src/components/expert/components/messages/components/ErrorBoundary.vue
cstns: at first glance the naming and behavior seems fishy. It's just introducing a rendering fallback to in order to not stop the word streamer if something happens. 

What is that something that could happen? 
Why not address this in the underlying component that handles the word streamer? (`@streaming-complete="onComponentComplete('flows-list')"` and `@failed="onComponentComplete('xxx')"` are called either way )

frontend/src/components/expert/components/ExpertChatInput.vue
cstns: yes, css would be the way to go

cstns (commented):
valid observations

frontend/src/components/expert/components/messages/components/ErrorBoundary.vue
andypalmi: Yeah, fair. Removed it entirely. Fixed the actual throws at the source instead (the optional streamable chains in the resource cards) and let anything genuinely unexpected fall through to the global errorHandler like the rest of the app. No more local boundary.

frontend/src/components/expert/components/messages/components/resources/QuestionsList.vue
andypalmi: Done, switched to ff-radio-group for single-select and ff-checkbox for multi.

frontend/src/components/expert/components/ExpertChatInput.vue
andypalmi: Switched to field-sizing: content with a max-height. Killed the measuring, the reflows and the flag. Only pins to a fixed height once you drag-resize.

andypalmi:
Thanks, all three addressed in the latest commits. @cstns @n-lark

andypalmi:
I've also changed to modal to be a ff-dialog in preparation of the MCP Tool Permissions Views

I've:
> Replaced the composer kebab menu with a settings gear that opens an ff-dialog. The follow-up-questions cadence control now lives in the dialog as an ff-radio-group, with a FormHeading per section so the panel can grow as more settings are added.

Screenshot 2026-06-30 at 15 08 40

n-lark:
Hey so while testing this - I am running into an issue where once I submit my answers, the chat just times out. This is on  https://7556.flowfuse.dev/. I get no console errors. Example messages, I've been able to recreate this multiple times - https://gist.github.com/n-lark/87be54af4c32064ff7223190a8545056


Screenshot 2026-06-30 at 6 35 35 AM

andypalmi:
@n-lark Unfortunately that's an issue with the Posthog feature flags to have the staging environment point to the dev agent and unrelated to this branch

EDIT: actually, let me look a bit more into this because I also got the AIMessage back but it did not render

andypalmi:
@n-lark this should now be fixed.

Root cause: rendering the reply lived in the composer's send handler, so a reply to a query submitted from the clarifying-questions card was fetched but never rendered when comms-beta (MQTT) is off and the chat falls back to HTTP. The response handling now lives in the shared handleQuery action, so every entry point (composer and the question/plan cards) renders the reply. It only surfaced where the MQTT push path is unavailable, which is why staging is hitting it right now.

I will test this myself once it lands in the staging environment, and you are welcome to give it another go too. Let me know if you still see it.

andypalmi:
@n-lark I can confirm it's working over HTTP now.

n-lark (approved):

PR #7573, opened 2026-06-18
[7324] Replace HTTP polling of instance / device status with MQTT-over-WS

### LINKED ISSUE #7324: Replace HTTP polling of instance / device status with MQTT-over-WS
# Replace HTTP polling of instance / device status with MQTT-over-WS

## Problem

Instance and device status are learned via HTTP — a fan-out on list load plus per-component poll timers. Doesn't scale, and externally-triggered state changes (crashes, autoscaler actions, admin actions) never appear in already-loaded lists.

On a 500-instance team, opening the list = **1 list call + 500 status calls + N per-row setTimeout loops**.

## Current polling

| Where | What | File |
| --- | --- | --- |
| Team instances list | N parallel `getStatus` calls on load (called *twice* — bug); per-row `<InstanceStatusPolling>` mounts | [pages/team/Instances.vue:300-326](frontend/src/pages/team/Instances.vue#L300-L326) |
| Open instance page | Transition-gated `<InstanceStatusPolling>` — misses external state changes | [pages/instance/index.vue:73](frontend/src/pages/instance/index.vue#L73) |
| Open instance logs | 5 s log poll | [pages/instance/components/InstanceLogs.vue:94](frontend/src/pages/instance/components/InstanceLogs.vue#L94) |
| Device lists | 10 s timer fetching every device (no pagination) | [components/DevicesBrowser.vue:612](frontend/src/components/DevicesBrowser.vue#L612) |
| Open device page | 5 s `getDevice` regardless of state | [pages/device/index.vue:414](frontend/src/pages/device/index.vue#L414) |

None of the lists ever reflect external state changes after load.

## Solution

Each browser opens one MQTT-over-WS connection per active team (same shape as `DeviceLog.vue` / `DevicePerformance.vue`, but a single connection scoped to the team rather than per-feature). That connection subscribes to:

- `ff/v1/<team>/p/+/status` — all instances in the team
- `ff/v1/<team>/d/+/status` — all devices in the team

Forge republishes a shaped payload (`{ id, meta: { state } }`) when state changes. Browser merges into the active list/store. No fan-out, no polling, no list refresh gap.

Forge already receives launcher and device status on `$share/platform/ff/v1/+/l/+/status` and `$share/platform/ff/v1/+/d/+/status` ([forge/comms/commsClient.js:115-127](forge/comms/commsClient.js#L115-L127)) — just need to forward, not collect.

### Backend

- New `BrokerClient` creator + credential endpoint for a per-team frontend client (per-tab session id, expert-client pattern — not the device pattern, which would revoke siblings).
- New ACL bucket gating `ff/v1/<team>/{p,d}/+/status` on team membership.
- Forwarder in `forge.comms` that reshapes launcher/device status onto the frontend topics.

### Frontend

- New team-channel module: connects on `setTeam`, disconnects on team-switch / logout.
- Subscribe to the two status wildcards on connect; route messages into the active instance / device maps.
- Delete: `getInstanceMeta()` fan-out, every `<InstanceStatusPolling>` mount, the file `components/InstanceStatusPolling.vue`, the 10 s timer in `DevicesBrowser.vue` / `DeviceActions.js`, the 5 s timer in `pages/device/index.vue`.

## Acceptance

- Opening a list of N items = 1 HTTP call. No fan-out, no per-row pollers.
- External state changes (crash, admin action, autoscaler) appear in every open browser within ~1 s.
- Open detail pages emit no recurring HTTP traffic for status.
- Team-switch tears down old subscriptions; prior-team messages don't leak.

## Out of scope

- Instance logs over MQTT (separate sub-task, mirror `DeviceLog.vue` pattern).
- Pagination of the instances list ([#7257](https://github.com/FlowFuse/flowfuse/issues/7257)).

n-lark:
## Description

See test plan: https://github.com/FlowFuse/flowfuse/issues/7324#issuecomment-4782679877
See details: https://github.com/FlowFuse/flowfuse/issues/7324#issuecomment-4783064456

> [!IMPORTANT]
> This PR will need an upstream nr-launcher fix prior to merge, see Failed restarts get stuck on restarting (live status) due to nr-launcher swallowing the start error #7606

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7324

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/comms/aclManager.js
```diff
@@ -333,6 +356,10 @@ module.exports = function (app) {
                 { topic: /^ff\/v1\/[^/]+\/t\/updated$/ },
                 // - ff/v1/<team>/u/<user>/membership
                 { topic: /^ff\/v1\/[^/]+\/u\/[^/]+\/membership$/ },
+                // - ff/v1/<team>/p/<instance>/state
+                { topic: /^ff\/v1\/[^/]+\/p\/[^/]+\/state$/ },
+                // - ff/v1/<team>/d/<device>/state
+                { topic: /^ff\/v1\/[^/]+\/d\/[^/]+\/state$/ },
```
n-lark, line 362: So devices already publish their raw status to the `…/status` topics, which forge subscribes to in order to collect it. 

If forge re-published the browser-facing copy onto that same `…/status` topic, it would receive its own message back and loop forever, so the browser updates use a separate `…/state` topic that nothing on the backend listens to.

Also named instances `.../state` for consistency. I was also considering `.../liveStatus` to imply FE, not sold on the name.

cstns: differentiating the topics to avoid loopbacks sounds sensible, the state name for consistency too

cstns: I also like the `liveStatus` approach but in the `live-state` form to adhere to the ui naming and topic format, up to you if you want to change it

forge/comms/aclManager.js
```diff
@@ -396,7 +423,11 @@ module.exports = function (app) {
                 // - ff/v1/<team>/t/updated
                 { topic: /^ff\/v1\/([^/]+)\/t\/updated$/, verify: 'checkUserIsTeamMember' },
                 // - ff/v1/<team>/u/<user>/membership
-                { topic: /^ff\/v1\/([^/]+)\/u\/([^/]+)\/membership$/, verify: 'checkUserIsTeamMember' }
+                { topic: /^ff\/v1\/([^/]+)\/u\/([^/]+)\/membership$/, verify: 'checkUserIsTeamMember' },
+                // - ff/v1/<team>/p/+/state
```
cstns, line 427: we could compact these under a single entry under `ff/v1/<team>/p|d/+/state` topic

forge/comms/devices.js
```diff
@@ -121,7 +121,11 @@ class DeviceCommsHandler {
             const teamId = this.app.db.models.Team.encodeHashid(device.TeamId)
             const startTime = Date.now()
             try {
+                const previousState = device.state
                 const payload = JSON.parse(status.status)
+                if (previousState === 'restarting' && payload?.state === 'stopped') {
+                    payload.state = 'restarting'
```
cstns: This works fine in our case in that we're not jumping back and forth between states when displaying to the user. However, I am concerned about unintended consequences it might have with status logging as a whole as it would slightly alter the way we log device states. Might be nothing but would feel better if we'd do this only when reporting the state for the UI

forge/containers/stub/index.js
```diff
@@ -119,8 +119,12 @@ module.exports = {
             } else {
                 const startTime = project.name === 'stub-slow-start' ? 6000 : module.exports.START_DELAY
                 return new Promise((resolve, reject) => {
-                    setTimeout(() => {
+                    setTimeout(async () => {
                         list[project.id].state = 'running'
+                        // mirror a real launcher reporting its settled state so inflight clears via the confirm path
+                        try {
+                            await this._app.db.controllers.Project.updateLatestProjectState(project.id, 'running')
```
cstns: this is solely for testing right?

forge/routes/api/projectActions.js
```diff
@@ -8,6 +8,15 @@
  * @namespace project
  * @memberof forge.routes.api
  */
+
```
cstns, line 11: looking at this file a bit closer and the hoops you had to make to get cohesive inflight status updates the more i think that we're doing something wrong to begin with

cstns: the current thought that's going through my mind right now is to move inflight status tracking into the container wrapper to have better control

forge/containers/stub/index.js
cstns: yes

forge/db/controllers/Project.js
```diff
@@ -810,6 +847,16 @@ module.exports = {
         } else {
             await this.setLatestProjectState(app, projectId, state)
         }
+        const project = await app.db.models.Project.byId(projectId, { barebone: true })
```
cstns: The `updateLatestProjectState` method does a `Project.byId(projectId, { barebone: true })` on every call, but the only caller outside of the test stub that i can find is the PUT /:instanceId/state endpoint, which already loads the full project in the shared prehandler. That's a redundant DB query on every state report from every instance. Might be worth passing request.project through instead of just the ID

forge/db/controllers/Project.js
```diff
@@ -810,6 +847,16 @@ module.exports = {
...
+        if (project) {
+            const inflight = await this.getInflightState(app, project)
+            const isTransition = inflight === 'starting' || inflight === 'restarting'
```
cstns: I hope a day will come when we do not use string literals for states

forge/routes/api/projectActions.js
```diff
@@ -50,8 +51,11 @@ module.exports = async function (app) {
                 await app.db.controllers.Project.setInflightState(request.project, 'starting')
                 const startResult = await app.containers.start(request.project)
                 startResult.started.then(async () => {
-                    await app.auditLog.Project.project.started(request.session.User, null, request.project)
-                    await app.db.controllers.Project.clearInflightState(request.project)
+                    if (request.project.state === 'suspended') {
```
cstns, line 54: it took me a while to get why you removed the `clearInflightState` from some of these endpoints. By clearing the inflight states within the `updateLatestProjectState` these updates become redundant and loud noise. Letting the nr-launcher be the authority and update the statuses seems to be a good shift

cstns (commented):
Having reviewed the backend changes so far, I feel I have to mention the slight paradigm shift:

Inflight state is no longer cleared optimistically by the endpoint that initiated the action. Instead, it's cleared on confirmation from two sources: nr-launcher reports its settled state (running/safe/crashed) back via the state endpoint, and updateLatestProjectState clears the inflight flag when that confirmation arrives. On the container side, the project controller clears inflight when issuing commands to the container wrapper (start, stop, restart), so any subsequent action on the instance resets the flag regardless. Error paths still clear inflight directly since neither the launcher nor a follow-up action is guaranteed if the initial call fails.
  
  Previously, the action endpoints (start, restart, rollback) would set inflight, issue the command, then immediately clear inflight in the same handler, before the launcher had actually confirmed anything. The UI could briefly show "running" for an instance that hadn't finished starting yet
  
  good stuff so far

forge/routes/api/projectActions.js
cstns: nvm, we talked it over, it wouldn't really solve anything for this particular use case

frontend/src/stores/live-status.ts
```diff
@@ -0,0 +1,31 @@
+import { defineStore } from 'pinia'
+import { computed, ref } from 'vue'
+
+export const useLiveStatusStore = defineStore('live-status', () => {
```
cstns, line 4: All the existing stores use the Options API (state, getters, actions). Could we stick with Options API here to keep things consistent with the rest of the codebase?

frontend/src/stores/live-status.ts
```diff
@@ -0,0 +1,31 @@
...
+    const instanceStatuses = ref<Record<string, string>>({})
+    const deviceStatuses = ref<Record<string, string>>({})
+    const live = ref(false)
+
+    const getInstanceStatus = computed(() => (id: string) => instanceStatuses.value[id])
```
cstns: This is an anti-pattern. A computed that returns a function defeats the purpose of computed properties. Vue can't cache the result because the actual value depends on the argument passed at call time, so it re-evaluates on every access. It's just a regular function with extra overhead.

frontend/src/utils/stateTransitions.ts
```diff
@@ -0,0 +1,7 @@
+export const TRANSITION_STATES = ['loading', 'installing', 'starting', 'stopping', 'restarting', 'suspending', 'importing', 'rollback'] as const
```
cstns: There's already an InstanceStates.js composable that classifies all instance states into running/error/stopped groups. The transition states defined here (starting, stopping, restarting, etc.) are a subset of those same lists. Could we add isTransitionState as another helper in InstanceStates.js instead of introducing a separate file? Keeps all the state classification logic in one place.

frontend/src/subscribers/team-channel.subscriber.ts
```diff
@@ -88,8 +92,12 @@ class TeamChannelSubscriber extends BaseSubscriber implements TeamChannelSubscri
         try {
             await transport.subscribe(connectionKey(teamId), [
                 `ff/v1/${teamId}/t/updated`,
-                `ff/v1/${teamId}/u/${userId}/membership`
+                `ff/v1/${teamId}/u/${userId}/membership`,
```
cstns, line 95: The subscriber pattern was designed so that each concern gets its own subscriber. Instance/device state tracking is separate from membership and team update events. Could we split this into a dedicated LiveStatusSubscriber that subscribes to the p/+/state and d/+/state topics, and keep TeamChannelSubscriber focused on membership and team updates? They can share the same connection key to reuse the MQTT connection, or use their own if needed.

  To avoid duplicating the connect/disconnect/message-routing boilerplate across both, it might be worth pulling that common logic up into BaseSubscriber, with abstract methods for the topic list and the regex-to-handler map. Each subscriber would then just declare what it cares about. That would also make adding future subscribers much simpler.

frontend/src/mixins/Instance.js
```diff
@@ -37,10 +39,22 @@ export default {
         instance (instance) {
             this.instanceChanged()
             this.setContextualInstance(instance)
-        }
+        },
+        liveInstanceStatuses: { handler: 'applyLiveStatus', deep: true }
     },
     methods: {
         ...mapActions(useContextStore, { setContextualInstance: 'setInstance' }),
+        applyLiveStatus () {
```
cstns, line 48: The applyLiveStatus method is duplicated across multiple components (Instances.vue, InstanceTile.vue, etc.), each manually reconstructing the instance object from a single state string. That logic should live in one place. Either the store itself could expose instances with live status already applied, or a composable could handle the mutation, so consumers just get a reactive object that's already up to date instead of each component reimplementing the same transformation.

cstns (changes requested):

frontend/src/mixins/Instance.js
n-lark: Waffled on this and decided against since the call sites are each slightly different but you're right the core merge is the same. Made an `applyLiveState` util with specs.

cstns: I've been complaining about the _slight_ differences between entities for a long time, it's been a blocker and pain for many useful abstractions/refactors. If it's too complicated let it be

frontend/src/utils/stateTransitions.ts
n-lark: Good call out didn't realize this was a thing, collapsed into that composable

frontend/src/stores/live-status.ts
n-lark: Ty - killed these and updated specs

frontend/src/stores/live-status.ts
n-lark: Discussed in slack - leaving as is.

frontend/src/subscribers/team-channel.subscriber.ts
n-lark: So I was under the impression that since these share a `fe-team` connection, they need to be in the same subscriber. Due to the setup logic they'd murder each other:

```
// createClient(key)
if (this.hasClient(key)) {
  await this._destroyClientUnlocked(key)   // existing key destroyed, not reused
}
// destroyClient(key) → _destroyClientUnlocked(key), no "is anyone else on this key?" check
```

So we'd either split this onto its own connection channel or change how we connect, which feels like a larger refactor we should do separate from this PR. What do you think?

forge/db/controllers/Project.js
n-lark: Great catch ty fixed

frontend/src/subscribers/team-channel.subscriber.ts
cstns: I forgot that we're destroying the client on each createClient invokation. We can separate this work in a follow up as long as it's not forgotten.

cstns: Adding more on top of the connect method in the mqttService seems sketchy, but we could go with a new `connectOrReuse` (naming up for debate)? The whole concept of subscribers was to differentiate between their scope and have them as minimal as possible.

Opening separate clients for each subscriber seems sloppy, I don't see why we wouldn't share a connection at the end of the day and just juggle with subscriptions and hooks. Something that I didn't think of when writing up the mqttService,

forge/comms/devices.js
n-lark: fixed!

forge/comms/aclManager.js
n-lark: discussed in slack leaving as is

forge/comms/aclManager.js
n-lark: gonna leave as is, ty for thoughts!

frontend/src/mixins/Instance.js
n-lark: Too late already fixed

frontend/src/subscribers/team-channel.subscriber.ts
n-lark: Follow up here: https://github.com/FlowFuse/flowfuse/issues/7608

cstns (approved):
survived a last poking session One thing i didn't get a chance to test on pre-staging is the failed palette lib use case because we're waiting on the nr-launcher version bump.

hardillb:
@cstns there is a process for adding a branch from nr-launcher to a pre-staging environment. Ask @ppawlowski  if you need help building a custom stack via the GH action

cstns:
I know there is but didn't bother since i was able to test that path locally, just not on the pre-staging and and it felt like mentioning

PR #7579, opened 2026-06-22
UI for device HTTP Tokens

### LINKED ISSUE #7566: Add httpToken UI to device/settings/security
### Description

The Instance version already has it. Need to consider refactoring to a common component.

### Epic/Story

_No response_

knolleary:
Adds UI for managing device tokens.

This is largely a copy/paste of the existing UI for Instances. But there is quite a lot of duplication now.

 - `frontend/src/pages/device/Settings/Security.vue`
 - `frontend/src/pages/instance/Settings/Security.vue`

The Hosted/Remote instance settings files already use a common component for the selection of auth type (`frontend/src/pages/admin/Template/sections/Security.vue`). These are generic options that apply to both types.

The Tokens table and all associated logic is ripe for making a common component as well; the challenge being the different endpoints needed between the two instance types. All doable, just more work to get right.

There is also duplication in the token create dialogs:
 - `frontend/src/pages/device/Settings/dialogs/TokenDialog.vue`
 - `frontend/src/pages/instance/Settings/dialogs/TokenDialog.vue`
 - `frontend/src/pages/account/Security/dialogs/TokenDialog.vue`

All three dialogs are visually identical - the difference is purely the api endpoint used to create/update the token. Again, ripe for refactoring to make a proper shared dialog.

I'm noting this here to reflect the state of the PR. I will look at how easy any of the above refactorings would be before deciding on this PR's next steps.

### DIFF
--- a/frontend/src/pages/device/Settings/Security.vue
+++ b/frontend/src/pages/device/Settings/Security.vue
@@ -13,34 +13,84 @@
             :device="device"
             :team="team"
         />
+        <div v-if="!!settings.features.httpBearerTokens && editable.settings.httpNodeAuth_type === 'flowforge-user'">
+            <FormHeading>HTTP Node Bearer Tokens</FormHeading>
+            <div v-if="hasTeamLevelTokenPermission">
+                <div v-if="deviceSupportsTokens">
+                    <ff-data-table
+                        data-el="tokens-table"
+                        :rows="tokens" :columns="columns" :show-search="true" search-placeholder="Search Tokens..."
+                        :show-load-more="false"
+                    >
+                        <template #actions>
+                            <ff-button data-action="new-token" @click="newToken()">
+                                <template #icon-left>
+                                    <PlusSmallIcon />
+                                </template>
+                                Add Token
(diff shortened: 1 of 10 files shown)


knolleary:
I'm not going to try deduplicating the code. It needs a larger refactor to make the api calls pluggable, or handled via events.

frontend/src/pages/device/Settings/Security.vue
```diff
@@ -100,6 +163,18 @@ export default {
...
+        hasTeamLevelTokenPermission () {
+            // The server checks device:edit at the team level (not application level),
+            // so we must match that here to avoid silent 403s
+            return this.hasPermission('device:edit')
+        },
+        deviceSupportsTokens () {
+            if (!this.device.agentVersion) {
+                return false
+            }
+            // return true
```
hardillb:
```suggestion
```

hardillb (approved):
Looks good and appears to work (inserts edit and removes tokens)

Just need to remove the commented out return that was obviously the same thing I asked about in slack

PR #7601, opened 2026-06-24
UI MCP tool definitions and automations service

### LINKED ISSUE #7561: Expert Agent Frontend Tool Wrappers
### Summary

Implement frontend tools (context queries, navigation, route discovery) on the first-party agent side. These wrap the agent's existing browser interactions into MCP tool definitions that are exposed via #7571 and proxied by the Forge backend (#7560).

### Why

3rd party MCP agents need situational awareness (what is the user looking at?) and the ability to direct users to relevant pages. The first-party agent already communicates with the browser over MQTT and has access to the frontend context store. These tools wrap that existing capability into structured MCP tool definitions.

By implementing these on the agent rather than the Forge backend, we keep all browser-dependent logic in one place. The Forge backend remains a pure aggregator/proxy with no direct browser communication.

### What to do

**Context tool**
* Agent queries the frontend context store via its existing MQTT channel to the browser
* Returns: current team, application, instance, device, page/route name, scope (immersive editor vs main app), and editor-specific state (selected nodes, NR version) when applicable
* Read-only: no setters, no store mutations

**Route discovery tool**
* Agent queries the frontend for the available route map
* Returns available routes with path patterns and required params
* Routes filtered by user permissions (the agent has the user's context)

**Navigation tool**
* Agent sends a navigation request to the browser via MQTT
* Browser-side handler calls `router.push()` with the provided route
* Read-annotated (not write): navigation changes what the user sees but doesn't mutate data
* Validates that the target route exists and is accessible before pushing

**Tool annotations**
* All three tools should include MCP annotations (`readOnlyHint: true` for all, since none mutate data)
* These annotations feed into the PAT enforcement layer (#7559) on the Forge backend

### Tests

* Context tool returns current route, entity IDs, and scope
* Context tool returns editor state when user is in the immersive editor
* Route tool returns available routes with path patterns and required params
* Navigation tool pushes the user to the specified route
* Navigation tool rejects invalid or inaccessible routes
* All three tools appear in the agent's tool list (#7571)
* No active browser session: structured error

### References

* Frontend context store: `frontend/src/stores/context.js` (the `expert` getter)
* Frontend route definitions: `frontend/src/routes.js`
* Expert agent MQTT tool exposure: #7571
* Forge backend proxy: #7560
* Parent story: #7424

cstns:
## Description

Implements the frontend-side UI tool definitions and dispatch service for the first-party agent. This covers the three tools described in #7561 (context, route discovery, navigation) but with an architectural change: tool definitions and execution live in the frontend, not on the agent.

  The original issue assumes tools are implemented on the agent side and exposed over MQTT. We moved them to the frontend because the 3rd party agent serves multiple Forge instances (cloud and self-hosted) running different frontend versions, each with potentially different UI capabilities. Having the frontend declare its own tools means the agent doesn't need to hardcode version-specific knowledge. It discovers tools dynamically per session.

  What's here:

  - frontend/src/mcp/tools/ with three tool files, each exporting an array of tool definitions using JSON Schema (serializable for MQTT transport)
    - context.ts - ui.get-context: reads from the context store's expert getter
    - routes.ts - ui.list-routes: returns available routes from the router with name, path, and metadata
    - navigation.ts - ui.navigate: validates route exists, calls router.push()
  - frontend/src/mcp/tools/index.ts aggregates all tool files
  - frontend/src/services/automations.service.ts - lifecycle service following the existing BaseService pattern. getToolDefinitions() returns definitions without handlers (for MQTT discovery). dispatch(toolName, args) looks up and executes the handler.
  - MCP types under frontend/src/types/mcp/
  - Wired into service registry and ServiceInstances type

  The service is self-contained and ready to be called when the MQTT inflight bridge lands. $services.automations.getToolDefinitions() for discovery, $services.automations.dispatch(toolName, args) for execution.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/7561

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/mcp/tools/navigation.ts
```diff
@@ -0,0 +1,37 @@
+import type { McpToolDefinition } from '@/types'
+
+const tools: McpToolDefinition[] = [
+    {
+        name: 'ui.navigate',
+        description: 'Navigate the user\'s browser to a specific page. Takes a route name and optional params. Use ui.list-routes to discover valid route names and their required parameters.',
+        annotations: { readOnlyHint: true, destructiveHint: false },
```
n-lark, line 13: Should this be `readOnlyHint: false` since it can navigate for the user??

frontend/src/mcp/tools/navigation.ts
```diff
@@ -0,0 +1,37 @@
...
+                params: {
+                    type: 'object',
+                    description: 'Route parameters (e.g. { id: "abc123" } or { team_slug: "my-team" })',
+                    additionalProperties: { type: 'string' }
+                }
+            },
+            required: ['route']
+        },
+        async handler (args, { router }) {
+            const { route: routeName, params } = args as { route: string, params?: Record<string, string> }
```
n-lark, line 30: This `as` is sus - the type is `unknown` then we force it here. Should we be validating this shape?

n-lark (approved):

Issue #409, opened 2022-03-22
Share templates


sammachin:
### Description

Related to #141 but this may have multiple separate stories to split out

**As a:** user 

**I want to:** share templates I have created with other teams.

**So that:** Others can build on my work

Considerations:
- Can any user share a template or is the owner only?
- There should be  2 types of sharing, public and private
- In public the user is given a url for their template that they can publish, send so any other user on the platform can open and use the template, to allow for creating tutorial blog posts etc.
- In private the user explicitly sends the template to another named team, the owner is then invited to accept or reject the template.
- Public templates would only be scoped to system boundary, so FlowForge.cloud template urls wouldn't be usable on a local instance.

PR #7635, opened 2026-06-29
feat(expert): plan mode — propose a plan before acting (#408, #409)

andypalmi:
## Summary

Frontend for Expert **plan mode** (FlowFuse/product#408, FlowFuse/product#409): the Expert proposes a plan before making any changes, and acts only once the user approves it.

- **Plan mode toggle:** an always-visible toggle chip in the composer. When enabled, the Expert proposes a plan instead of acting; the flag is shipped to the agent via the expert context.
- **Plan card:** the proposed plan renders as its own card (Markdown via RichContent) with four actions:
  - **Approve** — exits plan mode and proceeds with the plan (a follow-up acting turn).
  - **Edit** — loads the plan markdown into the composer for direct editing and resubmission.
  - **Request changes** — focuses an empty composer to describe a change in words; the agent re-proposes an updated plan.
  - **Reject** — abandons the plan.
- A plan card from a past turn is disabled once a newer message arrives (shared with the questions cards).
- Fixes a missing-prop console warning: the plan card now passes the message/answer uuids that RichContent requires.

## Stacking

This PR was stacked on #7556 (clarifying questions), which has now merged. It now targets `main` directly and the diff shown here is plan-mode only.

## Testing

- `npm run lint`, `npm run build` and `npm run test:unit:frontend` (570 tests) all pass.
- Manual: toggling plan mode, and Approve / Edit / Request changes / Reject on the plan card; no console warning on plan render.

Closes: 
* FlowFuse/product#406
* FlowFuse/product#408
* FlowFuse/product#409
* FlowFuse/product#410


frontend/src/stores/product-expert.js
```diff
@@ -181,7 +188,18 @@ export const useProductExpertStore = defineStore('product-expert', {
...
+            // a change to a proposed plan. Bumping a counter lets the composer react each
+            // time, including repeated requests.
+            this.planChangeRequest++
+        },
+        resetComposer () {
+            // Signal the chat composer to clear its input. Bumping a counter lets the
+            // composer react each time, including repeated resets.
+            this.composerReset++
+        },
+        async handleQuery ({ query, contextOverrides }) {
```
n-lark: What do we use `contextOverrides` for? It's piped through to `sendMqttQuery` but never used in that fn, while `sendHttpQuery` merges it. I don't see anywhere passing it to `handleQuery` either. Is this leftover, or meant to be wired up?

frontend/src/stores/product-expert.js
```diff
@@ -1085,7 +1107,7 @@ export const useProductExpertStore = defineStore('product-expert', {
         }
     },
     persist: {
-        pick: ['shouldWakeUpAssistant', 'questionCadence'],
+        pick: ['shouldWakeUpAssistant', 'questionCadence', 'planMode'],
```
n-lark: `onPlanApprove` calls `setPlanMode(false)`, and `planMode` is persisted. So approving any plan permanently turns off the user's plan-mode preference. Do we wanna persist this setting?

frontend/src/stores/product-expert.js
```diff
@@ -181,7 +188,18 @@ export const useProductExpertStore = defineStore('product-expert', {
...
+            // Signal the chat composer to focus an empty input so the user can describe
+            // a change to a proposed plan. Bumping a counter lets the composer react each
+            // time, including repeated requests.
+            this.planChangeRequest++
+        },
+        resetComposer () {
+            // Signal the chat composer to clear its input. Bumping a counter lets the
+            // composer react each time, including repeated resets.
+            this.composerReset++
+        },
```
n-lark: These `planChangeRequest++` / `composerReset++ `counters feel a bit sus. They're basically using store state as an event bus for the composer to watch, which nothing else in the stores does.

We already do this cleaner right here with `pendingInput` (set → consumed → cleared). Could we model these the same way instead of counters? Would kill a couple of the new watchers too.

n-lark (changes requested):
Hey flagged a couple things. <s>I wasn't able to test this since we don't have a staging env for non mainline PRs. Lmk when this is pointed at main and I will test. </s> nvm I am wrong, will test now as well and update. 

One general note - there are a lot of code comments here, and Serban got on me for this exact thing too so you're not alone. When they get this verbose they kinda make the code harder to read. My rule of thumb is if you can figure it out from reading the code, don't comment it. Might be worth dropping into a `CLAUDE.md`.

n-lark:
UI issue here - besides that functionality seems solid. 

1. Sending a plan to be edited exceeds the height of the expert chat, we probably need a max height and a scroll here.  

Screenshot 2026-06-30 at 11 07 05 AM

frontend/src/components/expert/components/ExpertChatInput.vue
```diff
@@ -321,6 +384,42 @@ export default {
     align-items: center;
 }
 
+// Reuses the shared DefaultChip for its bg, border, active state and theming; the only
+// styling here is sizing the toggle switch in the #icon slot, since the switch has no size
+// prop. It is a visual indicator only (pointer-events disabled); the chip handles the click.
+.plan-mode-chip {
```
n-lark, line 364: `:deep()` into `.ff-toggle-switch-button` to resize it feels fragile. This'll break quietly if `ff-toggle-switch` ever changes. Since it has no size option, could we add a small/size variant to the component itself instead of reaching into its guts from the chip?

frontend/src/stores/product-expert.js
andypalmi: Leftover, good catch. It was piped through but only sendHttpQuery ever merged it and nothing passed it to handleQuery, so it was dead. 
**Removed it** from handleQuery, sendQuery, sendHttpQuery and sendMqttQuery. Context comes solely from the context store getter now.

frontend/src/stores/product-expert.js
andypalmi: Agreed, we don't want approving a plan to wipe the user's preference. 
**Dropped planMode from the persisted keys, so it's session-only now** (persist is just shouldWakeUpAssistant and questionCadence). 
Toggling it off on approve no longer sticks across reloads.

frontend/src/stores/product-expert.js
andypalmi: Yeah, those counters were using store state as an event bus, which nothing else does. 
**Replaced** `planChangeRequest` / `composerReset` with a single `composerCommand` string set/consumed/cleared exactly like pendingInput, via `setComposerCommand`('request-plan-change' | 'reset' | null).
Collapsed the two watchers into one that reads the command and nulls it.

frontend/src/components/expert/components/ExpertChatInput.vue
andypalmi: Fair, reaching into .ff-toggle-switch-button was fragile. Added a proper size prop to ToggleSwitch (normal | small) with a ff-toggle-switch--small modifier in ff-components.scss. 
The chip now just uses <ff-toggle-switch size="small" />, no :deep into its internals.

andypalmi:
> One general note - there are a lot of code comments here, and Serban got on me for this exact thing too so you're not alone. When they get this verbose they kinda make the code harder to read. My rule of thumb is if you can figure it out from reading the code, don't comment it. Might be worth dropping into a CLAUDE.md.

Fair, and noted. Did a pass to cut the comments back to just the non-obvious stuff and left the code to speak for itself where it can.

andypalmi:
Pushed the review fixes:

- Removed the dead `contextOverrides` param.
- Stopped persisting `planMode`, so approving a plan no longer changes the saved preference.
- Replaced the composer counters with a single `composerCommand` modelled on `pendingInput`.
- Added a `size` variant to `ToggleSwitch` instead of `:deep`-resizing it.
- Trimmed the verbose comments.
- Capped the composer height when a long plan is loaded via "Edit" so it scrolls instead of growing (the `min-height: 0` was missing on the textarea's flex parent).

One change beyond the review: I removed the logic that limited plan mode to the immersive editor, so it's now available across the app rather than only inside an instance/device editor.

Plan mode now adapts to where the user is: it uses the `scope` already sent in the Expert context (`immersive` for an instance/device editor, `ff-app` otherwise) to change what the agent proposes and what it can act on depending on whether it's in the FlowFuse app or the editor. This is the groundwork for the follow-up PR that adds the platform automation tools and human-in-the-loop approvals across both the platform tools and the flow-building tools. See https://github.com/FlowFuse/flowfuse/pull/7639

Here is a short recording of the plan-mode flow in action:

https://github.com/user-attachments/assets/e134a620-bd1e-49da-aed5-3ec67c558bd9

@n-lark would you mind giving it a last look?

n-lark (approved):

andypalmi:
PR was approved but we will need to wait for agent deployment to production before merging, otherwise users will see plan mode but will not be working until agent is updated

cc: @Steve-Mcl @cstns @allthedoll

cstns (approved):

Issue #421, opened 2022-03-28
Licence Expires Calculation wrong


sammachin:
### Current Behavior

This is the info displayed for the license on staging at /admin/overview
![Screen Shot 2022-03-28 at 9 09 24 AM](https://user-images.githubusercontent.com/187645/160355445-3f83f8d1-4e55-4c29-9c52-0e6a94497bd0.png)
Its currently `2022-03-25T08:15` Which means that the 25th March 2023 is in 11 months 28 Days (14 hours)

### Expected Behavior

The Expires should display the correct amount of time remaining

### Steps To Reproduce

goto `/admin/overview` on a licensed system,
Ask Alexa "how long till 25th march 2023 at 23:59" :)

### Environment

- FlowForge version: 0.3.2 (may be 0.3.3)
- Node.js version: 16
- npm version: 8
- Platform/OS: Mac
- Browser: Chrome

PLCMercenary:
How do I get a license or access to a licensed platform manager instance?

knolleary:
For local development, you can use the license string here: https://github.com/flowforge/flowforge/blob/main/forge/licensing/index.js#L17

PLCMercenary:
is that in the handbook? I checked the docs and more than one readme.

knolleary:
No - you didn't miss anything. It needs adding.

PR #7639, opened 2026-06-30
feat(expert): human-in-the-loop tool permissions (#421)

andypalmi:
## Human-in-the-loop tool permissions for the Expert

Implements per-tool human-in-the-loop permissions for the Expert's flow-building tools, in the immersive editor, as described in FlowFuse/product#421. The builder (and their team role) controls which flow-building actions the Expert may run, which need approval, and which are off limits, so it never makes a change they would not have allowed.

Stacked on #7635 (`feat/408-expert-plan-mode`), which is the base of this PR and should merge first.

### What it does

- **Inline approval card** in chat when a tool's policy is *Ask*: friendly tool name, action type (Read / Write / Delete) and the concrete call parameters as prettified JSON, with **Allow / Always allow / Deny / Always deny**. "Always allow" and "Always deny" apply for the rest of the current chat and reset on Start Over and on refresh; after a choice the card shows exactly what was picked and collapses the payload. The agent pauses on the round-trip with no session timeout, however long the user takes; the chat stop button cancels it (treated as denied).
- **Per-team settings** (in the Expert settings dialog): permissions are saved per team. Each tool group (flow-building, platform) has its own always-visible **default permissions** — a per-action-type default (Always allow / Ask / Always deny) for read, write and delete — and collapses its individual-tool overrides behind an accordion. The policy control is a fast three-button toggle rather than a dropdown.
- **Per-tool overrides that stay put**: a permission set on an individual tool overrides its type default and keeps that setting until reset. Each type default shows an "N set individually" count of the tools in that scope carrying their own saved permission, and a **Reset** action beside it returns those tools to the default. Session-only grants are excluded from the count — they appear per-tool and reset on their own.
- **Make permanent**: a tool granted only for the current chat shows a "Make permanent" action to save that choice for the team.
- **Section ordering by context**: flow-building tools lead in the immersive editor; platform tools lead in the app.
- **Role inheritance, fail-closed**: read-only team members cannot enable or trigger write/delete tools and see why; the agent also fails closed server-side.
- **Version gating**: each tool carries a min/max nr-assistant version. Tools render as available, "update required", or deprecated against the instance's nr-assistant version. Versioned variants (e.g. Manage Groups v1/v2) collapse into one row and resolve to the in-range variant, nudging an update to the newest variant's min version when behind.

### Architecture

- The agent decides policy at the `toolsNode` seam (sibling of the plan-mode gate): role check first, then per-tool policy. `allow` runs, `deny` feeds the denial back to the model so it adapts and explains, `ask` publishes `expert:tool-approval` and awaits the browser's decision.
- The flow-building tool **catalog** is served over the agent's `GET /mcp/flow-tools` endpoint (friendly name + scope + version window only). Forge exposes `GET /api/v1/expert/mcp/tools`, which proxies that endpoint and returns the **merged** catalog: FlowFuse platform tools are curated into the same array (tagged as a platform group) — wired and commented out until the platform-tool work is merged, at which point it's a one-line switch. Every chat response carries a **hash** of the flow-building catalog; the browser refetches only when the hash diverges, so it stays correct across rolling deploys where instances can be on different versions.
- Saved per-team choices and per-chat session grants live in the existing product-assistant / product-expert Pinia stores.

### UI

The settings panel follows existing FlowFuse patterns: `FormHeading` for section titles, `ff-data-table` for the defaults and tool lists, `ff-accordion` to collapse the per-tool detail, and the shared three-button toggle for each policy control, so each tool name lines up with its own control across the row border.

### Out of scope (follow-ups)

- Admin-configurable team-wide default policy (DB migration + admin UI + server enforcement).
- Enabling platform (non-flow-building) tools in the catalog, once the platform-tool work is merged into the agent. The forge endpoint has the curation wired and commented, ready to switch on.

### Testing

- Build + color/eslint lint green.
- Automated unit tests:
  - `forge/ee/routes/expert/index_spec.js` (new `MCP tools Endpoint` block): `GET /mcp/tools` auth (401 for instance/device tokens), team-access (404 for non-members), missing `teamId` (400 from the querystring schema), the flow-tools catalog + hash proxy (asserting the upstream `/mcp/flow-tools` URL and service token), the empty-response defaults (`catalog: []`, `hash: null`), and upstream error-status propagation.
  - `frontend/src/stores/product-assistant.spec.js` (new tool-permissions block): the permission-resolution engine, i.e. the `classOf`/`groupOf` helpers, per-team class defaults, saved vs. session policy resolution, `resolvedToolPermissions`, version gating (`toolAvailabilityFor`), catalog/preference/override mutations, `resetGroupClassPreferences`, `promoteSessionOverride`, and the pending-approval registry.
  - `frontend/src/stores/product-expert-tool-permissions.spec.js` (new): catalog fetch (success / no-team / error) and the approval round-trip (session short-circuit, resolve, always-allow, always-deny, cancel).
- Manual: catalog populates on opening the immersive editor; *Ask* shows the card and pauses with no timeout (Allow applies to canvas, Deny explains gracefully); "Always allow"/"Always deny" apply for the chat and reset on Start Over / refresh; "Make permanent" saves the choice per team; a per-tool override holds against changing the type default, the "N set individually" count reflects it, and Reset returns those tools to the default; per-team settings stay separate across team switches and survive reload; read-only role sees write/delete disabled with a reason and cannot trigger them; leaving immersive hides the panel and stops sending permissions; chat stop while a card is open recovers cleanly.

Requires matching agent-side changes.

Refs FlowFuse/product#421

#### Expert Settings with Tools permissions

Screenshot 2026-07-01 at 17 24 10

#### Counter of how many tools have different permissions than their scope's permission

Screenshot 2026-07-01 at 17 24 32

#### Permissions reset and change behaviour per scope 

https://github.com/user-attachments/assets/bcb172b3-bafd-4138-b32f-ed06ab208434

#### Option to save a permission set for the current session

Screenshot 2026-07-01 at 17 32 07

https://github.com/user-attachments/assets/d0f75d55-bf7d-4e76-9a82-92fdba237264

#### Approval Cards

https://github.com/user-attachments/assets/cc27660f-e915-48d7-8ddd-9c0a8398bdab

Screenshot 2026-07-01 at 17 31 43


frontend/src/stores/product-expert.js
```diff
@@ -395,16 +436,92 @@ export const useProductExpertStore = defineStore('product-expert', {
...
+        },
+        // Deny every open approval (used when the user stops the chat) so the agent's
+        // approval wait unblocks instead of hanging on an abandoned prompt.
+        cancelPendingToolApprovals () {
+            const permStore = useProductAssistantStore()
+            if (!permStore.hasPendingApprovals()) return
+            for (const m of this._agentStore.messages) {
+                if (!Array.isArray(m.answer)) continue
+                for (const a of m.answer) {
+                    if (a.kind === 'tool-approval' && a.status === 'pending') a.status = 'denied'
```
n-lark: Couldn't test this without the agent side, but reading the code: on Stop, `cancelPendingToolApprovals` sets `status='denied'` on the store answer, but the card renders a shallow copy of it `useStreamingList({ shallow: true })`, so its status prop never update. Worth confirming, but looks like `Stop` won't resolve an open approval card.

frontend/src/stores/product-expert.js
```diff
@@ -395,16 +436,92 @@ export const useProductExpertStore = defineStore('product-expert', {
...
+            }
+        },
         async startOver () {
             const agentStore = this._agentStore
             agentStore.sessionId = uuidv4()
             agentStore.messages = []
 
+            // A new chat drops the per-session tool grants ("Always allow/deny for this chat").
+            useProductAssistantStore().clearSessionToolOverrides()
+
```
n-lark, line 594: Do we want a `cancelPendingToolApprovals()` here too?

n-lark (commented):
Hey so I cannot test the approve/deny part of this in that chat due to the staging env not having posthog synced up. The permissions page under settings UI looks fine to me but I don't feel comfortable approving this since I cannot test and am unfamiliar with this feature. I'd recommend @cstns or @Steve-Mcl to takes a look.

frontend/src/stores/product-expert.js
andypalmi: Good catch, confirmed. The card renders a detached streaming copy of the answer (AiMessage uses useStreamingList with shallow: true), so writing the status onto the store message never reached it. On Stop the buttons stayed live.

Fixed by recording the outcome in a reactive per-id map (toolApprovalStatuses) on the product-assistant store. AnswerWrapper now feeds the card its status from that map, so an external resolution (Stop / Start Over) updates a card the user never pressed. localStatus stays for instant feedback on the user's own press. Added store and product-expert tests covering the denied-on-cancel path.

frontend/src/stores/product-expert.js
andypalmi: Yes. startOver now calls cancelPendingToolApprovals() first, so any approval still awaiting a decision resolves (as denied) and the agent's paused tool call unblocks instead of hanging on a message we are about to drop. It also clears toolApprovalStatuses alongside the session overrides.

andypalmi:
Thanks for the review. Pushed a fix for the Stop issue you spotted.

Root cause: the approval card renders a detached streaming copy of its answer (AiMessage uses useStreamingList with shallow: true), so a status written onto the store message never reached the card. Clicking Allow/Deny worked only because the card tracks its own localStatus; the external Stop path had no way in, so the buttons stayed live.

Fix: approval outcomes are now recorded in a reactive per-id map (toolApprovalStatuses) on the product-assistant store, and AnswerWrapper feeds the card its status from that map. That covers external resolutions, Stop and Start Over, on a card the user never pressed. Start Over also cancels open approvals first so the paused tool call unblocks. This is in-memory session state only, not persisted, same lifecycle as the session overrides. Added store and product-expert tests for the denied-on-cancel path.

On testing: understood you cannot exercise approve/deny on staging without the agent side synced. @cstns or @Steve-Mcl, a second look would be welcome given the reactivity change.

forge/ee/routes/expert/index.js
```diff
@@ -556,6 +582,81 @@ module.exports = async function (app) {
             reply.code(error.response?.status || 500).send({ code: error.response?.data?.code || 'unexpected_error', error: error.response?.data?.error || error.message })
         }
     })
+
+    /**
+     * Retrieve the curated tool catalog for the Expert's human-in-the-loop permissions UI
+     * (#421). Returns the merged catalog for both sections the UI shows:
```
cstns, line 590: keep these concise

forge/ee/routes/expert/index.js
```diff
@@ -14,6 +14,30 @@ const { filterAccessibleMCPServerFeatures } = require('../../../services/expert.
 /** @type {typeof import('../../../comms/devices.js').DeviceCommsHandler} */
 const getDeviceComms = (app) => { return app.comms?.devices }
 
+/**
+ * Maps a platform automation tool's wire definition into a catalog entry for the
+ * Expert permissions UI (#421). Platform tools carry standard MCP annotations
```
cstns, line 19: a little bit more concise

forge/ee/routes/expert/index.js
```diff
@@ -556,6 +582,81 @@ module.exports = async function (app) {
...
+            const catalog = response.data?.catalog || []
+
+            // Merge in the FlowFuse platform tools. They are global (no per-team filtering)
+            // and served from the handler singleton already constructed on app.comms, so we
+            // reuse it rather than newing one up — constructing re-registers its MQTT event
+            // listener. getToolDefinitions() is synchronous and takes no args.
+            const platformHandler = app.comms?.platformAutomation
+            if (platformHandler) {
+                const platformDefs = platformHandler.getToolDefinitions() || []
+                catalog.push(...platformDefs.map(curatePlatformTool))
```
cstns, line 654: if we add titles on the mcp tools, do we need to map these?

frontend/src/components/elements/ToggleButtonGroup.vue
```diff
@@ -61,6 +61,13 @@ export default {
             type: String,
             required: false,
             default: 'value'
+        },
+        // Size of the underlying buttons (passed through to ff-button). Defaults to
+        // 'medium' to match existing usages; 'small' suits dense contexts like tables.
+        size: {
```
cstns, line 67: the prop name is self explanatory, adding possible options should be enough

frontend/src/components/expert/components/messages/components/resources/JsonViewer.vue
```diff
@@ -0,0 +1,170 @@
+<template>
+    <div class="json-viewer">
```
cstns, line 2: we should scope this as a generic component otherwise we'll end up replicating it. Please move it under frontend/src/components. It's not as abstracted as i would have liked but we can do that later on

frontend/src/components/expert/components/messages/components/resources/JsonViewer.vue
```diff
@@ -0,0 +1,170 @@
...
+            v-show="!collapsed"
+            class="json-viewer__scroll"
+            :class="{ 'json-viewer__scroll--wrap': wrapped }"
+        >
+            <pre class="json-viewer__content">{{ text }}</pre>
+        </div>
+    </div>
+</template>
+
+<script>
```
cstns, line 42: not a blocker but we should start moving towards ts support with the composables api to make our lives easier

frontend/src/components/expert/components/messages/components/resources/JsonViewer.vue
```diff
@@ -0,0 +1,170 @@
...
+// Read-only single-value JSON viewer with prettify, word-wrap and optional collapse,
+// following the presentation of SnapshotDiffChangePanel without its two-sided diff
+// machinery. Used for payloads such as the tool approval card's call parameters.
+import { ChevronRightIcon } from '@heroicons/vue/20/solid'
+
+const LONG_LINE_THRESHOLD = 50
+
+// Stringify that never throws — if the payload can't be serialised for any
+// reason, show a plain error instead of breaking the surrounding card.
+function safeStringify (value) {
```
cstns, line 52: this belongs in the string composable if there's no equivalent for it already.  see frontend/src/composables/strings/String.js

frontend/src/components/expert/components/messages/components/resources/ToolApprovalCard.vue
```diff
@@ -0,0 +1,179 @@
...
+        }
+    },
+    emits: ['approve', 'allow-always', 'deny', 'deny-always', 'streaming-complete'],
+    data () {
+        return {
+            // What the user pressed on this card, for instant feedback without waiting on a
+            // store round-trip. The `status` prop also reflects the outcome — AnswerWrapper
+            // feeds it from the store's reactive per-id map, which covers external
+            // resolutions (chat stop / Start Over) this card never pressed.
+            localStatus: null
```
cstns, line 85: we should be storing the decision as a data attribute and the localStatus should be a computed prop based on it. You're repeating yourself in the decide and outcomeLabel methods/props

frontend/src/components/expert/components/messages/components/resources/ToolApprovalCard.vue
```diff
@@ -0,0 +1,179 @@
...
+        // Collapse the payload once a decision is made; the header toggle stays live
+        // so the user can re-expand it at any time.
+        payloadCollapsed () {
+            return this.effectiveStatus !== 'pending'
+        }
+    },
+    mounted () {
+        // Interactive card, not streamed text — signal completion so the
+        // AnswerWrapper streaming order can advance.
+        this.$nextTick(() => this.$emit('streaming-complete'))
```
cstns, line 118: this is a workaround to allow the entire message to get streamed. When first implemented we didnt take into consideration that certain elements should just be shown (as the approval cards need to be) and would just block the message stream. 

The way forward should be to ad a concept of 'do not stream' or similar on message parts and just skip streamin in the orchestrator and not mock the streamed even. this will scale better with any futture non-streamable components we'l add in the future

frontend/src/components/expert/components/ExpertChatInput.vue
```diff
@@ -90,6 +90,7 @@
             confirm-label="Done"
             :can-be-canceled="false"
             data-el="expert-settings-dialog"
+            boxClass="max-w-[54rem]!"
```
cstns, line 93: why is this needed?

frontend/src/components/expert/components/ToolPermissionsSettings.vue
```diff
@@ -0,0 +1,468 @@
...
+            </div>
+
+            <ff-accordion v-if="group.tools.length" label="Individual tools" :set-open="false">
+                <template #content>
+                    <ff-data-table :columns="toolColumns" :show-search="false">
+                        <template #rows>
+                            <ff-data-table-row v-for="tool in group.tools" :key="tool.familyKey">
+                                <ff-data-table-cell class="tool-col">
+                                    <div class="tool-permissions__cell">
+                                        <span class="tool-permissions__title">{{ tool.displayName }}</span>
```
cstns, line 65: this will be hard to maintain going forward. Encapsulating smaller self-contained vue components add not only ease of maintenance but also local state that can be managed at the component level, allowing you to avoid  the sausage types of methods that are within this component.

frontend/src/stores/context.js
```diff
@@ -113,7 +115,18 @@ export const useContextStore = defineStore('context', {
                 supportsPlatformAutomation: useAccountSettingsStore().featuresCheck?.isExpertPlatformAutomationFeatureEnabled ?? false,
                 supportsPlatformUIAutomation: useAccountSettingsStore().featuresCheck?.isExpertPlatformAutomationFeatureEnabled ?? false,
                 questionCadence: useProductExpertStore().questionCadence,
-                planMode: useProductExpertStore().planMode
+                planMode: useProductExpertStore().planMode,
+                // Signals that this FlowFuse version implements human-in-the-loop tool
```
cstns: no need for such a long winded note, more concise

frontend/src/stores/product-assistant.js
```diff
@@ -6,6 +6,38 @@ import { useContextStore } from '@/stores/context.js'
 
 const MAX_DEBUG_LOG_ENTRIES = 100 // maximum number of debug log entries to keep
 
+// --- Expert tool permissions (human-in-the-loop, #421) -----------------------
```
cstns, line 9: these feel like they belong in a helper

frontend/src/stores/product-assistant.js
```diff
@@ -158,7 +190,26 @@ export const useProductAssistantStore = defineStore('product-assistant', {
         // internally; external consumers should read this.debugLog (the getter).
         debugLogEntries: [],
         editorState: { ...buildInitialEditorState() },
-        pendingRequests: new Map() // key is transactionId, value is { resolve, reject, timeout, timestamp, type, action, params }
+        pendingRequests: new Map(), // key is transactionId, value is { resolve, reject, timeout, timestamp, type, action, params }
+        // Expert tool permissions (HITL, #421). The catalog + hash are refreshed from
+        // the agent; defaults + preferences are the user's choices (persisted below).
+        toolCatalog: [],
+        toolCatalogHash: null,
```
cstns, line 197: these are expert concerns not assistant, or am i wrong?

andypalmi: I will create a follow up

frontend/src/stores/product-assistant.js
andypalmi: Noted

frontend/src/stores/context.js
andypalmi: Will update in follow up

frontend/src/components/expert/components/ToolPermissionsSettings.vue
andypalmi: Like you said in the call, we'll tackle this in a further iteration

frontend/src/components/expert/components/ExpertChatInput.vue
andypalmi: To make the expert dialog wider so that the permissions table has more breathing room

frontend/src/components/expert/components/messages/components/resources/ToolApprovalCard.vue
andypalmi: Noted

frontend/src/components/expert/components/messages/components/resources/ToolApprovalCard.vue
andypalmi: Will improve in follow up pr due to time constraints

frontend/src/components/expert/components/messages/components/resources/JsonViewer.vue
andypalmi: Follow up issue later will be raised

frontend/src/components/expert/components/messages/components/resources/JsonViewer.vue
andypalmi: Yep

frontend/src/components/elements/ToggleButtonGroup.vue
andypalmi: Are you suggesting this should be an enum?

Steve-Mcl (approved):
Based on Serbans review, I checked over remaining items and agreed follow ups with Andrea (he is to raise them asap)

PR #7645, opened 2026-07-01
Mcp over mqtt

cstns:
## Description

Final bits and bobs

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/comms/platform.js
```diff
@@ -0,0 +1,75 @@
+// /**
```
Steve-Mcl: delete this unused file

forge/comms/commsClient.js
```diff
@@ -46,12 +46,112 @@ class CommsClient extends EventEmitter {
...
+
+                if (topicParts[2] === 'expert') {
+                    const userId = topicParts[3]
+                    const sessionId = topicParts[4]
+                    const channel = topicParts[5] // platform
+                    const channelCommand = topicParts[6] // dynamic value, e.g. mcp:call-tool or mcp:read-resource
+                    const direction = topicParts[7] // request or response (only for inflight channels)
+
+                    // these are common for both insights and platform automations
+                    const payload = JSON.parse(message.toString())
```
Steve-Mcl: should we do safe try parse?

as this is a callback - risk app crash without try..catch?

Steve-Mcl (changes requested):
initial quick glance - major part of this is an isolated token as discussed.

forge/comms/commsClient.js
andypalmi: Small robustness question on this parse: there's no try/catch around it, so a malformed payload on the topic would throw out of the message handler rather than going through the error path (the agent would then just time out with no trace). Would it be worth guarding it? Something like:

```js
let payload
try {
    payload = JSON.parse(message.toString())
} catch (err) {
    app.log.warn(`Ignoring malformed expert payload on ${topic}: ${err.message}`)
    return
}
```

If the ACL and agent already guarantee well-formed JSON here, feel free to ignore.

forge/comms/devices.js
```diff
@@ -63,6 +79,141 @@ class DeviceCommsHandler {
         // Listen for any incoming device status events
         client.on('status/device', (status) => { this.handleStatus(status) })
         client.on('response/device', (response) => { this.handleCommandResponse(response) })
+
+        // Handle expert inflight requests sent from the FF Expert Agent - intended for an MCP server on a specific remote instance
+        client.on('request/device/expert/insight', async (userId, command, /** @type {MCPServerDetails} */ mcpServer, mcpDefinitionKind, mcpDefinition, data, onSuccess, onError) => {
```
andypalmi, line 84: This handler and the `request/instance/expert/insight` one in instances.js read very close to each other: the team/app/instance ownership checks, the `filterAccessibleMCPServerFeatures` call, the resource-template URI resolution, and the endpoint/token building all look the same. Would it make sense to pull the shared parts into a helper (perhaps next to `filterAccessibleMCPServerFeatures` in forge/services/expert.js) and pass in just the device-vs-instance transport? Or is the divergence between the two enough that keeping them separate is cleaner? Mainly thinking about a future change to the access checks needing to land in both places.

andypalmi (commented):
A few questions from reading through the MQTT handlers, nothing blocking. Mostly trying to understand the intent.

forge/comms/platformAutomation.js
```diff
@@ -0,0 +1,121 @@
...
+
+                const tool = this.findTool(toolName)
+                if (!tool) {
+                    return onError(
+                        `Unknown platform tool: ${toolName}`,
+                        'MCP_PLATFORM_TOOL_NOT_FOUND'
+                    )
+                }
+
+                const user = await this.app.db.models.User.byId(userId)
```
andypalmi, line 98: Quick question: if `User.byId(userId)` returns null here (for example if the user was deleted between the request being issued and handled), `getOrCreatePlatformToken` would receive a null user. Is that already ruled out upstream by the ACL/session checks, or would a small guard returning a clean 'unknown user' error before minting the token be worth adding?

forge/ee/lib/mcp/tools/instances.js
```diff
@@ -2,52 +2,54 @@ const { z } = require('zod')
...
     {
-        name: 'platform.get-instance',
-        description: 'Get details of a specific instance, including its current state, URL, and settings.',
+        name: 'platform_get_hosted_instance_status',
+        description: `FlowFuse platform automation tool:
+            Gets the live running status of a specific hosted instance (running, stopped, suspended, starting, etc.).
+            This is different from platform_get_hosted_instance: that tool gives you metadata and settings,
+            this tool tells you what the instance is doing right now.
+            Use this when the user asks if an instance is running, or when you need to check before performing an action that requires it to be online.`,
         annotations: { readOnlyHint: true, destructiveHint: false },
```
Steve-Mcl, line 29: optional - consider openworld hint

forge/ee/lib/mcp/tools/snapshots.js
```diff
@@ -2,23 +2,97 @@ const { z } = require('zod')
...
+        description: `FlowFuse platform automation tool:
+            Lists all snapshots that were taken from a hosted instance.
+            A snapshot is like a saved photo of everything running on the hosted instance at a point in time: the flows, the settings, and the configuration.
+            Use this when you need to see what snapshots exist for a hosted instance, for example to pick one to deploy or to check what changed between versions.`,
         annotations: { readOnlyHint: true, destructiveHint: false },
         inputSchema: {
-            instanceId: z.string().describe('The ID or hashid of the instance')
+            hostedInstanceId: z.string().describe('The ID or hashid of the hosted instance'),
+            cursor: z.string().optional().describe('Cursor for pagination (the hashid of the last item from the previous page)'),
+            limit: z.number().optional().describe('How many results to return per page')
```
Steve-Mcl: should max the limit to something like `50` (the agent recursion limit is 4x)

```suggestion
            limit: z.number().max(50).optional().describe('How many results to return per page')
```

forge/ee/lib/mcp/tools/applications.js
```diff
@@ -26,8 +29,119 @@ module.exports = [
...
+            A remote instance is a Node-RED that runs on the user's own hardware (like a Raspberry Pi or a server) rather than on the same environment as the FlowFuse platform.
+            Use this to see which remote instances are connected to an application, check if they are online or offline, or find one by name.
+            You can search by name using the query parameter and page through results using cursor or limit.
+            To get the full details of one specific remote instance, call platform_get_remote_instance with its ID.`,
+        annotations: { readOnlyHint: true, destructiveHint: false },
+        inputSchema: {
+            applicationId: z.string().describe('The ID or hashid of the application'),
+            query: z.string().optional().describe('Search remote instances by name'),
+            cursor: z.string().optional().describe('Cursor for pagination (the hashid of the last item from the previous page)'),
+            limit: z.number().optional().describe('How many results to return per page (default 100, max 100)')
```
Steve-Mcl:
```suggestion
            limit: z.number().max(50).optional().describe('How many results to return per page (default 100, max 100)')
```

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -2,59 +2,115 @@ const { z } = require('zod')
...
+            Use this when you need to see all the remote instances a team has, regardless of which application they belong to.
+            If you already know the application, use platform_get_application_remote_instances instead to get a narrower list.
+            To get the full details of one specific remote instance, call platform_get_remote_instance with its ID.`,
         annotations: { readOnlyHint: true, destructiveHint: false },
         inputSchema: {
-            teamId: z.string().describe('The ID or hashid of the team')
+            teamId: z.string().describe('The ID or hashid of the team'),
+            query: z.string().optional().describe('Search remote instances by name or type'),
+            cursor: z.string().optional().describe('Cursor for pagination (the hashid of the last item from the previous page)'),
+            limit: z.number().optional().describe('How many results to return per page (default 100, max 100)')
```
Steve-Mcl:
```suggestion
            limit: z.number().max(50).optional().describe('How many results to return per page (default 100, max 100)')
```

test/unit/forge/comms/devices_spec.js
```diff
@@ -10,6 +10,7 @@ const { DeviceCommsHandler } = FF_UTIL.require('forge/comms/devices')
 
 describe('DeviceCommsHandler', function () {
     let app
+    let eeApp
```
Steve-Mcl: these should not be here

Steve-Mcl (commented):
2nd pass - some suggestions

forge/comms/devices.js
Steve-Mcl: acknowledged

forge/ee/lib/mcp/tools/instances.js
Steve-Mcl: skip this for now until we use open world

Steve-Mcl (approved):
team review over slack

PR #7655, opened 2026-07-02
Validate instance ID type for MCP requests to prevent cache key errors

### LINKED ISSUE #7654: Expert Insights error when device is an MCP source
### Current Behavior

Getting a toast with `Internal Server Error: "arguments[2]" must be of type "string | Buffer", got number instead.`

This does not occur on local dev

After some digging it seems the redis cache can ONLY accept string keys

> redis driver → hGet/hSet(this.name, key, …). node-redis builds ['HGET', name, key] and rejects the numeric element at index [2] → "arguments[2]" must be of type "string | Buffer", got number instead.

The expected type of `key` in `hGet`/`hSet` is `RedisArgument` (which is defined as `String|Buffer`

### Expected Behavior

Should not error

### Steps To Reproduce

Include a device with MCP server tools/resource and make an insights request

### Environment

- FlowFuse version: FFC main
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

Steve-Mcl:
## Description

Ensure string key is used for cache to prevent redis cache throwing  `Internal Server Error: "arguments[2]" must be of type "string | Buffer", got number instead.`

## Related Issue(s) 

closes #7654

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


cstns (approved):
nice catch! We should use the redis cache driver locally to avoid these types of situations moving forward.

Approving in principle but there are some tests failing, re-merging main to trigger a re-run. I'll keep an eye on them

PR #7661, opened 2026-07-02
[7653] New instance creation fails to load immersive editor/FlowFuse Expert  

### LINKED ISSUE #7653: New instance creation fails to load immersive editor/FlowFuse Expert
### Current Behavior

When creating a new instance, there is a gap on first load that stops immersive editor from loading, falling back to non-immersive and failing to load FlowFuse Expert.

### Expected Behavior

FlowFuse Expert/Immersive should load automatically.

### Steps To Reproduce

Spin up a new instance, and click Open Editor.

### Environment

- FlowFuse version: 2.31
- Node.js version: 22.x
- npm version: Latest
- Platform/OS: Mac
- Browser: Chrome

n-lark:
## Description

Issue here was when we switched to ws/mqtt, we stopped polling which is what populated versions for a newly created instance. So the instance had no version by the time Open Editor checked, and we fell back to the direct link.

It doesn't reproduce locally because localfs reports versions immediately in the initial instances call. On Kubernetes they only appear once the pod is up, so we see the direct link issue.

**Fix**: the backend now includes versions in the published state message and the frontend merges them onto the instance.

**This needs to be tested in staging: https://7661.flowfuse.dev/**

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7653

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/db/controllers/Project.js
```diff
@@ -847,6 +851,22 @@ module.exports = {
...
+                        versionInfo[key] = versionInfo[key] || {}
+                        versionInfo[key].current = value
+                    }
+                    project.versions = versionInfo
+                    await project.save()
+                }
+            } catch (err) {
+                app.log.warn(`Failed to capture versions for ${project.id}: ${err.toString()}`)
+            }
+        }
```
n-lark, line 869: When an instance starts running, we grab its versions from the container and save them here so they get sent out in the state message. Same thing `liveState` already does on every `getInstance`, just triggered on the running transition too. Saving them (instead of passing them through inline) means any publish path picks them up, and the guard keeps it to one lookup per instance.

Would like a second opinion on one thing: this adds a `containers.details` call (HTTP to the launcher) on the running transition that wasn't there before. Tested in staging and didn't see any slowness, seems fine, just wanted to point it out.

forge/comms/index.js
```diff
@@ -82,9 +82,11 @@ module.exports = fp(async function (app, _opts) {
                     if (!teamHash || !id) return
                     client.publish(`ff/v1/${teamHash}/d/${id}/state`, JSON.stringify({ id, meta: { state } }))
                 },
-                notifyInstanceState: function (teamHash, id, state) {
+                notifyInstanceState: function (teamHash, id, state, versions) {
```
cstns: worth adding state and versions under a single obj argument

forge/db/controllers/Project.js
cstns: this lookup is made as a failover if the project received doesn't contain versions already. I'd say it's ok given that all it's doing is calling the underlying driver which at most is doing another http call

cstns (approved):

forge/comms/index.js
n-lark: done

Issue #7704, opened 2026-07-04
5.9-c Delete and destructive tools (phase 2)


andypalmi:
**Parent:** #7678 (5.9 Notifications and user self-service)
**Tool file:** new `forge/ee/lib/mcp/tools/user.js`

`destructiveHint: true`. Kept separate from ordinary writes so MCP clients gate them behind confirmation by default. One tool per delete endpoint.

| Tool | Endpoint | Scope | Annotation |
|---|---|---|---|
| `platform_delete_notification` | `DELETE /user/notifications/:notificationId` | `user:edit` (self) | destructive |

**Design notes:**
- `delete_notification` deletes a single notification by id, self-scoped via `byId(id, request.session.User)` (404 if not owner). There is **no bulk delete-all route**: `DELETE /user/notifications` exists only as commented-out code in `userNotifications.js`, so `notificationId` is required. If a bulk delete is wanted later, the backing route must be built first.

**Tool definitions (description + zod inputSchema):**
```js
platform_delete_notification: {
  description: 'Delete a single notification by id. Self-scoped: returns 404 if the notification is not owned by the caller. There is no bulk delete-all option, so notificationId is required.',
  inputSchema: z.object({
    notificationId: z.string().describe('Notification hashid to delete')
  })
}
```

**Excluded (sensitive self-service):** `PUT /user/change_password`, `DELETE /user` (account deletion), `PUT/DELETE /user/mfa*` (MFA), and all `/user/tokens*` routes. Account deletion is destructive and irreversible; token management via a PAT raises the circular-creation concern (#7515). These stay EXCLUDED and are not registered as tools.

**Scopes to allow-list + access-tag (#7445):** `user:edit` (write; the destructive notification permissions tagged so read-only PATs are rejected).

**Tests:**
- `delete_notification` carries `destructiveHint: true`; read-only PAT rejected.
- `delete_notification` operates only on the caller's own records (404 for foreign ids).
- Excluded sensitive routes (change_password, account deletion, MFA, tokens) are not registered as tools.

---

Issue #7703, opened 2026-07-04
5.9-b Write tools, non-destructive (phase 2)


andypalmi:
**Parent:** #7678 (5.9 Notifications and user self-service)
**Tool file:** new `forge/ee/lib/mcp/tools/user.js`

`readOnlyHint: false`, `destructiveHint: false`. Sibling actions that share one concept are consolidated into a single tool with an `action` enum (or an optional target id), following 5.1's pattern.

| Tool | Endpoint(s) | Scope | Annotation |
|---|---|---|---|
| `platform_set_notification_read_state` (single via `notificationId`, or bulk via `ids`) | `PUT /user/notifications/:notificationId`, `PUT /user/notifications` | `user:edit` (self) | write |
| `platform_respond_to_team_invitation` (`action`: `accept`\|`reject`) | `PATCH /user/invitations/:invitationId` (accept), `DELETE /user/invitations/:invitationId` (reject) | `user:edit` (self) | write |
| `platform_update_profile` | `PUT /user` | `user:edit` (self) | write |

**Design notes:**
- `set_notification_read_state` consolidates the single (`PUT /user/notifications/:notificationId`) and bulk (`PUT /user/notifications`) variants: with a `notificationId` the tool calls the single-notification route with `{ read }`; otherwise it calls the bulk route with `{ ids, read }`. There is **no bare "mark all" route** - the bulk route takes an explicit array of notification ids. `read` may be `true` or `false` (mark read or unread), so the tool is named for setting the state, not just marking read.
- `respond_to_team_invitation` consolidates accept and reject on the same invitation resource. `accept` maps to `PATCH`, `reject` maps to `DELETE`; both self-scope via `byId(id, request.session.User)` (404 if not owner). Reject only declines an invitation the caller received, so it is modelled as a non-destructive response rather than a destructive delete.
- `platform_update_profile` should expose only safe profile fields (name, defaultTeam), not privileged fields that only `/users/:userId` (admin) can set. (The route itself also accepts `username`/`email`/`tcs_accepted` as self-service, but v1 exposes only name/defaultTeam.)

**Tool definitions (description + zod inputSchema):**
```js
platform_set_notification_read_state: {
  description: 'Set the read state of the caller notifications. Provide notificationId to set the state of one notification, or ids to set several; there is no bare "mark all" shortcut, so bulk changes require an explicit list of ids. read=true marks read, read=false marks unread.',
  inputSchema: z.object({
[...]

PR #7718, opened 2026-07-04
feat(mcp): notification and self-service read tools

### LINKED ISSUE #7702: 5.9-a Read tools (phase 1)
**Parent:** #7678 (5.9 Notifications and user self-service)
**Tool file:** new `forge/ee/lib/mcp/tools/user.js`

`readOnlyHint: true`.

|---|---|---|
| `platform_get_current_user` | `GET /user` | read |
| `platform_list_notifications` | `GET /user/notifications` | read |
| `platform_list_own_invitations` | `GET /user/invitations` | read |

**Design notes:**
- `user:read`/`user:edit` are `role: Admin` **but** `self: true`, so `/user/*` own-account routes are eligible self-service; `/users/*` (arbitrary users) are admin-only and excluded (see 5.9-d).
- Notification/invitation routes self-scope via `byId(id, request.session.User)` (404 if not owner) and are gated by a file-level `addHook('preHandler', needsPermission('user:edit'))`.
- The list routes map to `user:edit`, so the tool annotations at the MCP layer mark them read-only even though the underlying permission is shared with the write endpoints in 5.9-b/c.

**Permissions:**

```js
platform_get_current_user: {
  description: 'Get the authenticated user own profile. No parameters or body.',
  inputSchema: z.object({})
}
platform_list_notifications: {
  description: 'List the authenticated user own notifications, with pagination.',
  inputSchema: z.object({
    ...basePagination // shared fragment, see issue 5.0
  })
}
platform_list_own_invitations: {
  description: 'List the team invitations the authenticated user has received. No parameters or body.',
  inputSchema: z.object({})
}
```

**Tests:**
- Each tool injects the correct route and returns the response unchanged.
- `platform_list_notifications` serialises cursor and limit onto the query string when given, and injects with no query string when omitted.
- All three tools pass through an error response unchanged.

---

andypalmi:
## Summary

Adds Phase 1 read-only MCP tools for user self-service #7702 .

**Tool file:** `forge/ee/lib/mcp/tools/user.js`

**Tools added:**
- `platform_get_current_user` - `GET /api/v1/user`
- `platform_list_notifications` - `GET /api/v1/user/notifications` (paginated)
- `platform_list_own_invitations` - `GET /api/v1/user/invitations`

All three are tagged `readOnlyHint: true, destructiveHint: false`.

**Permissions:**

Write, delete, and admin-only endpoints from the parent issue #7703 #7704 are out of scope for this PR, as are the excluded sensitive self-service routes (change_password, account deletion, MFA, tokens).

Closes #7702

## Test plan

- [x] `eslint` clean on changed files
- [ ] CI


forge/ee/lib/mcp/tools/user.js
```diff
@@ -0,0 +1,46 @@
...
+        }
+    },
+    {
+        name: 'platform_list_notifications',
+        title: 'List Notifications',
+        description: `FlowFuse platform automation tool:
+            Lists the authenticated user's own notifications, with pagination.
+            Use this to check for unread alerts or recent activity addressed to the current user.`,
+        annotations: { readOnlyHint: true, destructiveHint: false },
+        inputSchema: { ...basePagination },
```
cstns, line 25: this endpoint does not support pagination

cstns (changes requested):

forge/ee/lib/mcp/tools/user.js
```diff
@@ -0,0 +1,46 @@
...
+    },
+    {
+        name: 'platform_list_own_invitations',
+        title: 'List Own Team Invitations',
+        description: `FlowFuse platform automation tool:
+            Lists the team invitations the authenticated user has received but not yet accepted or rejected.
+            Use this to check whether the current user has any pending invitations to join a team.
+            No parameters or body required.`,
+        annotations: { readOnlyHint: true, destructiveHint: false },
+        inputSchema: {},
```
cstns, line 40: this one is paginated

cstns (changes requested):

forge/ee/lib/mcp/tools/user.js
andypalmi: `GET /api/v1/user/notifications` does paginate. It runs through `getNotifications` in `forge/services/notifications.js`, which calls `app.getPaginationOptions(request)` and passes it to `Notification.forUser`, and the response carries `meta` and `count`. Confirmed against a running instance: `platform_list_notifications { limit: 1 }` returned 1 of 4 notifications with `meta.next_cursor` set, and passing that cursor back returned the next entry. So `cursor` and `limit` are correct on this tool.

forge/ee/lib/mcp/tools/user.js
andypalmi: `GET /api/v1/user/invitations` does not paginate. It calls `Invitation.forUser(request.session.User)` with no pagination options and always returns `meta: {}` (that line carries a `// For future pagination` note in `forge/routes/api/userInvitations.js`), returning every invitation. Confirmed against a running instance: with two pending invitations, `platform_list_own_invitations { limit: 1 }` still returned both, and `{ limit: 1, cursor: "bogus" }` returned the same two with no error. So the params have no effect, which is why this tool has no pagination params; we can add them if the endpoint gains real pagination later.

andypalmi:
Verified the pagination behavior by invoking the tools against a local instance.

`platform_list_notifications` (account had 4 notifications):
- `{}` -> count 4, `meta: {}`, 4 items
- `{ limit: 1 }` -> count 4, `meta: { next_cursor: <id> }`, 1 item
- `{ limit: 1, cursor: <that id> }` -> next item, `meta: { next_cursor: <next id> }`
- `{ limit: 50 }` -> count 4, `meta: {}`, 4 items

`platform_list_own_invitations` (account had 2 pending invitations):
- `{}` -> count 2, `meta: {}`, 2 items
- `{ limit: 1 }` -> count 2, `meta: {}`, 2 items (limit ignored)
- `{ limit: 1, cursor: "bogus" }` -> identical, no error

Conclusion: notifications supports `cursor`/`limit`; invitations does not, matching the current tool schemas.

PR #7721, opened 2026-07-04
feat(mcp): catalog read tools

### LINKED ISSUE #7679: 5.10 Catalog completeness
# 5.10 Catalog completeness

**Parent:** Story 5 (#7232)
**Tool file:** `forge/ee/lib/mcp/tools/platform.js` (extend)

## Summary

Add the get-by-id read tools for the platform catalogs (templates, blueprints), plus the team-type catalog reads. The list variants already ship; agents need single-item detail (e.g. to inspect a template or blueprint before use). Team types are a global platform catalog with no team in the path, mirroring project-types, so they are filed with the platform catalog tools. This scope is read-only for non-admins: all catalog create/edit/delete are platform-admin and are tracked (not implemented) in 5.10-d.

## Read tools (phase 1)

`readOnlyHint: true`.

|---|---|---|
| `platform_get_template` | `GET /templates/:templateId` | read |
| `platform_get_blueprint` | `GET /flow-blueprints/:flowBlueprintId` | read |
| `platform_list_team_types` | `GET /team-types` | read |
| `platform_get_team_type` | `GET /team-types/:teamTypeId` | read |

**`platform_get_blueprint` flow content:** the raw route always inlines the full flow JSON. The tool adds an `includeFlow` param (default false) that replaces `flows` with a short instruction instead, since inspecting a blueprint's metadata doesn't need the flow content until it's actually being deployed.

**Design notes:**
- **No stack or instance-type get-by-id tools.** The existing `platform_list_hosted_instance_types` already returns every project-type nested with its full stacks and a `defaultStack`, decorated with team creatability flags, and it accepts a `projectType` argument to return one type and its stacks. The single-item routes `GET /project-types/:id` and `GET /stacks/:id` use the same `views.ProjectType.projectType` and `views.ProjectStack.stack` views that tool already returns, so dedicated `platform_get_hosted_instance_type` / `platform_get_stack` / `platform_list_stacks` tools would be redundant and were dropped.
- list/read on templates/blueprints are any-authenticated (role undefined); the `User.admin` flag only changes response detail. These are safe reads.
- `GET /flow-blueprints/:id` and the list route allow project/device token bypass; for user PATs they require `flow-blueprint:read`/`:list`.
- `templates` get blanks hidden env values in the response.
- `platform_list_team_types` and `platform_get_team_type` read the global team-type catalog (no team in the path), so they are filed with the platform catalog tools rather than the team tools. `list_team_types` supports pagination.
- Excluded here: `GET /flow-blueprints/export-public` (anonymous, config-gated read; not registered as a tool). The admin `/export` route is tracked in 5.10-d.

**Permissions:**

**Tests:**
- All tools allowed for read-only PAT.
- Get-by-id returns the same view the UI uses.
- Team-type list and get return the platform team-type catalog.
- `platform_get_blueprint` omits `flows` by default and returns it in full when `includeFlow: true`.

andypalmi:
## Summary

Read tools for the platform catalogs, added to `forge/ee/lib/mcp/tools/platform.js`. Adds get-by-id lookups (the list variants already exist) and the team-type catalog reads:

- `platform_get_template` - `GET /templates/:templateId`
- `platform_get_blueprint` - `GET /flow-blueprints/:flowBlueprintId`. The raw route always inlines the full flow JSON; the tool adds an `includeFlow` param (default false) that omits it in favour of a short instruction, since agents inspecting blueprint metadata don't need the flow content until they actually deploy it.
- `platform_list_team_types` - `GET /team-types`
- `platform_get_team_type` - `GET /team-types/:teamTypeId`

No stack or instance-type get-by-id tools: the existing `platform_list_hosted_instance_types` already returns every project-type nested with its full stacks and a `defaultStack`, and accepts a `projectType` argument to narrow to one type. `GET /project-types/:id` and `GET /stacks/:id` use the same `views.ProjectType.projectType` and `views.ProjectStack.stack` views that tool already returns, so `platform_get_hosted_instance_type`, `platform_get_stack`, and a standalone `platform_list_stacks` would be redundant and are not added.

Team types are a global platform catalog with no team in the path, mirroring project-types, so they are filed with the platform catalog tools following the one-file-per-resource convention. All tools are `readOnlyHint: true, destructiveHint: false`.

**Permissions:**

Closes #7679

## Test plan
- [x] `eslint` on changed files passes
- [x] `mocha` unit tests for the tool file


andypalmi:
@cstns Please see commit for the latest changes as a last round review https://github.com/FlowFuse/flowfuse/pull/7721/changes/d4f7a9cf6025a256f93839983c72a8e991b28fb3

andypalmi:
### Testing results of new/changed tools

@cstns for visibility.

The four tools this PR adds (`platform_get_template`, `platform_get_blueprint`, `platform_list_team_types`, `platform_get_team_type`) were invoked against a running local platform.

The context the calls ran against:
- One template ("Default"), temporarily given one visible and one hidden environment variable to exercise the masking guarantee, then restored to its original state afterward.
- Twenty flow blueprints.
- Three active team types (Enterprise, Starter, Pro).

Payloads below are redacted for ids, names and secrets; structural values (counts, statuses, and which fields are present) are verbatim.

## platform_get_template

| Case | Input params | Verdict | Returned payload (redacted) |
| --- | --- | --- | --- |
| existing template | {templateId: `TEMPLATE_ID`} | PASS | full template detail, byte-for-byte match with the raw route |
| unknown template | {templateId:"doesnotexist"} | PASS | 404 isError, {code:"not_found"} |
| template with visible + hidden env vars | {templateId: `TEMPLATE_ID`} | PASS | settings.env: the visible var's value returned as set, the hidden var's value blanked (`""`) with `hidden:true` retained - matches the raw route exactly |

## platform_get_blueprint

| Case | Input params | Verdict | Returned payload (redacted) |
| --- | --- | --- | --- |
| existing blueprint, default | {flowBlueprintId: `BLUEPRINT_ID`} | PASS | full blueprint detail, but `flows: "Flow content omitted to keep the response small. Call this tool again with includeFlow: true to fetch the full flow JSON."`; `modules` still returned as-is |
| existing blueprint, includeFlow: true | {flowBlueprintId: `BLUEPRINT_ID`, includeFlow:true} | PASS | `flows` returned as the full flow JSON (42 nodes), byte-for-byte match with the raw route |
| unknown blueprint | {flowBlueprintId:"doesnotexist"} | PASS | 404 isError, {code:"not_found"} |
| unknown blueprint, includeFlow: true | {flowBlueprintId:"doesnotexist", includeFlow:true} | PASS | 404 isError, {code:"not_found"} - error path is unaffected by `includeFlow` |

## platform_list_team_types

| Case | Input params | Verdict | Returned payload (redacted) |
| --- | --- | --- | --- |
| default (filter omitted) | {} | PASS | 3 active team types returned |
| filter: "all" | {filter:"all"} | PASS | same 3 types (no inactive types exist locally, so this doesn't distinguish from default, but the filter value was accepted and passed through) |
| query search | {query:"pro"} | PASS | 1 matching type returned |
| pagination limit:1 | {limit:1} | PASS | 1 type returned, `meta.next_cursor` set |

## platform_get_team_type

| Case | Input params | Verdict | Returned payload (redacted) |
| --- | --- | --- | --- |
| existing team type | {teamTypeId: `TYPE_ID`} | PASS | full team type detail, byte-for-byte match with the raw route |
| unknown team type | {teamTypeId:"doesnotexist"} | PASS | 404 isError, {code:"not_found"} |

## Findings

- `platform_get_template`, `platform_list_team_types`, and `platform_get_team_type` pass the route response through unchanged: their output matched the corresponding raw REST route response byte-for-byte for the same parameters. `platform_get_blueprint` matches the raw route byte-for-byte only when `includeFlow: true`; by default it deliberately replaces `flows` (see below).
- `platform_get_template` correctly surfaces the hidden environment variable's name and hidden flag while blanking its value, matching the underlying route's masking behaviour.
- `platform_get_blueprint` gained an `includeFlow` param (default false) since the raw route otherwise always returns the full flow JSON inline, which is large and unnecessary for agents that just want the blueprint's metadata; verified both the omitted-by-default and opt-in-full-flow paths above.
- Unrelated to this PR: while restoring the template fixture, `PUT /api/v1/templates/:templateId` (`forge/routes/api/template.js:235`) was found to throw a 500 whenever the request body's `settings` omits `env`, since the handler calls `.map()` on it unconditionally. This is a pre-existing issue on `main`, outside this PR's diff, noted here for visibility and not otherwise pursued.

## Conclusion

All four tools verified against their descriptions, including pagination, search, and the environment variable secret-masking guarantee. No open defects in this PR's diff.

PR #7726, opened 2026-07-04
feat(mcp): remote instance (device) read tools

### LINKED ISSUE #7708: 5.12-a Remote instance (device) read tools (phase 1)
**Parent:** #7681 (5.12 Team data and packages)
**Tool file:** `forge/ee/lib/mcp/tools/devices.js` (extend)

Read tools for the remote instance (device) resource. Filed in `devices.js` alongside the existing device tools, following the one-file-per-resource convention (a tool lives in the file for the noun it is about, regardless of the URL's first path segment). The team-data and package tools that previously sat under this issue now live with the team resource in `teams.js`.

`readOnlyHint: true`.

|---|---|---|
| `platform_list_team_provisioning_tokens` | `GET /teams/:teamId/devices/provisioning` | read |

**Design notes:**
- `platform_list_team_provisioning_tokens` describes device provisioning, so it is filed with the device tools. The summary view omits the token secret (unlike the create route), so it is safe to expose read-only.
- The remote instance audit-log, history, and HTTP-tokens reads are generalized on #7727 as `platform_get_instance_audit_log`, `platform_get_instance_history`, and `platform_list_instance_http_tokens`. Each takes an `instanceType: 'hosted' | 'remote'` argument and lives in the new `forge/ee/lib/mcp/tools/shared-instances-devices.js`, so they serve remote instances (devices) too and no remote-only variants are needed here.
- The CSV audit-log export is dropped. The `GET /devices/:deviceId/audit-log/export` route only produces CSV, which is not a useful shape for an automation tool; the shared audit-log tool returns entries as JSON.

**Permissions:**

**Tests:**
- Read tool allowed for read-only PAT.
- Provisioning-token list never surfaces the token secret.

andypalmi:
## Summary

Phase 1 read-only MCP tool for the remote instance (device) resource, added to `forge/ee/lib/mcp/tools/devices.js`:

- `platform_list_team_provisioning_tokens` - `GET /teams/:teamId/devices/provisioning`. Lists a team's device provisioning tokens. The summary view omits the token secret, so it is safe to expose read-only.

The tool describes the device resource, so it is filed with the existing device tools, following the one-file-per-resource convention. It is annotated `readOnlyHint: true, destructiveHint: false`.

## Consolidation notes

- The remote instance audit-log, history, and HTTP-tokens reads that this PR originally carried are generalized on #7727. `platform_get_instance_audit_log`, `platform_get_instance_history`, and `platform_list_instance_http_tokens` each take an `instanceType: 'hosted' | 'remote'` argument and live in the new `forge/ee/lib/mcp/tools/shared-instances-devices.js`, so they serve remote instances (devices) too and no remote-only variants are needed here.
- The CSV audit-log export is dropped. The `GET /devices/:id/audit-log/export` route only produces CSV, which is not a useful shape for an automation tool; the shared `platform_get_instance_audit_log` returns audit entries as JSON.

## Permissions

No permissions changes. #8185 removed the hardcoded `user:expert-mcp` scope allow-list, so expert MCP tokens now inherit the user's permissions gated by team role and no allow-list entries are needed.

Closes #7708

## Test plan

- [x] `eslint` clean on the changed files
- [x] `mocha` unit tests


forge/ee/lib/mcp/tools/devices.js
```diff
@@ -1,5 +1,13 @@
 const { z } = require('zod')
 
+const { teamId, remoteInstanceId, basePagination, basePaginationKeys, searchQuery, searchQueryKeys, auditLogFilters, auditLogFilterKeys, appendQuery } = require('../schemas')
+
+// Audit-log routes accept cursor+limit pagination, free-text query, event
+// (single name or array) and username. The device audit-log route has no scope
+// parameter, so these tools only compose the base audit-log fragments.
+const auditLogInput = { ...basePagination, ...searchQuery, ...auditLogFilters }
```
cstns: these are only used in one tool, why extract them?

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -189,5 +197,41 @@ module.exports = [
             const response = await inject({ method: 'PUT', url: `/api/v1/devices/${args.remoteInstanceId}`, payload: { application: args.applicationId } })
             return response
         }
+    },
+    {
+        name: 'platform_get_remote_instance_audit_log',
+        title: 'Get Remote Instance Audit Log',
+        description: `FlowFuse platform automation tool:
+            Reads the audit log for a remote instance/device, showing events like connection changes,
```
cstns: exclude /device

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -189,5 +197,41 @@ module.exports = [
...
+            deployments, and configuration changes for that device.
```
cstns: stick with remote instance

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -189,5 +197,41 @@ module.exports = [
...
+            Set format to "csv" to export the log as a downloadable file instead of reading entries directly.`,
+        annotations: { readOnlyHint: true, destructiveHint: false },
+        inputSchema: {
+            remoteInstanceId,
+            ...auditLogInput,
+            format: z.enum(['json', 'csv']).optional().describe('Output format. "json" (default) reads entries directly; "csv" exports the log as a downloadable CSV file.')
+        },
+        handler: async (args, { inject }) => {
+            const suffix = args.format === 'csv' ? '/audit-log/export' : '/audit-log'
+            const url = appendQuery(`/api/v1/devices/${args.remoteInstanceId}${suffix}`, args, auditLogKeys)
```
cstns: this endpoint only does csv

cstns: this endpoint is also only syntactic sugar over the audit log

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -189,5 +197,41 @@ module.exports = [
...
+            const response = await inject({ method: 'GET', url })
+            return response
+        }
+    },
+    {
+        name: 'platform_list_team_provisioning_tokens',
```
cstns, line 196: remote_instance

cstns (changes requested):

forge/ee/lib/mcp/tools/devices.js
andypalmi: Thanks, all fair points. As discussed in the recent call, we have decided to move tools that are shared amongst hosted and remote instances into a new file, `forge/ee/lib/mcp/tools/shared-instances-devices.js`, so it is clear those are shared. The audit-log, history, and HTTP-tokens reads are now generalized there on #7727 as `platform_get_instance_audit_log`, `platform_get_instance_history`, and `platform_list_instance_http_tokens`, each taking an `instanceType: 'hosted' | 'remote'` argument, so the remote-only variants are removed from this PR. The CSV export is dropped too, since that route only produces CSV and the shared audit-log tool returns JSON. This PR now adds only `platform_list_team_provisioning_tokens`, which resolves the extracted-fragment point as well.

andypalmi:
### Testing results of new/changed tools

@cstns for visibility.

This PR now adds a single tool, `platform_list_team_provisioning_tokens`, which was invoked against a running local platform. The context the calls ran against:

- Three teams were available; one was seeded with a single device provisioning token, the other two had none.
- The seeded token was created through the platform's own create route, whose response returns the token secret; the list summary is expected to omit that secret.

Payloads below are redacted for ids, names and secrets; structural values (counts, statuses, error codes, and which fields are present) are verbatim. No credential secret is reproduced.

## platform_list_team_provisioning_tokens

Lists a team's device provisioning tokens. The summary view omits the token secret.

| Case | Input params | Verdict | Returned payload (redacted) |
| --- | --- | --- | --- |
| team with no tokens | {teamId: `TEAM_ID`} | PASS | {meta:{}, count:0, tokens:[]} |
| team with one token | {teamId: `TEAM_ID`} | PASS | {meta:{}, count:1, tokens:[{id, name, team, expiresAt:null}]} - no `token` secret field |
| per-token field check | {teamId: `TEAM_ID`} | PASS | token entry keys are exactly `id`, `name`, `team`, `expiresAt`; `token` is absent |
| nonexistent team | {teamId:"doesnotexist"} | PASS | 404 isError, {code:"not_found"} |

## Findings

- The tool passes the route response through unchanged: the tool output matched the raw `GET /teams/:teamId/devices/provisioning` response byte-for-byte for the same team.
- The summary view omits the token secret. The create route returns a `token` field, whereas each list entry exposes only `id`, `name`, `team`, and `expiresAt`, confirming the tool is safe to expose read-only.
- An unknown team surfaces a clean 404 rather than an empty list.

## Conclusion

The single tool added in this PR verified against its description, including the secret-omission guarantee. No open defects.

PR #7733, opened 2026-07-06
Add granular RBAC SSO support

### LINKED ISSUE #7396: SSO: enable SAML group management of granular RBAC
FlowFuse currently supports managing user roles at the team level via SSO groups.

We have since enabled granular RBAC within the platform, where a user can be assigned a different role per application in the team.

This feature request is to extend our SAML group support to enable management of the granular RBAC roles.

## Current Support

If an SSO configuration has group management enabled then, when as user logs in via the configuration, it checks their groups that match the pattern: (with an optional prefix/suffix that I won't mention again).
 
 - `ff-<team>-<role>` 
    - The `<team>` matches the team slug
    - The `<role>` matches the user role for that team

## Proposed Support

We need to extend this group support to express a role for an individual application.

 - `ff-<team>-<role>` - sets the user role in the team
 - `ff-<team>[<application>]-<role>`
    - `<application>` - *either* the application name, or application id
    - example: `ff-development[building control]-members`
        - Applies `member` role to the user in the `building control` application of team `development`.

Notes:
 - Setting an application scoped role can only happen if a team role has been set
 - Application names can contain `[]`, so defensive parsing needed
 - Application names can be changed (as can team slugs) - that will break the link
 - Application IDs are immutable, but not known ahead of time, so harder to use in an automation - but we can support both
 - As with team role, these are only applied at login. Changes made to the idp are only reflected if the user logs out and back in.

hardillb:
fixes #7396

## Description

Includes both SMAL and LDAP

Also includes a fix for the direct SSO login buttons on login page

## Related Issue(s)

#7396 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/db/models/Team.js
```diff
@@ -780,6 +780,19 @@ module.exports = {
                 unsuspend: async function () {
                     this.suspended = false
                     await this.save()
+                },
+                hasApplication: async function (nameOrId) {
+                    const applicationId = M.Application.decodeHashid(nameOrId)
+                    const application = await app.db.models.Application.findOne({
+                        where: {
```
knolleary, line 787: Missing a where clause to ensure its an Application in this team.

forge/ee/lib/sso/index.js
```diff
@@ -351,25 +352,46 @@ module.exports.init = async function (app) {
...
+                        const teamSlug = applicationRolesMatch[1]
+                        const applicationId = applicationRolesMatch[2]
+                        const applicationRoleName = match[2]
+                        const applicationRole = Roles[applicationRoleName]
+                        const team = await app.db.models.Team.bySlug(teamSlug)
+                        const application = await team.hasApplication(applicationId)
+                        if (application) {
+                            if (providerOpts.groupAllTeams || (providerOpts.groupTeams || []).includes(teamSlug)) {
+                                if (desiredTeamApplicationroles[team.hashid]) {
+                                    desiredTeamApplicationroles[team.hashid][application.hashid] = Math.max(desiredTeamApplicationroles[teamSlug][application.hashid] || 0, applicationRole)
```
knolleary: There is a mix of `team.hashid` and `teamSlug` when keying into  `desiredTeamApplicationroles` to read from it, but its only written to via `team.hashid`. Need to verify if that's intentional.

knolleary (changes requested):
I've done a first pass and spotted a couple things.

It would be helpful to have some more inline comments (similar to the existing code) to help make the logic more explicit.

I need to spend some more time testing it to verify the behaviour.

forge/db/models/Team.js
hardillb: will fix

forge/ee/lib/sso/index.js
hardillb: yeah, they probably should all be `teamSlug`

Given it's at the team level and we only ever use slugs for team ids in group names it should be OK. But I didn't test with 2 groups for the same application (one name, one hash id) with different role values, I will do that and push needed changes.

knolleary: @hardillb I'm still seeing a mix of teamSlug and hashid on this line after your commit.

hardillb: @knolleary I think I've got all the ones I missed.

knolleary:
@hardillb I've pushed a small update
 - Put the work to do the granular rbac management inside a `if` so it only runs if there's something to be done.
 - Added a couple debug log statements about existing/desired g-rbac permissions

I *think* it's good to go. Can you give that last commit a once over before merging?

knolleary (approved):

hardillb:
It failing one of the SSO SAML tests, will have a look at it in the morning

hardillb:
@knolleary The new if block causes a test failure.

Adding the if block means the test that removes an application override when the group is removed from the user doesn't fire if there are no other application overrides.

The new if block needs removing

forge/ee/lib/sso/index.js
```diff
@@ -461,6 +483,44 @@ module.exports.init = async function (app) {
             }
 
             await Promise.all(promises)
+            // Now all group membership updated apply application overrides
+
+            const desiredTeamApplicationArray = Object.entries(desiredTeamApplicationroles)
+            if (desiredTeamApplicationArray.length > 0) {
```
hardillb: This now means that users are not removed from Application level groups if all they are no longer in any Application level groups

knolleary: Does that mean if you have group management enabled you can't do manual application level management as well? This whole thing may need to be an option on the same provider if that's the case.

hardillb:
If a user has any Application Group assignments in a Team then they can only be Group managed, but other users are free to be manually configured. You can't mix the 2

knolleary:
I'm seeing some incorrect behaviour in my local testing.

1. SAML config is set to only apply to TeamA, with the 'allow to be a member of other teams' option checked
2. I add user to TeamB via an invite. Then apply an application override for something in TeamB
3. User logs in. Their membership and overrides of TeamA are correct. Their override in TeamB has been deleted. TeamB should not have been touched.

hardillb:
I can see how that would happen, will work out how to fix

hardillb:
@knolleary fix pushed

hardillb:
We will need to regenerate the types if this merges after https://github.com/FlowFuse/flowfuse/pull/7825

PR #7766, opened 2026-07-07
[7446] Scoped PATs: API routes and controllers

### LINKED ISSUE #7446: Scoped PATs - Auth Middleware: Populate PAT Metadata
### Summary

When a PAT is used to authenticate a request, load the token's scope restrictions and make them available to the rest of the request lifecycle. This is the bridge between the foundation (#7445) and the enforcement tasks that follow. No behavior changes for existing tokens.

### Why

Today, when a named PAT authenticates a request, the auth middleware deletes the token's scope from the session (`forge/routes/auth/index.js` lines 123-126), giving the token full user access. This was an intentional "temp hack" that we're now replacing.

This task keeps the existing behavior (PATs still get full access) but adds PAT metadata to two places: `request.session.pat` (for `needsPermission`, which has the request object) and the `@fastify/request-context` store (for `hasPermission`, which does not). Later tasks will read from these to enforce restrictions.

### What to do

**Replace the scope deletion in `verifySession`**

`forge/routes/auth/index.js` (lines 123-126)

Replace the "temp hack" that deletes scope for named tokens. Instead, keep the scope deletion (preserves existing behavior) but also load the token's team scopes from the `AccessTokenTeamScope` join table and build a PAT metadata object containing `readOnly`, `adminOptIn`, and `teamScopes`. Store it on both `request.session.pat` and in the `@fastify/request-context` store.

When a token has no team scope entries, `teamScopes` should be `null` (meaning team-agnostic, all teams accessible). When entries exist, it should be an array of team identifier objects `{ [<team-id> ]: <team-role> }` (which will allow us in the future to set granular restrictions on each team - similar to membershib object passed in hasPermission).

**Consider eager-loading optimization**

The team scope query adds a second DB hit per PAT-authenticated request. Consider eager-loading `AccessTokenTeamScope` (with nested `Team`) in `getOrExpire` when the token has a `name` field set, to avoid the separate query in the auth middleware.

### Tests

* PAT-authenticated request populates `request.session.pat` with correct structure
* `request.session.pat.teamScopes` is `null` when the token has no team scope entries
* `request.session.pat.teamScopes` is an array when team scope entries exist
* `request.session.pat.readOnly` and `request.session.pat.adminOptIn` reflect the token's stored values
* `@fastify/request-context` store contains the same PAT data
* Existing PATs continue to work identically to before
* Non-PAT auth (cookie sessions, device tokens, project tokens) is unaffected

### LINKED ISSUE #7515: Scoped PATs - API Routes and Controllers
### Summary

Update the PAT CRUD endpoints to accept, validate, and return the new scoping fields (readOnly, adminOptIn, teamIds - see #7446 for enforced format). No enforcement yet, just wiring the data through the API layer.

### Why

The database and auth middleware now support PAT metadata, but there's no way for users to actually set it. This task makes the new fields available through the existing PAT management endpoints so the frontend (#7482) and API consumers can create and update scoped tokens.

### What to do

**POST /api/v1/user/tokens (create)**
* Accept `readOnly` (boolean), `adminOptIn` (boolean), `teamIds` (array of team hashids)
* Validate: each teamId must be a team the user belongs to
* Validate: `adminOptIn` only allowed if the user is a platform admin
* Validate: a token created via another PAT cannot escalate privileges beyond what the originating token allows (e.g. a read-only PAT cannot create a read-write token, a team-scoped PAT cannot create a token with broader team access, a non-admin-opted-in PAT cannot create one with adminOptIn)
  - decision: should the PAT creation endpoint be excluded from the PAT accessible routes to avoid circular creation? 
* Create `AccessTokenTeamScope` entries for each valid team ID after token creation

**PUT /api/v1/user/tokens/:id (update)**
* Accept the same new fields
* Same validation rules, including the escalation checks
* When `teamIds` is provided, replace existing `AccessTokenTeamScope` entries (delete old, create new, inside a transaction)

**GET /api/v1/user/tokens (list)**
* Return `readOnly`, `adminOptIn`, and `teams` (array of {id, name}) in the token summary view

**Controller changes**
* `createPersonalAccessToken` in `forge/db/controllers/AccessToken.js`: extend to accept and persist the new fields, create team scope entries
* `updatePersonalAccessToken`: accept and save the new fields, handle team scope replacement

**View changes**
* `personalAccessTokenSummary` in `forge/db/views/AccessToken.js`: include the new fields and the eager-loaded team scope data

### Tests

* Create a token with team scope, read-only, and admin opt-in, verify the response includes all fields
* Non-admin user cannot set adminOptIn
* teamIds referencing teams the user doesn't belong to are rejected
* Updating a token's team scope replaces the old entries
* Listing tokens returns the new fields and team associations
* Creating/updating a token without the new fields works as before (backwards compatible defaults)
* Escalation prevention: a read-only PAT cannot create a read-write token
* Escalation prevention: a team-scoped PAT cannot create a token with teams outside its own scope
* Escalation prevention: a PAT without adminOptIn cannot create one with adminOptIn
* Cookie session (no PAT restrictions) can create tokens with any valid combination

cstns:
## Description

Wires `readOnly`, `adminOptIn`, and `teamIds` through the PAT CRUD layer.

- `POST /api/v1/user/tokens` and `PUT /api/v1/user/tokens/:id` now accept `readOnly`, `adminOptIn`, and `teamIds` in the request body
- `GET /api/v1/user/tokens` returns `readOnly`, `adminOptIn`, and `teams` (array of `{ id, name }`) for each token
- Both mutating endpoints are blocked for PAT-authenticated requests (`403 pat_cannot_create_pat`) via a new `app.blockPAT` preHandler (reusable on any route that needs the same guard)
- Team membership is validated on create/update — unknown or non-member teams are rejected with `400 invalid_team`
- Non-admin users cannot set `adminOptIn: true` (`403`)
- Named PAT authentication now populates `request.session.isPAT`, `request.session.pat`, and mirrors the same object into `request.requestContext`. laying the groundwork for enforcement in later tasks without changing any existing behavior
- Token create and update operations use transactions to keep  `AccessToken` and `AccessTokenTeamScope` rows consistent
- getOrExpire now eager-loads AccessTokenTeamScope (with nested Team) for all token lookups, avoiding a second DB hit during auth                       
- request.session.pat now includes teamScopes: null when no team scope entries (all teams accessible), or an array of { [hashid]: role } objects (with the user's actual team role from TeamMember)
- Both request.session.pat and request.requestContext receive the full PAT metadata including teamScopes

### No functional changes yet

- No UI exists for scoped PATs yet, so `teamIds` will almost never be sent in practice (existing token creation/update flows are unaffected)
- `AccessTokenTeamScope` rows will not be created until the UI ships
- `readOnly` and team scope restrictions are **not enforced** on any other routes yet (that's a follow-up task)

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/7446
closes https://github.com/FlowFuse/flowfuse/issues/7515

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


knolleary (approved):

knolleary:
I do wonder if tokens should be immutable in terms of their auth properties. Renaming is okay. Not a thought for this PR.

cstns:
> I do wonder if tokens should be immutable in terms of their auth properties. Renaming is okay. Not a thought for this PR.

I have to admit, I was surprised when I noticed that the expiry date can be updated. We can create a follow up for it

Issue #7765, opened 2026-07-07
chore: Remove runtime dependency to `@node-red/util`


Steve-Mcl:
### Description

`@node-red/util` is imported into FlowFuse runtime dependencies solely for the helper functions `getObjectProperty` and `settObjectProperty` in `forge/containers/stub/index.js` ([ref](https://github.com/FlowFuse/flowfuse/blob/cababe3110240b9c3e8c983f54d784360a4d1543/forge/containers/stub/index.js#L14))

`@node-red/util` brings several dependencies itself, one of them being `jsonata` which is being flagged by trivy as a high vuln.

We could ignore this but in the interest of sanitisation, we could simply create our own getObjectProperty/settObjectProperty functions 

@hardillb @ppawlowski 
Not knowing how trivy deals with alerts and if they can/should be silenced is this the right course of action?

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

hardillb:
If that is the only use, then just lift/shift the required function into the stub file and drop the dependency.

ppawlowski:
According to Dependabot, the `jsonata` dependency is introduced by two separate dependencies: https://github.com/FlowFuse/flowfuse/security/dependabot/328
https://github.com/FlowFuse/flowfuse/security/code-scanning?query=is%3Aopen+branch%3Amain+jsonata

Steve-Mcl:
> According to Dependabot, the `jsonata` dependency is introduced by two separate dependencies: [jsonata: Malicious inputs to "$toMillis" function can cause resource exhaustion](/FlowFuse/flowfuse/security/dependabot/328) https://github.com/FlowFuse/flowfuse/security/code-scanning?query=is%3Aopen+branch%3Amain+jsonata

Yes it is, and I have raised issue against Node-RED 
and will be raising against launcher too (to pick up new version once published)

PR #7767, opened 2026-07-07
chore: dependency updates

Steve-Mcl:
### Update transitive deps

Bumps a set of transitive dependencies up to their fixed lines via npm `overrides`. All changes are scoped to `package.json` + `package-lock.json` - no source changes.

Every affected package is a **transitive** dependency (none are declared directly), so the fixes are applied through `overrides` rather than direct version bumps.

### What changed

| Package | Result |
|---|---|
| serialize-javascript | 6.0.2 → **7.0.7** (two copies deduped onto one) |
| minimatch | 9.0.1 → **9.0.9** |
| uuid | 8.3.2 / 9.0.1 → **11.1.1** |
| vite | 6.4.2 → **6.4.3** |

```jsonc
"overrides": {
    "serialize-javascript@<7.0.7": "7.0.7",
    "minimatch@>=9 <9.0.9": "9.0.9",
    "uuid@8": "11.1.1",
    "uuid@9": "11.1.1",
    "vite": "6.4.3"
}
```

### Why these choices

**Surgical, version-scoped keys where possible.** `serialize-javascript` and `minimatch` use range-keyed overrides (`@<7.0.7`, `@>=9 <9.0.9`) so only the scoped copies are rewritten. The already-patched copies in the tree (e.g. the nine `minimatch@10.x` instances, `serialize-javascript@7.0.5`) are left untouched. This results in minimal lockfile churn minimal.

**uuid pinned to 11.1.1, not the latest 14.** uuid 13/14 are effectively ESM-only for the `require()` path and require `node ^20.19 || >=22.12`. Our runtime consumer, `sequelize`, loads uuid via CommonJS (`require("uuid").v1/.v4`), and our `engines` field allows `node >=20.10.0`. uuid **11.1.1** is the lowest fixed line that still ships a genuine CommonJS build (via the `require` export condition), so it works across our entire supported Node range. The `uuid@8` / `uuid@9` keys target the transitive copies (sequelize, sockjs, @sentry/webpack-plugin) without colliding with our direct `uuid@^14` dependency.

**vite held at 6.4.3 (not 7/8).** vite ≥7 requires `node ^20.19 || >=22.12`, which would break our `engines: node >=20.10.0` on the dev/test toolchain. 6.4.3 is the patched release that stays compatible. It's pinned as a blanket override rather than a range-keyed one because vite is a *peer* dependency of `@vitejs/plugin-vue` and `vitest`; a range-keyed override there trips an `ERESOLVE` peer conflict, whereas an exact pin keeps every vite edge in agreement. NOTE the peer dep supports v4, v5, v6, v7 and v8 so a blanket override is not an issue here. 

**Pinned exact (no `^`/`~`).** Overrides force a version tree-wide, so they're pinned to avoid silent drift on the next `npm install` / dependabot regeneration. If a pin causes issues later, it's a one-line change to relax.

### Validation

- `npm install` resolves cleanly (no `ERESOLVE`).
- Lockfile diff is limited to the six target entries plus two removals (`randombytes`, the duplicate `serialize-javascript@7.0.5`) -no unrelated transitive drift.
- Smoke-tested the one runtime-affecting change: `sequelize` defining and creating a `UUIDV4`-default model against uuid 11.1.1.

### Follow-up

`jsonata` is also flagged but is intentionally **not** addressed here. It will be handled separately in the **nr-launcher** once Node-RED is updated in and **flowfuse** (#7765).

## Related Issue(s)

dependency updates

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb (approved):
I don't like using overrides, but if it clears the old CVEs then so be it.

Steve-Mcl:
> I don't like using overrides, but if it clears the old CVEs then so be it.

I dont disagree. It creates a maintenance burden (i.e. we need to check & remove as new updates arrive) but there is no other way to update transient dependencies. To mitigate this, I did tightly scoped (where possible) (some could not be so tight due to peer deps)

The alternative is we live with wayward transient _issues_ until the parent of those transients update. 

Lets get @knolleary take before we merge.

Alternatively, we hold until Eng Meeting?

hardillb:
@Steve-Mcl what are we doing with this? it's been sitting for a while so we need to check that the changes are still valid with regard to any other dependabot PRs that have been merged.

Steve-Mcl:
> **vite held at 6.4.3 (not 7/8).** vite ≥7 requires `node ^20.19 || >=22.12`, which would break our `engines: node >=20.10.0` on the dev/test toolchain. 6.4.3 is the patched release that stays compatible. It's pinned as a blanket override rather than a range-keyed one because vite is a _peer_ dependency of `@vitejs/plugin-vue` and `vitest`; a range-keyed override there trips an `ERESOLVE` peer conflict, whereas an exact pin keeps every vite edge in agreement. NOTE the peer dep supports v4, v5, v6, v7 and v8 so a blanket override is not an issue here.

With the move to Node 24 in FF 3.0 we need to address this - lets get this merged for the time being & I'll raise an issue

PR #7778, opened 2026-07-08
Scoped PATs: clean up team scope entries on member removal

### LINKED ISSUE #7481: Scoped PATs - Token Lifecycle: Cleanup on Team Membership Removal
### Summary

  When a user is removed from a team, clean up the corresponding `AccessTokenTeamScope` entries so their scoped tokens no longer reference a team they no longer belong to.

  ### Why

  The join table approach was chosen specifically to make this cleanup straightforward. Without it, stale scope entries would reference teams the user has left, creating a confusing state where a token claims team access the user no longer has. The denormalized `UserId` on the join table makes this a single indexed query.

  ### Behavior

* **User removed from a team:** remove all `AccessTokenTeamScope` rows matching that user and team. The tokens themselves are NOT deleted. A token that was scoped only to this team ends up with zero team scopes (no team access, but the token still exists and can be re-scoped by the user).
  * **Edge case, last team scope removed:** if the removed team reference is the *last* one in a token's scope list and the user still has access to multiple teams, the token becomes globally applicable across those other teams. This was not the original intent of the token (it was meant to be limited to the now-removed team).
  * **Open decision:** we need to decide whether to (a) completely delete the token once its last team scope is removed, or (b) allow it to fall back to a "team-global" token. Each has trade-offs:
    * **Delete the token:** safest; no unintended access escalation, but breaks any integration relying on that token (silent failure for the user).
    * **Let it become team-global:** preserves the token, but grants broader access than was originally scoped, which is a potential security/privilege-escalation concern.
  * User added to a new team: do nothing. Team-agnostic tokens (no team scopes) automatically work with any team the user belongs to. Team-scoped tokens are not extended. The user must explicitly add new teams to existing tokens.

  ### What to do

  File: `forge/db/controllers/Team.js` (`removeUser` function, after the membership is destroyed)

  Add a single cleanup query using the `(UserId, TeamId)` index on the join table.

  ### Tests

  * Removing a user from a team deletes AccessTokenTeamScope entries for that user/team combination
  * The token itself is not deleted, only the team association
  * Tokens scoped to other teams for the same user are unaffected
  * Adding a user to a new team does not extend any existing scoped tokens
  * Team-agnostic tokens (no scope entries) are unaffected by membership changes

### Epic/Story

_No response_

cstns:
## Description

When a user is removed from a team, delete their `AccessTokenTeamScope` entries for that team so scoped tokens no longer eference a team they don't belong to. 

If removing the entry leaves a token with zero remaining team scopes, the token itself is deleted to prevent silent escalation to team-global access. The scope deletion and any token cleanup run inside a transaction.                                                                                          
                                                                                                                                                            
Team and user deletions are already handled by CASCADE constraints on the foreign keys.                                                                     

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/7481

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/controllers/Team.js
+++ b/forge/db/controllers/Team.js
@@ -134,6 +134,37 @@ module.exports = {
             }
             await userRole.destroy()
 
+            // Clean up scoped PAT entries that reference this team.
+            // Find which tokens had scopes for this team, remove those
+            // entries, then delete any tokens left with zero team scopes
+            // (they were scoped exclusively to the removed team and would
+            // otherwise silently escalate to team-global access).
+            const affectedScopes = await app.db.models.AccessTokenTeamScope.findAll({
+                where: { UserId: user.id, TeamId: team.id },
+                attributes: ['AccessTokenId']
+            })
+            const affectedTokenIds = affectedScopes.map(s => s.AccessTokenId)
+            if (affectedTokenIds.length > 0) {
+                await app.db.sequelize.transaction(async (t) => {
+                    await app.db.models.AccessTokenTeamScope.destroy({
+                        where: { UserId: user.id, TeamId: team.id },
+                        transaction: t
+                    })
+                    for (const tokenId of affectedTokenIds) {
+                        const remaining = await app.db.models.AccessTokenTeamScope.count({
+                            where: { AccessTokenId: tokenId },
+                            transaction: t
+                        })
+                        if (remaining === 0) {
+                            await app.db.models.AccessToken.destroy({
+                                where: { id: tokenId },
+                                transaction: t
+                            })
+                        }
+                    }
+                })
+            }
+
             await app.db.controllers.StorageSession.removeUserFromTeamSessions(user, team)
 
             return true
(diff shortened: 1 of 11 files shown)


forge/db/controllers/Team.js
```diff
@@ -134,6 +134,37 @@ module.exports = {
             }
             await userRole.destroy()
 
+            // Clean up scoped PAT entries that reference this team.
+            // Find which tokens had scopes for this team, remove those
+            // entries, then delete any tokens left with zero team scopes
+            // (they were scoped exclusively to the removed team and would
+            // otherwise silently escalate to team-global access).
+            const affectedScopes = await app.db.models.AccessTokenTeamScope.findAll({
```
cstns, line 142: Not the friendliest approach in terms of UX, users can find themselves with their tokens removed to avoid silently escalating them. Good enough to start with, we can revisit the approach if/when needed

knolleary (approved):

PR #7783, opened 2026-07-10
fix: normalise Expert inbound origins to bare origin

### LINKED ISSUE #7782: Expert panel drops nr-assistant editor messages as "unknown origin" on instances with a non-root httpAdminRoot
### Description

The Expert panel silently drops all inbound `postMessage`s from the Node-RED editor (sent by the nr-assistant editor plugin) when the instance runs the editor under a non-root `httpAdminRoot`. The browser console shows repeated:

```
Received message from unknown origin. Ignoring.
```

As a result the Expert never receives `assistant-ready` (and follow-up palette/selection/action/debug messages), so the panel does not populate.

### Cause

`allowedInboundOrigins` (`frontend/src/stores/product-assistant.js`) pushes `instance.url` into the allow-list and compares it against the incoming `MessageEvent.origin`. `instance.url` is the full editor URL and can carry the editor path (`httpAdminRoot`) or a trailing slash, whereas `MessageEvent.origin` is always a bare `scheme://host:port`. When they differ, the strict `includes()` check fails.

Example (instance with `httpAdminRoot: '/admin'`):

- allow-list entry: `https://my-instance.flowfuse.cloud/admin`
- `event.origin`: `https://my-instance.flowfuse.cloud`

These do not match, so every message is rejected. Instances served at the bare origin (no `httpAdminRoot`) are unaffected, which is why it only reproduces on some instances. The HTTP Node Security setting is unrelated.

### Steps to reproduce

1. Create/configure an instance with a non-root editor path (`httpAdminRoot`, e.g. `/admin`).
2. Open the editor and the Expert panel.
3. Observe `Received message from unknown origin. Ignoring.` in the console and an unpopulated panel.

### Fix

Normalise each allow-list entry to its origin (via `new URL(url).origin`) before comparison so the editor path and trailing slashes are stripped.

andypalmi:
## Description

Fixes #7782.

The Expert panel dropped every inbound `postMessage` sent by the nr-assistant editor plugin on instances that run the editor under a non-root `httpAdminRoot`, logging `Received message from unknown origin. Ignoring.` on repeat. With the messages rejected, the Expert never received `assistant-ready` and the panel did not populate.

`allowedInboundOrigins` compared the incoming `MessageEvent.origin` against `instance.url` / `device.editor.url` directly. Those are full editor URLs that can carry the editor path (`httpAdminRoot`) or a trailing slash, but a `MessageEvent` origin is always a bare `scheme://host:port`, so the strict `includes()` check never matched.

Example (instance with `httpAdminRoot: '/admin'`):

- allow-list entry: `https://my-instance.flowfuse.cloud/admin`
- `event.origin`: `https://my-instance.flowfuse.cloud`

This normalises each allow-list entry to its origin via `new URL(url).origin` before comparison. Instances served at the bare origin are unchanged.

## Testing

`test/unit/frontend/stores/product-assistant.spec.js` gains two regression cases: an `instance.url` carrying an editor path and one with a trailing slash both resolve to the bare origin. Full suite passes (89 tests).

### DIFF
--- a/frontend/src/stores/product-assistant.js
+++ b/frontend/src/stores/product-assistant.js
@@ -296,16 +296,26 @@ export const useProductAssistantStore = defineStore('product-assistant', {
             return state.debugLogEntries || []
         },
         allowedInboundOrigins () {
-            const allowedOrigins = [window.origin]
+            const allowedOrigins = new Set([window.origin])
             const instance = useContextStore().instance
             const device = useContextStore().device
-            if (instance?.url) allowedOrigins.push(instance.url)
-            if (device?.editor?.url) {
-                // todo this might not be needed because it's just the path to the editor tunnel, not an actual origin
-                //   and the only origin we might receive messages is the current window origin
-                allowedOrigins.push(device.editor.url)
+            // instance.url / device.editor.url can carry an editor path (httpAdminRoot) or a
+            // trailing slash, but a MessageEvent's origin is always a bare scheme://host:port.
+            // Reduce each entry to its origin so the comparison in handleMessage matches.
+            const addOrigin = (url) => {
+                try {
+                    allowedOrigins.add(new URL(url).origin)
+                } catch {
+                    // not a parseable URL - ignore
+                }
             }
-            return allowedOrigins
+            if (instance?.url) {
+                addOrigin(instance.url)
+            }
+            if (device?.editor?.url) {
+                addOrigin(device.editor.url)
+            }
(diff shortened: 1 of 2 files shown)


cstns (approved):
The fix is sensible and only battle hardens the allowed origins between domains. 

What I can't figure out is why we didn't experience this sooner. It would appear to be the kind of thing that would be caught during initial implementation.. Did something change recently on what we store on hosted/remote instance urls?

PR #7784, opened 2026-07-10
Fix navigation for custom editor paths

cstns:
## Description

extends on https://github.com/FlowFuse/flowfuse/pull/7783

Adds a useUrlHelper composable (safeOrigin, originsMatch) that wraps new URL().origin in a try/catch so a malformed URL can't throw and break message handling.                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                   
  Wires it into the two places that compare postMessage origins:                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                   
  - HostedInstanceEditorWrapper.vue was doing a raw === between event.origin (bare origin) and instance.url (can include a path from httpAdminRoot), so the comparison would silently fail for non-root editor paths. Same root cause as #7782 but in the iframe message listener rather than the Expert panel.      
  - product-assistant.js already had the fix from #7783 with an inline new URL().origin, now uses the shared composable instead.

## Related Issue(s)

<!-- What issue does this PR relate to? -->

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


andypalmi (commented):
Looks good. One thing to fix before merge:

- `UrlHelper.ts` is missing a trailing newline, which fails the `eol-last` lint rule.

Non-blocking suggestion: `useUrlHelper` has no reactive state or lifecycle, so it's a plain function factory rather than a composable. Might fit better under `utils/` alongside the other URL helpers.

andypalmi (approved):

cstns:
> Non-blocking suggestion: useUrlHelper has no reactive state or lifecycle, so it's a plain function factory rather than a composable. Might fit better under utils/ alongside the other URL helpers.

that's a good observation! I'm still undecided if we should keep both composables and utils or merge them into one

Issue #6872, opened 2026-03-16
Need to rate limit instance restart as part of Scheduled maintenance


hardillb:
### Current Behavior

All instances in a given hour are suspened/updated/restated with no delay between tasks or instances.

This can overwelm the backend container driver when large numbers of instances need restarting

### Expected Behavior

There should be a pause between suspend and restart and between each instance to reduce load on container driver.

### Steps To Reproduce

schedule large number of instances for upgrade.

### Environment

- FlowFuse version: 2.28.0
- Node.js version: 20
- npm version: 10
- Platform/OS: all
- Browser: all

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Issue #6766, opened 2026-02-26
Enable scheduled maintenance by default for new Pro/Enterprise instances


dimitrieh:
### Description

Currently, scheduled maintenance is enforced for Starter tier instances but is not enabled by default for Pro and Enterprise tiers. When a new instance is created in a Pro or Enterprise team, scheduled maintenance is off unless manually enabled.

This issue proposes making scheduled maintenance **opt-out** (enabled by default) for newly created instances across Pro and Enterprise tiers, reducing friction when we ship new functionality via stack updates.

#### Key requirements:
- New instances created in **Pro** teams should have scheduled maintenance enabled by default
- New instances created in **Enterprise** teams should have scheduled maintenance enabled by default
- Team owners/managers can opt out (disable) scheduled maintenance per instance
- Customers with strict change-control requirements must be able to turn it off — this needs clear communication
- The scheduled maintenance indicator on the instance overview should be prominent (not just 1 line) so users are aware of the setting

#### Out of scope (covered elsewhere):
- Retroactive application to existing instances → #6349, #6755
- Starter tier enforcement → FlowFuse/CloudProject#982 (done)
- Admin-level toggle for enabling the setting per tier → FlowFuse/CloudProject#1158
- Automatic NR node updating during maintenance → #6756

### Context

From internal discussion: enabling auto-update by default for new instances removes friction for exposing new functionality we develop. Starter tier already enforces this. Pro should allow opt-out. Enterprise needs careful handling as some customers cannot accept automatic updates. Clear communication to users is essential.

### Epic/Story

#4608

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

dimitrieh:
@hardillb Could you give this a quick review when you have the time? I am not entirely sure where the option defined in https://github.com/FlowFuse/CloudProject/issues/1158 shows up. Is it per instance or only for admins (so self hosted)?

hardillb:
@dimitrieh This is configured on a Team Type basis, so the settings are here:

<img width="2260" height="946" alt="Image" src="https://github.com/user-attachments/assets/c94c6fab-5e36-4b9d-9c84-26ae98edf82f" />

And then on each team type

So Start is configured as follows:

<img width="1170" height="496" alt="Image" src="https://github.com/user-attachments/assets/aa9b1853-04ca-4b49-a4d0-bd5906d9fcfb" />

[...]

PR #7822, opened 2026-07-14
fix: enable default maintenance schedule for opt-out tiers on instance creation

### LINKED ISSUE #7821: New Pro/Enterprise instances don't get scheduled maintenance enabled by default (#6766 not effective on FFC)
### Current Behavior

New Pro and Enterprise instances are **not** getting scheduled maintenance enabled by default. This is the behaviour that #6766 was meant to deliver ("Enable scheduled maintenance by default for new Pro/Enterprise instances", closed as deployed on 8 May 2026).

The team-type config side did ship: on FFC prod, both Pro and Enterprise team types have `autoStackUpdate: { enabled: true, allowDisable: true }`. But no code path actually applies a schedule to new instances in those tiers, so scheduled maintenance stays off unless a user turns it on manually. That is the old, pre-#6766 opt-in behaviour.

### Evidence (FFC prod, read-only mirror)

Looking at instances created since the 8 May deploy and whether they have a `stackUpgradeHour_*` setting (the per-instance schedule marker):

- Starter (enforced, `allowDisable:false`): effectively 100% have a schedule.
- Pro (`allowDisable:true`): effectively 0% have a schedule (a tiny handful, consistent with manual opt-in).
- Enterprise (`allowDisable:true`): effectively 0% have a schedule (same pattern).

The monthly trend (Jan to Jul 2026) shows no change at the deploy date: the proportion of new Pro/Enterprise instances with a schedule is unchanged before and after 8 May. Starter enforcement works as intended. (Absolute figures omitted here as they are internal usage numbers; happy to share them via an internal channel.)

### Live reproduction

1. Created a fresh Medium hosted instance in an Enterprise team (`allowDisable:true`).
2. Instance got its `settings`, `credentialSecret` and `k8sType` settings written at creation, but no `stackUpgradeHour_*` row.
3. Re-checked after a delay to rule out replication lag: still no schedule.

### Root cause

Two places gate the schedule write, and both exclude opt-out tiers:

1. Instance creation, `forge/db/controllers/Project.js:558`:
   ```js
   if (autoStackUpdate.enabled && !autoStackUpdate.allowDisable) {
   ```
   The `!autoStackUpdate.allowDisable` clause means a schedule is only applied when the tier cannot disable it. Pro/Enterprise have `allowDisable: true`, so this is never true for them and no schedule is written at creation.

2. Daily enforcement job, `forge/ee/lib/autoUpdateStacks/tasks/enforce-team-rules.js:18`:
   ```js
   if (typeProperties.autoStackUpdate?.allowDisable === false) {
   ```
   Also only targets `allowDisable === false` tiers, so it does not backfill Pro/Enterprise either.

Net effect: only enforced tiers (Starter) ever receive a schedule anywhere. The "enabled by default, opt-out" requirement for Pro/Enterprise was never implemented in code; only the team-type config that surfaces the opt-out control shipped.

### Expected Behavior

A new Pro/Enterprise instance should have scheduled maintenance enabled by default (a random day/hour picked from the tier's `autoStackUpdate.days`/`hours`), with the owner able to opt out afterwards.

### Suggested fix

At `Project.js:558`, apply the default whenever the tier has it enabled, regardless of `allowDisable`:

```js
if (autoStackUpdate.enabled) {
```

`allowDisable` should only govern whether the user can subsequently turn it off, not whether it is on by default at creation. The daily enforcement job should stay scoped to `allowDisable === false`, so opt-out tiers are not force-re-enabled after a user opts out. Retroactive application to existing Pro/Enterprise instances is a separate question (see #6349 / #6755).

### Related

- #6766 (the feature this was meant to deliver, closed as deployed)
- Epic #4608
- #6349 (retroactive application to existing instances)

### Environment

- FlowFuse Cloud (app.flowfuse.com), observed 14 July 2026
- Verified against the read-only prod DB mirror and a live instance creation in an Enterprise team

robmarcer:
## Description

New instances only got a scheduled maintenance window applied at creation when the team type had `autoStackUpdate.enabled` **and** `allowDisable === false`. That combination only matches enforced tiers (Starter), so opt-out tiers (Pro/Enterprise, `allowDisable: true`) were created with no schedule even though the tier has maintenance enabled by default.

This changes the creation-time check in `forge/db/controllers/Project.js` to apply the default schedule whenever the tier has `autoStackUpdate.enabled`, regardless of `allowDisable`. `allowDisable` governs whether a user can later turn maintenance off, not whether it is on by default at creation.

```diff
-                if (autoStackUpdate.enabled && !autoStackUpdate.allowDisable) {
+                if (autoStackUpdate.enabled) {
```

No tier regresses: for `allowDisable: false` tiers the condition is unchanged in effect (`enabled` was already required), and the daily enforcement job (`enforce-team-rules.js`) is intentionally left scoped to `allowDisable === false` so opt-out tiers are not force-re-enabled after a user opts out.

## Scope

This is the forward fix only: it stops new Pro/Enterprise instances being created without a schedule. Retroactively applying a schedule to existing instances that were created without one is deliberately **out of scope** here and tracked separately (see #7821 and #6349), as it needs customer comms and restart rate-limiting (#6872).

## Related

- Fixes #7821
- Delivers the behaviour intended by #6766
- Epic #4608

## Checklist

- [x] Minimal, targeted change
- [ ] Regression test for the creation path on an opt-out tier (sensible follow-up; existing tests cover only the enforcement job)

### DIFF
--- a/forge/db/controllers/Project.js
+++ b/forge/db/controllers/Project.js
@@ -555,7 +555,7 @@ module.exports = {
             await team.ensureTeamTypeExists()
             if (team.TeamType.getProperty('autoStackUpdate')) {
                 const autoStackUpdate = team.TeamType.getProperty('autoStackUpdate')
-                if (autoStackUpdate.enabled && !autoStackUpdate.allowDisable) {
+                if (autoStackUpdate.enabled) {
                     const days = autoStackUpdate.days
                     const hours = autoStackUpdate.hours
                     // generate random day and hour in ranges


hardillb (approved):
It looks OK, and appears to do what was intended.

Issue #7563, opened 2026-06-18
3rd Party MCP - Inject Audit Headers in MCP Tool Calls


cstns:
### Summary

Wire up the MCP endpoint to inject custom audit headers when calling existing routes via `app.inject()`, so the audit system can distinguish MCP actions from regular API calls.

### Why

The previous task (#7562) added the infrastructure: a source column, request context propagation, and audit logger updates. This task connects the MCP endpoint to that infrastructure by adding the right headers to every `app.inject()` call.

The first-party expert agent also needs to inject these headers when it uses the MCP path. Distinguishing `mcp` from `mcp:expert` lets administrators see which actions came from their own agent vs external agents, which is important for trust and debugging.

### What to do

* In the MCP endpoint's shared `inject()` helper (`_helpers.js` from #7430), add `X-FF-Audit-Source` and `X-FF-Audit-Token` headers to every `app.inject()` call
* Source value: `mcp` for 3rd party agents, `mcp:expert` for the first-party expert agent. The MCP endpoint knows which it is based on the authentication context.
* Token value: the PAT identifier from `request.session.pat.id`
* For Phase 2 MQTT-bridged tools: the same headers should be injected if/when the bridge makes any `app.inject()` calls internally. For tools that only go through MQTT (no inject), the audit entry is created at the MCP HTTP layer before bridging.

### Tests

* Platform tool call via 3rd party MCP: audit entry has `source: 'mcp'`
* Platform tool call via first-party expert: audit entry has `source: 'mcp:expert'`
* Token identifier in audit matches the PAT used
* Existing non-MCP audit entries are unaffected

### References

* MCP inject helper: #7430 (`_helpers.js`)
* Audit source infrastructure: #7562
* Parent story: #7425
* Depends on: #7430 (MCP tool definitions), #7562 (audit source column)

hegu-1:
Distinguishing mcp from mcp:expert is useful, but I would treat these headers as transport context, not as the audit fact itself.

The source value must be derived from authenticated server-side context and stripped or overwritten if a caller supplies the same header. The PAT id is a credential reference, not actor identity, so the persisted audit record should also bind:

- tenant, user/service actor, and agent/runtime identity
- run and tool-call correlation ids
- canonical tool and target
- policy decision plus policy version
- requested, attempted, and actual outcome
- timestamp, request digest, and any resulting resource id

[...]

Issue #7425, opened 2026-06-11
Story 4: MCP Audit Trail


cstns:
### Summary

Ensure that actions performed through the MCP endpoint are clearly distinguishable from regular user actions in the audit trail, so administrators can tell "user did this in the UI" from "agent did this via MCP."

### Context

The MCP platform tools use `app.inject()` to call existing HTTP routes. Audit logging happens inside those route handlers. From the route's perspective, an MCP-originated request looks like a normal authenticated request. Without explicit signaling, there's no way for the audit system to know the action came from an MCP agent rather than a user clicking a button.

The audit system is centralized: all audit calls go through `app.auditLog.[Scope].[category].[action]()` which writes to the `AuditLogs` table via controller functions in `forge/db/controllers/AuditLog.js`. The formatters in `forge/auditLog/formatters.js` produce a `trigger` object, but it only identifies who (user vs system), not how (UI vs API vs MCP). There is currently no "source" field in the database.

### Decisions

**Source categories:** A new `source` column on the AuditLog model with these values:
- `null` (or `ui`): cookie session, browser action (default/legacy)
- `api`: PAT-authenticated direct API call (no MCP)
- `mcp`: 3rd party MCP agent
- `mcp:expert`: first-party expert agent via MCP path

A queryable column (rather than JSON body field) lets administrators filter audit logs by source directly.

**Signaling mechanism:** Custom headers on `app.inject()` calls from MCP:
- `X-FF-Audit-Source: mcp` (or `mcp:expert`)
- `X-FF-Audit-Token: <tokenId>`

These headers are read after `verifySession` and stored in `@fastify/request-context` (already used for PAT metadata from #7411). The centralized audit loggers read from the request context without changing their function signatures or touching individual route handlers.

For non-MCP PAT requests, the source is auto-detected: if `request.session.pat` exists and no MCP header is present, source is `api`.

**Token traceability:** The audit entry includes the token identifier when available (from `X-FF-Audit-Token` or `request.session.pat.id`). This lets admins trace which specific PAT/integration caused the action, implicitly distinguishing different agents (each uses a different token).

### Sub-tasks

- #7562 Audit source column + request context propagation
- #7563 Inject audit headers in MCP tool calls

### What this story does NOT cover
- Audit for non-MCP actions (existing system, unchanged)
- PAT scopes (#7411), handled externally
- MCP endpoint implementation details (Story 1)

### References
- Audit log controller: `forge/db/controllers/AuditLog.js`
[...]

Issue #7411, opened 2026-06-10
Story 0: Add PAT (Personal Acces Tokens) scopes


cstns:
## Problem
                                                                                                                                                                                                                                                                                                                              
  PATs currently inherit the full access of the user who created them. For an admin, that means every token is an admin token with read/write across the entire platform. With AI agents (MCP clients, Claude Desktop, Cursor, etc.) increasingly using PATs to interact with FlowFuse, the blast radius is unacceptable.   
  There is no way to issue a reduced-privilege token, and the only mitigation is policy, which doesn't scale.                                                                                                                                                                                                                 

  There will be no separate MCP token type. MCP clients will authenticate with scoped PATs, same as any other API consumer.                                                                                                                                                                                                   

##  What we're adding                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                            
  Three restriction levers on PATs, all optional, all narrowing:                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                            
  1. **Multi-team scoping**
  A token can be scoped to one or more teams. If no teams are selected, the token has access to all teams the user belongs to (backwards compatible). Implemented via a join table (AccessTokenTeamScopes) between tokens, teams, and users. The join table includes a denormalized UserId column alongside AccessTokenId and TeamId. This is intentional: AccessToken.ownerId is a polymorphic STRING field with no real FK constraint, so routing cleanup queries through it would require string coercion and a subquery. A proper INTEGER FK on UserId gives us a direct indexed lookup for the primary use case (clean up scope entries when a user  
  is removed from a team). This follows the same pattern as Device.TeamId, which is stored directly even though it's derivable from the Application relationship.                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                              
  2. **Read-only toggle**
  A single global boolean per token. When enabled, the token can only perform read operations across all accessible resources. Each permission in the system gets tagged as read or write to support this. Per-team read/write granularity might come later.

  3. **Admin opt-in**
  Off by default. Admin users' tokens are treated as regular users (using their team role, not the admin bypass) unless the admin checkbox is explicitly checked at creation time. Existing admin PATs will lose admin bypass after this ships, which is intentional.                                                         

## How it works
  The effective permissions of a scoped PAT are always the intersection of the user's real permissions (including application-level granular RBAC) and the token's restrictions. A token can only narrow access, never widen it.                                                                                              

[...]

PR #7825, opened 2026-07-15
Add source column to AuditLog with request context propagation

### LINKED ISSUE #7562: 3rd Party MCP - Audit Source Column + Request Context Propagation
### Summary

Add a `source` column to the AuditLog model and propagate request source information through `@fastify/request-context` so audit loggers can record how an action was triggered.

### Why

MCP platform tools use `app.inject()` to call existing HTTP routes. Audit logging happens inside those route handlers. From the route's perspective, an MCP-originated request looks identical to a regular API request. Without explicit signaling, administrators have no way to tell "user did this in the UI" from "3rd party agent did this via MCP" from "first-party expert agent did this."

A queryable database column (rather than burying it in the JSON body) lets administrators filter audit logs by source directly, which is the primary use case: "show me everything an MCP agent did in the last 24 hours."

`@fastify/request-context` is the carrier because it's already used for PAT metadata (#7411). The audit loggers can read from it without changing their function signatures or touching every route handler that calls them. The source is set once (during auth / header processing) and available everywhere downstream.

### What to do

**Database migration**
* Add `source` column to `AuditLogs` table: STRING, nullable, default null
* Null represents the legacy/UI path (cookie session, no PAT, no MCP header)

**Source detection and propagation**
* After `verifySession`, detect the request source and store it in `@fastify/request-context`:
  * Cookie session (no PAT): source remains null (or `ui` if we want to be explicit)
  * PAT present, no MCP header: `api`
  * MCP custom header `X-FF-Audit-Source: mcp`: `mcp` (3rd party agent)
  * MCP custom header `X-FF-Audit-Source: mcp:expert`: `mcp:expert` (first-party agent)
* Also store the token identifier from `X-FF-Audit-Token` header when present, for traceability to the specific PAT/integration

**Audit logger updates**
* Update the centralized audit logging functions (`platformLog`, `userLog`, `teamLog`, `projectLog`, `applicationLog`, `deviceLog` in `forge/db/controllers/AuditLog.js`) to read source from `@fastify/request-context` and pass it to the model
* Update the AuditLog view formatter to include `source` in the output
* Update the `triggerObject()` in `formatters.js` to include the token identifier when available

### Tests

* MCP-originated action (via `app.inject()` with custom headers) records `source: 'mcp'`
* First-party expert action records `source: 'mcp:expert'`
* Direct PAT API call records `source: 'api'`
* Cookie session action records `source: null` (or `ui`)
* Token identifier is included in the audit body when present
* Existing audit entries (before migration) have null source
* Audit log API returns the source field

### References

* Audit log controller: `forge/db/controllers/AuditLog.js`
* Audit formatters: `forge/auditLog/formatters.js`
* `@fastify/request-context` setup: #7411 (T1, T2)
* Parent story: #7425

cstns:
## Description

Adds infrastructure so audit log entries can record how an action was triggered, not just who triggered it. A new source column on the AuditLog table distinguishes cookie sessions (null), direct PAT API calls (api), first party expert agent calls (mcp:expert), and eventually third party MCP agents (mcp).

- New source column on AuditLogs (nullable, null = legacy/cookie session)
- Generic in-memory nonce store (app.nonceStore) for process-local trust boundaries. Nonces are single-use and short-lived (30s TTL). Includes a createSourceNonce() factory that validates required fields and auto-generates a correlationId when not provided.
- resolveSourceContext in verifySession consumes nonces and populates sourceContext in @fastify/request-context. Falls back to source: 'api' with tokenId for plain PAT calls, null for cookie sessions.
- Audit log controllers read from request context and write the source column. When source context includes a correlationId (MCP calls), a sourceContext object is merged into the audit body with toolName, correlationId, and tokenId.
- Formatter and view updates to include source in API output.

Anti-spoofing: external requests can't set a fake source because only in-process code can mint nonces via app.nonceStore. No valid nonce means no MCP source, regardless of what headers a caller sends.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/7562

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/lib/nonceStore.js
+++ b/forge/lib/nonceStore.js
@@ -53,10 +53,11 @@ class NonceStore {
         if (!metadata?.source) {
             throw new Error('source is required for source context nonces')
         }
-        if (!metadata.correlationId) {
-            metadata.correlationId = crypto.randomUUID()
+        const meta = { ...metadata }
+        if (!meta.correlationId) {
+            meta.correlationId = crypto.randomUUID()
         }
-        return this.create(metadata, ttl)
+        return this.create(meta, ttl)
     }
 
     /**
--- a/forge/lib/nonceStore.js
+++ b/forge/lib/nonceStore.js
@@ -1,0 +1,103 @@
+/**
+ * Generic in-memory nonce store.
+ *
+ * Each nonce is single-use (consumed on first read) and expires after a
+ * configurable TTL. Callers attach arbitrary metadata when creating a nonce
+ * and receive it back on consumption.
+ *
+ * In-memory only: intended for process-local trust boundaries such as
+ * Fastify's app.inject(), where the producer and consumer always live in
+ * the same Node.js process.
+ *
+ * Registered as a Fastify plugin on `app.nonceStore`.
+ */
+
+const crypto = require('crypto')
+
+const fp = require('fastify-plugin')
+
+const DEFAULT_TTL_MS = 30_000 // 30 seconds
+const SWEEP_INTERVAL_MS = 60_000 // 1 minute
+
+class NonceStore {
+    constructor () {
+        /** @type {Map<string, { metadata: object, expiresAt: number }>} */
+        this._store = new Map()
+        this._sweepTimer = setInterval(() => this._sweep(), SWEEP_INTERVAL_MS)
+        this._sweepTimer.unref()
+    }
+
+    /**
+     * Create a nonce and associate metadata with it.
+     * @param {object} metadata - Arbitrary data returned on consume()
+     * @param {number} [ttl] - Time-to-live in ms (default 30 000)
+     * @returns {string} The nonce token
+     */
+    create (metadata, ttl = DEFAULT_TTL_MS) {
+        const nonce = crypto.randomBytes(16).toString('hex')
+        this._store.set(nonce, {
+            metadata,
+            expiresAt: Date.now() + ttl
+        })
+        return nonce
+    }
+
+    /**
+     * Create a nonce for source context propagation.
+     * Validates that source is present. Generates a correlationId if not provided.
+     * @param {{ source: string, correlationId?: string, toolName?: string, tokenId?: number }} metadata
+     * @param {number} [ttl] - Time-to-live in ms (default 30 000)
+     * @returns {string} The nonce token
+     */
+    createSourceNonce (metadata, ttl) {
+        if (!metadata?.source) {
+            throw new Error('source is required for source context nonces')
+        }
+        if (!metadata.correlationId) {
+            metadata.correlationId = crypto.randomUUID()
+        }
+        return this.create(metadata, ttl)
+    }
+
+    /**
+     * Consume a nonce, returning its metadata exactly once.
+     * Returns null if the nonce does not exist or has expired.
+     * @param {string} nonce
+     * @returns {object|null}
+     */
+    consume (nonce) {
+        const entry = this._store.get(nonce)
+        if (!entry) {
+            return null
+        }
+        this._store.delete(nonce)
+        if (entry.expiresAt < Date.now()) {
+            return null
+        }
+        return entry.metadata
+    }
+
+    /** Remove all expired entries. */
+    _sweep () {
+        const now = Date.now()
+        for (const [key, entry] of this._store) {
+            if (entry.expiresAt < now) {
+                this._store.delete(key)
+            }
+        }
+    }
+
+    /** Stop the periodic sweep timer. */
+    close () {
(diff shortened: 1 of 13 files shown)


andypalmi:
**Test coverage gap: the nonce trust boundary is untested**

`resolveSourceContext` and the `x-ff-source-nonce` path in `forge/routes/auth/index.js` have no test coverage. The controller tests in `AuditLog_spec.js` monkeypatch `requestContext.get`, which bypasses the actual source-resolution logic in auth.

The most security-relevant case is never exercised: a request with a bogus or expired nonce must not receive an MCP source. Suggested auth-level cases:

- bogus/expired nonce -> source stays null
- valid minted nonce -> sourceContext populated from the nonce metadata
- PAT session without a nonce -> source `api` with tokenId

Since this is the anti-spoofing boundary, it is worth covering here rather than deferring to #7563.

andypalmi:
**Follow-up: reconcile the downstream issues with the nonce approach**

The PR description documents the nonce mechanism (`x-ff-source-nonce` plus `app.nonceStore`) and its anti-spoofing rationale well. Flagging only that the downstream artifacts still describe the earlier header scheme (`X-FF-Audit-Source` / `X-FF-Audit-Token`) and will need updating so the story stays consistent:

- #7563 is written against the header scheme and will need to mint via `app.nonceStore.createSourceNonce()` and send the nonce header instead
- the Decisions section in the parent story #7425 still describes the header mechanism

Not a change to this PR, just a heads up for the follow-up work.

andypalmi:
**`createSourceNonce` mutates the caller's metadata object**

In `forge/lib/nonceStore.js`, `createSourceNonce` assigns `metadata.correlationId` back onto the passed-in object when one is not provided. A caller that reuses or inspects that object afterwards will find a correlationId it did not set. Cloning first avoids the side effect:

```js
const meta = { ...metadata }
if (!meta.correlationId) {
    meta.correlationId = crypto.randomUUID()
}
return this.create(meta, ttl)
```

andypalmi:
**Minor: `body.sourceContext` could collide with an existing body key**

`applySourceContext` in `forge/db/controllers/AuditLog.js`, and the passthrough in `forge/auditLog/formatters.js`, merge `sourceContext` into the audit body. If an audit event ever carried its own `sourceContext` key it would be overwritten. This is practically impossible for current audit bodies, so noting only for awareness. No change needed unless you want a defensive guard.

andypalmi:
**What is solid**

- The nonce approach is a stronger anti-spoofing model than trusting headers: external callers cannot mint a nonce, so no valid nonce means no MCP source regardless of what headers are sent.
- The nonce is checked first on both the cookie and PAT auth paths, is single-use (deleted before the expiry check), TTL bound, swept periodically, and the sweep timer is unref'd.
- Cross `app.inject()` propagation is correct: the nonce is re-resolved inside the inner request, so the source context lands in that request's context store and the audit controller reads it back. This mirrors the existing `pat` / `isPAT` mechanism from #7411.
- Reading source context is safe outside a request: `requestContext.get` returns undefined when no store is active, so audit writes from background tasks stay `source: null` rather than throwing.
- The migration is nullable and backwards compatible, and is correctly the latest migration by date.
- Good unit coverage for the nonce store, formatters, and controller source handling.

andypalmi (approved):
Looks great, thanks for the quick turnaround on the earlier notes. The nonce clone fix is in, and the new `resolveSourceContext` integration test covers exactly the anti-spoofing cases that matter: bogus nonce, expired nonce, valid nonce, PAT fallback, and single-use reuse, all checked end to end against the audit source column. That was the main thing I wanted to see.

The nonce approach reads as a solid anti-spoofing model, and the migration is nicely backwards compatible. Approving. Nothing blocking from me.

forge/lib/nonceStore.js
```diff
@@ -0,0 +1,104 @@
...
+ * Generic in-memory nonce store.
+ *
+ * Each nonce is single-use (consumed on first read) and expires after a
+ * configurable TTL. Callers attach arbitrary metadata when creating a nonce
+ * and receive it back on consumption.
+ *
+ * In-memory only: intended for process-local trust boundaries such as
+ * Fastify's app.inject(), where the producer and consumer always live in
+ * the same Node.js process.
+ *
```
hardillb, line 11: @cstns can you explain this to me please

hardillb:
This will need types re-generated if merged after https://github.com/FlowFuse/flowfuse/pull/7733

hardillb (approved):

cstns:
will merge it now

PR #7830, opened 2026-07-15
Inject source context headers in MCP tool calls

### LINKED ISSUE #7563: 3rd Party MCP - Inject Audit Headers in MCP Tool Calls

cstns:
## Description

Wires up the MCP platform automation endpoint to use the source context infrastructure from #7562, so expert agent tool calls are recorded as mcp:expert in the audit trail.                                                                                                                                                

- platformAutomation.js mints a nonce per app.inject() call via app.nonceStore.createSourceNonce() with source: mcp:expert and the tool name.                                                 
- Audit log UI (AuditEntry.vue) shows a sparkles icon next to the username for MCP-sourced entries and a terminal icon for API calls. Tooltip includes the source and tool name (e.g. "via Expert (Tool name: create-instance)").

image

image

image

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/7563

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/config/cypress-shared.config.js
+++ b/config/cypress-shared.config.js
@@ -5,6 +5,10 @@ const downloadsFolder = 'test/e2e/frontend/cypress/downloads'
 module.exports = {
     viewportWidth: 1024,
     viewportHeight: 768,
+    retries: {
+        runMode: 2,
+        openMode: 0
+    },
     e2e: {
         downloadsFolder,
         fixturesFolder: 'test/e2e/frontend/cypress/fixtures',
--- a/forge/auditLog/formatters.js
+++ b/forge/auditLog/formatters.js
@@ -141,6 +141,7 @@ const formatLogEntry = (auditLogDbRow) => {
         User: auditLogDbRow.User, // TODO: Kept for compatibility. Remove once Audit Log UI overhaul complete
         event: auditLogDbRow.event,
         createdAt: auditLogDbRow.createdAt,
+        source: auditLogDbRow.source || null,
         scope: {
             id: auditLogDbRow.entityId,
             type: auditLogDbRow.entityType
@@ -199,6 +200,12 @@ const formatLogEntry = (auditLogDbRow) => {
             if (body?.key && body?.scope && body?.store && (body.store === 'memory' || body.store === 'persistent')) {
                 formatted.body = formatted.body || {}
                 formatted.body.context = { key: body.key, scope: body.scope, store: body.store }
+            }
+
+            // pass through PAT audit context if present
+            if (body?.sourceContext) {
+                formatted.body = formatted.body || {}
+                formatted.body.sourceContext = body.sourceContext
             }
 
             // format log entries for know Node-RED audit events
--- a/forge/comms/commsClient.js
+++ b/forge/comms/commsClient.js
@@ -232,8 +232,12 @@ class CommsClient extends EventEmitter {
                 'ff/v1/+/d/+/resources/heartbeat',
                 // Platform sync messages
                 'ff/v1/platform/sync',
-                // Listen for Expert platform requests
-                'ff/v1/expert/+/+/platform/+/request'
+                // Listen for Expert platform requests.
+                // Uses a dedicated shared subscription group. The group name defines the set
+                // of consumers that share the workload, so keeping Expert separate from the
+                // "platform" group prevents unrelated features from sharing a consumer pool and
+                // allows them to scale independently.
+                '$share/expert/ff/v1/expert/+/+/platform/+/request'
             ])
         }
     }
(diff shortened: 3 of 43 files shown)


andypalmi:
Would it be worth adding some test coverage here? Assuming the `resolveSourceContext` tests land in #7825, this PR could probably cover the two new pieces:

- a unit test for the inject behavior in `platformAutomation.js`, checking a tool call mints a nonce with `source: 'mcp:expert'` and the right `toolName` and sends it as `x-ff-source-nonce`
- an integration test for the full path: drive a platform automation tool call and assert the resulting audit entry has `source: 'mcp:expert'` and the expected `toolName`

That would close the loop on the audit path end to end. Happy to help write these if it's useful.

andypalmi:
Small question on the nonce metadata: was leaving out `tokenId` intentional here? The nonce carries `source` and `toolName` but not the token id:

```js
const nonce = this.app.nonceStore.createSourceNonce({
    source: 'mcp:expert',
    toolName
})
```

#7563 mentions the token identifier as part of the source context, and the #7562 infra already carries `tokenId` through into the audit body, so the plumbing is there. It looks like the main blocker is that `getOrCreateMcpPlatformToken` only returns `{ token }` and not the AccessToken id, so there's nothing to pass yet.

If that's the reason, would it be worth surfacing the id from that helper and setting `tokenId` on the nonce? It becomes more useful once third-party `mcp` agents (each with their own PAT) arrive and you want to tell them apart. Not blocking, just wanted to check the thinking.

andypalmi:
Minor one on the SCSS: dropping `display: block` makes sense for the flex icon layout. I traced the grid and it still renders correctly in both the mobile and large layouts, since the DOM order (time, info, trigger) lines up with the `grid-template-areas` order and `.ff-audit-entry-info` is the one explicitly placed cell.

The only thing is that removing the `grid-area` assignments means placement now relies on auto-placement matching DOM order, so a future reorder could shift things without warning. Would it be worth keeping the two `grid-area` declarations just to make it explicit? They don't conflict with the new flex. Totally optional.

andypalmi:
Not related to this PR, just spotted it while reading nearby: in the tamper check just above, `tool.annotations` is read before the `if (!tool)` guard, so an unknown tool name would throw a raw error instead of the friendly `MCP_PLATFORM_TOOL_NOT_FOUND`. Might be worth reordering the null check at some point. Feel free to ignore for now since it predates this change.

andypalmi:
Nice work on this, the nonce wiring looks clean. A few things I liked while reviewing:

- The nonce is minted fresh per `inject()` call, which is right for single-use, and it's the header the base branch's `resolveSourceContext` consumes.
- The old `x-ff-automation-source: expert` header is fully removed with no leftover readers.
- The metadata passed to `createSourceNonce` is a fresh object each call, so the mutation note from #7825 doesn't apply here.
- Frontend reads `entry.source` and `entry.body?.sourceContext?.toolName`, matching the view and formatter output from #7562.
- The `source === 'mcp'` UI branch being ready for third-party agents ahead of time is a nice touch.

Nothing blocking from me, just the couple of questions above, but I really think we should add tests.

cstns:
Logging/exposing the token doesn't make sense for the first-party expert agent (short-lived, nameless, user-invisible tokens). The scaffolding for token name is already in place via the PAT path in resolveSourceContext for when third-party MCP agents arrive.

andypalmi (approved):
Looks good to me

PR #7834, opened 2026-07-15
[7833] Flaky HA unit test + update cypress config

### LINKED ISSUE #7833: Flaky HA unit test (probably)
### Current Behavior

2 consecutive same-test-fails on PR [#7831](https://github.com/FlowFuse/flowfuse/pull/7831):

* https://github.com/FlowFuse/flowfuse/actions/runs/29411291983/attempts/2?pr=7831
* https://github.com/FlowFuse/flowfuse/actions/runs/29411291983/attempts/3?pr=7831

I suspect this is actually a flaky test - dependant on runtime umph - but its just a guess based on the nature of the code in the test:
```
call /api/v1/projects/${TestObjects.project.id}/ha => replicas: 2
await 250ms
await project reload()
project.state.should.equal('suspended')
```

IMO, the "suspended" assertion (`TestObjects.project.state.should.equal('suspended')`) is testing the stub driver's timing, not the HA feature. So a potential fix is to ditch that!

But to be safe - in case this is a real (time sensitive) bug, I am logging it.

Below is the trace that happened on test run 2 & 3 (in case above links are deleted & to be searchable)

```
1) HA Instance API
       applies valid ha option:

      AssertionError: expected 'running' to be 'suspended'
      + expected - actual

      -running
      +suspended

      at Assertion.value (node_modules/should/cjs/should.js:356:19)
      at Context.<anonymous> (test/unit/forge/ee/routes/ha/index_spec.js:101:42)
      at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

### Expected Behavior

Should pass tests regardless of _grunt_ on the test runner

### Steps To Reproduce

NA

### Environment

- FlowFuse version: FFC
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

n-lark:
## Description

Several unrelated CI flakes, green locally, red on runners. Different causes, not one bug.

| File | Fix |
|---|---|
| `ha/index_spec.js` | Dropped a `state === 'suspended'` check that raced the restart's re-write; kept the reliable `inflight` check. |
| `InstanceStatusPolling.vue` | Caught the 404 when an instance is deleted mid-poll (was an unhandled rejection failing random tests). |
| `tokens.spec.js` | Count baseline now read via retryable `cy.get()` (not sync `Cypress.$` pre-render); expiry assertion made timezone-safe. |
| `admin.spec.js` | Wait for the team redirect to settle on `/overview` before clicking the Applications nav. The click was racing the pending redirect and getting bounced back. |
| `cypress-shared.config.js` | `retries: { runMode: 2 }` absorbs CI variance. Retried passes still log `Attempt 2 of 3`. See below. |

**Retries:** on CI a failing test is re-run up to 2 more times (3 attempts total); it only counts as failed if all 3 fail, so a one-off hiccup no longer reds the build. Local `cypress open` keeps 0 retries for instant feedback. 

Downside: a genuinely flaky test can now pass silently, but will torture us less. Check run logs for `Attempt 2 of 3` when that matters.

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7833

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/config/cypress-shared.config.js
+++ b/config/cypress-shared.config.js
@@ -5,6 +5,10 @@ const downloadsFolder = 'test/e2e/frontend/cypress/downloads'
 module.exports = {
     viewportWidth: 1024,
     viewportHeight: 768,
+    retries: {
+        runMode: 2,
+        openMode: 0
+    },
     e2e: {
         downloadsFolder,
         fixturesFolder: 'test/e2e/frontend/cypress/fixtures',
--- a/test/e2e/frontend/cypress/tests/account/tokens.spec.js
+++ b/test/e2e/frontend/cypress/tests/account/tokens.spec.js
@@ -86,10 +86,11 @@ describe('FlowFuse - Personal Access Tokens', () => {
     })
 
     it('can be added with an expiry date', () => {
-        // count how many tokens we already have
-        const rows = Cypress.$('[data-el="tokens-table"] tbody').find('tr.ff-data-table--row').length
-
         cy.intercept('POST', '/api/*/user/tokens').as('addPersonalAccessToken')
+
+        // retryable read - a synchronous Cypress.$ count can beat the table render on slow CI, undercounting the baseline
+        let rows
+        cy.get('[data-el="tokens-table"] tbody').find('tr.ff-data-table--row').then($rows => { rows = $rows.length })
 
         cy.get('[data-action="new-token"]').click()
 
@@ -108,10 +109,12 @@ describe('FlowFuse - Personal Access Tokens', () => {
         cy.get('[data-el="add-token-confirmation"] [data-action="token-confirmation-done"]').click()
 
         // check we have one more token than before
-        cy.get('[data-el="tokens-table"] tbody').find('tr').should('have.length', rows + 1)
-        cy.get('[data-el="tokens-table"] tbody').find('tr').last().should('contain', '31')
-        cy.get('[data-el="tokens-table"] tbody').find('tr').last().should('contain', '12')
-        cy.get('[data-el="tokens-table"] tbody').find('tr').last().should('contain', '2050')
+        cy.get('[data-el="tokens-table"] tbody').find('tr.ff-data-table--row').should($rows => {
+            expect($rows).to.have.length(rows + 1)
+        })
+        // the cell renders Date.parse(expiresAt).toLocaleDateString(); derive the label the same way so the assertion is timezone-independent
+        const expiryLabel = new Date(Date.parse('2050-12-31')).toLocaleDateString()
+        cy.get('[data-el="tokens-table"] tbody').find('tr').last().should('contain', expiryLabel)
     })
 
     it('can be added with read-only enabled', () => {
--- a/test/e2e/frontend/cypress/tests/admin.spec.js
+++ b/test/e2e/frontend/cypress/tests/admin.spec.js
@@ -65,11 +65,15 @@ describe('FlowFuse platform admin users', () => {
         cy.get('[data-nav="admin-teams"]').click()
         cy.wait('@getTeams')
 
+        // check that we're on the admin/teams page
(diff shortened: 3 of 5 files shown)


test/e2e/frontend/cypress/tests/account/tokens.spec.js
```diff
@@ -108,10 +109,12 @@ describe('FlowFuse - Personal Access Tokens', () => {
         cy.get('[data-el="add-token-confirmation"] [data-action="token-confirmation-done"]').click()
 
         // check we have one more token than before
-        cy.get('[data-el="tokens-table"] tbody').find('tr').should('have.length', rows + 1)
-        cy.get('[data-el="tokens-table"] tbody').find('tr').last().should('contain', '31')
-        cy.get('[data-el="tokens-table"] tbody').find('tr').last().should('contain', '12')
-        cy.get('[data-el="tokens-table"] tbody').find('tr').last().should('contain', '2050')
```
Steve-Mcl, line 114: nit - are these exact presence checks handled in other test? If not, I would be tempted to include (a version of) them in the `should =>` callback on line 113?

not a blocker.

Steve-Mcl (approved):
Approved with nit - you decide 😸

config/cypress-shared.config.js
```diff
@@ -5,6 +5,10 @@ const downloadsFolder = 'test/e2e/frontend/cypress/downloads'
 module.exports = {
     viewportWidth: 1024,
     viewportHeight: 768,
+    retries: {
```
cstns, line 8: I was certain we had this enabled for this runner.. I must have been thinking of the dashboard repo

test/e2e/frontend/cypress/tests/account/tokens.spec.js
cstns: This is just checking that the correct expiry date was set on the token, I'd say it's good enough

test/e2e/frontend/cypress/tests/account/tokens.spec.js
```diff
@@ -86,11 +86,12 @@ describe('FlowFuse - Personal Access Tokens', () => {
...
 
     it('can be added with an expiry date', () => {
-        // count how many tokens we already have
-        const rows = Cypress.$('[data-el="tokens-table"] tbody').find('tr.ff-data-table--row').length
-
         cy.intercept('POST', '/api/*/user/tokens').as('addPersonalAccessToken')
 
+        // retryable read - a synchronous Cypress.$ count can beat the table render on slow CI, undercounting the baseline
+        let rows
+        cy.get('[data-el="tokens-table"] tbody').find('tr.ff-data-table--row').then($rows => { rows = $rows.length })
```
cstns, line 93: counting the number of rows in a table was never a good idea to begin with. If there's a next time, we should rethink how we validate entries exist in the table

test/e2e/frontend/cypress/tests/admin.spec.js
```diff
@@ -68,7 +68,8 @@ describe('FlowFuse platform admin users', () => {
         // Not a member of BTeam
         cy.get('[data-el="teams-table"]').contains('BTeam').click()
 
-        cy.url().should('match', /\/team\/[^/]+/)
+        // wait for the team redirect to settle on /overview, else the Applications click below can race the pending redirect
+        cy.url().should('match', /\/team\/[^/]+\/overview/)
```
cstns: I think this was just a gross oversight.. the url should never have matched `team` after navigating away from the admin teams list. Running the suites locally were just fast enough to catch the old href before navigation. 

We should only be checking if the page matches `overview` after nav

cstns (commented):
I'll push a fix for the flaky admin test in order to get this in sooner

PR #7841, opened 2026-07-16
Fix MCP feature enumeration with with cap, deadline and per instance fault containement

### LINKED ISSUE #7832: Parallelise enumeration of MCP Features
### Description

Current logic in the MCP Features enumeration routine (see mcp/features endpoint in forge/ee/routes/expert/index.js) loops the candidate MCP servers to check they are online. 

This should be Parallelised  to reduce compound impact from a slow responding instance or device. There are other considerations:

### TL;DR

The MCP features endpoint checks each candidate instance/device for liveness one at a time (pass 2), so one slow or unreachable instance holds up all the others (worst case `N × timeout`). We should make these checks run concurrently but bounded. While we're in here, pass 3 should switch to `allSettled` so a single device timeout can't sink the whole response (the new `Device.status` check reduces how often that happens, but doesn't remove it). 

### Context 
- Pass 2 is sequential and does the timeout prone calls: `instance.liveState()` (hosted), `deviceComms.sendCommandAwaitReply(..., { timeout: 3000 })` (device), and `getOrCreateToken()`. Worst case latency is **N × timeout** - an unreachable instance stalls everything behind it.
- Pass 3 (the feature fetch) is **already** `Promise.all` parallel, so it already carries some of the risks below.

### What should be looked at
- [ ] Convert pass 2 from a sequential `for` loop to bounded concurrency execution (batch/pool, e.g. 10–20 in flight) instad of an unbounded Promise.all
- [ ] Keep each parallel task self-contained with its own `try..catch` (or use `allSettled`) so one slow/failing instance is skipped, never rejecting the whole batch.
- [ ] Harden the existing pass 3 Promise.all  to **`allSettled`** for the same reason (at scale, the probability of  timeout will sink the whole response).
- [ ] Add an **overall request deadline** (AbortController) - return partial results with an explicit "unchecked/skipped" list rather than waiting for the slowest chain.
- [ ] Consider a **per-request instance cap** (backstop) - query the first N instances, report the remainder as not-checked.

### Why this should be considered
- **DB connection pool exhaustion** - `liveState()` and `getOrCreateToken()` touch the DB; Sequelize's pool is relevant here. Unbounded concurrency queues on pool and can hit acquire timeout errors - thrashing, not parallelism. 
- **Broker / comms pressure** - e.g. 1000 simultaneous `sendCommandAwaitReply` = 1000 in-flight MQTT requests + pending reply-handlers - risk of overwhelming the broker and the pending-reply map.
- **Container driver / k8s API rate limits** -  `liveState()` on hosted instances hits the driver / k8s API.
- **Batch size vs tail latency** - with concurrency `C` and timeout `T`, worst case ≈ `ceil(N/C) × T`. At N=1000, C=20, T=3s that's ~150s - so concurrency alone is not enough at 1000; the overall deadline + partial results are required.
- **Timeout behaviour of `sendCommandAwaitReply` rejects** On timeout it rejects with `Error('Command timed out')` (`forge/comms/devices.js`), and rejects on other error paths too.
  - Pass 2 is safe today - the await is inside a per-iteration `try/catch { continue }`.
  - **Pass 3 has a latent failure**: the rejecting promises feed a bare `Promise.all` with no per-item catch, so a single device timeout rejects the whole batch and fails the endpoint. Switching pass 3 to `allSettled` (or per-item `.catch()`) is required, not optional.
  - **Partially mitigated by the new `Device.status` virtual column**: pass 1 now skips devices whose status is not `online` (i.e. `lastSeenAt` older than 30 min), so known-stale devices never reach pass 3 and can't trigger the timeout-driven batch rejection. This shrinks the problem but doesnt remove it altogether. A device that is `online` (seen within 30 min) but unresponsive right now will still time out. i.e. a good argument for `allSettled`.

### Epic/Story

_No response_

Steve-Mcl:
### TL;DR
The `POST /api/v1/expert/mcp/features` endpoint checked each candidate instance/device for liveness sequentially, so one slow or unreachable instance held up all the others (worst case `N x timeout`). It also fetched features via a bare `Promise.all` with no per-item error handling - and since a device MQTT call rejects on timeout, a single unresponsive device could reject the whole batch and 500 the endpoint.

This PR replaces that with a single bounded, pipelined per-instance scan: each instance flows liveness -> token -> features independently under a concurrency limit, wrapped so one failure can never sink the batch. The scan is bounded by a total server cap and an overall deadline, and the response now carries a `summary` describing coverage. All limits are configurable via `app.config.expert.insights.*`.

### Details
- Pipelined operations (instead of 2 barriers!).  wall clock is now the slowest single instance chain, not the sum of scans.
- Fault containment (bug fix): each per-instance scan is wrapped in `try..catch`, so a rejecting device call is skipped rather than failing the whole request.
- Total cap `MAX_SERVERS` (default 20); the remainder is reported as unenumerated.
  - This is a change in behaviour agreed in AI meeting today (2026-07-16) - deemed necessary to keep things snappy AND avoid timeouts / db pool exhaustion. Agreement was A) we have very few insights users today (no SH have insights) and B) it is still marked as a BETA feature and as such, our BETA testing has revealed this need. 
- Bounded concurrency `ENUM_CONCURRENCY` (default 5, aligned with the DB pool).
- Overall deadline `ENUM_DEADLINE_MS` (default 10s) returns partial results; in-flight MQTT calls run out their own timeouts (no abort for MQTT).
- Per device feature timeout reduced 10s -> 5s (`DEVICE_FEATURES_TIMEOUT`); liveness stays at 3s.
- All limits configurable via `app.config.expert.insights.{ MAX_SERVERS, ENUM_CONCURRENCY, ENUM_DEADLINE_MS, DEVICE_LIVENESS_TIMEOUT, DEVICE_FEATURES_TIMEOUT }`, each falling back to the defaults above.
   - NOTE: These are not advertised in docs for now. Only used as emergency internal tweaks (and for test overrides)
- New `summary` in the response (and schema): `{ instanceCount, mcpServerCount, enumeratedCount, unenumeratedCount, limit, limited, deadlineReached }`.
   - Following discussion in AI meeting, it was stated a summary would be included so that we can debug and at some point inform users of limits. 

Builds on the recently merged `Device.status` gate, which already skips offline devices before they reach the scan.

### Tests
`test/unit/forge/ee/routes/expert/index_spec.js`:
- The existing endpoint tests passed against the refactor without modification - the behaviour is unchanged for the common cases. The only edit was extra assertions in the "return early with no registrations" test, to cover the always-present `summary`.
- Added a `buildDeviceRegistrations(count)` helper producing N distinct online device registrations for fan-out/cap coverage.
- Cap: 25 registrations against the cap of 20 - asserts only 20 are enumerated, the `summary` reflects the cap (`enumeratedCount 20`, `unenumeratedCount 5`, `limited true`), and instances beyond the cap are never contacted.
- Deadline + fault containment: four devices (rejects at liveness, rejects at features, exceeds a shortened deadline, responds in time). Asserts the request still returns 200, only the healthy device is enumerated, `deadlineReached: true`, and failed instances are skipped rather than reported incompatible.

## Related Issue(s)

closes #7832

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/ee/routes/expert/index.js
+++ b/forge/ee/routes/expert/index.js
@@ -15,6 +15,29 @@ const { filterAccessibleMCPServerFeatures } = require('../../../services/expert.
 const getDeviceComms = (app) => { return app.comms?.devices }
 
 /**
+ * Run `fn` over `items` with at most `limit` concurrent executions.
+ * `fn` must not throw (each task should wrap its own errors) so a single failure cannot sink the
+ * pool - this gives Promise.allSettled-like semantics. `isStopped()` is polled before pulling each
+ * next item: when it returns true the workers stop taking new work (used to enforce the overall
+ * deadline). It does NOT cancel in-flight work - there is no abort signal for MQTT calls.
+ * @param {any[]} items
+ * @param {number} limit
+ * @param {(item:any, index:number) => Promise<void>} fn
+ * @param {() => boolean} [isStopped]
+ */
+async function mapWithConcurrency (items, limit, fn, isStopped) {
+    let cursor = 0
+    const worker = async () => {
+        while (cursor < items.length) {
+            if (isStopped?.()) { return }
+            const index = cursor++
+            await fn(items[index], index)
+        }
+    }
+    await Promise.all(Array.from({ length: Math.min(limit, items.length) }, worker))
+}
+
+/**
  * Maps a platform automation tool's wire definition into a catalog entry for the
  * Expert permissions UI. The read/write/delete class comes from the MCP annotations
  * (readOnlyHint / destructiveHint), and `group: 'platform'` routes it to the platform
@@ -340,6 +363,18 @@ module.exports = async function (app) {
                                 },
                                 required: ['instance', 'instanceType']
                             }
+                        },
+                        summary: {
+                            type: 'object',
+                            properties: {
+                                instanceCount: { type: 'integer' },
+                                mcpServerCount: { type: 'integer' },
+                                enumeratedCount: { type: 'integer' },
+                                unenumeratedCount: { type: 'integer' },
+                                limit: { type: 'integer' },
+                                limited: { type: 'boolean' },
+                                deadlineReached: { type: 'boolean' }
+                            }
                         }
                     }
                 },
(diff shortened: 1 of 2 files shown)


cstns (approved):
Changes look good to me. One thing I'm unclear about is that since you changed the signature of the /features endpoint, the type drift check didn't fail..

Not going to block it for that thogh

Issue #7608, opened 2026-06-26
Split live-status tracking into its own subscriber (and share one MQTT connection)


n-lark:
Follow-up from [PR review](https://github.com/FlowFuse/flowfuse/pull/7573#discussion_r3473882776). Live instance/device state tracking currently rides inside
`TeamChannelSubscriber` alongside membership and team-update events. By the subscriber design, each
concern should be its own minimal subscriber.

## Goal

- Add a dedicated `LiveStatusSubscriber` for the `p/+/state` and `d/+/state` topics.
- Keep `TeamChannelSubscriber` focused on membership and team updates.
- Hoist the shared connect/disconnect/message-routing boilerplate into `BaseSubscriber`, with abstract
  members for the topic list and the regex→handler map. New subscribers then just declare what they care
  about.

## Blocker to solve first

The two subscribers share the fe-team connection key, but `mqttService.createClient(key)` destroys any
existing client on that key — there's no "is anyone else using this key?" check:

```ts
// createClient(key)
if (this.hasClient(key)) {
  await this._destroyClientUnlocked(key)   // existing client torn down, not reused
}
// destroyClient(key) → _destroyClientUnlocked(key) — no refcount guard
```

So two subscribers on the same key tear each other down. Sharing the connection needs a
`connectOrReuse`-style entry point (name TBD) that reuses an existing client and tracks subscribers, so
teardown only happens once the last one detaches. Opening a separate client per subscriber is the
fallback but wastes a connection.

n-lark:
# #7608 — Manual test plan: split live-status subscriber + shared MQTT connection

<details><summary> ✅ All tests passed! </summary>

## Verified

### Shared connection & lifecycle

- [x] Entering a team opens one shared MQTT connection per open tab (creds key on a
      per-tab session id) — verified in EMQX.
- [x] Switching teams keeps the connection count stable — no growth, no orphaned
      connections left behind for the previous team.
- [x] Logout tears down the team connection(s) and clears their topics from EMQX.

### Live status

- [x] Starting / restarting a **hosted** instance updates status live (no refresh).
- [x] Starting / restarting a **remote** instance updates status live.
- [x] No stale status across a team switch — second team shows only its own instance/
      device statuses (store cleared on disconnect).

### Team-channel events

- [x] Renaming the team triggers a live update.
- [x] Revoking a user's team membership triggers a live update.

### Regression — standalone MQTT consumers (share `mqtt.service.createClient`)

[...]

PR #7843, opened 2026-07-16
Split live-status tracking into its own subscriber (and share one MQTT connection)  - Part 1: Add `attachClient`/`detachClient`

n-lark:
## Description

Adds `attachClient`/`detachClient` to the MQTT service so multiple subscribers can share one connection per key. Reused while anyone's attached, torn down only when the last detaches. `createClient`/`destroyClient` are unchanged and nothing uses attach/detach yet.

## Related Issue(s)

Relates to https://github.com/FlowFuse/flowfuse/issues/7608

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/services/mqtt.service.ts
+++ b/frontend/src/services/mqtt.service.ts
@@ -14,7 +14,10 @@ import { BaseService } from './service.contract'
 import { Maybe } from '@/types/common/types'
 import type {
     BinaryPayload,
+    ManagedMqttAttachment,
     ManagedMqttClient,
+    MqttAttachmentHandle,
+    MqttConnectionHandlers,
     MqttConnectionOptions,
     MqttConnectionWaiter,
     MqttCredentialProvider,
@@ -48,6 +51,8 @@ class MqttService extends BaseService implements MqttServiceI {
 
     protected $destroyed: boolean
 
+    protected $attachmentSeq: number
+
     constructor ({ app, router, services }: CreateServiceOptions) {
         super({
             name: 'mqtt',
@@ -60,6 +65,7 @@ class MqttService extends BaseService implements MqttServiceI {
         this.$clients = new Map()
         this.$clientOperations = new Map()
         this.$destroyed = false
+        this.$attachmentSeq = 0
 
         this.init()
     }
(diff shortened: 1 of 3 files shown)


frontend/src/types/services/mqtt.types.ts
```diff
@@ -87,6 +87,16 @@ export interface MqttConnectionWaiter {
     timer: ReturnType<typeof setTimeout> | null
 }
 
+export interface ManagedMqttAttachment {
```
cstns: I'd go with Observers as a name for this concept. My first thought was subscribers but that muddies the waters (with subscriptions). Observers is distinct from the MQTT terminology, it's a well defined pattern and it's descriptive of what's happening (multiple observers watching events on a shared connection)

frontend/src/services/mqtt.service.ts
```diff
@@ -99,75 +105,135 @@ class MqttService extends BaseService implements MqttServiceI {
...
 
+        return this.runClientOperation(key, async () => {
             if (this.hasClient(key)) {
                 await this._destroyClientUnlocked(key)
             }
+            return this._createAndConnect(key, options, attachment)
+        })
+    }
+
+    async attachClient (key: string, options: Partial<MqttConnectionOptions> = {}): Promise<MqttAttachmentHandle> {
```
cstns: I'd make this a bit more descriptive method call. In it's current format it would become `attachClientAttachment` but with the observer proposal it would become `attachClientObserver` (it also reads better IMO)

cstns: I know it's a frivolous ask, but now that i'm reading this bit of code after ~2 months of writing it the first argument in these methods is vague and I have to keep reminding myself what it is.. it should have been `clientKey` in the first place.. the cognitive load is too much to bare

cstns (changes requested):
Looking great! Please rename the `attached` to something more explicit

frontend/src/types/services/mqtt.types.ts
n-lark: Updated!

frontend/src/services/mqtt.service.ts
n-lark: Updated name and changed key -> clientKey for entire file.

cstns (approved):

PR #7844, opened 2026-07-16
Split live-status tracking into its own subscriber (and share one MQTT connection) - Part 2: Add `LiveStatusSubscriber`

### LINKED ISSUE #7608: Split live-status tracking into its own subscriber (and share one MQTT connection)

n-lark:
## Description

Gives live status its own `LiveStatusSubscriber` instead of bundling it into the `team-channel` subscriber, and has the two share one team MQTT connection (using the `attach/detach` from PR 1) rather than each needing its own.

Cut from https://github.com/FlowFuse/flowfuse/pull/7843

See test plan: https://github.com/FlowFuse/flowfuse/issues/7608#issuecomment-4994776222

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7608

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/subscribers/team-subscriber.contract.ts
```diff
@@ -0,0 +1,173 @@
+import type { App } from 'vue'
```
n-lark, line 1: So I renamed this from the `BaseSubscriber` we had since I needed to hoist some shared team logic into it. As we expand, I figured it was better to make a team specific contract then if/when we get more channels outside `fe-team`, we can make a new one for that instead of team specific logic in the `BaseSubscriber`.

cstns: good call

cstns (commented):
Looks good to me. The only thing i'd like to touch base on is the naming issue I mentioned on the pr this is based off of. 

Other than that, happy to merge

cstns (approved):

PR #7847, opened 2026-07-17
Refactor common code

knolleary:
Extracts the common code between the LDAP and SAML paths as the logic is fiddly and we don't want two copies to maintain.

### DIFF
--- a/forge/ee/lib/sso/index.js
+++ b/forge/ee/lib/sso/index.js
@@ -318,6 +318,116 @@ module.exports.init = async function (app) {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


knolleary:
I have tested this locally with SAML/Azure for various combinations to ensure it only modifies groups its meant to.
Pulling the logic into one place will avoid bugs only being fixed in one path (ldap vs saml). I think it also helps to reason on the logic.

hardillb:
Testing with LDAP looks good, just trying a couple other situations

PR #7848, opened 2026-07-17
[7827] Improve UI for device status

### LINKED ISSUE #7827: Improve UI for device status
### Description

TLDR - showing a "Last Seen" of "1 Year" and "Last Known Status" of "Running" is not good. Ideally, this would be a single "Status" column. 

Once https://github.com/FlowFuse/flowfuse/issues/7826 in place, we can utilise that property to better inform users on the state of their devices.

There are 2 competing options/proposals laid out in the [parent issue](https://github.com/FlowFuse/flowfuse/issues/7790#issuecomment-4979231012) - once consensus is achieved, we should _make it so_

### Epic/Story

_No response_

n-lark:
## Description

Routes the new computed device status to `onlineStatus` (online/offline/never-seen). Collapses the device tables' "Last Seen" + "Last Known Status" columns into a single "Status" pill with last-seen and last-reported state in a tooltip.

Screenshot 2026-07-17 at 1 12 56 PM

### Live update changes 

Live status updates piggyback on the team-channel state message, which now carries `onlineStatus`. 

We only ever push `online` from `handleStatus`, which runs when a device checks in so we send it on any reconnect that flips it back. We can't push `offline` because going `offline` is silence, so `offline` is computed from `lastSeenAt` on page load instead.

### Items to note 

- I left the top "Last Seen" + "Last Known Status" grid things alone, lmk if we want any changes for those. 
- The device detail page now feels a little inconsistent with the list view. I didn't make any changes but I can make a follow up if we'd like. Lmk! 

Screenshot 2026-07-17 at 12 58 47 PM

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7827

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
Love it. It is exactly how I hoped (personal opinion) 
- Single "Status" header text is much nicer than multi word
- Tooltip reveals all the info I would want/expect


> * I left the top "Last Seen" + "Last Known Status" grid things alone, lmk if we want any changes for those.

I think that is fine for now. Moving forward, for me, the separate pills/badges (where there is more space) is more useful for "at a glance" info.  One thing I would probably want tho is the colour of the device overview pill to reflect online/offline status.



> * The device detail page now feels a little inconsistent with the list view. I didn't make any changes but I can make a follow up if we'd like. Lmk!

Similar to last point, where there is more space, it feels (to me) better to make information more glanceable. I think the task here is to align with whatever we do to the header. 

---

Perhaps a single follow up issue to update the top "Last Seen" + "Last Known Status" and device details to utilise the online/offline status for colour and appearance would be in order?

n-lark:
Hey @Steve-Mcl thank you for your thoughts! I went ahead and created Make device detail page and list summary bars consistent with new online status design https://github.com/FlowFuse/flowfuse/issues/7876 as a follow up around these points.

forge/db/views/DeviceGroup.js
```diff
@@ -126,7 +126,7 @@ module.exports = function (app) {
                 description: item.description,
                 application: item.Application ? app.db.views.Application.applicationSummary(item.Application) : null,
                 deviceCount: parseInt(item.deviceCount) || 0,
-                devices: item.Devices ? item.Devices.map(app.db.views.Device.device) : [],
+                devices: group.Devices ? group.Devices.map(app.db.views.Device.device) : [],
```
cstns, line 129: why?

frontend/src/pages/device/components/DeviceOnlineStatusCell.vue
```diff
@@ -0,0 +1,49 @@
...
+        <MinusCircleIcon v-else class="w-4 h-4" />
+        <span class="ml-1 whitespace-nowrap">{{ label }}</span>
+    </div>
+</template>
+
+<script setup lang="ts">
+import { MinusCircleIcon, SignalIcon, SignalSlashIcon } from '@heroicons/vue/24/outline'
+import { computed } from 'vue'
+
+import { capitalize } from '../../../composables/strings/String.js'
```
cstns: why not use @/ for these imports?

cstns (changes requested):

frontend/src/pages/device/components/DeviceOnlineStatusCell.vue
```diff
@@ -0,0 +1,49 @@
+<template>
+    <div
+        v-ff-tooltip:left="tooltip"
+        class="forge-badge"
```
cstns: we should use a different kind of pointer when hovering

cstns:
created https://github.com/FlowFuse/flowfuse/issues/7879 as a follow up

cstns:
Also, device sorting by status seems to be way off

forge/db/views/DeviceGroup.js
n-lark: `item` is `group.toJSON()`, so `item.Devices` are plain objects. `device.status` (a Sequelize virtual) is `undefined` on those, which drops `onlineStatus` and fails the required schema check. `group.Devices` are the real instances, so it resolves.

frontend/src/pages/device/components/DeviceOnlineStatusCell.vue
n-lark: Added `cursor-help`.

n-lark:
> Also, device sorting by status seems to be way 

Discussed in slack, since `onlineStatus` is a computed column we cannot sort by it in the db - get column does not exist errors. Leaving as `lastSeenAt` since that behavior is mostly correct. Will revisit if it becomes an issue.

cstns (approved):

PR #7850, opened 2026-07-20
Fix pipeline stage validations

knolleary:
Part of https://github.com/FlowFuse/security/issues/122
Fixes validation of PipelineStage properties to ensure proper pipeline/team ownership of all resources.

### DIFF
--- a/forge/ee/db/controllers/Pipeline.js
+++ b/forge/ee/db/controllers/Pipeline.js
@@ -275,7 +275,8 @@ module.exports = {
             delete options.source
         }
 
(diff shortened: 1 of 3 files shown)


knolleary:
The issue around resource validation only exists on the Create Stage path as the Update Stage path uses a transaction across all of its work to ensure proper rollback if an error occurs.

I *didn't* take that approach for the Create Stage path; the `add*Id` functions really needed to be validating what they are asked to do - rather than relying on validation happening later on.

hardillb (approved):

Steve-Mcl (approved):

Issue #6756, opened 2026-02-23
Automatic NR Node updating


hardillb:
### Description

Currently Node-RED nodes need to be manually upgraded by 

- going into the editor and using palette manager
- manually editing the Settings -> Palette -> modules

Then restarting the instance.

This epic is to investigate how to automate this as part of the scheduled maintenance 

### Which customers would this be available to

Everyone - CE/Starter/Team/Enterprise

allthedoll:
@knolleary this continues to get bumped. Thoughts on doing or not doing it?

knolleary:
I think we can push to the backlog. Implementing automatic updating begs the question of having a way for a user to pin a version because they know the latest version breaks something for them etc.

I think there is a more sensible iteration to help the user know if there are updates available and allow them to take action - the SBOM view was a start, but operates Team-wide. We don't provide the same functionality when looking at an individual instance. Where that should sit in the overall prioritisation of life, the universe, and everything, I'm not sure.

hardillb:
Taking milestones and spints off this item

Esmewieringa:
[Customer link](https://app-eu1.hubspot.com/contacts/26586079/record/0-2/60136232184)

ignitedPotato:
@allthedoll 

We are currently using the "Scheduled Restarts/Upgrades" feature for our test-instances to be able to test new Node-RED versions before upgrading our production instances by hand. This provides us a hassle-free upgrade path while we can also make sure everything works as expected.

We'd like the same for node updates. But @knolleary is right, we then would need a way to pin versions, same as with the Stacks for instances. In my understanding this is already possible by editing the module version in the instance palette settings and removing the ~. If something breaks on a test instance we could pin a module version which would then be propagated through the pipeline.

Alternatively you could provide update buttons for every instance in the SBOM view  - which is already a great help for identifying old versions. But this would still require manual intervention even for all our test instances, which is exactly what we don't want.

knolleary:
Refinement needed - plan to refinement in wk1 of Sprint 1/2.33.

hardillb:
- "daily" task to pull catalogues into a local cache with latest versions
- extract dependences table from settings object into it's own table to make searching/comparing displaying quicker/eastier
    - should also have latest version column (updated when task 1 runs)

First iteration is gathering the data on what's available, second task improve BOM view using this data,  third task to examine how to apply those.

Need to turn ^ into tasks

Esmewieringa:
[...]

PR #7880, opened 2026-07-21
Installed node settings hook

### LINKED ISSUE #7736: Add hook to instance settings update to extract installed nodes and versions
### Description

This should populate a new table with
- instance type
- instance id
- package name
- current version
- latest version from #7735 

### Epic/Story

#6756

hardillb:
fixes #7736
## Description

Uses DB hooks to keep latest version of Node-RED nodes list up to date

This PR merges into https://github.com/FlowFuse/flowfuse/pull/7772

## Related Issue(s)

#7736

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/db/models/Project.js
+++ b/forge/db/models/Project.js
@@ -260,6 +260,14 @@ module.exports = {
                         app.log.error(`Error removing MCPRegistrations for deleted instance ${project.id}: ${err.message}`)
                     }
                 }
+                if (app.db.models.NodeREDNodeVersions?.destory) {
+                    await app.db.models.destory({
+                        where: {
+                            ownerType: 'instance',
+                            ownerId: project.id
+                        }
+                    })
+                }
             }
         }
     },
--- a/forge/db/models/Project.js
+++ b/forge/db/models/Project.js
@@ -260,14 +260,13 @@ module.exports = {
                         app.log.error(`Error removing MCPRegistrations for deleted instance ${project.id}: ${err.message}`)
                     }
                 }
-                if (app.db.models.NodeREDNodeVersions?.destory) {
-                    await app.db.models.destory({
-                        where: {
-                            ownerType: 'instance',
-                            ownerId: project.id
-                        }
-                    })
-                }
+                // Remove version info when Project removed
+                await app.db.models.NodeREDNodeVersions.destory({
+                    where: {
+                        ownerType: 'instance',
+                        ownerId: project.id
+                    }
+                })
             }
         }
     },
--- a/forge/db/migrations/20260709-01-add-node-red--node-version.js
+++ b/forge/db/migrations/20260709-01-add-node-red--node-version.js
@@ -1,0 +1,37 @@
+const { DataTypes } = require('sequelize')
+
+module.exports = {
+    /**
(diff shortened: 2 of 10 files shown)


Steve-Mcl:
I see I was tagged as reviewer for this but Is it ready for review Ben? Just from a very quick glance I see things like swallowed errors (`// console.log(err)`), no support for devices (even though associations to device is setup), no tests, and the target branch PR is still in draft/has no reviewer. 🤷  

note also, I believe you might need to drop the association for device and switch to application logic as we have done for things like access_token  (casting from int to string for the device.id/ownerId will likely fail for PG) - but in saying that I see existing tests did not fail

hardillb:
> I see I was tagged as reviewer for this but Is it ready for review Ben? Just from a very quick glance I see things like swallowed errors (// console.log(err)), no support for devices (even though associations to device is setup), no tests, and the target branch PR is still in draft/has no reviewer. 🤷

Yes all intentional, especially the devices as we have no way to gather that information as it is not shared. The current table is built that way to make it work if/when it ever is available.

The errors get dropped because this is a async hook fired in the background and the only possible errors are back end db so failing to update is a small problem

forge/db/models/Project.js
```diff
@@ -260,6 +260,14 @@ module.exports = {
                         app.log.error(`Error removing MCPRegistrations for deleted instance ${project.id}: ${err.message}`)
                     }
                 }
+                if (app.db.models.NodeREDNodeVersions?.destory) {
+                    await app.db.models.destory({
```
knolleary: I doubt the guard is necessary as (unlike MCP) this model is in the core set and will always be available. But we do need to fix the reference...
```suggestion
                if (app.db.models.NodeREDNodeVersions?.destory) {
                    await app.db.models.NodeREDNodeVersions.destory({
```

knolleary (changes requested):

hardillb:
@Steve-Mcl The AccessTokens still has `belongsTo` for Device in the `associations`. But we are limited to just Projects for this pass, just need to make sure we don't do anything that will break later if/when we can add Device installed nodes.

hardillb:
Something strange is going on with package flies and lint, I'm going to merge this down to the other PR so I only need to fix it once and it can all be reviewed as one

Steve-Mcl (commented):
I have a some thoughts on the design...

1. Why we are not using the existing controller methods `getProjectModule`? - but then I found your proposal in a grandparent issue [#6756 comment](https://github.com/FlowFuse/flowfuse/issues/6756#issuecomment-4891499832)
   > extract dependences table from settings object into it's own table to make searching/comparing displaying quicker/eastier

   Fair enough - a flat/normalised table could be useful to improve performance of bom operations
2. As there is only a single table, every row has a `latestVersion` meaning potentially hundreds/thousands of duplicates. I would prefer a separate table e.g. `NodeREDPackageVersions` where each row simply represents a single package - then a simple join with this table `NodeREDNodeVersions` would remove the need for a static `updateAllLatest` method.

However, this is not blocking and could be later retrod so I wont block.

PR #7887, opened 2026-07-21
Expert bridge self healing via end-2-end heartbeat monitoring

### LINKED ISSUE #7818: Add heartbeat MQTT flow between FF Broker and Expert Broker
### Description

In order to detect any bridging issues between the forge broker and expert broker, we need to introduce an application-level MQTT heartbeat flow.

The forge app should publish to a suitable topic every minute (timing TBD) and validate a response is received.

If no response is received - perhaps after N attempts - breakdown the bridge and recreate.

This will require something on the Expert broker side to bounce back the heartbeats to a suitable topic. This needs to be aligned to the Expert request/response topics.

### Epic/Story

_No response_

Steve-Mcl:
# Do not merge before https://github.com/FlowFuse/CloudProject/issues/1519

### TLDR

Adds an end-to-end heartbeat check for the Expert Agent bridge. On a schedule, the platform publishes a heartbeat through the MQTT bridge to the Expert Broker and back; if the agent misses enough heartbeats in a row, the platform automatically tears down and re-creates the bridge. New ACL rules gate the (deliberately reversed) heartbeat topics so only the platform or the agent can use them, in the correct direction.

### Key Points

- **New `ExpertCommsHandler`** (`forge/comms/expert.js`) owns the heartbeat request/response lifecycle: publishes a heartbeat with a fresh `correlationData`, tracks the outcome in a small _**(LRU/in memory or Redis for installs with replicas depending on the installation)**_ cache (`expert-bridge-heartbeat-cache`, configurable via `app.config.expert.centralBroker.heartbeat.{bridgeCacheMax,bridgeCacheTTL}`), and reports `{ errorCount }` back to the caller via callback once the response arrives or `maxResponseTime` elapses. Wired into `forge/comms/index.js` as `app.comms.expert`.
- **New housekeeper task** (`forge/ee/lib/expert/tasks/heartbeat.js`), registered in `forge/ee/lib/expert/index.js` only when the Expert Bridge is enabled (`app.config.expert.centralBroker.server` set). Validates its own cron schedule at registration time (must be recurring, and must not run more often than `2 * maxResponseTime` apart, to prevent overlapping runs) and fails safe (logs and skips registration) if the schedule is invalid. On repeated failure (`errorCount` hits a multiple of `maxSuccessiveFailureCount`, default 3), it force re-syncs the bridge via `syncBridge(app, { force: true })`.
- **`forge/comms/commsClient.js`**: recognizes the agent's echoed heartbeat message on the bridge and re-emits it as `response/platform/expert/bridge/heartbeat`, which `ExpertCommsHandler` listens for.
- **`forge/comms/aclManager.js`**: adds four hardcoded, exact-topic ACL rules for the heartbeat handshake. The topic naming is intentionally reversed — the platform's heartbeat "request" is published on the topic literally named `.../response`, and vice versa — because the heartbeat reuses the existing platform command channel rather than adding new topics. Each direction is locked to the correct client type (`forge_platform` vs `expert-agent`) and pub/sub direction.

### Tests added to `test/unit/forge/comms/expert_spec.js`
New file covering `ExpertCommsHandler`: cache defaults and config overrides (`centralBroker.bridgeCache.max`/`.ttl`), the response-listener storing a payload in cache, the outgoing heartbeat request shape (topic/qos/correlationData), success/failure/reset of `errorCount` across repeated calls, ignoring a response with a mismatched `correlationData`, clearing a still-pending timer when called again, tolerating no callback, and forwarding a cache-read error to the callback.

### Tests added to `test/unit/forge/comms/authRoutesV2_spec.js`
Added to the existing "Expert Agent" ACL block, plus a new "Platform (forge_platform)" block: allow/deny pairs for all four heartbeat topics, confirming each client type can only use its topic in the correct direction (e.g. the agent can subscribe to the heartbeat response topic but not publish to it, and vice versa for the platform).

### Tests added to `test/unit/forge/ee/lib/expert/tasks/heartbeat_spec.js`
New file covering the task factory in isolation (fake `app`/stubbed `syncBridge`, no full app boot): schedule/startDelay/maxResponseTime/maxSuccessiveFailureCount validation (including the `startDelay: 0` fix and the "too frequent schedule" guard), the `startDelay` gate in `run()`, and the re-sync callback logic (no re-sync on success or below the failure threshold, re-sync at each multiple of `maxSuccessiveFailureCount`, and that a `syncBridge` failure is logged rather than thrown).

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/7818

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
@Steve-Mcl I'm unlikely to get to this before Monday

Steve-Mcl:
Thanks for letting me know @hardillb - will reassign review to @cstns

Steve-Mcl:
@cstns added as reviewer in place of Ben since you have knowledge of the pathways involved.

What you may not be aware of is the outage we had on Expert a couple of Saturdays ago.

Long story short, the bridge became uni-directional. Stopping and starting it recovered the 2 way communications. 

The owner issue https://github.com/FlowFuse/flowfuse/issues/7818 was raised to add heartbeat and self healing (have a read and decide if the Pr fulfils that)


NOTE:
The pathway chosen (the command/inflight path way) was a deliberate 1st iteration choice.
I toyed with making this 100% coverage (i.e. sen HB to /chat/, route it back over /platform/request to the platform, bounc it back to Expert on platform/response then bounce it back to /chat/response - however it was quite messy and considerations about source origin, app replicas, and adding extra fiorge app topics to ACLs meant it was a LOT bigger than it already is.  

In short, I chose the least path of resistance (the command/inflight channels) and used them in reverse as this mean no new acls and at least covered 2 directions - we can iterate if we find the bridge breaks down for a single channel.

forge/comms/expert.js
```diff
@@ -0,0 +1,97 @@
...
+        const payload = { timestamp: Date.now(), platformId: this.client.platformId, correlationData }
+        this.bridgeHeartbeatCache.get('errorCount') // a get hit keeps the errorCount key alive in the cache (no need to await it)
+        clearTimeout(this.timer) // clear any existing timer before setting a new one
+
+        const topic = 'ff/v1/expert/forge_platform/bridge/platform/heartbeat/response' // /response is the correct topic
+        const mqttOptions = { qos: 1, retain: false, properties: { correlationData } }
+        const payloadData = JSON.stringify(payload)
+
+        // Send the heartbeat request and set a timer to check for a response after the maxResponseTime has elapsed
+        this.client.publish(topic, payloadData, mqttOptions)
```
cstns, line 59: The heartbeat request is published with only correlationData (no userProperties) & the heartbeat handler at line ~159 in commsClient is placed after this early return which ignores messages without userProperties.

forge/ee/lib/expert/tasks/heartbeat.js
```diff
@@ -0,0 +1,81 @@
...
+
+            if (Date.now() < skipUntil) {
+                return // skip this run, as we are still in the startup delay period
+            }
+
+            // small random delay to further minimise contention on the bridge
+            await sleep(randomInt(0, 999))
+
+            // Request a heartbeat from the Expert Agent via the bridge
+            expertCommsHandler.requestBridgeHeartbeat(maxResponseTime, async (err, result) => {
```
cstns, line 69: Consider using an object parameter here, e.g. `async ({ err, result } = {}) => { ... }`. That would make the callback a bit more flexible, since callers that only need the result wouldn’t have to pass `null` as the first argument just to skip the error

forge/comms/expert.js
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Handle bridge heartbeat response
+     * @param {Object} payload The received message
+     * @param {import('mqtt').IClientPublishOptions['properties']} properties The message properties
+     */
+    bridgeHeartbeatResponseHandler (payload, properties) {
+        // This event is triggered (emitted) in the commsClient when a valid heartbeat response is received from the Expert Broker via.
+        // the bridge (Expert Broker → FF App Instance Broker) in response to a heartbeat request made in `this.requestBridgeHeartbeat()`.
+        // Here we simply update the cache with the response. The cache is required since there may be replica instances of the FF App
+        // and the expert platform topic is a shared topic, so the response may be received by a different instance than the one that initiated the request.
```
cstns, line 90: Wouldn't it make more sense to keep the heartbeat cache in a shared chase then? this would mitigate cross talk from the initiator replica and responder replica

forge/comms/index.js
```diff
@@ -34,6 +35,7 @@ module.exports = fp(async function (app, _opts) {
         const deviceCommsHandler = DeviceCommsHandler(app, client)
         const instanceCommsHandler = InstanceCommsHandler(app, client)
         const platformAutomationHandler = PlatformAutomationHandler(app, client)
+        const expertCommsHandler = new ExpertCommsHandler(app, client)
```
cstns, line 38: we should we conditionally create/instantiate the expert commsHandler only if there's a bridge is enabled

forge/comms/expert.js
Steve-Mcl: in the heartbeat reply (generated by the other side, userProperties is added).

forge/comms/expert.js
Steve-Mcl: using app.caches does resolve to an in memory or redis cache. On FFC, we use redis (so it is not a local cache)

forge/comms/index.js
Steve-Mcl: fair point. will consider.

cstns (approved):

forge/ee/lib/expert/tasks/heartbeat.js
Steve-Mcl: can be easily retroed - lets leave for now

PR #7892, opened 2026-07-21
123 f6 cross team mcp registration

hardillb:
fixes FlowFuse/security#125

see FlowFuse/security#123 for details

### DIFF
--- a/forge/ee/routes/mcp/registrations.js
+++ b/forge/ee/routes/mcp/registrations.js
@@ -116,7 +116,7 @@ module.exports = async function (app) {
                 if (!device) {
                     throw new Error(`Device '${request.params.typeId}' not found`)
                 }
-                typeId = device.id
+                typeId = '' + device.id
                 if (device.Team.id !== request.team.id) {
                     reply.code(403).send({ code: 'unauthorized', error: 'Unauthorized' })
                     return
@@ -202,7 +202,7 @@ module.exports = async function (app) {
                 reply.code(404).send({ code: 'not_found', error: 'Device not found' })
                 return
             }
-            typeId = device.id
+            typeId = '' + device.id
             if (device.Team.id !== request.team.id) {
                 reply.code(403).send({ code: 'unauthorized', error: 'Unauthorized' })
                 return
--- a/forge/ee/routes/mcp/registrations.js
+++ b/forge/ee/routes/mcp/registrations.js
@@ -134,7 +134,7 @@ module.exports = async function (app) {
                 throw new Error(`Unknown MCP target type '${request.params.type}'`)
             }
 
-            if (request.params.typeId !== request.session.ownerId) {
+            if (String(typeId) !== String(request.session.ownerId)) {
                 reply.code(403).send({ code: 'unauthorized', error: 'Unauthorized' })
                 return
             }
@@ -195,20 +195,22 @@ module.exports = async function (app) {
             }
         }
     }, async (request, reply) => {
+        let typeId = request.params.typeId
         if (request.params.type === 'device') {
             const device = await app.db.models.Device.byId(request.params.typeId)
-            if (device.Team.id !== request.team.id) {
-                reply.code(403).send({ code: 'unauthorized', error: 'Unauthorized' })
+            if (!device) {
+                reply.code(404).send({ code: 'not_found', error: 'Device not found' })
                 return
             }
+            typeId = device.id
             if (device.Team.id !== request.team.id) {
                 reply.code(403).send({ code: 'unauthorized', error: 'Unauthorized' })
                 return
             }
         } else if (request.params.type === 'instance') {
             const project = await app.db.models.Project.byId(request.params.typeId)
-            if (project.Team.id !== request.team.id) {
-                reply.code(403).send({ code: 'unauthorized', error: 'Unauthorized' })
+            if (!project) {
+                reply.code(404).send({ code: 'not_found', error: 'Instance not found' })
                 return
             }
             if (project.Team.id !== request.team.id) {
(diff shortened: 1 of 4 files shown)


forge/ee/routes/mcp/registrations.js
```diff
@@ -117,15 +117,28 @@ module.exports = async function (app) {
...
                 }
+                if (project.Team.id !== request.team.id) {
+                    reply.code(403).send({ code: 'unauthorized', error: 'Unauthorized' })
+                    return
+                }
             } else {
                 throw new Error(`Unknown MCP target type '${request.params.type}'`)
             }
 
+            if (request.params.typeId !== request.session.ownerId) {
```
andypalmi: By this point `typeId` is already the resolved id (`typeId = device.id` for devices a few lines up, and the UUID param for instances), so the owner check can compare against that with both sides stringified. As written it uses `request.params.typeId`, which for a device is the hashid while `ownerId` is the numeric id, so a device registering itself never matches and 403s (instances are unaffected since both sides are the UUID). Suggestion:

```suggestion
            if (String(typeId) !== String(request.session.ownerId)) {
```

test/unit/forge/ee/routes/mcp/index_spec.js
```diff
@@ -233,4 +233,80 @@ describe('MCP Server Registration', function () {
...
+        const cToken = (await cInstance.refreshAuthTokens()).token
+
+        const deleteResponse = await app.inject({
+            method: 'DELETE',
+            url: `/api/v1/teams/${app.team.hashid}/mcp/instance/${app.instance.id}/abcde`,
+            headers: {
+                Authorization: `Bearer ${cToken}`
+            }
+        })
+        deleteResponse.statusCode.should.equal(403)
```
andypalmi, line 295: Both cross-team tests here 403 via the `typeId !== ownerId` check, so the `Team.id !== request.team.id` branch is not actually exercised. A case that hits the team mismatch directly (own token, own asset id, a different `teamId` in the URL) would cover it, and a device happy-path test would surface the owner-id comparison noted in the handler.

forge/ee/routes/mcp/registrations.js
```diff
@@ -182,6 +195,33 @@ module.exports = async function (app) {
...
+                reply.code(403).send({ code: 'unauthorized', error: 'Unauthorized' })
+                return
+            }
+        } else {
+            throw new Error(`Unknown MCP target type '${request.params.type}'`)
+        }
+        if (request.params.typeId !== request.session.ownerId) {
+            reply.code(403).send({ code: 'unauthorized', error: 'Unauthorized' })
+            return
+        }
```
andypalmi: Two device-specific things in the DELETE path, kept separate from the suggestion above since they overlap that block:

1. Unlike POST, this handler never resolves `typeId` to the numeric id (`typeId = device.id`), so there is no resolved id for this owner check to compare against. It would need that resolution step added before the same fix as POST could apply.
2. `byTypeAndIDs` below is called with `request.params.typeId` (the device hashid), but POST stores `targetId = device.id` (the numeric id). So for a device the delete looks the row up by hashid and never matches what POST wrote. Worth resolving the device id and using it for the lookup too.

andypalmi:
~~Two things that are not line-specific:~~

- ~~The DELETE lookup (`byTypeAndIDs`) is not scoped by team, so it still matches rows across teams. With the new owner and team checks in front it should not be reachable cross-team, but scoping the lookup by `request.team.id` would line up with the original finding.~~
- ~~The register and delete paths do not emit an audit event, so MCP registration changes are silent. That was called out as a secondary point in the finding.~~

All handled in commit [4558d25](https://github.com/FlowFuse/flowfuse/pull/7892/commits/4558d25062bd05d5eb9eee6f0b80930501165861)

forge/ee/routes/mcp/registrations.js
andypalmi: Outdated by commit I made.

test/unit/forge/ee/routes/mcp/index_spec.js
andypalmi: Handled in [4558d25](https://github.com/FlowFuse/flowfuse/pull/7892/commits/4558d25062bd05d5eb9eee6f0b80930501165861)

forge/ee/routes/mcp/registrations.js
andypalmi: Both points are now fixed in commit [4558d25](https://github.com/FlowFuse/flowfuse/pull/7892/commits/4558d25062bd05d5eb9eee6f0b80930501165861)

1. `DELETE` resolves `typeId = device.id` before the owner check
2. `byTypeAndIDs` is called with that resolved id (plus the team)

Steve-Mcl (commented):
I don't like this. The pre-operation code in the POST and DELETE handlers is almost identical, the only real difference being that POST throws (which gets swallowed by the try/catch and returned as a generic 500) where it should be replying with a proper status code. A "device not found" or "unknown type" is a 404/400, not a 500, so the two handlers are out of alignment on how they report the same failures.

I believe we should refactor the pre-op checks (resolve device/instance, verify team ownership, and confirm the target matches the session owner) into a common function used by both POST and DELETE. That keeps the behaviour consistent and avoids this kind of misalignment creeping in again as the code changes.

@andypalmi

andypalmi:
Agreed, refactored the pre-op checks into a shared `resolveTarget` function used by both POST and DELETE. It resolves the device/instance, checks team ownership, and confirms the target matches the session owner, replying with the same status code for the same failure in both routes (404 not found, 400 unknown type, 403 unauthorized). POST no longer swallows these into a generic 500. Pushed in d8940e552.

Updated the three POST tests that asserted the old 500 behavior to expect 404/404/400 to match DELETE.

andypalmi:
To be clear, the 500 was already there before your review, I just hadn't considered it worth changing. Agreed 400/404 is more useful for callers than a blanket 500, thanks for flagging it.

Steve-Mcl (approved):

PR #7893, opened 2026-07-21
fix: use a shared subscription for the Expert platform request bridge source

andypalmi:
## Problem
The bridge connector from the FF App Instance Broker to the Expert Broker is provisioned on every EMQX node. The platform-request source subscription was a plain MQTT subscription, so with N EMQX nodes each platform action ran N times, visible as duplicate snapshots, and as a 400 on create-instance when a second, racing execution collided with the `Project.safeName` uniqueness check.

## Fix
Subscribe to the platform-request topic as a shared subscription (`$share/expert/...`), matching the group name the app broker client already uses for the same topic (`forge/comms/commsClient.js`). This ensures only one node ingests each request, so the platform executes it exactly once. Per-tenant isolation is unaffected, it's enforced by the Expert Broker's licence-id mountpoint, not by the group name.

## Testing
- Reproduced locally with a 2-node EMQX broker: before the fix, a single snapshot request produced two identical snapshots; after patching the running broker's subscription to the shared form, the same request produced exactly one.
- `test/unit/forge/ee/lib/expert/tasks/emqx-bridge/setup_spec.js` passes unchanged.

Closes FlowFuse/engineering#190

### DIFF
--- a/forge/ee/lib/expert/emxq-bridge/templates.js
+++ b/forge/ee/lib/expert/emxq-bridge/templates.js
@@ -46,7 +46,7 @@ const sourceChat = {
     enable: true,
     parameters: {
         qos: 1,
-        topic: 'ff/v1/expert/+/+/+/+/support/chat/response'
+        topic: '$share/expert/ff/v1/expert/+/+/+/+/support/chat/response'
     },
     resource_opts: {
         health_check_interval: '15s'
@@ -62,7 +62,7 @@ const sourceInflight = {
     enable: true,
     parameters: {
         qos: 1,
-        topic: 'ff/v1/expert/+/+/+/+/support/inflight/+/request'
+        topic: '$share/expert/ff/v1/expert/+/+/+/+/support/inflight/+/request'
     },
     resource_opts: {
         health_check_interval: '15s'
--- a/forge/ee/lib/expert/emxq-bridge/templates.js
+++ b/forge/ee/lib/expert/emxq-bridge/templates.js
@@ -78,7 +78,7 @@ const sourcePlatform = {
     enable: true,
     parameters: {
         qos: 1,
-        topic: 'ff/v1/expert/+/+/platform/+/request'
+        topic: '$share/expert/ff/v1/expert/+/+/platform/+/request'
     },
     resource_opts: {
         health_check_interval: '15s'


Steve-Mcl:
> Reproduced locally with a 2-node EMQX broker

~~Can you explain exactly how you setup?~~ 

~~We have 1 app broker and 1 expert (central/shared) broker. 
Therefore I believe there is no need for the bridge topic to be setup as a share.~~

~~There is however a duplicate forge instance (replica) and that is why CommsClient subscribes to `'$share/expert/ff/v1/expert/+/+/platform/+/request'`~~


OK, scratch that, just spoke with Ben - it seems there is a 2 node cluster for the core forge app Broker. Ben suspects the Expert Broker may actually be the same - with that in mind, we may need to actually update the others:
* '$share/expert/ff/v1/expert/+/+/+/+/support/chat/response'
* '$share/expert/ff/v1/expert/+/+/+/+/support/inflight/+/request'
* '$share/expert/ff/v1/expert/+/+/platform/+/request'

Steve-Mcl (changes requested):
See my comment https://github.com/FlowFuse/flowfuse/pull/7893#issuecomment-5037908076 

Since it was confirmed there is a cluster, we should look at the other subs that the bridge makes.

This will need to be tested locally before deploying

andypalmi:
Confirmed setup: 2 app-broker EMQX nodes and 2 forge replicas locally, connected to the real cloud Expert broker.

On the clustering question: the duplication is driven by the app broker node count, not whether the Expert broker itself is clustered. The bridge connector is provisioned once per app-broker EMQX node, and each instance opens its own independent plain subscription against the Expert broker. N app-broker nodes means N separate pulls of every message, regardless of how many nodes sit behind the Expert broker (a subscriber sees one logical broker either way). Our 2-node app-broker repro already reproduces this fully.

So yes, `sourceInflight` and `sourceChat` are exposed to the same duplication as `sourcePlatform`. I'd missed that when scoping this PR to just the confirmed bug. It didn't show up as an observed symptom on those two paths because the consumers there already have their own protection against duplicate delivery, but this fix shouldn't rely on that for correctness.

I'll push a follow-up commit adding `$share/expert/` to both, matching `sourcePlatform`, so the fix is consistent across all three source topics.

Steve-Mcl (approved):
pulled and tested locally (including a non clustered env to ensure effect)

all good.

Issue #7891, opened 2026-07-21
Frontend: Remove feature flag + simplify shared instances component 


n-lark:
We need to check that the iframe renders correctly in PROD: 

✅ If yes - remove feature flag 
❌ If no - debug 

✅ Confirmed this was working in prod. 

---

## Acceptance criteria 

- [x] Confirm the iframe renders in PROD
- [x] Remove posthog feature flag from code `EMBEDDED_DASHBOARD_ENABLED`
- [x] Retire the dashboardRoleOnly list - Once `EMBEDDED_DASHBOARD_ENABLED` is removed, the flag-off <TeamInstances :dashboard-role-only="true" /> fallback and its mode in `Instances.vue` are dead code. Drop them for a single dashboards list.

n-lark:
# Dashboard Cleanup (flag removal + dashboardRoleOnly retirement) — Test Plan

<details><summary> ✅ All tests passed! </summary>

`EMBEDDED_DASHBOARD_ENABLED` is gone; the embedded dashboards feature is always on. The `TeamInstances` dashboard-only mode is retired.

## Feature always on (flag removed)

- [x] With the PostHog flag absent/off, team "Dashboards" nav entry still shows
- [x] Application "Dashboards" tab still shows (after Remote Instances)
- [x] Instance "Dashboard" tab shows when the instance has Dashboard 2 installed, hidden when it doesn't
- [x] No console errors referencing `isEmbeddedDashboardEnabled` / `featuresCheck`

## Dashboard-only user landing

- [x] A dashboard-role-only user visiting `/team/:slug` lands on the Dashboards list (title "Dashboards"), not Hosted Instances
- [x] Empty state ("There are no dashboards in this team.") shows when they have no dashboards
- [x] Row click opens the immersive viewer; the viewer's drawer lists their dashboards

## Hosted Instances regression (Instances.vue)

- [x] Title reads "Hosted Instances"; full helptext (all three paragraphs) renders
- [x] Rows selectable → row click opens the instance
- [x] Search, sort, and pagination work
- [x] Empty state ("Get Started with your First Node-RED Instance") + Create Instance button show when the team has no instances


</details>

Issue #7875, opened 2026-07-20
Frontend: application dashboards tab


n-lark:
# Frontend: application dashboards tab

Epic FlowFuse/product#459. Gated on `EMBEDDED_DASHBOARD_ENABLED` (subtask 0).

Application-level dashboards tab, mirroring the team pattern, backed by subtask 2's endpoint.

## Scope

- [ ] Dashboards tab on the application view, listing only that app's dashboard instances
- [ ] Clicking a instance opens the immersive dashboard editor with a list of application scoped dashboards
- [ ] <s>Remove page padding via an intercepted URL param at the page/layout level (no prop threading)</s>

## Dependencies

Subtasks 2 and 3.

n-lark:
# Application Dashboards Tab — Manual Test Plan

<details><summary> ✅ All tests passed! </summary>

Gated on `EMBEDDED_DASHBOARD_ENABLED`. Team list/viewer were generalized into shared components, so team flows need a regression pass too.

## Flag gating

- [x] Flag off: no "Dashboards" tab on the application view
- [x] Flag on: "Dashboards" tab appears (after Remote Instances)

## Dashboards tab (list)

- [x] Lists only that application's dashboard instances; instances without a dashboard, and other apps' dashboards, excluded
- [x] Search + sort by Name and Status work; no Application column (redundant in app scope)
- [x] Empty state when the app has no dashboards
- [x] Row click opens the immersive viewer; ⌘/Ctrl-click opens a new tab
- [x] Renders in-tab (app header + tabs stay visible)

## Immersive viewer (`/…/applications/:id/dashboards/:instanceId`)

- [x] Selected dashboard embeds; full immersive, no app chrome / side nav
- [x] Drawer lists only this app's dashboards; selecting one switches the embed and updates the URL
- [x] Direct load / refresh of a viewer URL loads both the application and the team (PageHeader/team switcher populated), and selects that dashboard
- [x] Home returns to the app Dashboards tab
- [x] Not-running instance shows the unavailable state; drawer pin/fullscreen/resize work
- [x] Expert button opens the assistant stacked in the drawer (back/close returns); it renders, not blank

## Access control

- [x] Only dashboards the user is permitted to access appear

## Team regression (shared components)

- [x] Team Dashboards list still works (list, search, row-click)
- [x] Team viewer still works (drawer switch, Home → team list, selection, no-flicker)
- [x] Immersive editor drawer unaffected (resize, pin/fullscreen, tabs)

</details>

PR #7894, opened 2026-07-21
[7874] Frontend: team dashboards nav entry + view

### LINKED ISSUE #7874: Frontend: team dashboards nav entry + view
# Frontend: team dashboards nav entry + view

Epic FlowFuse/product#459. Gated on `EMBEDDED_DASHBOARD_ENABLED` (subtask 0).

Team nav entry aggregating all team dashboards, backed by the existing `/teams/:teamId/dashboard-instances` endpoint. Replaces the dashboard-only user's bare instances list.

## Scope

- [x] Team nav entry (Viewer+), listing all team dashboards; exclude instances without a dashboard
- [x] Host the wrapper (subtask 3) with the selector shown; switch between dashboards in-view
- [x] Dashboard-only role lands here on login, no dead ends
- [x] Add the drawer and list of instances as documented under #7873 

## Dependencies

Subtask 3.

n-lark:
## Description

Adds a team-level Dashboards view, a nav entry and list of all the team's hosted-instance dashboards, each opening in an embedded immersive viewer with a drawer switcher. Gated behind the `EMBEDDED_DASHBOARD_ENABLED` flag.

See test plan: https://github.com/FlowFuse/flowfuse/issues/7874#issuecomment-5038692519

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7874

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/components/dashboard/index.vue
```diff
@@ -9,52 +9,48 @@
                 <p>The instance must be running to view the dashboard.</p>
             </template>
         </EmptyState>
-        <iframe v-else :src="dashboardURL" />
+        <iframe v-else :src="dashboardURL" :style="{ pointerEvents: disableEvents ? 'none' : 'auto' }" />
```
n-lark, line 12: This is so the resize can work for the drawer.

frontend/src/stores/product-expert.js
```diff
@@ -159,6 +159,18 @@ export const useProductExpertStore = defineStore('product-expert', {
...
+            if (contextStore.route?.name === 'team-dashboards-view') {
+                const drawersStore = useUxDrawersStore()
+                if (!drawersStore.editorImmersiveDrawer.state) {
+                    drawersStore.openEditorImmersiveDrawer()
+                }
+                return import('../components/expert/Expert.vue')
+                    .then(({ default: ExpertPanel }) => drawersStore.openRightDrawer({
+                        component: markRaw(ExpertPanel),
+                        header: { title: 'Expert' }
+                    }))
```
n-lark, line 171: I know this is sus but we do the same thing below 🔪

frontend/src/pages/team/index.vue
```diff
@@ -12,7 +12,8 @@
         <router-view :key="team.id" />
     </template>
     <template v-else-if="!canAccessTeam">
-        <TeamInstances :dashboard-role-only="true" />
+        <TeamDashboards v-if="isEmbeddedDashboardEnabled" />
+        <TeamInstances v-else :dashboard-role-only="true" />
```
n-lark, line 16: I added a follow up to scale down `<TeamInstances />` once the FF is off [here](https://github.com/FlowFuse/flowfuse/issues/7891#issue-4940608466) since this is the only place that uses `:dashboard-role-only="true" `.

cstns (commented):
We should guard against this and show the empty state container when the instance is running but the dashboard itself isn't.

Scenario: I tried to access the dashboard of an instance that had started but wasn't fully running. The logs showed it stuck waiting for missing node types:

```
23/07/2026 12:35:18  [info]    Waiting for missing types to be registered:
23/07/2026 12:35:18  [info]     - mcp-server
23/07/2026 12:35:18  [info]     - mcp-prompt
23/07/2026 12:35:18  [info]     - mcp-resource
23/07/2026 12:35:18  [info]     - mcp-response
23/07/2026 12:35:18  [info]     - mcp-tool
23/07/2026 12:36:06  [info]    FlowFuse Expert is Loading Advanced Completions...
23/07/2026 12:36:07  [info]    FlowFuse Expert Completions Loaded
```

How I got here: I had installed the MCP server nodes and deployed a flow using them. Later, the nodes were removed (I unlinked my npm dependencies), which left the instance in this awkward state: started, but with unregistered types blocking the dashboard.

Admittedly this is an edge case, but the current behavior gives no useful feedback to the user, so showing the empty state would be a clear improvement.

Image

cstns (commented):
Image

Image

The empty state component uses the .well type styling, but it's currently missing padding. The content sits too close to the container edges. Compare with the "Edge Devices" empty state (screenshot 1), where the well has comfortable breathing room around the illustration, heading, and CTA. We should match that spacing here so the two empty states feel consistent.

Additionally, the containing element should stretch to 100% height. Right now the well only wraps its content, which leaves the grey area exposed below it (visible in screenshot 2 under the "Dashboard not available" message). Stretching the container to fill the available vertical space, and ideally vertically centering the empty state content within it like the Edge Devices example does, would eliminate the grey strip and make the page feel intentional rather than half-rendered.

NB:
I'll fix my work as well for the instance dashboard view as I made the same mistake on my end

frontend/src/components/dashboard/index.vue
cstns, line 12: my empty state padding fix would imply adding a wrapper for the EmptyState component:
```js
        <div v-if="!isRunning" class="p-7">
            <EmptyState>
```
that would also affect the team dashboards page though

frontend/src/components/dashboard/index.vue
```diff
@@ -9,52 +9,48 @@
...
     </div>
 </template>
 
-<script>
+<script setup>
```
cstns, line 16: thank you again

frontend/src/components/drawers/dashboard/DashboardDrawer.vue
```diff
@@ -0,0 +1,121 @@
+<template>
```
cstns, line 1: there is a `components/drawers` directory that might be better suited for this component

frontend/src/pages/team/Dashboards/index.vue
```diff
@@ -0,0 +1,78 @@
...
+
+const contextStore = useContextStore()
+const { navigateTo } = useNavigationHelper()
+const { instances, loading, statusChannelLive, fetchData, instanceUpdated } = useTeamDashboards()
+
+const columns = [
+    { label: 'Name', class: ['grow'], key: 'name', sortable: true, component: { is: markRaw(DeploymentName), map: { url: 'url' }, extraProps: { copyable: true } } },
+    { label: 'Status', class: ['w-44'], component: { is: markRaw(InstanceStatusBadge), map: { instanceId: 'id', status: 'status' }, extraProps: { instanceType: 'instance' } } },
+    { label: 'Application', class: ['grow-[0.25]'], key: 'application.name', sortable: true, component: { is: markRaw(ApplicationLink), map: { id: 'application.id', name: 'application.name' } } },
+    { label: 'Last Updated', class: ['w-60'], key: 'flowLastUpdatedAt', sortable: true, component: { is: markRaw(SimpleTextCell), map: { text: 'flowLastUpdatedSince' } } }
```
cstns, line 67: we should add the kebab menu with the ability to start/stop/suspend an instance as well, it should be an easy drop in.

cstns: it should be gated by granular rbac though, added only if the user has access to start/stop it

frontend/src/components/drawers/dashboard/DashboardDrawer.vue
n-lark: moved

frontend/src/pages/team/Dashboards/index.vue
n-lark: Good idea. It's a bit more than a drop-in here, this list is `<script setup>`, so the Options-API `instanceActionsMixin` can't be reused directly, and the RBAC/row-state flags belong in the shared composable that #7875 generalizes. We also hate mixins so this could be a nice opportunity to kill it. 

I added this as a follow up here: `Frontend: status kebab on the dashboards list`
https://github.com/FlowFuse/flowfuse/issues/7927

frontend/src/components/dashboard/index.vue
n-lark: Fixed for both locations: 
Screenshot 2026-07-23 at 7 11 53 AM
Screenshot 2026-07-23 at 7 11 58 AM

cstns (approved):

PR #7895, opened 2026-07-22
Enabled Expert Insights by default for self-hosted

cstns:
## Description

Fixes the partial enablement of the insights feature for self hosted.

## Related Issue(s)

fixes the partial implementation done in https://github.com/FlowFuse/flowfuse/pull/7773

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/ee/lib/index.js
+++ b/forge/ee/lib/index.js
@@ -39,7 +39,7 @@ module.exports = fp(async function (app, opts) {
         await app.register(require('./expert'))
 
         // Set the AI Features Flag (global gate for all AI features)
-        const isAiEnabled = app.config?.ai?.enabled ?? true
+        const isAiEnabled = !!(app.config?.ai?.enabled ?? true)
         app.config.features.register('ai', isAiEnabled, true)
 
         // Set the Generate Snapshot Description Feature Flag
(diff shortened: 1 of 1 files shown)


cstns:
FYI @dimitrieh, it turns out the insights default enablement was not properly implemented in the first place. This PR fixes it. 

To move forward with full customer adoption, we'll need a point release so self-hosted customers can pick it up. Given the recent security fixes, that point release will most likely happen this week.

forge/ee/lib/index.js
```diff
@@ -56,7 +56,10 @@ module.exports = fp(async function (app, opts) {
         app.config.features.register('expertAssistant', isAiEnabled && (app.config?.expert?.enabled ?? false), true)
 
         // Set the Expert Insights flag
-        const isInsightsEnabled = isAiEnabled && app.config?.expert?.enabled && app.config?.expert?.insights?.enabled
+        const isInsightsEnabled = isAiEnabled &&
+            !!app.config?.expert?.enabled &&
+            (Object.prototype.hasOwnProperty.call(app.config?.expert ?? {}, 'insights') ? !!app.config?.expert?.insights?.enabled : true)
```
Steve-Mcl, line 61: (nit) 
Pretty sure this is functionally the same (but less verbose)
```suggestion
            (app.config?.expert?.insights?.enabled ?? true)
```

Steve-Mcl (approved):
Approved with nit

forge/ee/lib/index.js
Steve-Mcl: To be clear, this is 100% a nit - the `Object.prototype.hasOwnProperty.call` version is functionally the same as the `??` just more verbose.

dimitrieh:
@cstns from our dm's: Will you work with Piotr and/or Ben to set up a new point release in that case? If so, can you make me understand the ETA of that :)?

Our reach out to existing customers is underway but hit a bit of a snag due to esme going OOO. Resolving that in this thread: https://flowfuse.slack.com/archives/C05GYH95NJZ/p1784709785416589

cstns:
from what I can tell, a point release won't be needed because the `expert.insights.enabled` setting was set to true in the configs (@ppawlowski correct me if I'm wrong)

ppawlowski:
> the `expert.insights.enabled` setting was set to true in the configs 

* True for the docker-based deployments (https://github.com/FlowFuse/docker-compose/blob/main/docker-compose.yml#L65)
* False for the Kubernetes-based deployments - this must be explicitly enabled when configuring the Expert

PR #7906, opened 2026-07-22
Await team broker topic recording in ACL route

### LINKED ISSUE #7905: Team Broker ACL route does not await topic write, causing race condition
## Summary

In `forge/comms/v2AuthRoutes.js`, the `/acls` route handler calls `app.teamBroker.addUsedTopic(...)` without awaiting it, even though `addUsedTopic` performs a real database write (an `upsert` against `MQTTTopicSchema`, defined in `forge/ee/lib/teamBroker/index.js`).

Because the call is not awaited, the ACL response can be sent back to the client before the topic write has actually completed. This creates a race between the write and any immediately following read of the team's topic list.

## Impact

This is visible in `test/unit/forge/ee/routes/teamBroker/index_spec.js`, in the "Add topics to list" test under "Test Topic listing by Team". That test posts to `/api/comms/v2/acls` (which triggers the unawaited topic write) and then immediately calls `GET /api/v1/teams/:hashid/brokers/team-broker/topics`, asserting the just-published topic is already present.

Because the write is not awaited before the ACL response is sent, the subsequent read can run before the write lands, causing the test to fail intermittently. This is more likely to surface under the Postgres CI job than under SQLite, since Postgres write/commit latency is higher, making the race window more likely to be hit.

## Fix

Add `await` to both call sites of `app.teamBroker.addUsedTopic(...)` in `forge/comms/v2AuthRoutes.js`. Both call sites are already inside `async` route handlers, so this requires no other control-flow changes.

andypalmi:
## Summary

`app.teamBroker.addUsedTopic(...)` (defined in `forge/ee/lib/teamBroker/index.js`) performs a database upsert against `MQTTTopicSchema`, but the two call sites in the `/acls` route handler (`forge/comms/v2AuthRoutes.js`) never awaited it before sending the ACL response.

Because the write was never awaited, the handler could reply before the write actually landed, racing against any immediately following read of the team's topic list. This sometimes caused the "Add topics to list" test in `test/unit/forge/ee/routes/teamBroker/index_spec.js` to fail, since that test posts to `/api/comms/v2/acls` and then immediately reads back the topics list, expecting the just-published topic to already be present. The race is more likely to surface under the Postgres CI job than under SQLite, due to Postgres's higher write/commit latency.

## Fix

Added `await` to both `app.teamBroker.addUsedTopic(...)` call sites. Both are already inside `async` route handlers, so no other control-flow changes were needed.

Closes #7905

## Test plan

- [x] `npx mocha test/unit/forge/ee/routes/teamBroker/index_spec.js` passes, including "Add topics to list" (run repeatedly to confirm no flakiness)
- [x] `npx mocha test/unit/forge/comms/authRoutesV2_spec.js` passes
- [x] `npx eslint forge/comms/v2AuthRoutes.js` passes with no errors

### DIFF
--- a/forge/comms/v2AuthRoutes.js
+++ b/forge/comms/v2AuthRoutes.js
@@ -155,7 +155,7 @@ module.exports = async function (app) {
                 if (action === 'publish') {
                     const m = /^ff\/v1\/([^/]+)\/c\/(.+)$/.exec(topic)
                     if (m) {
-                        app.teamBroker.addUsedTopic(m[2], m[1])
+                        await app.teamBroker.addUsedTopic(m[2], m[1])
                     }
                 }
                 reply.logProperties = { username, topic, action, result: 'allow' }
@@ -202,7 +202,7 @@ module.exports = async function (app) {
                         } else {
                             if (mqttMatch(acl.pattern, request.body.topic)) {
                                 if (acl.action === 'both' || acl.action === 'publish') {
-                                    app.teamBroker.addUsedTopic(request.body.topic, teamId)
+                                    await app.teamBroker.addUsedTopic(request.body.topic, teamId)
                                     reply.logProperties = { username, topic, action, result: 'allow' }
                                     reply.send({ result: 'allow' })
                                     return


knolleary:
Given this is a high-traffic route, we chose not to await those calls so the http request isn't blocked.

If this is causing a timing issue in a test, can we modify the test to add a short delay to account for it?

andypalmi:
Given the decision behind this, I will close the pull request. If the test fails again, I will add some delay to the test. 

Thank you @knolleary !

Steve-Mcl:
> Given the decision behind this, I will close the pull request. If the test fails again, I will add some delay to the test.

@andypalmi if you do add delay to the test, can I suggest you also add an explicit comment to those un-awaited lines stating that it is deliberate & why - to avoid future confusion?

PR #7908, opened 2026-07-22
fix(mcp): resolve stacks per instance type and expose an instance's specification

andypalmi:
## Summary
- Scope `platform_list_stacks` to a given instance type, and surface a `defaultStack` (the latest recommended stack) on both `platform_list_hosted_instance_types` and `platform_list_stacks`, so the agent has a way to resolve "the latest available stack" instead of picking from an undifferentiated, cross-type list.
- Point `platform_get_hosted_instance` / `platform_get_application_hosted_instances`' descriptions at reading an instance's specification (instance type, stack, template) before creating another instance matching it, instead of guessing.

## Test plan
- [x] Live-validated against a running instance: asked the agent to create 4 more instances "of the same specification" as an existing one — it called `platform_get_hosted_instance` first, read the real `projectType`/`stack`/`template` ids, and reused them exactly rather than guessing

Closes FlowFuse/engineering#163
Closes FlowFuse/engineering#194
Closes FlowFuse/engineering#180

### DIFF
--- a/forge/ee/lib/mcp/tools/applications.js
+++ b/forge/ee/lib/mcp/tools/applications.js
@@ -36,8 +36,8 @@ module.exports = [
         description: `FlowFuse platform automation tool:
             Gets all the hosted instances that live inside an application.
             A hosted instance is a Node-RED that runs on the same environment as the FlowFuse platform.
-            Use this to see which hosted instances an application has. Each result includes the instance name, URL, and basic settings.
-            To get the full details of one specific hosted instance, call platform_get_hosted_instance with its ID.
+            Use this to see which hosted instances an application has. Each result includes the instance id, name, URL, and basic settings.
+            This list does not include an instance's specification (its instance type, stack, or template). To read an instance's specification, for example to create another instance matching it, call platform_get_hosted_instance with the instance id.
             To check if hosted instances are currently running or stopped, call platform_get_application_instances_status instead.`,
         annotations: { readOnlyHint: true, destructiveHint: false },
         inputSchema: {
--- a/forge/ee/lib/mcp/tools/applications.js
+++ b/forge/ee/lib/mcp/tools/applications.js
@@ -37,7 +37,7 @@ module.exports = [
             Gets all the hosted instances that live inside an application.
             A hosted instance is a Node-RED that runs on the same environment as the FlowFuse platform.
             Use this to see which hosted instances an application has. Each result includes the instance id, name, URL, and basic settings.
-            This list does not include an instance's specification (its instance type, stack, or template). To read an instance's specification, for example to create another instance matching it, call platform_get_hosted_instance with the instance id.
+            This list does not include an instance's specification (its instance type, stack, or template). To read an instance's specification, for example to duplicate it, call platform_get_hosted_instance with the instance id.
             To check if hosted instances are currently running or stopped, call platform_get_application_instances_status instead.`,
         annotations: { readOnlyHint: true, destructiveHint: false },
         inputSchema: {
--- a/forge/ee/lib/mcp/tools/instances.js
+++ b/forge/ee/lib/mcp/tools/instances.js
@@ -90,7 +90,7 @@ module.exports = [
             Creates a new hosted Node-RED instance inside an application. The instance starts automatically after creation.
             Before calling this tool, gather the required parameters:
             1. Call platform_list_hosted_instance_types first to see what instance types are available on this platform, then ask the user which one they want.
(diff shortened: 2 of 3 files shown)


forge/ee/lib/mcp/tools/instances.js
```diff
@@ -8,7 +8,8 @@ module.exports = [
             Gets the full details of one specific hosted instance.
             A hosted instance is a Node-RED that runs on the same environment as the FlowFuse platform.
             Use this when you already have a hosted instance ID and need to know everything about it:
-            its name, URL, settings, what application and team it belongs to, and its current state.
+            its name, URL, settings, what application and team it belongs to, its current state, and its specification (the instance type, stack, and template it uses).
+            Read the specification from this tool whenever you need to know or compare what an existing instance is running, for example to answer a question about it or to create another instance matching it.
```
cstns: it would be more concise to use `duplicate` instead of `...question about it or to create another instance matching it`

forge/ee/lib/mcp/tools/instances.js
```diff
@@ -90,7 +91,7 @@ module.exports = [
             Creates a new hosted Node-RED instance inside an application. The instance starts automatically after creation.
             Before calling this tool, gather the required parameters:
             1. Call platform_list_hosted_instance_types first to see what instance types are available on this platform, then ask the user which one they want.
-            2. If they want a specific Node-RED version or stack, or just the latest. Call platform_list_stacks to get the options. If the user has no preference, use the latest available stack.
+            2. If they want a specific Node-RED version or stack, or just the latest. Call platform_list_stacks for the chosen instance type to get the options. If the user has no preference, use the instance type's defaultStack, which is the latest recommended version.
```
cstns: I think this should work together with a step-by-step flow where the agent asks for details one at a time. I had this problem several times: I asked the expert to create an instance, and it correctly showed me a multi-option selection. The problem is that the instance types and stack versions in the options did not match each other. So when I selected a combination that was not compatible, the instance creation failed on the first try. After that, the agent called the tool again with different parameters that I never approved.

In my opinion, the solution is to make the agent understand that every instance type has a fixed set of stacks that work with it. So when a user asks to create an instance, the agent should first ask which instance type the user wants. Only after getting that answer, it should ask which stack version the user wants, showing only the stacks that work with that instance type.

forge/ee/lib/mcp/tools/platform.js
```diff
@@ -1,8 +1,13 @@
+const { z } = require('zod')
+
 module.exports = [
     {
         name: 'platform_list_hosted_instance_types',
         title: 'List Hosted Instance Types',
-        description: 'FlowFuse platform automation tool: List all available hosted instance types. Use this to find valid projectType values when creating a hosted instance.',
+        description: `FlowFuse platform automation tool:
```
cstns, line 27: Maybe a better option is to combine these two tools into one tool. It would return each instance type together with the stacks that are available for it.

andypalmi: Merge into one tool that returns the nested list of types and stacks for each type.

Each type is returned with the flags:
1. available
2. creatable

A possible argument is to pass the projectType to only see its stacks.

Also limit the project types to the creatable only (by default it's enabled)

forge/ee/lib/mcp/tools/instances.js
andypalmi: Not needed anymore

forge/ee/lib/mcp/tools/instances.js
andypalmi:
```suggestion
            Read the specification from this tool whenever you need to know or compare what an existing instance is running, for example to duplicate it.
```

forge/ee/lib/mcp/tools/applications.js
```diff
@@ -36,8 +36,8 @@ module.exports = [
         description: `FlowFuse platform automation tool:
             Gets all the hosted instances that live inside an application.
             A hosted instance is a Node-RED that runs on the same environment as the FlowFuse platform.
-            Use this to see which hosted instances an application has. Each result includes the instance name, URL, and basic settings.
-            To get the full details of one specific hosted instance, call platform_get_hosted_instance with its ID.
+            Use this to see which hosted instances an application has. Each result includes the instance id, name, URL, and basic settings.
+            This list does not include an instance's specification (its instance type, stack, or template). To read an instance's specification, for example to create another instance matching it, call platform_get_hosted_instance with the instance id.
```
andypalmi:
```suggestion
            This list does not include an instance's specification (its instance type, stack, or template). To read an instance's specification, for example to duplicate it, call platform_get_hosted_instance with the instance id.
```

Steve-Mcl (approved):

PR #7911, opened 2026-07-23
fix(mcp): stop empty UI-tool errors and tighten tool-usage guardrails

andypalmi:
Stacked on #7908.

## Summary

- `ui_navigate` (and any other UI automation tool) could return `Tool "X" failed: ` with nothing after the colon, because vue-router throws an `Error` with no `message` in production for an unknown route name, and the dispatcher's catch collapsed straight to `err.message`. The dispatcher now falls back to the error's own properties so the reason is never lost. `ui_navigate` also now catches `resolve`/`push` directly and returns a clean, structured failure. `ui_list_routes` no longer hides named redirect-only routes (like an application's landing route) from discovery.
- Tightened a few tool descriptions so an id param can't be mistaken for a display name, and so `platform_create_application` / `platform_create_remote_instance` / `platform_create_hosted_instance` are explicit about checking for name collisions before creating and asking the user instead of guessing.

## Closes

- Closes FlowFuse/engineering#188
- Closes FlowFuse/engineering#185
- Closes FlowFuse/engineering#182
- Closes FlowFuse/engineering#183

## Test plan

- [x] `npx vitest run --config ./config/vitest.config.ts test/unit/frontend/mcp/tools/navigation.spec.js test/unit/frontend/mcp/tools/routes.spec.js test/unit/frontend/services/automations.service.spec.js` - all passing
- [x] `npx eslint` on all touched files - clean

### DIFF
--- a/forge/ee/lib/mcp/tools/applications.js
+++ b/forge/ee/lib/mcp/tools/applications.js
@@ -36,8 +36,8 @@ module.exports = [
         description: `FlowFuse platform automation tool:
             Gets all the hosted instances that live inside an application.
             A hosted instance is a Node-RED that runs on the same environment as the FlowFuse platform.
-            Use this to see which hosted instances an application has. Each result includes the instance name, URL, and basic settings.
-            To get the full details of one specific hosted instance, call platform_get_hosted_instance with its ID.
+            Use this to see which hosted instances an application has. Each result includes the instance id, name, URL, and basic settings.
+            This list does not include an instance's specification (its instance type, stack, or template). To read an instance's specification, for example to create another instance matching it, call platform_get_hosted_instance with the instance id.
             To check if hosted instances are currently running or stopped, call platform_get_application_instances_status instead.`,
         annotations: { readOnlyHint: true, destructiveHint: false },
         inputSchema: {
--- a/forge/ee/lib/mcp/tools/applications.js
+++ b/forge/ee/lib/mcp/tools/applications.js
@@ -148,6 +148,7 @@ module.exports = [
         description: `FlowFuse platform automation tool:
             Creates a new application in a team.
             An application is a container that groups together hosted instances and remote instances that work together.
+            Before creating, call platform_list_applications for this team to check whether an application with this name already exists. If one does, tell the user and ask whether to use the existing one or create a new one with the same name - do not create a duplicate without asking.
             After the application is created, ask the user if they want to be taken to it. If they do, use the ui_navigate tool with the route name "Application" and params { id: <the new application id> }.`,
         annotations: { readOnlyHint: false, destructiveHint: false },
         inputSchema: {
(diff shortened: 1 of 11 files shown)


forge/ee/lib/mcp/tools/applications.js
```diff
@@ -148,6 +148,7 @@ module.exports = [
         description: `FlowFuse platform automation tool:
             Creates a new application in a team.
             An application is a container that groups together hosted instances and remote instances that work together.
+            Before creating, call platform_list_applications for this team to check whether an application with this name already exists. If one does, tell the user and ask whether to use the existing one or create a new one with the same name - do not create a duplicate without asking.
```
cstns:
```suggestion
            Before invoking this tool, call platform_list_applications for this team to check whether an application with this name already exists. If one exists, ask the user whether to use the existing one or create a new one with the same name - DO NOT create a duplicate application without asking first.
```

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -96,7 +96,7 @@ module.exports = [
             A remote instance is a Node-RED that runs on the user's own hardware rather than on the same environment as the FlowFuse platform.
             This only registers the device on the platform, it does not install anything on the user's hardware.
             The response includes credentials that the user will need to configure on their device to connect it to the platform.
-            If the user wants to assign it to an application, call platform_assign_remote_instance_to_application after creation.
+            If the user names or references an application for this device, call platform_assign_remote_instance_to_application immediately after creation. Otherwise, ask the user whether they want it assigned to an application.
```
cstns:
```suggestion
            If the user named an application for this device, call platform_assign_remote_instance_to_application immediately after creation. If not, ask before assigning. A best practice is to always assign hosted instances to applications.
```

frontend/src/mcp/tools/navigation.ts
```diff
@@ -9,6 +9,7 @@ const tools: McpToolDefinition[] = [
             Use ui_list_routes to discover valid route names and the parameters they need.
             Before navigating, call ui_get_context to remember what page the user is currently on, so you can go back if something goes wrong.
             After calling this tool, call ui_get_context again to verify the navigation actually worked and the user ended up on the right page.
+            success: true only means the route name matched and the browser was pushed to that URL - it does not confirm that an id param refers to a real entity, which is why the ui_get_context check above is required, not optional.
```
cstns:
```suggestion
            { success: true } only confirms the route name matched and the browser navigated. It does not confirm that an id param points to a real entity, which is why the ui_get_context check above is required.
```

forge/ee/lib/mcp/tools/devices.js
andypalmi:
```suggestion
            If the user named an application for this device, call platform_assign_remote_instance_to_application immediately after creation. If not, ask before assigning. A best practice is to always assign the remote instance to an application.
```

andypalmi:
@cstns I've already applied your suggestions in this PR.

The only additional change is the commit I just pushed: extracted the describeError fallback out of automations.service.ts into a shared frontend/src/mcp/tools/errors.ts helper, and reused it in ui_navigate so a NavigationFailure with no message surfaces its details instead of a generic string.

Could I get your review on this again, and if it looks good, an approve?

cstns (approved):

PR #7925, opened 2026-07-23
[7849] Carried-over search term from Home status tiles breaks pagination on Hosted Instances list 

### LINKED ISSUE #7849: Carried-over search term from Home status tiles breaks pagination on Hosted Instances list
### Current Behavior

- The table shows **"No Data Found. Try Another Search."** - even though the team has running instances
- Meanwhile the pagination footer still reflects the *unfiltered* dataset: **"1–25 of 95"** with page buttons 1–4, contradicting the empty result set
- Clearing the search restores the full list (95 items) and pagination behaves normally

Recording: https://flowfuse.slack.com/archives/C032Q63FGG1/p1784539035152989

### Expected Behavior

- Clicking a status tile should land on a list actually filtered to those instances (e.g. the running ones)
- Pagination (item count and page buttons) should always reflect the filtered result set, not the total dataset

### Steps To Reproduce

**Steps to reproduce**
1. On the team **Home** page, click one of the status tiles in the *Hosted Instances* summary (e.g. **Running**).
2. You are navigated to **Hosted Instances**, with a pre-populated search term carried over from the tile (a long pipe-separated status query: `importing | connected | info | success | pushing | pulling | loading | updating | installing | safe | protected | running | warning | starting`).

### Environment

- FlowFuse version: 2.32 (self-hosted deployment)
- Node.js version: 	v24.18.0
- npm version:
- Platform/OS: MacOS
- Browser: Firefox 151.0.2

n-lark:
## Description

See test plan here: https://github.com/FlowFuse/flowfuse/issues/7849#issuecomment-5064714421

Image

### What changed & why

- **Search box is name-only; status moved to a "Status" dropdown.** The search previously did double duty (name + a hidden status query), which returned no results and broke pagination. Now search filters by name, and the Status dropdown (Running / Error / Not Running) filters by state — both run server-side and combine.
- **Loading states consolidated into a skeleton loader.** The old full-page spinner + empty-state checks could flash "No Data Found" mid-load (worst on landing with a filter). Now there's a single loading signal and the table shows a skeleton while loading, so empty states only appear once data has actually loaded.
- **Fetch requests are sequence-guarded.** Typing fires several requests; responses can arrive out of order, and a stale (often empty) one would overwrite the correct results. Each fetch now carries a sequence number and only the latest response is applied.
- **Global-search deep-links still filter.** Server-side search disabled the table's old client-side handling of a carried `searchQuery`, so the page now reads that param and drives the server query — clicking an instance in global search lands on the list filtered by that name.

## Related Issue(s)

Resolves  #7849

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/pages/team/Instances.vue
```diff
@@ -236,7 +271,7 @@ export default {
                 },
                 {
                     label: 'Application',
-                    class: ['grow-[0.25]'],
+                    class: ['w-70'],
```
n-lark: Made this a fixed width so the skeleton loader wouldn't jump column sizes - seems okay to me but lmk.

frontend/src/pages/team/Instances.vue
```diff
@@ -194,24 +221,32 @@ export default {
...
         const { hasPermission } = usePermissions()
 
-        return { hasPermission, isRunningState, navigateTo }
+        return { hasPermission, isRunningState, navigateTo, statesMap }
     },
     data () {
         return {
-            loading: false,
+            loading: true,
+            fetchSeq: 0,
```
cstns: this looks like we're patching up some strong code smells

frontend/src/pages/team/Instances.vue
```diff
@@ -31,22 +31,48 @@
...
+                    :loading="loading"
+                    loading-type="skeleton"
                     @row-selected="openInstance"
                     @update:search="updateSearch"
                     @update:sort="updateSort"
                     @update:page="onPageChange"
                     @update:page-size="onPageSizeChange"
                 >
                     <template #actions>
+                        <ff-popover
```
cstns, line 52: there's an outline on the button when first getting clicked which we should remove

frontend/src/pages/team/Instances.vue
```diff
@@ -31,22 +31,48 @@
...
+                            button-kind="secondary"
+                            data-el="status-filter"
+                        >
+                            <template #panel>
+                                <section>
+                                    <popover-item
+                                        v-for="filter in statusFilters" :key="filter.key"
+                                        :title="filter.label"
+                                        :data-action="'filter-' + filter.key"
+                                        @click="toggleStatusGroup(filter.key)"
```
cstns, line 63: we should keep the url params in sync with the stats group when toggled. Ex: landing on the instances page with a filter from the dashboard, changing the filter works but upon page refresh were taken back to the original filter we landed with

frontend/src/pages/team/Instances.vue
```diff
@@ -31,22 +31,48 @@
...
+                                    >
+                                        <template #icon>
+                                            <ff-checkbox
```
cstns, line 66: very good initiative! we can expand these later to also filter by application, dashboard etc

frontend/src/pages/team/Instances.vue
cstns: an abortController would be the better approach here

cstns: by using the fetchSeq leaves the requests to get fulfilled even if they should have been canceled

cstns (changes requested):

frontend/src/pages/team/Instances.vue
n-lark: This is the keyboard-focus outline, it's on every ff-btn app-wide, not unique to this PR. It has to stay for keyboard focus (WCAG 2.4.7) but we can probably hide it just on mouse-click. It would be a shared button change so if we want that we should create a follow up since it has a wide scope.

frontend/src/pages/team/Instances.vue
n-lark: done

frontend/src/pages/team/Instances.vue
n-lark: swapped to abortController

cstns (approved):

PR #7932, opened 2026-07-24
Add session persistence, session timer recovery, and store reset methods

### LINKED ISSUE #7915: Persist active expert chat session in local storage
### Current Behavior

Page refreshes currently discard the active chat session entirely.

As a quick win, the active chat session should be stored in local storage so it survives refreshes. The stored session should be cleared when:

- The user starts a new chat session
- The user logs back in
- The user logs out

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

cstns:
## Description

store expert chats in localStorage

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/7915

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/stores/account-auth.js
+++ b/frontend/src/stores/account-auth.js
@@ -42,6 +42,12 @@ export const useAccountAuthStore = defineStore('account-auth', {
             return this.sessionId
         },
         login (user) {
+            // Clear persisted expert chat when the authenticated user changes
+            // (covers SSO account switches, re-login as a different user, etc.)
+            useProductExpertSupportAgentStore().$reset()
+            useProductExpertInsightsAgentStore().$reset()
+            useProductExpertStore().$reset()
+
             this.user = user
             this.loginInflight = false
         },
@@ -161,9 +167,6 @@ export const useAccountAuthStore = defineStore('account-auth', {
                         console.error('posthog error resetting user')
                     }
                 }
-
-                // Clear any persisted chat session so a subsequent login starts fresh
-                this.clearStores()
 
                 if (router.currentRoute.value.meta.requiresLogin !== false) {
                     const currentPath = router.currentRoute.value.fullPath === '/'
--- a/frontend/src/stores/account-auth.js
+++ b/frontend/src/stores/account-auth.js
@@ -42,12 +42,6 @@ export const useAccountAuthStore = defineStore('account-auth', {
             return this.sessionId
         },
         login (user) {
-            // Clear persisted expert chat when the authenticated user changes
-            // (covers SSO account switches, re-login as a different user, etc.)
-            useProductExpertSupportAgentStore().$reset()
-            useProductExpertInsightsAgentStore().$reset()
-            useProductExpertStore().$reset()
-
             this.user = user
             this.loginInflight = false
         },
@@ -167,6 +161,10 @@ export const useAccountAuthStore = defineStore('account-auth', {
                         console.error('posthog error resetting user')
                     }
                 }
+
+                useProductExpertSupportAgentStore().$reset()
+                useProductExpertInsightsAgentStore().$reset()
+                useProductExpertStore().$reset()
 
                 if (router.currentRoute.value.meta.requiresLogin !== false) {
                     const currentPath = router.currentRoute.value.fullPath === '/'
--- a/frontend/src/stores/account-auth.js
+++ b/frontend/src/stores/account-auth.js
@@ -162,6 +162,9 @@ export const useAccountAuthStore = defineStore('account-auth', {
                     }
                 }
 
+                // Clear any persisted chat session so a subsequent login starts fresh
+                this.clearStores()
+
                 if (router.currentRoute.value.meta.requiresLogin !== false) {
                     const currentPath = router.currentRoute.value.fullPath === '/'
                         ? window.location.pathname + window.location.search + window.location.hash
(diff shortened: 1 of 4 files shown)


andypalmi:
A couple of things I noticed:

1. `clearStores()` (`account-auth.js`) is called inside `checkState()`'s catch block, the "not logged in" branch, not the successful-login branch. `checkState()` runs both at app boot and right after `loginWithCredentials()` succeeds, but on a successful login `userApi.getUser()` takes the try path, so this call never fires there. In the usual flow (logout or session expiry, then login page loads, then submit credentials) it still works out, since reaching the login page itself goes through the not-logged-in branch first. But it means the clear is tied to "detected as not logged in" rather than "just logged in," so a flow that goes straight from one authenticated identity to another without an intervening logged-out state (an SSO account switch, for example) would skip it. Was this intentional, or would it make more sense to move this into the success path, maybe `loginWithCredentials` itself?

2. `resumeSessionTimer()` in `product-expert.js` duplicates the interval callback from `startSessionTimer()` almost line for line (same threshold checks, same `addSystemMessage` calls), just without resetting `sessionStartTime`. Might be worth pulling the shared interval logic into one helper both functions call, so there's only one copy to keep in sync.

cstns:
good catch, fixing now

andypalmi:
Thanks for the quick turnaround, the session timer refactor into `_startSessionCheckInterval()` looks good, that resolves the duplication.

One more thing on the login-clearing fix though: `login(user)` now unconditionally `$reset()`s the expert stores, but `this.login(user)` (`account-auth.js:78`) is called from `checkState()`'s try branch, which runs on every app boot, including an ordinary page refresh while already logged in, not just on an actual new login. `App.vue` calls `checkState()` on mount, so this fires every time. That means the expert stores get reset, and the reset gets persisted, wiping localStorage, on every refresh, which reintroduces the exact bug #7915 was meant to fix, just through this new code path.

The trouble is `login()` fires for two different situations that need different handling: a genuinely new authenticated identity should clear, but the same user's session just being reconfirmed on refresh should not. Since `this.user` isn't persisted, there's nothing to diff against in memory after a reload to tell these apart.

Also worth noting, moving the clear into `loginWithCredentials()` instead wouldn't fully cover this either, since both SSO paths (Google and direct/SAML) do a full `window.location` redirect and never call `loginWithCredentials()`. They land back on the same boot-time `checkState()` path.

Would persisting a `userId` on the account-auth store and only resetting when the incoming `user.id` differs from the persisted one work here? That would cover password login, MFA, and both SSO flows the same way, while leaving an ordinary refresh by the same user alone.

andypalmi:
**`afterHydrate` receives the wrong argument** (`product-expert-support-agent.js:33-40` and the identical block in `product-expert-insights-agent.js:53-60`)

```js
afterHydrate (store) {
    store.messages.forEach(msg => {
```

I checked `pinia-plugin-persistedstate`'s actual runtime and confirmed `afterHydrate` is called with the Pinia plugin context (`{ store, app, pinia, options }`), not the store itself, so `store.messages` is `undefined` here and `.forEach` throws every time a store hydrates. The plugin's own try/catch swallows this silently (no `debug: true` set), so it never surfaces anywhere. Was this tested with a persisted conversation that actually contains a multi-part answer (tool calls, resource lists) after a refresh? A single-answer text reply won't show any visible symptom, which might be why this slipped through.

```suggestion
        afterHydrate ({ store }) {
            store.messages.forEach(msg => {
                msg._streamed = true
                if (msg.answer) {
                    msg.answer.forEach(a => { a._streamed = true })
                }
            })
        }
```

(same fix applies to `product-expert-insights-agent.js`)

**`resumeSessionTimer` keeps polling after a session has already expired** (`product-expert.js:723`)

```js
if (!agentStore.sessionStartTime || agentStore.sessionCheckTimer) return
```

If `sessionExpiredShown` is already `true` before the refresh, this still spins up a 30s interval that can never do anything useful (both flags are already set). Small early-return would avoid it:
```suggestion
            if (!agentStore.sessionStartTime || agentStore.sessionCheckTimer || agentStore.sessionExpiredShown) return
```

**`localStorage` scopes chats across tabs, not per-tab** — as discussed, since `localStorage` is shared across all tabs of the same origin, having multiple chats open in different tabs means each one overwrites the same key, and a refresh leaves only the last-written chat visible, silently replacing the others. Switching `persist.storage` from `localStorage` to `sessionStorage` in the three stores fixes this (still survives a refresh, but scoped to that tab).

Also found (unrelated to this diff, but only reproducible because of it): `FlowsList.vue` is missing a `!shouldStream ||` fallback that its sibling list components have, so "Related Flows" cards stay permanently hidden after this PR restores an already-answered message. Filed as #7949, fix is up in #7948, no action needed here.

cstns:
very nice catch! fixing now

andypalmi (approved):
Looks solid, all the earlier feedback (`afterHydrate` argument, session timer duplication, `sessionExpiredShown` guard, localStorage to sessionStorage) has been addressed.

One residual, non-blocking edge case: `clearStores()` only fires from `checkState()`'s catch branch, so an SSO identity switch that goes authenticated to authenticated without an intervening logged-out state would still skip the reset. Both SSO flows do a full-page redirect through `checkState()` at boot, so the realistic exposure is narrow (shared-device SSO handoff without a logout in between). Worth a follow-up if you want full coverage, not blocking here.

PR #7933, opened 2026-07-24
Fix expert chat input focus on drawer open and after MQTT response

### LINKED ISSUE #7931: Expert chat input loses focus
### Current Behavior

Two related focus issues with the expert chat input:

1. **Opening the drawer manually** (clicking the expert button, not the auto-open after login) focuses the drawer container element, not the chat textarea. Users have to click into the input before they can start typing.                                                                                                                                     
2. **After receiving a response**, the chat input loses focus. The user has to click the input again to continue the conversation.                                             

## Why it happens

The drawer's `mounted()` hook calls `this.$refs.drawer.focus()`, which focuses the outer `<div>`, not the chat input component.                                                  

The response focus issue is specific to the MQTT path. The original flow was built around HTTP, where `handleQuery()` awaits the full response before resolving. `handleSend()` chains `.then(() => textarea.focus())` on that promise, so focus restores after the response renders. With MQTT, `sendMqttQuery()` returns `Promise.resolve()` immediately after publishing. The `.then()` fires right away (before any response arrives), and when the actual response later comes in through `_onMqttMessage`, DOM updates blur the input with no refocus to follow.                                                                                                                                                          

The HTTP version also relied on the `AbortController` for lifecycle management (cancellation, cleanup in `finally`). That scaffolding carried over structurally (the controller is still created and nulled) but doesn't serve the same role over MQTT, where cancellation is a separate `abort: true` publish. The gap means response lifecycle hooks like post-response refocus don't have a natural place to run in the MQTT flow.                                                                                                        

### Expected Behavior

the chat input should get auto-focused to aid expert interaction and ux

### Steps To Reproduce

_No response_

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

cstns:
## Description

- Focus the chat textarea (not the drawer container) when the expert drawer is manually opened, so the user can start typing immediately
- Refocus the chat input after a response completes, using a watcher on `isWaitingForResponse` instead of chaining `.then()` on `handleQuery()`. The old approach relied on the HTTP promise lifecycle where `handleQuery` resolves after the full response. Over MQTT, `sendMqttQuery` returns immediately, so the `.then()` refocus fired before the response arrived and any DOM updates from the actual response would blur the input with nothing to restore it.                                                                          
                   

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/7931

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -158,9 +158,12 @@ module.exports = function (env, argv) {
             new DotenvPlugin(),
             new DefinePlugin({
                 __VUE_OPTIONS_API__: true,
-                __VUE_PROD_DEVTOOLS__: devMode
+                __VUE_PROD_DEVTOOLS__: devMode,
+                __VUE_PROD_HYDRATION_MISMATCH_DETAILS__: devMode,
+                'process.env.hotReloading': devMode && process.env.NODE_RUN_HOT === 'hot'
             })
         ],
+        cache: { type: 'filesystem' },
         optimization: {
             moduleIds: 'deterministic',
             runtimeChunk: 'single',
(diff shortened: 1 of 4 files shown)


andypalmi (approved):
Nice fix, traced through isWaitingForResponse and the MQTT _inFlightRequests cleanup and confirmed the watcher transition lines up correctly with the actual response arriving for both HTTP and MQTT.

PR #7936, opened 2026-07-24
fix(mcp): forward device comms to platform automation tool calls

### LINKED ISSUE #7912: platform_get_remote_instance_status failing even if the remote instance is online
### Current Behavior

The platform_get_remote_instance_status tool always returns a failure response even when the target device is up and running:
```json
  {
    "success": false,
    "error": "Device communications not available",
    "tool": "platform_get_remote_instance_status"
  }
```
  This causes the LLM to fall back to platform_get_remote_instance, which returns the last state registered by the application. That state can be stale, so the agent ends up reporting outdated info.

### Expected Behavior

  platform_get_remote_instance_status should work over MQTT and device comms so it returns live status when the device is   reachable.

### Steps To Reproduce

_No response_

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

andypalmi:
Closes #7912

`platform_get_remote_instance_status` always returned `Device communications not available`,
even when the target device was online, because the platform-automation MQTT bridge
(`forge/comms/platformAutomation.js`) only passed `inject` to tool handlers, never `comms`.
The regular MCP tool loader (`forge/ee/lib/mcp/toolLoader.js`) already forwarded `comms`
correctly, so this only affected calls made through the Expert.

Forwards `comms: this.app.comms?.devices` alongside `inject`, matching the tool loader.

Added test coverage in `test/unit/forge/comms/platformAutomation_spec.js`:
- comms is forwarded to the tool so it can return live state instead of the old error
- a missing comms configuration still degrades gracefully with a structured error

Tested against a running device agent: before the fix, the tool always returned
`Device communications not available` regardless of device state. After the fix,
it returns the real live status, e.g.:

    {
      "state": "running",
      "health": { "uptime": 694, "snapshotRestartCount": 0 },
      "snapshot": "0"
    }

### DIFF
--- a/forge/comms/platformAutomation.js
+++ b/forge/comms/platformAutomation.js
@@ -115,7 +115,7 @@ class PlatformAutomationHandler {
                     }
 
                     const { formatResponse } = require('../ee/lib/mcp/toolLoader')
-                    const response = await tool.handler(args, { inject })
+                    const response = await tool.handler(args, { inject, comms: this.app.comms?.devices })
                     result = formatResponse(response)
                 }
                 break
--- a/test/unit/forge/comms/platformAutomation_spec.js
+++ b/test/unit/forge/comms/platformAutomation_spec.js
@@ -134,6 +134,49 @@ describe('PlatformAutomationHandler', function () {
         })
     })
 
+    describe('platform_get_remote_instance_status', function () {
+        afterEach(function () {
+            delete app.comms
+        })
+
+        it('forwards device comms to the tool so it can query live state', async function () {
+            app.comms = {
+                devices: {
+                    sendCommandAwaitReply: sinon.stub().resolves({ state: 'running', health: 'ok', snapshot: 'snapshot-1' })
+                }
+            }
+            const tool = handler.findTool('platform_get_remote_instance_status')
+
+            const res = await invokeToolCall({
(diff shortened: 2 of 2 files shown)


andypalmi:
One thing worth flagging for review: this passes the full `app.comms.devices` (`DeviceCommsHandler`) into the tool handler, not just the read-only `get-liveState` call it uses today. `platform_get_remote_instance_status` is currently the only handler touching `comms`, and it only reads state, but nothing at the plumbing level scopes that down - the handler also has `sendCommand`, `enableEditor`/`disableEditor`, and broadcast-to-all-devices methods available on the same object. The `readOnlyHint`/`destructiveHint` annotations are descriptive for the LLM, not an enforced boundary.

Worth discussing whether to narrow what gets passed here (e.g. a scoped `{ getLiveState }` wrapper instead of the whole handler) before more handlers start depending on `comms`, or if this is an acceptable tradeoff for now.

cstns:
I passed the device comms handler to not end up with a scattered mess of individual helpers get passed into the tool handlers. I did not see anything wrong with doing that as the execution/usage of those arguments are under our control.

adding this as a talking point for today's briefing

cstns (approved):

PR #7951, opened 2026-07-27
fix(tables): include schema in tables list, fix data queries and show schema in UI

andypalmi:
## Summary
- The tables list endpoint never returned a table's schema, even though the response schema documents it. The Postgres drivers only selected `tablename`, not `schemaname`.
- The table data endpoint 500'd for any table outside the `public` schema, because the query used an unqualified identifier that only resolves via Postgres's default `search_path`.
- `getTable`, `getTableData`, and `dropTable` now accept an optional trailing `schemaName` path segment, so a specific table can be targeted when the same name exists in more than one schema. This is backwards compatible (omitting it keeps the existing auto-detection), so no v1 to v2 migration was needed. Schema resolution never assumes `public` as a fallback.
- The Tables UI now surfaces schema throughout: a muted label next to each table name in the list, `schema_name.table_name` (with a copy button) in the detail view header, and `tableSelection` in the frontend store is now a `{name, dbSchema}` identity so the right table is always addressed even when names collide across schemas.
- The create-table drawer no longer implies a schema picker is coming; it notes the table will be created in the database's default schema, since there's no endpoint yet to list a database's schemas.

Closes FlowFuse/engineering#208
Closes FlowFuse/engineering#209
Closes FlowFuse/engineering#210

## Test plan
- [x] Unit tests added/updated for both Postgres drivers (getTables schema field, getTable/getTableData/dropTable schema resolution with and without an explicit schemaName, and the no-hardcoded-schema fallback path)
- [x] Unit tests added/updated for the frontend store (dbSchema mapping, schema-aware tableSelection, schema-scoped getTableSchema/getTableData/deleteTable)
- [x] Verified manually against a real Postgres instance: created a table in a non-public schema, confirmed it appears in the tables list with the correct schema and that its sample data loads without error

### DIFF
--- a/forge/ee/lib/tables/drivers/postgres-localfs.js
+++ b/forge/ee/lib/tables/drivers/postgres-localfs.js
@@ -126,7 +126,7 @@ module.exports = {
             const teamClient = libPg.newClient({ ...this._options.database, database: team.hashid })
             try {
                 await teamClient.connect()
-                const res = await teamClient.query('SELECT "tablename" FROM "pg_catalog"."pg_tables" WHERE "schemaname" != \'pg_catalog\' AND "schemaname" != \'information_schema\'')
+                const res = await teamClient.query('SELECT "tablename", "schemaname" FROM "pg_catalog"."pg_tables" WHERE "schemaname" != \'pg_catalog\' AND "schemaname" != \'information_schema\'')
                 if (res.rows && res.rows.length > 0) {
                     const tables = res.rows.map(row => {
                         return {
@@ -202,8 +202,12 @@ module.exports = {
             const teamClient = libPg.newClient({ ...this._options.database, database: team.hashid })
             try {
                 await teamClient.connect()
+                // look up the table's schema instead of relying on the default search_path ("public")
+                const schemaRes = await teamClient.query('SELECT "schemaname" FROM "pg_catalog"."pg_tables" WHERE "tablename" = $1 AND "schemaname" != \'pg_catalog\' AND "schemaname" != \'information_schema\' LIMIT 1', [table])
+                const schemaName = schemaRes.rows[0]?.schemaname || 'public'
+                const escapedSchema = libPg.pg.escapeIdentifier(schemaName)
                 const escapedTable = libPg.pg.escapeIdentifier(table)
-                const query = `SELECT * FROM ${escapedTable} LIMIT $1`
+                const query = `SELECT * FROM ${escapedSchema}.${escapedTable} LIMIT $1`
                 const res = await teamClient.query(query, [rows || 10])
                 if (res.rows && res.rows.length > 0) {
                     return {
--- a/forge/ee/lib/tables/drivers/postgres-localfs.js
+++ b/forge/ee/lib/tables/drivers/postgres-localfs.js
@@ -154,7 +154,7 @@ module.exports = {
             throw new Error(`Failed to retrieve tables for team ${team.hashid}: ${err.message}`)
         }
     },
-    getTable: async function (team, database, table) {
+    getTable: async function (team, database, table, schemaName) {
         // SELECT column_name, data_type, is_nullable, column_default
         // FROM information_schema.columns
         // WHERE table_name = 'your_table_name';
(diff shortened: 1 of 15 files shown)


hardillb:
IIRC we only setup the database user with access to the public schema, not any others

hardillb:
@andypalmi just making sure you say my comment, that FF Tables users **ONLY** have access to the public schema

https://github.com/FlowFuse/flowfuse/blob/980da5f93995b243ded8b20a391a424809a15af1/forge/ee/lib/tables/drivers/postgres-localfs.js#L74-L76

forge/ee/lib/tables/drivers/postgres-localfs.js
```diff
@@ -202,8 +202,12 @@ module.exports = {
             const teamClient = libPg.newClient({ ...this._options.database, database: team.hashid })
             try {
                 await teamClient.connect()
+                // look up the table's schema instead of relying on the default search_path ("public")
+                const schemaRes = await teamClient.query('SELECT "schemaname" FROM "pg_catalog"."pg_tables" WHERE "tablename" = $1 AND "schemaname" != \'pg_catalog\' AND "schemaname" != \'information_schema\' LIMIT 1', [table])
+                const schemaName = schemaRes.rows[0]?.schemaname || 'public'
+                const escapedSchema = libPg.pg.escapeIdentifier(schemaName)
```
hardillb: Not needed as the ff tables users don't have access to any other schema

forge/ee/lib/tables/drivers/postgres-supavisor.js
```diff
@@ -268,8 +268,12 @@ module.exports = {
             const teamClient = libPg.newClient({ ...this._options.backend, database: team.hashid })
             try {
                 await teamClient.connect()
+                // look up the table's schema instead of relying on the default search_path ("public")
+                const schemaRes = await teamClient.query('SELECT "schemaname" FROM "pg_catalog"."pg_tables" WHERE "tablename" = $1 AND "schemaname" != \'pg_catalog\' AND "schemaname" != \'information_schema\' LIMIT 1', [table])
+                const schemaName = schemaRes.rows[0]?.schemaname || 'public'
+                const escapedSchema = libPg.pg.escapeIdentifier(schemaName)
```
hardillb: Same as above, FF tables users can't access any other schema

frontend/src/pages/team/Tables/Table/TableExplorer/components/RowsHeader.vue
```diff
@@ -11,6 +11,10 @@
         >
             <ArrowPathIcon class="ff-icon ff-icon-md" aria-hidden="true" />
         </button>
+        <span v-if="selectedTable" class="table-title truncate">
+            {{ selectedTable.name }}
+            <span class="schema-label">{{ selectedTable.dbSchema }}</span>
+        </span>
```
hardillb, line 16: No point in showing it if you can't use any other schema

andypalmi:
@hardillb This came up while testing the tables-query node with @steve-mcl and @cstns, specifically poking at edge cases. Thanks for flagging it, wanted to share what we found.

The role isn't actually restricted to `public` in practice. `GRANT ALL PRIVILEGES ON DATABASE ... TO ...` includes `CREATE`, so the role can run `CREATE SCHEMA` itself and becomes owner of whatever it creates there. We reproduced it on FlowFuse Cloud with a tables-query node:

```sql
CREATE SCHEMA "named-tables";
CREATE TABLE "named-tables"."fred" (
    "id" serial4 NOT NULL,
    "description" text,
    CONSTRAINT "fred_pkey" PRIMARY KEY ("id")
);
```

followed by an insert and a `SELECT * FROM "named-tables"."fred"`, both worked fine.

Personally I think it's actually useful to let FF Tables users create their own schemas this way, so I'd lean towards keeping that possible. But if the decision is that users shouldn't be able to do that, happy to close this PR and file a follow-up to restrict the role to `public` only instead. As things stand today though, users can create tables outside `public`, so I think this fix is still needed for that case.

andypalmi:
To add a bit more detail on that last point: once a user creates a table outside `public`, two concrete problems show up that this PR addresses.

- Tables with the same name in different schemas get mixed up, since the API/UI had no way to distinguish them.
- The table data endpoint 500s outright for any table that isn't in `public`, since the query relied on an unqualified identifier.

So both issues are already reachable today, not just hypothetical.

forge/ee/lib/tables/drivers/postgres-localfs.js
```diff
@@ -166,7 +166,18 @@ module.exports = {
             const teamClient = libPg.newClient({ ...this._options.database, database: team.hashid })
             try {
                 await teamClient.connect()
-                const res = await teamClient.query('SELECT column_name, udt_name, is_nullable, column_default, character_maximum_length, is_generated FROM information_schema.columns WHERE table_name = $1', [table])
+                let resolvedSchema = schemaName
+                if (!resolvedSchema) {
+                    const schemaRes = await teamClient.query('SELECT "schemaname" FROM "pg_catalog"."pg_tables" WHERE "tablename" = $1 AND "schemaname" != \'pg_catalog\' AND "schemaname" != \'information_schema\' LIMIT 1', [table])
+                    resolvedSchema = schemaRes.rows[0]?.schemaname
```
hardillb, line 172: Doesn't this still fail if there are 2 tables with the same name in different schema and no schemaName is passed in? It will only return the schema name for the first one. Should it not also have a `schemaName != 'public'` in the query?

andypalmi: Worth noting first: the frontend already avoids this case. Once a table is selected, it always passes the new `schemaName` param, since `getTables` now returns each table's schema. So this only comes up if someone calls `GET /tables/:tableName(/data)` directly, without that param.

For that direct-call case: today there's no `ORDER BY` in the fallback query, so it just returns whichever schema Postgres happens to return first for that table name.

Should we make that consistent? A couple of options if so:

- Sort alphabetically by schema name, so it's at least the same every time (though it wouldn't specifically pick `public`).
- Or sort so `public` comes first, then alphabetically for the rest, so a collision always resolves to `public`.

One thing either way: if we ever change the default schema away from `public`, or let users pick their own, we'd need to remember to update this too.

What do you think makes sense here?

hardillb: Leave it as is for now

hardillb (approved):

PR #7990, opened 2026-07-28
feat(mcp): add FlowFuse Tables platform automation tools

### LINKED ISSUE #7989: 5.4-a.2 FlowFuse Tables tools
Splitting up Team Read tools to implement the FlowFuse Tables as a separate PR

**Parent:** #7673 (5.4 Teams and membership)
**Tool file:** `forge/ee/lib/mcp/tools/teams.js` (extend)

Read tools for the FlowFuse Tables.

`readOnlyHint: true`.

|---|---|---|---|
| `platform_list_team_databases` | `GET /teams/:teamId/databases` | `team:database:list` | read |
| `platform_get_team_database` | `GET /teams/:teamId/databases/:databaseId` | `team:database:list` | read |
| `platform_list_database_tables` | `GET /teams/:teamId/databases/:databaseId/tables` | `team:database:list` | read |
| `platform_get_database_table` | `GET /teams/:teamId/databases/:databaseId/tables/:tableName` | `team:database:list` | read |
| `platform_query_database_table_data` | `GET /teams/:teamId/databases/:databaseId/tables/:tableName/data` | `team:database:list` | read |

**Design notes:**
- **Database reads leak credentials - redact before returning.** `GET /databases` and `GET /databases/:databaseId` return the `DatabaseCredentials` view, whose `credentials` object includes a required `password` field (`forge/ee/db/views/Table.js`). These read tools strip the `credentials` object from the tool output, consistent with the #7513 credential-escalation stance. Table and table-data reads return only schema/row data, no credentials.
- Plan-gated per team (not per user role): `tables`. When the plan does not include the feature the route returns a 404 independent of the PAT's permissions; the tool surfaces a clear "feature not enabled for this team" message. The response body differs per surface: `tables` returns the literal `Not Found - not available on team`.

**Scopes to allow-list + access-tag (#7445):**
`team:database:list` (read)

andypalmi:
## Summary
- Adds `platform_list_team_databases`, `platform_get_team_database`, `platform_list_database_tables`, `platform_get_database_table`, and `platform_query_database_table_data` to the MCP platform automation tool set, covering FlowFuse Tables databases, tables, and row data.
- Database responses have their `credentials` object stripped via a shared helper (`forge/ee/lib/mcp/utils.js`) before being returned, since the underlying API includes a password in that object.
- Adds `team:database:list` to the `expert-mcp` token's implicit scope allow-list.

Closes FlowFuse/flowfuse#7989
Closes FlowFuse/engineering#213

## Test plan
- [x] Unit tests added for all 5 tools covering URL construction, credential redaction on success, error passthrough, and the `limit` query param


forge/ee/lib/mcp/tools/teams.js
```diff
@@ -24,5 +26,108 @@ module.exports = [
             const response = await inject({ method: 'GET', url: `/api/v1/teams/${args.teamId}` })
             return response
         }
+    },
+    {
+        name: 'platform_list_team_databases',
+        title: 'List Team Databases',
+        description: `FlowFuse platform automation tool:
+            Lists the FlowFuse Tables databases for a team.
+            FlowFuse Tables is a plan-gated feature; if it is not enabled for the team's plan, the underlying API's error response is returned as-is.
```
Steve-Mcl: Is there a way for the expert to discover this? What happens if this tool is called and the feature is unavailable to the user/team? what is returned?

IMHO, we should do 1 of the following (my preference is no1) 
1. a low cost feature check  and return a message about "not enabled for this team"
2. not bind this tool if the feature is not enabled.

forge/ee/lib/mcp/tools/teams.js
```diff
@@ -24,5 +26,108 @@ module.exports = [
...
+            The underlying API response includes a credentials object with connection details, including a password. This tool strips that object before returning results, so no credentials are ever exposed.`,
```
Steve-Mcl: Isnt this just bound to confuse the LLM? Why even mention it?

forge/ee/lib/mcp/tools/teams.js
```diff
@@ -24,5 +26,114 @@ module.exports = [
...
+        },
+        handler: async (args, { inject }) => {
+            const response = await inject({ method: 'GET', url: `/api/v1/teams/${args.teamId}/databases` })
+            if (response.statusCode >= 400) {
+                return response
+            }
+            const databases = response.json().map(redactDatabaseCredentials)
+            return {
+                statusCode: response.statusCode,
+                json: () => databases
```
Steve-Mcl: Consider converting the response to an object containing { databases } (since it is a non-extensible array) - future proofing for when we need to augment this response.

forge/ee/lib/mcp/tools/teams.js
```diff
@@ -24,5 +26,114 @@ module.exports = [
...
+            }
+        }
+    },
+    {
```
Steve-Mcl: Is this tool even necessary? the one above returns same info and only every contains 1 item. In a future where MANY databases are returned, I can see the benefit of this tool, just not right now.

forge/ee/lib/mcp/tools/teams.js
```diff
@@ -24,5 +26,114 @@ module.exports = [
...
+        inputSchema: {
+            teamId: z.string().describe('The ID or hashid of the team'),
+            databaseId: z.string().describe('database hashid'),
+            tableName: z.string().describe('Name of the database table'),
+            schemaName: z.string().describe('Schema the table lives in, as returned by platform_list_database_tables')
+        },
+        handler: async (args, { inject }) => {
+            const url = `/api/v1/teams/${args.teamId}/databases/${args.databaseId}/tables/${args.tableName}/${encodeURIComponent(args.schemaName)}`
+            const response = await inject({ method: 'GET', url })
+            return response
```
Steve-Mcl: Consider converting the response to an object containing { database: databaseId, tableName, schemaName, columns } (since it is a non-extensible array) - - will aid the AI and future proofing for when we need to augment this response.

forge/ee/lib/mcp/tools/teams.js
```diff
@@ -24,5 +26,114 @@ module.exports = [
...
+        handler: async (args, { inject }) => {
+            const response = await inject({ method: 'GET', url: `/api/v1/teams/${args.teamId}/databases/${args.databaseId}/tables` })
+            return response
+        }
+    },
+    {
+        name: 'platform_get_database_table',
+        title: 'Get Database Table',
+        description: `FlowFuse platform automation tool:
+            Gets the schema definition of a single table in a FlowFuse Tables database (column names, types, and constraints). Does not return row data or credentials.
```
Steve-Mcl: Is it really necessry to add "Does not return row data or credentials" when it starts with "Gets the schema definition of a single table"

forge/ee/lib/mcp/tools/teams.js
```diff
@@ -24,5 +26,114 @@ module.exports = [
...
+            schemaName is required, since the same table name can exist in more than one schema; get it from platform_list_database_tables.
```
Steve-Mcl: is thought `schemaName` was to be optional (i.e. to be backwards compatible with the v1 API)?

forge/ee/lib/mcp/tools/teams.js
```diff
@@ -24,5 +26,114 @@ module.exports = [
...
+            FlowFuse Tables is a plan-gated feature; if it is not enabled for the team's plan, the underlying API's error response is returned as-is.
```
Steve-Mcl: IMHO, if the feature is not enabled, we should do 1 of the following (my preference is no1)

a low cost feature check and return a message about "not enabled for this team"
not bind this tool if the feature is not enabled.

forge/ee/lib/mcp/tools/teams.js
```diff
@@ -24,5 +26,114 @@ module.exports = [
...
+        }
+    },
+    {
+        name: 'platform_query_database_table_data',
+        title: 'Query Database Table Data',
+        description: `FlowFuse platform automation tool:
+            Reads the row data of a table in a FlowFuse Tables database. There are no column-filter parameters; this returns rows as stored.
+            At most 10 rows are returned per call (the limit is capped at 10 by the platform).
+            schemaName is required, since the same table name can exist in more than one schema; get it from platform_list_database_tables.
+            FlowFuse Tables is a plan-gated feature; if it is not enabled for the team's plan, the underlying API's error response is returned as-is.
```
Steve-Mcl: IMHO, if the feature is not enabled, we should do 1 of the following (my preference is no1)

a low cost feature check and return a message about "not enabled for this team"
not bind this tool if the feature is not enabled.

forge/ee/lib/mcp/tools/teams.js
```diff
@@ -24,5 +26,114 @@ module.exports = [
...
+            return response
+        }
+    },
+    {
+        name: 'platform_query_database_table_data',
+        title: 'Query Database Table Data',
+        description: `FlowFuse platform automation tool:
+            Reads the row data of a table in a FlowFuse Tables database. There are no column-filter parameters; this returns rows as stored.
+            At most 10 rows are returned per call (the limit is capped at 10 by the platform).
+            schemaName is required, since the same table name can exist in more than one schema; get it from platform_list_database_tables.
```
Steve-Mcl: is thought schemaName was to be optional (i.e. to be backwards compatible with the v1 API)?

forge/ee/lib/mcp/tools/teams.js
andypalmi: Fair question. If we drop platform_get_team_database, how would the agent get a databaseId in the first place, starting from just a teamId? Today the answer is platform_list_team_databases, which already includes the id - so the get-by-id tool doesn't add anything right now, since a team only ever has one database.

The reason I kept it is future-proofing: if a team can ever have more than one database, the agent would need a way to fetch one directly by id instead of always listing and filtering. Worth keeping now for that, or drop it and reintroduce it if/when multi-database support actually ships?

forge/ee/lib/mcp/tools/teams.js
andypalmi: Leaning toward keeping the tool bound rather than gating discovery or adding an extra check, that way the agent can see Tables exists and nudge the user toward enabling it, which could double as a soft upsell to a plan that includes it. The current behavior (pass the API's error straight through) would stay as-is. Does that match what you'd want, or is there a reason to gate discovery instead?

forge/ee/lib/mcp/tools/teams.js
andypalmi: Leaning toward keeping the tool bound rather than gating discovery or adding an extra check, that way the agent can see Tables exists and nudge the user toward enabling it, which could double as a soft upsell to a plan that includes it. The current behavior (pass the API's error straight through) would stay as-is. Does that match what you'd want, or is there a reason to gate discovery instead?

forge/ee/lib/mcp/tools/teams.js
andypalmi: Leaning toward keeping the tool bound rather than gating discovery or adding an extra check, that way the agent can see Tables exists and nudge the user toward enabling it, which could double as a soft upsell to a plan that includes it. The current behavior (pass the API's error straight through) would stay as-is. Does that match what you'd want, or is there a reason to gate discovery instead?

forge/ee/lib/mcp/tools/teams.js
andypalmi: Fair point, removed that line. The agent only needs the database id from this response, the credential-stripping detail was internal implementation and didn't need to be surfaced.

forge/ee/lib/mcp/tools/teams.js
andypalmi: Same call here, removed "Does not return row data or credentials" - the description already says it returns just the schema, so the extra clarification wasn't adding anything.

forge/ee/lib/mcp/tools/teams.js
andypalmi: Went with this, it now returns `{ databases: [...] }`. Also added a zod outputSchema describing the shape, and did the same across all five Tables tools, including platform_list_database_tables and platform_query_database_table_data, which already came back as objects.

forge/ee/lib/mcp/tools/teams.js
andypalmi: Went with this, it now returns `{ database: databaseId, tableName, schemaName, columns }`, using the values from the request rather than re-deriving them, since the underlying endpoint only returns the bare column array. Output schema added too.

forge/ee/lib/mcp/tools/teams.js
andypalmi: Table-level Tables tools only ship for FlowFuse 2.33+, since that's the release where the endpoints always return the table's schema (the earlier bug this fixes). So there's no live 2.32-and-earlier caller for these two tools to stay backward compatible with, mandatory schemaName was a deliberate choice to stop the agent from guessing the wrong table across schemas rather than a compatibility gap.

forge/ee/lib/mcp/tools/teams.js
andypalmi: Table-level Tables tools only ship for FlowFuse 2.33+, since that's the release where the endpoints always return the table's schema (the earlier bug this fixes). So there's no live 2.32-and-earlier caller for these two tools to stay backward compatible with, mandatory schemaName was a deliberate choice to stop the agent from guessing the wrong table across schemas rather than a compatibility gap.

Steve-Mcl: cool cool!

forge/ee/lib/mcp/tools/teams.js
Steve-Mcl: ok, fine with that, but what is returned when tables is not enabled? Is the response to LLM clear that this optional feature is not enabled for this team?  Thats what i mostly wanted (and why I said I favour option 1)

forge/ee/lib/mcp/tools/teams.js
Steve-Mcl: ok, fine with that, but what is returned when tables is not enabled? Is the response to LLM clear that this optional feature is not enabled for this team?  Thats what i mostly wanted (and why I said I favour option 1)

forge/ee/lib/mcp/tools/teams.js
Steve-Mcl: 👍

forge/ee/lib/mcp/tools/teams.js
Steve-Mcl: 👍

forge/ee/lib/mcp/tools/teams.js
Steve-Mcl: ok, fine with that, but what is returned when tables is not enabled? Is the response to LLM clear that this optional feature is not enabled for this team?  Thats what i mostly wanted (and why I said I favour option 1)

andypalmi: Now thanks to #7994  the returned message is `Not Found - FlowFuse Tables is not enabled for this team` which is more explicit.

forge/ee/lib/mcp/tools/teams.js
andypalmi: Now thanks to #7994  the returned message is `Not Found - FlowFuse Tables is not enabled for this team` which is more explicit.

forge/ee/lib/mcp/tools/teams.js
andypalmi: Now thanks to #7994  the returned message is `Not Found - FlowFuse Tables is not enabled for this team` which is more explicit.

forge/ee/lib/mcp/tool-schemas/tables.js
```diff
@@ -0,0 +1,24 @@
+const { z } = require('zod')
+
+// Shared field schemas reused across the FlowFuse Tables tools, since the same
+// database/table identifiers appear in both the input and output schemas.
+const teamIdSchema = z.string().describe('The ID or hashid of the team')
```
Steve-Mcl:
```suggestion
const teamIdSchema = z.string().describe('The hashid of the team')
```

forge/ee/lib/mcp/tools/tables.js
```diff
@@ -0,0 +1,157 @@
...
+            database: databaseIdSchema,
+            tableName: tableNameSchema,
+            schemaName: schemaNameSchema,
+            columns: z.array(z.object({
+                name: z.string(),
+                type: z.string()
+            }).loose())
+        },
+        handler: async (args, { inject }) => {
+            const url = `/api/v1/teams/${args.teamId}/databases/${args.databaseId}/tables/${args.tableName}/${encodeURIComponent(args.schemaName)}`
```
Steve-Mcl: should encode table name too?

forge/ee/lib/mcp/tools/tables.js
```diff
@@ -0,0 +1,157 @@
...
+            schemaName: schemaNameSchema,
+            limit: z.number().int().min(1).max(10).default(10).describe('Maximum number of rows to return (1-10, default 10)')
+        },
+        outputSchema: {
+            count: countSchema,
+            rows: z.array(recordSchema),
+            meta: recordSchema
+        },
+        handler: async (args, { inject }) => {
+            const url = `/api/v1/teams/${args.teamId}/databases/${args.databaseId}/tables/${args.tableName}/data/${encodeURIComponent(args.schemaName)}${args.limit !== undefined ? `?limit=${encodeURIComponent(args.limit)}` : ''}`
```
Steve-Mcl: encode table name?

Steve-Mcl (approved):

Steve-Mcl:
Tested locally - worked perfectly when combined with a skill in the expert.

PR #7999, opened 2026-07-29
feat(mcp): consolidate instance and device listing tools

### LINKED ISSUE #7919: Agent truncates results when listing large sets of instances
### Current Behavior

A prompt like "show me all the applications in this team" followed by "list the remote instances connected to each application" returned an incomplete response.

  - Teams with fewer than 5 instances were listed correctly
  - One team with ~79 hosted instances only had 5 listed, even though the API response from platform_get_application_remote_instances returned the correct total count
  - The tool response payload had a limit of 10, but only 5 were actually listed in the agent's output

  It looks like the agent is ignoring the total count and only working with the paginated response it received. Might be worth considering whether the tool should paginate through all results automatically, or whether the agent needs guidance to do so.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

- FlowFuse version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

andypalmi:
## Summary

Consolidates 4 overlapping platform automation tools into 2: `platform_list_hosted_instances` and `platform_list_remote_instances`. Each works both team-wide (paginated) and scoped to a single application.

- No effect on end users. This only changes the tools available to the assistant, not any user-facing API or UI.
- Fewer tool calls per chat. "What instances are running" previously needed a list call plus a separate status call; now one call returns both together. Team-wide listing also includes each instance's spec (instance type, stack, template), which the old endpoint never returned.
- Adds a `hostedInstanceId` filter to list the devices assigned to a specific hosted instance's device group. There was no way to do this before.

<details>
<summary>Details</summary>

**Removed:**
- `platform_get_application_hosted_instances`
- `platform_get_application_instances_status`
- `platform_get_application_remote_instances`
- `platform_list_team_remote_instances`

**Added:**
- `platform_list_hosted_instances` — `teamId` (required), optional `applicationId`, `query`, `state` (`running` / `error` / `notRunning`, matching the dashboard's Running/Error/Not Running filter), `includeLiveStatus`, `page`, `limit`
- `platform_list_remote_instances` — `teamId` (required), optional `applicationId`, `hostedInstanceId`, `query`, `mode` (`autonomous` / `developer`, matching the dashboard's Fleet Mode/Developer Mode filter), `page`, `limit`

Team-wide instance listing returns `{ count, meta: { page, pageSize, total, pageCount }, instances }`; application-scoped listing returns `{ count, instances }` with no pagination, since the underlying `/applications/:id/instances` endpoint has none. It also uses a lighter summary view with no instance type/stack/template, so application-scoped results don't include spec either. Unifying that would need a backend change and isn't part of this PR.

Remote instance listing returns the same `{ count, meta, devices }` shape at every scope (team, application, or a specific hosted instance's device group), since the devices API supports pagination consistently across all three.
</details>

## Test plan

- [x] `npx mocha test/unit/forge/ee/lib/mcp/tools/instances_spec.js test/unit/forge/ee/lib/mcp/tools/devices_spec.js` — 32/32 passing

<details>
<summary>Manually verified via chat</summary>

Asking naturally and checking both the arguments the assistant chose and the response shape:

**`platform_list_hosted_instances`**
- [x] "What hosted instances are running?" (team-wide) → `state: ["running"]`, `includeLiveStatus: true`
- [x] "Which instances are not running?" (team-wide) → `state: ["notRunning"]`
- [x] "Show me the instances that have crashed" (team-wide) → `state: ["error"]`
- [x] "Show me instances that are stopped or crashed" (team-wide) → `state: ["error", "notRunning"]`
- [x] "List the hosted instances in this application" (scoped to one application) → `applicationId` set, response has no pagination

**`platform_list_remote_instances`**
- [x] "What devices are in developer mode?" (team-wide) → `mode: "developer"`
- [x] "What devices are in developer mode in this application?" (scoped to one application) → `applicationId` + `mode: "developer"`
- [x] "Are any devices in fleet mode in this application?" (scoped to one application) → `applicationId` + `mode: "autonomous"`
- [x] "What devices are assigned to this instance?" (scoped to one hosted instance) → `hostedInstanceId` set
- [x] "List the devices in this team" (team-wide, no filters) → defaults to `page: 1`, `limit: 10`
</details>

Closes FlowFuse/engineering#184
Closes Flowfuse#7919


forge/ee/lib/mcp/tools/devices.js
```diff
@@ -2,39 +2,68 @@ const { z } = require('zod')
...
-            limit: z.number().min(1).max(10).describe('How many results to return per page')
+            mode: z.enum(['autonomous', 'developer']).optional()
+                .describe('Filter by mode: "autonomous" (Fleet Mode, running its assigned snapshot independently) or "developer" (Developer Mode, connected to the editor for live development). Matches the dashboard\'s own mode filter.'),
+            page: z.number().min(1).optional().describe('Page number to fetch (1-based). Defaults to 1.'),
+            limit: z.number().min(1).max(10).default(10).describe('How many results to return per page')
         },
         handler: async (args, { inject }) => {
-            let url = `/api/v1/teams/${args.teamId}/devices`
-            const params = []
+            const params = [`page=${args.page || 1}`, `limit=${args.limit || 10}`]
```
cstns: Lets use UrlSearch params or better yet the URL class for this logic, it's a pain too look at

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -2,39 +2,68 @@ const { z } = require('zod')
...
-            If you already know the application, use platform_get_application_remote_instances instead to get a narrower list.
+            Pass applicationId to list only the remote instances assigned to one application, or hostedInstanceId to list only the remote instances assigned to one hosted instance's device group. Omit both to list every remote instance in the team.
+            You can search by name using the query parameter, filter by mode ("autonomous", i.e. Fleet Mode, or "developer", i.e. Developer Mode), and page through results using page and limit.
             To get the full details of one specific remote instance, call platform_get_remote_instance with its ID.`,
         annotations: { readOnlyHint: true, destructiveHint: false },
         inputSchema: {
             teamId: z.string().describe('The ID or hashid of the team'),
+            applicationId: z.string().optional().describe('Restrict results to remote instances assigned to this application. Omit to list every remote instance in the team.'),
+            hostedInstanceId: z.string().optional()
```
cstns, line 18: wouldn't it be better to add a quick phrase on these input params to 'only use when you need to filter by..' or similar than adding `            Pass applicationId to list only the remote instances assigned to one application, or hostedInstanceId to list only the remote instances assigned to one hosted instance's device group. Omit both to list every remote instance in the team.`  in the tool description?

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -2,39 +2,68 @@ const { z } = require('zod')
...
-            if (params.length > 0) {
-                url += '?' + params.join('&')
+
+            const body = response.json()
+            const devices = (body.devices || []).map((device) => ({
+                id: device.id,
+                name: device.name,
+                ownerType: device.ownerType,
+                mode: device.mode,
+                status: device.status,
```
cstns: device status is the entry stored in the database (which is either the last known or desired state) not the actual live state status. That's why we have two api endpoints for it

forge/ee/lib/mcp/tools/instances.js
```diff
@@ -1,6 +1,48 @@
 const { z } = require('zod')
 
+// Mirrors the runningStates/errorStates/stoppedStates groups in frontend/src/composables/InstanceStates.js,
+// the same grouping the dashboard's own Running/Error/Not Running status filter uses (frontend/src/pages/team/Instances.vue).
+const STATE_GROUPS = {
```
cstns, line 5: we need a better way of doing this, portable across FE/BE to not mix up domain boundaries and avoid copy pasting (and ultimately maintaining) multiple entries of the same thing.

Not asking to do anything about it now, just saying.

forge/ee/lib/mcp/tools/instances.js
```diff
@@ -122,3 +164,84 @@ module.exports = [
...
+    }))
+
+    return {
+        statusCode: listResponse.statusCode,
+        json: () => ({ count: instances.length, instances })
+    }
+}
+
+async function listTeamHostedInstances (args, { inject }) {
+    const params = [`page=${args.page || 1}`, `limit=${args.limit || 10}`]
```
cstns: UrlSearch params or the URL class for these actions please

forge/ee/lib/mcp/tools/devices.js
cstns: we should add descriptions for this because it can confuse agents the same way it's confusing us

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -2,39 +2,68 @@ const { z } = require('zod')
...
+                onlineStatus: device.onlineStatus,
```
cstns: the status field on line 51 and onlineStatus on line 53 of the MCP tool response are two different stored/computed DB values, and neither is the
actual live state.

The distinction in the tool description for platform_get_remote_instance_status is accurate, but the list tool doesn't make this clear.

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -2,39 +2,68 @@ const { z } = require('zod')
...
             Remote instances are sometimes referred to as devices.
             A remote instance is a Node-RED that runs on the user's own hardware (like a Raspberry Pi or an edge server) rather than on the same environment as the FlowFuse platform.
-            Use this when you need to see all the remote instances a team has, regardless of which application they belong to.
-            If you already know the application, use platform_get_application_remote_instances instead to get a narrower list.
+            Pass applicationId to list only the remote instances assigned to one application, or hostedInstanceId to list only the remote instances assigned to one hosted instance's device group. Omit both to list every remote instance in the team.
+            You can search by name using the query parameter, filter by mode ("autonomous", i.e. Fleet Mode, or "developer", i.e. Developer Mode), and page through results using page and limit.
             To get the full details of one specific remote instance, call platform_get_remote_instance with its ID.`,
```
cstns, line 13: we should explain the following: `The status field reflects the last-known
or desired state stored on the platform. To get the actual live runtime state, call platform_get_remote_instance_status.`

forge/ee/lib/mcp/tools/devices.js
cstns: onlineStatus is not a live MQTT query. It's computed from lastSeenAt using a 30-minute threshold.

cstns: the `platform_get_remote_instance_status` tool is the only true retriever of a device's status; we should consider merging that in. we should talk about this entire thing as it's filled with nuances

cstns (changes requested):

cstns (approved):

PR #8019, opened 2026-07-31
Enforce cache key is a string

### LINKED ISSUE #7666: Enforce string keys on cache drivers
Enforce cache keys to be strings to avoid exceptions in production (by catching them in dev/testing time)

See below linked issue for reasons

> Clarified situation with @hardillb 
> Since pre-staging does not use redis, and the drivers behave differently there were 2 options came to mind:
> - switch to redis on pre-stagin (so we have parity and can catch future similar issues in testing
>   OR 
> - enforce string keys in both drivers for consistency 
> 
> The consensus (and economical choice) was to ensure operational consistency in the memory and redis cache drivers.
> 
> Issue to be raised. 

 _Originally posted by @Steve-Mcl in [#7654](https://github.com/FlowFuse/flowfuse/issues/7654#issuecomment-4876659163)_

Steve-Mcl:
### What does this PR do?

Makes the app's cache system reject non-string keys instead of silently misbehaving.

### Why was this needed?

The app can store its short-lived cache data in one of two ways: an in-memory cache (used locally and in some environments) or a Redis/Valkey cache (used in production). These two behave differently if you accidentally use something other than a plain string as a cache key, e.g. a number instead of a string. The in-memory cache will quietly accept it and appear to work fine, but Redis handles keys differently and can fail in confusing ways.

This mismatch already caused a hard-to-track-down bug: something worked fine locally but broke in an environment using Redis, and it took a long time to figure out why, because the local setup didn't use Redis and couldn't have caught it.

### What changed?

Both cache drivers now check that a key is a string before using it, and immediately throw a clear, descriptive error if it isn't. This means:

- The same mistake now fails the same way everywhere, whether running locally or in production.
- Problems like this get caught early, during development or testing, instead of surfacing later as a confusing production issue.

### Testing

Added tests to confirm both cache drivers reject non-string keys consistently.

## Related Issue(s)

closes #7666 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb (approved):

hardillb:
@Steve-Mcl I've hit the update branch button and then enabled auto-merge, disable if you disagree

PR #8024, opened 2026-08-03
Re-enable the documentation checks

dimitrieh:
Deletes the two `if: false` lines, so the doc link checks run again and broken links get caught before publishing.

**Merge after https://github.com/FlowFuse/website/pull/5473.** `test-with-website` needs that PR in order to build a docs PR against its own docs. Before it, this fails on `Missing script: "docs"`.

Should also restore the `needs` on `publish` that #8029 drops.

`test` is blocked by something unrelated and repo-wide: `npm ci` rejects the lockfile with `lock file's globals@17.8.0 does not satisfy globals@17.9.0`. `Build and contenerize` fails the same way.

Also fixes five broken `(optional)` anchors in `docs/`. An unescaped `(` ends a markdown link target, so they point at nothing. They are also what makes this PR's own checks run, since the workflow only triggers on `docs/**`.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,7 +9,6 @@ on:
       - 'docs/**'
 jobs:
   test:
-    if: false
     runs-on: ubuntu-latest
     name: Test Documentation links
     steps:
@@ -24,7 +23,6 @@ jobs:
       - name: Test Documentation Links
         run: npm run test:docs
   test-with-website:
-    if: false
     runs-on: ubuntu-latest
     name: Test Documentation with website
     steps:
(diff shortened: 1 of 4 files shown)


dimitrieh:
comment with context at https://github.com/FlowFuse/website/issues/5432#issuecomment-5164751264

hardillb:
@dimitrieh we can't just re-enable them, those steps need fixing first to use the new website process.

dimitrieh:
True, this needs to be merged in order after https://github.com/FlowFuse/website/pull/5473 as defined in the comment and in the body of this pr

ZJvandeWeg (approved):

hardillb (approved):

PR #8030, opened 2026-08-03
[7652] Data Farm > Applications 3 - subscriber

n-lark:
## Description

Adds an applications subscriber routing `a/+/created|updated|deleted` MQTT events into the data-farm store, so cross-session `create`/`rename`/`delete` reflect in the list without a reload. 

**Known limitation**: a newly-appeared card's count badges stay blank until refresh (no summary counts in the payload) - this will be addressed when we farm the data for hosted/remote instances. 

See test plan: https://github.com/FlowFuse/flowfuse/issues/7652#issuecomment-5168411248

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7652

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production
 - [ ] Link to Changelog Entry PR, or note why one is not needed.

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

PR #8052, opened 2026-08-03
[7839] Data Farm > Applications 4 - remaining vue + loaders

### LINKED ISSUE #7839: Applications 4 - remaining vue + loaders
## PR — remaining vue + loaders

- `pages/application/index.vue` + `mixins/Application.js` read the store (`activeApplication`; `context.application` becomes a pass-through like `context.team`) instead of local fetch; add `activeApplicationId`/`activeApplication`/`setActiveApplication`.
- Page Loader for the application detail; Component Loaders for sub-widgets (e.g. the instance table) so the shell stays interactive.
- Delete dead fetching in `mixins/Application.js`.
- Instances still load via the existing path until the Hosted Instances store lands; add the joined getter then.

- [ ] Single-application views consume the store.
- [ ] Page/Component Loaders during load.

n-lark:
## Description

See test plan: https://github.com/FlowFuse/flowfuse/issues/7839#issuecomment-5169986098

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/7839

- Single-application views now read the app from the `data-farm` store instead of local fetch.
- Removes `mixins/Application.js`, replaced by a `useActiveApplication` composable.
- Detail-page Page Loader is now a shell skeleton `ApplicationDetailSkeleton`.

The loading work for the ux store will be done in a follow up - Applications 5 - UXStore Loader Registry https://github.com/FlowFuse/flowfuse/issues/8053

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


andypalmi:
Nice migration, the `context.application` getter deriving from the store is much cleaner than the old mixin watch.

One thing on `createInstance.vue`: the breadcrumb at lines 7-8 reads `application.id` and `application.name` directly, but `application` now comes from the store's `activeApplication`, which is `null` until `loadActiveApplication` resolves in `created` (the old mixin defaulted it to `{}`). That breadcrumb sits in the always-rendered header and is only guarded by `v-if="team"`, and `team` is usually already present from session storage, so could it throw during the first render before the app loads? Line 28 in the same file already uses `application?.id`, so mirroring that (`application?.id` / `application?.name`, or `v-if="team && application"`) would probably cover it. `index.vue` and `CreateInstanceMultiStep.vue` already look guarded.

Would it be worth a small render test for that page with an unloaded store, just so this path is covered? The store additions themselves have good coverage.

andypalmi:
Two small non-blocking things for possible follow-up:

**Stale app data in the list cache.** `context.setApplication` now delegates to `setActiveApplication`, which merges its argument into `applicationsById`. That argument now includes the app object nested in an instance or device (`setInstance`/`setDevice` pass `instance.application` / `device.application`). If that nested object is partial or stale relative to the list, opening an instance would overwrite the list card's `name`/`description` with the instance payload's values until the next full refetch. Is `instance.application` guaranteed to carry the current, complete name and description, or could this surface stale values on the Applications list?

**Delete-while-viewing, as a seam with #8030.** If a `deleted` event arrives for the app you're currently on, `removeApplication` drops it from `applicationsById`, so `activeApplication` becomes null and `index.vue`'s `!application?.id` guard flips the page back to the loading skeleton permanently. Is that the intended outcome, or should a delete-while-viewing redirect to the Applications list (or show a "removed" state) instead of an indefinite skeleton?

n-lark:
Hey @andypalmi ty for the review - I guarded the breadcrumb with `v-if="team && application"`, and added a watcher so a delete redirects to the Applications list with a notice instead of stranding on the skeleton.

frontend/src/composables/useActiveApplication.ts
```diff
@@ -0,0 +1,33 @@
...
+            })
+            return null
+        }
+    }
+
+    function clearActiveApplication (): void {
+        applicationsStore.setActiveApplication(null)
+    }
+
+    return { application: activeApplication, isLoadingActiveApplication, loadActiveApplication, clearActiveApplication }
```
n-lark: In https://github.com/FlowFuse/flowfuse/pull/8063 I removed these since they are just a pass through to the store, unnecessary duplicated state. 

```
application: activeApplication, 
isLoadingActiveApplication, 
clearActiveApplication
```

andypalmi (approved):
Both fixes look great. The `v-if="team && application"` guard resolves the breadcrumb null-deref, and the delete watcher is nicely guarded — `previous && !application` keeps it off the initial load, and `!this.loading.deleting` cleanly separates a remote delete from a self-initiated one so there's no double alert or double redirect. The redirect-to-Applications-with-notice is a good seam with #8030's realtime path.

Approving. Two optional follow-ups, neither blocking merge:
- `setActiveApplication`'s `{...existing, ...application}` merge can still let a partial `instance.application`/`device.application` overwrite a list card's present keys until the next full refetch — worth a look at some point, but fine to defer.
- The store additions have good coverage; if it's ever convenient, a small render test for `createInstance.vue` with an unloaded store (and one for the delete watcher) would lock in these two paths.

cstns:
I would have preferred to hold off on the skeleton loaders a bit, because there are some prerequisites that depend on them.

1. Clicking an application from the applications list flashes an empty state container between the skeleton loader and the loaded application.
   - Throttling to GPRS shows this clearly.

   https://github.com/user-attachments/assets/8d182037-840f-4529-a09b-a814eb0f0fa8

1. Switching between tabs triggers the skeleton for the tabs section too, even though the tabs are static and don't change.
   - The `SectionNavigationHeader` skeleton should be independent of the rest of the page.
   - the rest of the page should be within its own dedicated wrapper as well

   https://github.com/user-attachments/assets/cae41661-4c4c-477d-aa93-81d43234d245

1. Switching to the remote instances page:
   - Displays the skeleton loader while loading.
   - Finishes loading and hides the skeleton, but then shows the old loader while the devices are loading.
   - It should keep showing the skeleton until the devices API call completes.
   - We should drop the old loader in favor of the skeleton.

   https://github.com/user-attachments/assets/629b3996-8cdc-470b-a9e6-84f05448ec82

1. Switching to the hosted instances page:
   - Same as the first point, it flashes the empty state container.
   - There's no reason to keep the empty state container while loading; it should only appear when there actually are no instances.

1. Double loader:
   - Dashboard, device groups, snapshots, and pipelines all have an intermediate stage where they show the old loader while gathering data.

1. The skeleton loader doesn't match the page structure:
   - This applies to the pipelines, logs, audit log, and settings pages.
   - Not all application pages are a datatable list.

1. Device group pages:
   - These pages don't match the format of the applications page.
   - They don't occupy the full available height and have a gray underlay.

Fixing all of this would extend the scope of this task, which is why I would have preferred to go with the page loader approach first and tackle these as a follow-up feature focused on skeleton loaders. That would also have given us a chance to identify these niche cases up front and land on a cohesive approach when rolling skeleton loaders out, not just for applications, but across the other entities and pages in the app as well.

frontend/src/mixins/Application.js
```diff
@@ -1,113 +0,0 @@
-import { mapActions } from 'pinia'
-
-import ApplicationApi from '../api/application.js'
-import alerts from '../services/alerts.js'
-
-import { useContextStore } from '@/stores/context.js'
-import { useDataFarmApplicationsStore } from '@/stores/data-farm-applications'
-
-export default {
```
cstns, line 9: love seeing mixins disappear

frontend/src/pages/application/index.vue
```diff
@@ -35,38 +35,68 @@
...
+            isLoadingActiveApplication,
+            loadActiveApplication,
+            clearActiveApplication,
+            deleteApplicationEntity: applicationsStore.deleteApplication
+        }
+    },
+    data () {
+        return {
+            applicationInstances: new Map(),
+            loading: {
```
cstns: We shouldn't move this forward, even less so now with the skeleton loader added.

I also don't see how an application could be suspended in the first place; this may have been copy-pasted from somewhere else, which is another reason to hold off.

The cleanest approach would be to simply redirect to the applications list page after a deletion. The MQTT subscriber will take care of removing the app from the list once the operation succeeds.

frontend/src/pages/application/index.vue
```diff
@@ -17,7 +17,7 @@
                 :application="application"
                 :instances="instancesArray"
```
cstns, line 15: can't wait for the instances data farm

frontend/src/pages/application/index.vue
```diff
@@ -141,12 +171,62 @@ export default {
...
+                this.$router.push({ name: 'Applications', params: { team_slug: this.team?.slug } })
+            }
+        },
         liveInstanceMetadata: { handler: 'applyLiveStatus', deep: true }
     },
+    beforeUnmount () {
+        this.clearActiveApplication()
+    },
     methods: {
+        async loadApplicationData () {
```
cstns, line 201: shouldn't these be in the application data-farm?

cstns: or are they moved in a follow-up? I didn't read the follow-ups yet

cstns (changes requested):

cstns:
I stand corrected, I just noticed that you removed the skeleton loaders in the follow-up

frontend/src/pages/application/index.vue
n-lark: The app load already uses the store `loadActiveApplication`. The instances will get moved under the hosted instances work https://github.com/FlowFuse/flowfuse/issues/8076

n-lark:
Hey @cstns! For [this comment](https://github.com/FlowFuse/flowfuse/pull/8052#issuecomment-5216274165) the Page Loader PR is merged in, dropping skeletons entirely, so the skeleton-specific points no longer apply.

The rest is pre-existing behavior in files this PR doesn't touch, handled in the per-entity loader follow-ups:

1. Double loaders (dashboards, device groups, snapshots, pipelines, remote instances) - each tab loads its own data, the page loader only gates the application.
2. Hosted instances empty-state flash - instances load after the app hydrates, same as before this PR.
3. Device group layout (height / gray underlay) - that page is untouched here.

These land as each child entity gets its own store + loader. This PR just swaps the old per-page ff-loading for the shared page loader. If there is anything you specifically want me to address here please lmk!

frontend/src/pages/application/index.vue
```diff
@@ -58,16 +59,43 @@ export default {
...
+            hasPermission,
+            isVisitingAdmin,
+            loadActiveApplication,
+            deleteApplicationEntity: applicationsStore.deleteApplication
+        }
+    },
+    data () {
+        return {
+            applicationInstances: new Map(),
+            loadingInstanceStatuses: false
```
n-lark, line 78: Will address this under #8076

cstns (approved):

PR #8069, opened 2026-08-05
docs: declare sidebar group order in section frontmatter

dimitrieh:
The docs sidebar groups sections under a `navGroup` heading, but the order of those headings is a hardcoded array in the website repo (`GROUP_ORDER` in `nuxt/composables/useDocsNav.ts`), even though the group names themselves are declared here. Adding, renaming or reordering a group needed a website change.

This adds `navGroupOrder` alongside each `navGroup`. The website ranks a heading by the lowest `navGroupOrder` of its sections, so the whole docs structure (paths, page order, group names, group order) now lives in this repo.

Values reproduce the order the site renders today. Merge this before the website side, which drops the array: https://github.com/FlowFuse/website/pull/5513

### DIFF
--- a/docs/admin/README.md
+++ b/docs/admin/README.md
@@ -1,5 +1,6 @@
 ---
 navGroup: FlowFuse Self-Hosted
+navGroupOrder: 4
 navOrder: 4
 navTitle: Administering FlowFuse
 redirect:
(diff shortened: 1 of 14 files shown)


dimitrieh:
@hardillb this is about moving logic of how the navigation is defined from the website to the documentation so there is an SSOT. With the docs to nuxt migration this got tangled up, and currently part of the nav is hardcoded in https://github.com/FlowFuse/website.

After this PR merges, we can merge https://github.com/FlowFuse/website/pull/5513

hardillb (approved):

PR #8070, opened 2026-08-05
Preview the website on documentation pull requests

dimitrieh:
Contributes to [Website preview on documentation pull requests](https://github.com/FlowFuse/engineering/issues/237).

Replaces the copy of the website build in `docs.yml` with a call to `build_website/v1` in FlowFuse/github-actions-workflows, and uploads the result to Netlify under a `docs-pr-<number>` alias so documentation pull requests get a preview URL.

The deploy is aliased and unpublished, so netlify.toml's build command never runs against it and flowfuse.com keeps serving what it served before. Fork pull requests receive no secrets, so they build and link-check without uploading.

Draft until:

- [ ] `NETLIFY_AUTH_TOKEN` and `NETLIFY_SITE_ID` are available to this repository
- [ ] main's required checks name `Build the website with these docs / Build and check` in place of `Test Documentation with website`


dimitrieh:
Context at https://github.com/FlowFuse/engineering/issues/237

dimitrieh:
Slack thread: https://flowfuse.slack.com/archives/C032Q63FGG1/p1785930190689399

.github/workflows/docs.yml
```diff
@@ -22,56 +31,41 @@ jobs:
...
+      preview_alias: docs-pr-${{ github.event.pull_request.number }}
+      pr_number: ${{ github.event.pull_request.number || 0 }}
+    secrets:
+      ci_app_id: ${{ secrets.GH_BOT_APP_ID }}
+      ci_app_key: ${{ secrets.GH_BOT_APP_KEY }}
+      netlify_auth_token: ${{ secrets.NETLIFY_AUTH_TOKEN }}
+      netlify_site_id: ${{ secrets.NETLIFY_SITE_ID }}
+  # A reusable workflow reports its checks as "caller job / called job". This gate keeps the
+  # original context name and the publish dependency below intact.
   test-with-website:
```
ppawlowski: What is the purpose of the `test-with-website` job now?

.github/workflows/docs.yml
```diff
@@ -7,6 +7,15 @@ on:
...
     paths:
       - 'docs/**'
+      # A change to this workflow should be exercised by this workflow.
+      - '.github/workflows/docs.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number || github.ref }}
+  # Only the newest push to a pull request is worth previewing. Pushes to main are never
+  # cancelled, so a publish cannot be interrupted by the next docs merge.
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
ppawlowski, line 17:
```suggestion
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

ppawlowski (changes requested):

.github/workflows/docs.yml
dimitrieh: It keeps the `Test Documentation with website` check name reporting. A reusable workflow reports as "caller job / called job", so the build shows up as `Build the website with these docs / Build and check` and the old context stops existing. The required context on main would then be a name nothing publishes, and `publish`'s `needs` would break too.

The alternative is to update the required contexts on main and point `publish` at `needs: website`, and then this job can go. Both would have to land together.

I'd keep it: one job, a few seconds, and no ruleset change tied to the merge. Let me know if you see it differently, you have the view on the protection settings that I don't.

.github/workflows/docs.yml
dimitrieh: Good catch, `github.ref` is already unique per pull request (`refs/pull/<n>/merge`), so the number prefix added nothing. Committed.

dimitrieh:
thread at https://flowfuse.slack.com/archives/C032Q63FGG1/p1786701915790499?thread_ts=1786685810.202179&cid=C032Q63FGG1

.github/workflows/docs.yml
ppawlowski: Please remove it and adjust the job's dependency - we do not need to spin another runner just to report the status of a previous job.

dimitrieh: Removed, and `publish` now needs `website` directly. Default `needs` semantics already stop `publish` on a failed or skipped build, so the gate's `always()` plus result check was only ever buying the name.

That name goes with it: the build reports as `Build the website with these docs / Build and check` and `Test Documentation with website` stops existing. If the old context is still in main's required checks it has to be swapped before this merges, otherwise main waits on a context nothing publishes. I cannot read the protection settings, so that part is on your side.

dimitrieh: @ppawlowski can you check this?

ZJvandeWeg:
@dimitrieh Seems convoluted, shouldn't we just move the docs to the website repo?

dimitrieh:
@knolleary Steps still needed: https://flowfuse.slack.com/archives/C032Q63FGG1/p1786707548693539?thread_ts=1786685810.202179&cid=C032Q63FGG1

PR #8071, opened 2026-08-05
docs: document the project node limitations

dimitrieh:
The page presents calling a shared instance's flows as the headline use case for these nodes, but states none of the constraints a builder hits.

Adds a Limitations section covering:
- only hosted instances can be named as a target or a specific source
- a remote instance can send and receive broadcasts, but can never be sent to directly, so shared logic has to live on a hosted instance
- a call waits 30 seconds for a target that must be running with a matching `Project In` node, and travels through the broker, so it is unavailable to a caller that is offline
- load is shared between copies only in HA mode, via shared subscriptions

Also fixes "links to extended" and the implication that remote instances are symmetric peers.

Paired with FlowFuse/nr-project-nodes#255, which puts the same facts in the node help and the error messages.

### DIFF
--- a/docs/user/projectnodes.md
+++ b/docs/user/projectnodes.md
@@ -9,12 +9,15 @@ very quick and easy to securely send and receive messages between different
 instances in a team.
 
 The nodes act in a similar way to the Node-RED Link nodes, but by allowing the
-links to extended between different instances and devices, they open up a wide
-range of possibilities.
+links to extend between different instances, they open up a wide range of
+possibilities. Remote instances can take part too, with the limitations described
+below.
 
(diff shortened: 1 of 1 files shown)


dimitrieh:
Small fix, based on research done into link nodes resuling in that we can better inform users of link nodes needing logic to reside on hosted instances

Slack thread: https://flowfuse.slack.com/archives/C03FAUBPPJ7/p1785930859746159?thread_ts=1785861957.912329&cid=C03FAUBPPJ7

hardillb (approved):

sumitshinde-84 (approved):

Issue #8092, opened 2026-08-06
Rename application route names to kebab-case


cstns:
Part of #8084

## What

Rename all 16 application route names from PascalCase to kebab-case.

| Current | New |
|---------|-----|
| `Application` | `application` |
| `ApplicationInstances` | `application-instances` |
| `ApplicationDashboards` | `application-dashboards` |
| `ApplicationDevices` | `application-devices` |
| `ApplicationDeviceGroups` | `application-device-groups` |
| `ApplicationSnapshots` | `application-snapshots` |
| `ApplicationPipelines` | `application-pipelines` |
| `CreatePipeline` | `application-pipeline-create` |
| `EditPipeline` | `application-pipeline-edit` |
| `CreatePipelineStage` | `application-pipeline-stage-create` |
| `EditPipelineStage` | `application-pipeline-stage-edit` |
| `ApplicationDeviceGroupIndex` | `application-device-group` |
| `ApplicationDeviceGroupDevices` | `application-device-group-devices` |
| `ApplicationDeviceGroupSettings` | `application-device-group-settings` |
| `ApplicationDeviceGroupSettingsGeneral` | `application-device-group-settings-general` |
| `ApplicationDeviceGroupSettingsEnvironment` | `application-device-group-settings-environment` |

## Where to look

Route definitions in `frontend/src/pages/application/routes.js`.

This is the largest batch (~95 refs across ~30 files), but highly cohesive. The bulk of references are within `pages/application/` and closely related components.

Key files outside the application directory:
- `pages/instance/index.vue` (link to `Application`)
- `pages/device/Overview.vue`, `pages/device/Settings/General.vue` (links to `Application`, `ApplicationDeviceGroupIndex`, `ApplicationDevices`)
- `pages/team/components/ApplicationSummaryLabel.vue` (links to `ApplicationInstances`, `ApplicationDevices`, `ApplicationDeviceGroups`, `ApplicationSnapshots`, `ApplicationPipelines`)
- `pages/team/Applications/components/compact/InstancesWrapper.vue`, `DevicesWrapper.vue`
- `pages/team/Applications/components/ApplicationHeader.vue`
- `pages/team/Pipelines/components/TeamPipeline.vue`, `TeamPipelineStage.vue`
- `pages/team/DeviceGroups/index.vue`
- `components/audit-log/AuditEntry.vue`
- `components/bill-of-materials/InstancesItem.vue`
- `components/global-search/components/ResultSection.vue`
[...]

Issue #8085, opened 2026-08-06
Rename account route names to kebab-case


cstns:
Part of #8084

## What

Rename 5 account-related route names from mixed casing to kebab-case.

| Current | New |
|---------|-----|
| `Sign up` | `sign-up` |
| `PasswordReset` | `password-reset` |
| `VerifyPendingEmailChange` | `verify-pending-email-change` |
| `ForgotPassword` | `forgot-password` |
| `Sign out` | `sign-out` |

## Where to look

Route definitions in `frontend/src/pages/account/routes.js`.

References:
- `mixins/Navigation.js` (signOut method uses `Sign out`)
- `pages/Login.vue` (link to `ForgotPassword`)

This is the smallest blast radius batch, mostly self-contained within the account directory.

## Verification

After renaming, grep the codebase for each old name as a quoted string to make sure nothing was missed.

Issue #8084, opened 2026-08-06
Standardize Vue Router route names to kebab-case


cstns:
## What

All our Vue Router route names should follow kebab-case (`team-members`, `application-overview`, etc.). Right now about 43 of ~133 route names use PascalCase (`ApplicationInstances`), title case with spaces (`Admin Settings`, `Sign up`), or bare words (`Home`, `Billing`). We want to clean this up before rolling out routes as MCP resources.

## Why

We're planning to expose Vue routes as MCP resources. Having consistent, predictable naming is a prerequisite for that. Beyond MCP, it just makes the codebase easier to work with when every route name follows the same pattern.

## How

The work is broken down into sub-issues, ordered by blast radius (smallest changes first, biggest last). Each sub-issue is a self-contained PR that renames a batch of routes and updates all references.

After each PR, grep for the old route name strings to confirm nothing was missed.

The final sub-issue adds a lint rule to prevent regressions.

**Out of scope:** Vue component `defineOptions({ name })` and component `name:` properties are left as-is. We're only changing the route `name` values and their references.

## Sub-issues

- [x] #8085 — Rename account route names to kebab-case
- [x] #8086 — Rename help, admin parent, and user settings route names to kebab-case
- [x] #8087 — Rename team library route names to kebab-case
- [x] #8088 — Rename low-ref team route names to kebab-case
- [x] #8089 — Rename team action route names to kebab-case
- [x] #8090 — Rename Home and team navigation route names to kebab-case
- [x] #8091 — Rename Team and Applications route names to kebab-case
- [x] #8092 — Rename application route names to kebab-case
- [x] #8093 — Add lint rule to enforce kebab-case route names

## Naming conventions

Routes follow a `{domain}-{page}-{action}` pattern:
- Domain prefix: `team-`, `application-`, `instance-`, `device-`, `admin-`, `user-`
- Nesting with hyphens: `instance-settings-general`, `team-members-invitations`
- Actions as suffixes: `-create`, `-edit`, `-view`

Notable naming decisions:
- `Instances` becomes `team-hosted-instances` (matches the "Hosted Instances" nav label, distinct from remote instances)
- `TeamDevices` becomes `team-remote-instances` (matches the "Remote Instances" nav label)
- `Admin Settings` becomes `admin` (the child `admin-settings` already exists)
- `LibraryTeamLibrary` becomes `team-library-files`
- `ApplicationDeviceGroupIndex` becomes `application-device-group` (drop the "Index" suffix)

PR #8107, opened 2026-08-06
feat: add lint rule to enforce kebab-case route names

### LINKED ISSUE #8093: Add lint rule to enforce kebab-case route names
Part of #8084

## What

Add a lint rule (or custom lint script) that enforces all Vue Router route `name` values match the kebab-case pattern: `/^[a-z][a-z0-9]*(-[a-z0-9]+)*$/`.

This should run as part of `npm run lint` so it catches regressions in CI.

## Why

After the full route rename effort (#8084), we want to make sure no one accidentally introduces a PascalCase or spaced route name again.

## Approach

Might be worth considering a small custom ESLint rule, a vitest snapshot test that reads all registered routes, or a standalone lint script. Whatever fits best with the existing tooling.

cstns:
Closes #8093
Part of #8084

## Summary

* Adds a vitest test that scans all `routes.js` files under `frontend/src/` and validates every `name:` value matches the kebab-case pattern (`/^[a-z][a-z0-9]*(-[a-z0-9]+)*$/`)
* Runs as part of `npm run test:unit:frontend`, so CI will catch any non-kebab-case route names going forward
* Uses static file analysis (reads files as text, extracts names via regex) to avoid the complexity of importing Vue components in the test environment

## Note

This PR should be merged after all the rename PRs (#8085 through #8092) are merged, since the test will fail until all route names are kebab-case.

## Test plan

- [ ] `npx vitest run test/unit/frontend/routes/route-naming.spec.js --config config/vitest.config.ts` fails on `main` (expected, names not renamed yet)
- [ ] After all rename PRs are merged, the test passes
- [ ] Adding a PascalCase route name causes the test to fail


n-lark (approved):
I pushed a fix for the linter issues - looks good

Issue #8076, opened 2026-08-05
Entity: Hosted Instances — data-farm


n-lark:
# Entity: Hosted Instances — data-farm

**Parent:** Data Farm · **Depends on:** Applications pilot (store shape, subscriber file, `notifyEntityLifecycle` helper, Page/Component Loaders).

Hosted instances = "projects" (`p/` namespace). Status already flows into `live-status`; entity data does not. Reuses the Applications pattern — this doc notes where instances differ: three-source reconciliation, the entity↔status join, pagination, and (unlike Applications, which left secondary callers direct) the store owning **all** instance reads.

Four PRs (shape in 00): backend · store+consumer · subscriber · remaining-vue. Backend is independent; the rest depend on the store.

## Current state

- `mixins/Instance.js` fetches `getInstance()` in `created()` + optimistic `InstanceStateMutator`.
- `pages/team/Instances.vue` watches `team`, calls `getInstances()`, fetches statuses separately; polls (`InstanceStatusPolling`) when `!statusChannelLive`.
- Status is live (`p/+/state` → `live-status`); entity CRUD is not.
- `context.setInstance` cascades the owning application/device.
- Three sources of truth: API, `live-status`, optimistic mutator.

## Store: `stores/data-farm-hosted-instances.ts`

Route-lazy + team-scoped like Applications → list load-state drives a non-blocking Page Loader. Normalized (`byId`) — unlike Teams' flat list — because realtime CRUD and detail lookups are by id. **Single source for all instance reads:** every fetch (table, detail, pickers, dialogs) upserts into `instancesById`; views select/filter from it instead of fetching their own copy. Owns the entity↔status join (getters merge `live-status.instanceStatuses[id]`), and folds the optimistic mutator in so all three sources reconcile in one place.

| Kind | Name | Caller | PR |
| --- | --- | --- | --- |
| state | `instancesById` | normalized map | store |
| state | `teamInstanceIds` | loaded team's list (order + pagination) | store |
| state | `loadedTeamId` | ensure-guard | store |
| state | `isLoadingTeamInstances` | Page Loader | store |
| state | `activeInstanceId` | route's instance | remaining-vue |
| getter | `teamInstances` | Instances table | store |
| getter | `teamInstancesWithStatus` / `instanceWithStatus(id)` | joins `live-status` | store |
| getter | `activeInstance` | `context.instance` pass-through | remaining-vue |
[...]

PR #8111, opened 2026-08-06
[8110] Carried-over status search term breaks pagination on Devices and Application instance/device lists

### LINKED ISSUE #8110: Carried-over status search term breaks pagination on Devices and Application instance/device lists
# Carried-over status search term breaks pagination on Devices and Application instance/device lists 

Follow-up to #7849 / #7925, which only fixed the team **Hosted Instances** list. The same broken deep-link pattern is still present in three other places: a status count navigates with a pipe-separated status string stuffed into `searchQuery` (name search), so the target list returns "No Data Found" while the pagination footer still shows the unfiltered totals.

Affected call sites:

- **Home → Remote Instances tile** — `frontend/src/pages/team/Home/index.vue:251` → `TeamDevices` with `searchQuery: states.join(' | ')`
- **Application → Hosted Instances counters** — `frontend/src/pages/team/Applications/components/compact/InstancesWrapper.vue` → `ApplicationInstances`
- **Application → Remote Instances counters** — `frontend/src/pages/team/Applications/components/compact/DevicesWrapper.vue` → `ApplicationDevices`

**Expected:** clicking a status count lands on a list actually filtered to those states, with pagination reflecting the filtered result set.

**Fix:** apply the #7925 approach — a server-side `state` filter (status query param + backend live-state resolution) plus a status filter control — to `TeamDevices`, `ApplicationInstances`, and `ApplicationDevices`.

n-lark:
## Description

See test plan: https://github.com/FlowFuse/flowfuse/issues/8110#issuecomment-5208482101

- Status tiles/counters now deep-link with `?status=<group>` instead of the pipe search term that broke filtering + pagination.
- Devices: status filter added to the Filters dropdown (combines with mode, shared count); backend accepts a state group.
- App hosted instances: status filter + skeleton while statuses load.

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/8110

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/mixins/Application.js
+++ b/frontend/src/mixins/Application.js
@@ -11,6 +11,7 @@ export default {
         return {
             application: {},
             applicationInstances: new Map(),
+            loadingInstanceStatuses: false,
             loading: {
                 deleting: false,
                 suspend: false
@@ -62,6 +63,7 @@ export default {
                 })
 
                 // Not waited for, as loading status is slightly slower
+                this.loadingInstanceStatuses = true
                 ApplicationApi
                     .getApplicationInstancesStatuses(applicationId)
                     .then((instanceStatuses) => {
@@ -74,6 +76,9 @@ export default {
                     })
                     .catch((err) => {
                         console.error(err)
+                    })
+                    .finally(() => {
+                        this.loadingInstanceStatuses = false
                     })
             } catch (err) {
                 this.$router.push({
--- a/frontend/src/pages/application/Overview.vue
+++ b/frontend/src/pages/application/Overview.vue
@@ -33,6 +33,8 @@
                 :search="searchTerm"
                 search-placeholder="Search Instances"
                 :rows-selectable="true"
+                :loading="tableLoading"
+                loading-type="skeleton"
                 @row-selected="selectedCloudRow"
             >
                 <template #actions>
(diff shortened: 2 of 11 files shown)


frontend/src/mixins/Application.js
```diff
@@ -11,6 +11,7 @@ export default {
         return {
             application: {},
             applicationInstances: new Map(),
+            loadingInstanceStatuses: false,
```
n-lark, line 14: This is a temporary work around for preventing an empty list when we deep link a status filter into the applications instances list. 

The status comes in a separate query so it showed `no data found` until we loaded. Will address in a more proper way under https://github.com/FlowFuse/flowfuse/issues/8076

frontend/src/pages/application/Overview.vue
```diff
@@ -186,8 +229,19 @@ export default {
                 { label: '', component: { is: markRaw(InstanceEditorLinkCell), map: { instance: '_self' } } }
             ]
         },
+        statusFilter () {
```
cstns, line 232: Thinking out loud here, we might want to formalize this pattern at some point. I feel like I've seen it scattered across components before. Not going to block this on it though.

cstns (approved):
Looking good!

I created https://github.com/FlowFuse/flowfuse/issues/8118 as a natural follow-up for when we'll have some wiggle room

PR #8151, opened 2026-08-07
[8080] Data Farm >  Hosted Instances 2 - store + first consumer

### LINKED ISSUE #8080: Hosted Instances 2 - store + first consumer
## PR — store + first consumer

- `pages/team/Instances.vue` reads the store via `ensureTeamInstancesLoaded(teamId)` behind a Page Loader; drop local list + status re-fetch. Keep the polling fallback when `!live-status.live`.
- Route create/delete + lifecycle through the store (`upsert`/`remove`, optimistic reconciled) — **fixes the same-session bug.**

- [x] Store + vitest tests.
- [x] Table reads store behind a Page Loader; polling fallback preserved.
- [x] Optimistic reconciled in the store, not the mixin.
- [x] Status via the join getter, not re-fetched.

n-lark:
## Description

See test plan https://github.com/FlowFuse/flowfuse/issues/8080#issuecomment-5220257986

## Related Issue(s)

Resolves https://github.com/FlowFuse/flowfuse/issues/8080

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/pages/team/Instances.vue
```diff
@@ -356,36 +366,20 @@ export default {
             this.abortController = controller
             this.loading = true
```
n-lark, line 367: Had to keep unique loading state for this component since it fires when pagination/sorting/filters/etc fires. 

We don't want this on the `ff-page` since it would reload the entire screen each time the pagination/sorting/filters/etc changed.

frontend/src/pages/team/Instances.vue
```diff
@@ -176,28 +176,25 @@
             </template>
         </div>
         <template v-if="!statusChannelLive">
-            <InstanceStatusPolling v-for="instance in instances" :key="instance.id" :instance="instance" @instance-updated="instanceUpdated" />
+            <InstanceStatusPolling v-for="instance in currentPageInstanceRefs" :key="instance.id" :instance="instance" @instance-updated="instanceUpdated" />
```
n-lark, line 179: Had to add `currentPageInstanceRefs` for the polling loop. 

`InstanceStatusPolling` watches its instance prop by reference, and the decorated rows are new objects every recompute, so it kept resetting its poll timer and stalling. Using a stable ref fixes.

frontend/src/pages/team/Instances.vue
```diff
@@ -456,22 +435,48 @@ export default {
...
         },
-        instanceUpdated: function (newData) {
-            const mutator = new InstanceStateMutator(newData)
-            mutator.clearState()
-            newData.running = this.isRunningState(newData.meta.state)
-            newData.notSuspended = newData.meta.state !== 'suspended'
-            this.instancesMap.set(newData.id, {
-                ...this.instancesMap.get(newData.id),
-                ...newData
+        async instanceStart (instance) {
```
n-lark, line 438: All these are from the mixin that I'll remove in a follow up PR.

frontend/src/stores/data-farm-hosted-instances.ts
```diff
@@ -0,0 +1,157 @@
+import { defineStore } from 'pinia'
```
n-lark, line 1: This store is a little different from the other `data-farm` stores since the instances list is server-paginated/searched/sorted, so it can't cache-and-serve. Each query refetches and the store only holds the current page. 

The more traditional farming patterns will be established in follow-ups for the other use cases.

cstns: i feel that's a good approach to the problem and we should go with it for all paginated requests

cstns (approved):

PR #8155, opened 2026-08-10
Move round feature enablement

hardillb:
part of FlowFuse/engineering#230

## Description

Adds new license tiers and enables features accordingly as laid out in FlowFuse/engineering#230.

The `enterprise` tier should continue to add all new features so they are covered by tests

## Related Issue(s)

FlowFuse/engineering#230

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


hardillb:
I had to check in a new `package-lock.json` to get `npm ci` to run in the , so something is off, but I've merged main and updated package-lock.json to match

knolleary:
Still need to double check, but with the new product tiers it is possible for a customer to buy a license for both hub and edge to unlock the full feature set. So we need to handle such license types.

I propose we allow the type to be a `,` separated list of tier types. This will require a slightly different approach for enabling features - a feature in hub & edge shouldn't get enabled twice because the license enables both.

I suggest we have a codified version of the tickbox table that maps feature to tiers (or tiers to features... not such which way around makes the most sense...)

```
const tierFeature = {
 featureName: { hub: true, edge: true, enterprise: true },
 ...
}
```

And then a utility function `isFeatureEnabledForTier(string) => boolean` - which either handles the comma array of tier types, or gets called multiple times and the result aggregated by the caller.

hardillb:
Can't that just be the `enterprise` license that we need to keep to ensure we test everything?

There is also the stuff about maybe including the cert nodes tokens in the license and the new feature listed about access to different BluePrint libraries (that we have not done any planing/design for yet)

knolleary:
I don't think it can be. We'd want the platform to be able to report what tier they are licensed for - using `enterprise` as a catch-all will mask what they have actually licensed.

hardillb:
OK, so we'll need to start again here, multi or mixed tier licenses was not in the original spec issue.

forge/ee/lib/index.js
```diff
@@ -1,72 +1,103 @@
...
+// common features across all higher tier (enterpise, hub, edge, fleet)
+async function commonFeatures (app, opts) {
+    app.decorate('sso', await require('./sso').init(app))
+    await require('./teamBroker').init(app)
+    // Set the MFA Feature Flag
+    app.config.features.register('mfa', true, true)
+    // Set the Project History timeline Feature Flag
+    app.config.features.register('projectHistory', true, true)
+    // Set the Bill of Materials Feature Flag
+    app.config.features.register('bom', true, true)
```
knolleary: BoM has been moved to `hub` only

knolleary (changes requested):
Need to move bom to hub only.

I do think it would be better to have a map of tier to available features that can be more readily compared to the pricing table to spot any inconsistencies. As it is, we need to step through the code to work out what features have been enabled for each tier and risks things slipping through.

I will do a deeper review to check for any other updates needed.

forge/ee/lib/index.js
hardillb: OK, BOM isn't even in the list I was working from

https://github.com/FlowFuse/engineering/issues/230

forge/ee/lib/index.js
```diff
@@ -1,72 +1,106 @@
...
+            // HA
+            require('./ha').init(app)
+            // Protected Instances
+            require('./protectedInstance').init(app)
+            // Git Opps
+            app.decorate('gitops', await require('./gitops').init(app))
+            // Set the Bill of Materials Feature Flag
+            app.config.features.register('bom', true, true)
+        } else if (tiers.includes('edge') || tiers.includes('fleet')) {
+            await commonFeatures(app, opts)
```
knolleary:
```suggestion
```

Looks like `commonFeatures` gets called twice in the edge/fleet path.

knolleary:
With the new packaging, we are back to needing to providing separate limits on hosted/remote - the license asks for a combined limit. I'll work on the updates needed for that now.

knolleary:
I have pushed some updates:

1. fixed up the license version checks
2. added a `remoteInstances` feature flag (defaults to true, if the license is hub only, gets togged to false)
3. a number of front-end feature flag checks to disable remote instance UX as needed
4. centralised the overage check logic - a hub only license will have a limit of `0` remote instances; but we allow overages for licensed platforms. We don't want to allow overage for hub only on remote instances - so added that logic here.

knolleary (approved):
Having pushed the updates to the UX, I think this is sufficient for now.

Issue #8159, opened 2026-08-10
MCP header button: opt-in browser session exposure for third-party agents


cstns:
Third-party MCP agents can only operate on browser tabs that the user has explicitly exposed. Add an MCP button to the page header (similar to the existing Expert button in `ExpertButton.vue` / `PageHeader.vue`) that toggles third-party session exposure for the current tab.

**When toggled ON:**
* Register the tab's session in the Redis cache (#8158) with the first heartbeat
* Start the heartbeat interval and context-update watchers
* Open an MQTT connection using the tab session ID (#8157) and subscribe to in-flight topics so third-party agents can reach this tab
* Show a visual indicator that the session is exposed

**When toggled OFF (or tab closes):**
* Clear the session from the cache
* Stop heartbeats and context watchers
* Unsubscribe from in-flight topics
* The MQTT connection for first-party chat (if any) is unaffected

The button should be gated by the same AI platform feature flags that gate the Expert button. The toggle state can be persisted to `sessionStorage` so a page refresh within the same tab restores the exposure without requiring the user to click again.

Depends on #8157 and #8158.

PR #8162, opened 2026-08-10
Persist session ID in sessionStorage for tab identity

### LINKED ISSUE #8157: Add a tab session ID to the context store for third-party MCP targeting
There are two distinct session ID concepts that need to coexist:

* **Chat session IDs** (existing): one per agent mode (support/insights), stored in each agent's Pinia store, used for MQTT chat topic routing and broker credentials. These stay as-is.
* **Tab session ID** (new): one per browser tab, stored in the context store, used for third-party MCP agent targeting and browser session presence.

The context store (`frontend/src/stores/context.js`) should own the tab session ID. It initializes once per tab (persisted to `sessionStorage`) and is already the central place for tab-scoped state (team, application, instance, device, route).

The tab session ID is not used until the user explicitly opts in to third-party MCP exposure via the MCP header button (#8159). It is not consumed by the first-party chat flow.

This is foundational for browser session presence and third-party targeting work.

cstns:
## Description

  - tabSessionId added to state (separate from the agent stores' chat sessionId)                                                                                                                                                                                                                                              
  - Persisted to sessionStorage so it survives page refreshes within the same tab
  - Auto-generated on first hydration if not present                                                                                                                                                                                                                                                                          
  - Not consumed by anything yet. The MCP button (#8159) will be the first consumer.                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                              
  The agent stores, expert store, and post-message service are all untouched. Chat session IDs remain as they are. 

### Addendum: reuse `account-auth` session ID instead of a new `tabSessionId`

The original approach added a new `tabSessionId` to the context store. After further discussion, we realized the `account-auth.js` store already has a per-tab `sessionId` (generated via `getSessionId()`, used by the `fe-team` MQTT connection for broker client uniqueness). The only problem was that it was stored in-memory, so it died on page refresh.

By persisting it to `sessionStorage`, it becomes exactly what `tabSessionId` was supposed to be: per-tab, stable across refreshes, unique across tabs. This also means we can reuse the existing `fe-team` MQTT connection for presence topics on the backend, so there's no need for a separate MQTT connection or a new client identity for MCP presence. The `tabSessionId` addition to the context store was reverted.

The change simplifies to: persist the existing `account-auth` session ID in `sessionStorage` (split the `persist` config into an array with `localStorage` for `redirectUrlAfterLogin` and `sessionStorage` for `sessionId`).

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/8157

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/stores/account-auth.js
+++ b/frontend/src/stores/account-auth.js
@@ -38,7 +38,6 @@ export const useAccountAuthStore = defineStore('account-auth', {
         isAdminUser: (state) => !!state.user?.admin
     },
     actions: {
-        // In-memory, per page-load — duplicate tabs each mint their own
         getSessionId () {
             if (!this.sessionId) this.sessionId = uuidv4()
             return this.sessionId
@@ -245,8 +244,14 @@ export const useAccountAuthStore = defineStore('account-auth', {
             useProductExpertStore().$reset()
         }
     },
-    persist: {
-        pick: ['redirectUrlAfterLogin'],
-        storage: localStorage
-    }
+    persist: [
+        {
+            pick: ['redirectUrlAfterLogin'],
+            storage: localStorage
+        },
+        {
+            pick: ['sessionId'],
+            storage: sessionStorage
+        }
+    ]
 })
--- a/frontend/src/stores/context.js
+++ b/frontend/src/stores/context.js
@@ -16,8 +16,7 @@ export const useContextStore = defineStore('context', {
         route: null,
         application: null,
         instance: null,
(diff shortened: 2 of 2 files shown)


cstns:
see https://github.com/FlowFuse/flowfuse/issues/8157#issuecomment-5248711633

n-lark (approved):

Issue #8161, opened 2026-08-10
list_browser_sessions platform automation tool


cstns:
Third-party MCP agents need a way to discover which browser tabs a user has exposed. This is a platform automation tool that the gateway invokes via the standard platform automation channel (MQTT to Forge).

Forge reads from the Redis session cache (#8158), filters by the requesting user's ID, and returns the list of active sessions with their context payloads. The agent uses this to pick a `session_id` for subsequent in-flight tool invocations.

If no active sessions exist, the tool response should be descriptive enough to nudge the agent into prompting the user to enable the MCP button on a browser tab.

Depends on #8158.

PR #8182, opened 2026-08-11
Add tab presence support to track user activity across sessions

### LINKED ISSUE #8158: Store browser session presence in Forge's Redis cache
Third-party MCP agents need to discover a user's active browser tabs and their context in order to target in-flight requests. The current design in FlowFuse/engineering#233 has a thin TTL map on the gateway, but that's not sufficient. The gateway needs the full context payload (equivalent to `contextStore.expert`) to reconstitute in-flight MQTT topics with the correct `entityType`/`entityId`, and third-party agents need the same context the first-party agent gets.

## Backend

Introduce a new cache bucket using the existing Redis cache driver (`forge/caches/`), keyed per user + session, to store browser session state. Per-field TTL auto-expires entries when heartbeats stop (browser crash, network drop). Forge does not relay presence to the gateway. The gateway retrieves sessions via a platform automation call to Forge when needed.

The backend exposes two endpoints:
* A heartbeat endpoint that extends TTL and accepts visibility/focus state
* A context update endpoint that accepts the full `contextStore.expert` payload on route changes

## Frontend

Sessions are not registered automatically on tab load. The user must explicitly opt in via the MCP header button (#8159). The presence lifecycle is driven by the button toggle, not the team lifecycle.

Once opted in, the frontend should use the existing subscriber pattern (`frontend/src/subscribers/`) to manage the presence lifecycle. A new subscriber would:

* Start/stop when the user toggles the MCP button on/off
* Use the transport layer and subscriber factory/registry for singleton management
* Send periodic heartbeats (~45s) carrying current visibility/focus state (`document.visibilityState`, `document.hasFocus()`)
* Watch for Vue Router navigation changes and push updated context to Forge immediately (event-driven, not polled)
* Clear the session on tab close (`beforeunload`), component unmount, or when the user toggles the MCP button off

Depends on #8157.

cstns:
## Description

 - Add MQTT-based browser session presence so third-party MCP agents can discover which tabs a user has exposed                                                                                                                                                                                                              
 - Forge stores presence in a Redis cache bucket (browserSessions) with a 135s TTL, auto-expired when heartbeats stop                                                                                                                                                                                                        
 - Browser tabs publish on ff/v1/tab-presence/<userId>/<sessionId>/heartbeat and .../context using the existing fe-team MQTT connection                                                                                                                                                                                      
 - Forge subscribes via a shared subscription ($share/platform/...) so only one replica handles each message, avoiding redundant Redis writes

 The browser sends two types of messages over MQTT:                                                                                                                                                                                                                                                                          
 - Heartbeat (periodic, ~45s): extends TTL, carries visibility/focus state                                                                                                                                                                                                                                                 
 - Context (on route change): carries the full contextStore.expert payload so the gateway can reconstitute in-flight MQTT topics with the correct entity info                                                                                                                                                                
                                                                                                                                                              
 Both message types call cache.set() which resets the TTL. If heartbeats stop (browser crash, tab close), the entry auto-expires after ~135s.                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                              
` app.comms.browserSessions.getSessionsByUser(userId)` retrieves all active sessions for a user, for the future list_browser_sessions MCP tool (#8161).    

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/8158

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/comms/aclManager.js
+++ b/forge/comms/aclManager.js
@@ -492,7 +492,7 @@ module.exports = function (app) {
                 // Uses [^/]+ for the message-type segment because the subscription wildcard (+)
                 // is matched as a literal character. The publish-side ACL on teamFrontend
                 // already restricts to heartbeat|context.
-                { topic: /^ff\/v1\/browser\/tab-presence\/[^/]+\/[^/]+\/[^/]+$/ }
+                { topic: /^ff\/v1\/browser\/tab-presence\/[^/]+\/[^/]+\/[^/]+$/, shared: true }
             ],
             pub: [
                 // Send commands to project launchers
--- a/forge/comms/aclManager.js
+++ b/forge/comms/aclManager.js
@@ -489,7 +489,10 @@ module.exports = function (app) {
                 { topic: /^ff\/v1\/expert\/([^/]+)\/([^/]+)\/platform\/([^/]+)\/request$/, verify: 'checkExpertPlatformTopic', allowWildcard: { user: true, session: true, command: true }, isPlatform: true, isSub: true, agent: 'platform' },
                 // platform can listen for browser tab presence (shared subscription)
                 // - ff/v1/tab-presence/<userId>/<sessionId>/<heartbeat|context>
-                { topic: /^ff\/v1\/tab-presence\/[^/]+\/[^/]+\/(heartbeat|context)$/, shared: true }
+                // Uses [^/]+ for the message-type segment because the subscription wildcard (+)
+                // is matched as a literal character. The publish-side ACL on teamFrontend
+                // already restricts to heartbeat|context.
+                { topic: /^ff\/v1\/tab-presence\/[^/]+\/[^/]+\/[^/]+$/, shared: true }
             ],
             pub: [
                 // Send commands to project launchers
--- a/forge/comms/aclManager.js
+++ b/forge/comms/aclManager.js
@@ -488,11 +488,11 @@ module.exports = function (app) {
                 // platform can listen for Expert Agent requests
                 { topic: /^ff\/v1\/expert\/([^/]+)\/([^/]+)\/platform\/([^/]+)\/request$/, verify: 'checkExpertPlatformTopic', allowWildcard: { user: true, session: true, command: true }, isPlatform: true, isSub: true, agent: 'platform' },
                 // platform can listen for browser tab presence (shared subscription)
-                // - ff/v1/tab-presence/<userId>/<sessionId>/<heartbeat|context>
+                // - ff/v1/browser/tab-presence/<userId>/<sessionId>/<heartbeat|context>
                 // Uses [^/]+ for the message-type segment because the subscription wildcard (+)
                 // is matched as a literal character. The publish-side ACL on teamFrontend
                 // already restricts to heartbeat|context.
-                { topic: /^ff\/v1\/tab-presence\/[^/]+\/[^/]+\/[^/]+$/, shared: true }
+                { topic: /^ff\/v1\/browser\/tab-presence\/[^/]+\/[^/]+\/[^/]+$/, shared: true }
             ],
             pub: [
                 // Send commands to project launchers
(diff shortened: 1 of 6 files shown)


forge/comms/aclManager.js
```diff
@@ -171,6 +171,25 @@ module.exports = function (app) {
                 return false
             }
         },
+        checkPresenceTopic: async function (requestParts, usernameParts) {
+            // requestParts = [ fullTopic, <userId>, <sessionId>, <messageType> ]
+            // usernameParts = [ 'fe-team', <userHash>, <teamHash>, <sessionId> ]
+            const topicUserId = requestParts[1]
+            const usernameUserHash = usernameParts[1]
+            if (topicUserId !== usernameUserHash) {
+                return false
```
n-lark, line 180: The pub ACL checks `userId` but not `sessionId`, so a tab can publish presence for any `sessionId` under its own user. Should it also check the `sessionId` matches?

n-lark (approved):

forge/comms/aclManager.js
cstns: The sessionId is client-generated (UUID v4 in sessionStorage) and passed to the backend when requesting MQTT credentials, where it gets embedded in the username. Both values are available in the ACL check, but there's no server-side source of truth to validate against, so comparing them only checks internal consistency within the same client session.                                                                                                                                                                                                                                                                             

The userId check is what actually matters. A spoofed sessionId under your own user only pollutes your own presence data, not an escalation

forge/comms/commsClient.js
```diff
@@ -249,7 +265,9 @@ class CommsClient extends EventEmitter {
                 // of consumers that share the workload, so keeping Expert separate from the
                 // "platform" group prevents unrelated features from sharing a consumer pool and
                 // allows them to scale independently.
-                '$share/expert/ff/v1/expert/+/+/platform/+/request'
+                '$share/expert/ff/v1/expert/+/+/platform/+/request',
+                // Browser tab presence - shared subscription
+                '$share/platform/ff/v1/tab-presence/+/+/+'
```
Steve-Mcl: There are a few things I dislike about this topic.
1. tab-presence is very limiting to specific behaviour (not really extensible if we need other transmissions from browser to forge)
2. ends with a `+` - a subscription that's already fixed at exactly 3 wildcard levels can't grow a 4th literal segment without risk of overlapping - that said, if we scope it under browser/operation (e.g. `ff/v1/browser/tab-presence/+/+/+`) then this is less of an issue
3. using the "platform" share group name - we should prob avoid sharing a consumer pool with unrelated features so each can scale independently?

What if we went with something a little boarder that allows more than than just "tab-presence" (even if we never do, it is cheap to do now)

e.g.  `'$share/browser/ff/v1/browser/tab-presence/+/+/+'` or `'$share/ui/ff/v1/ui/tab-presence/+/+/+'`

cstns: sensible request, updating

forge/comms/commsClient.js
```diff
@@ -52,6 +52,22 @@ class CommsClient extends EventEmitter {
                 const ownerId = topicParts[4]
                 const messageType = topicParts[5]
 
+                if (topicParts[2] === 'browser') {
```
Steve-Mcl, line 55:
```suggestion
                if (topicParts[2] === 'browser' || topicParts[3] === 'tab-presence') {
```

forge/comms/aclManager.js
```diff
@@ -467,7 +486,13 @@ module.exports = function (app) {
...
                 { topic: /^ff\/v1\/platform\/leader$/ },
                 // platform can listen for Expert Agent requests
-                { topic: /^ff\/v1\/expert\/([^/]+)\/([^/]+)\/platform\/([^/]+)\/request$/, verify: 'checkExpertPlatformTopic', allowWildcard: { user: true, session: true, command: true }, isPlatform: true, isSub: true, agent: 'platform' }
+                { topic: /^ff\/v1\/expert\/([^/]+)\/([^/]+)\/platform\/([^/]+)\/request$/, verify: 'checkExpertPlatformTopic', allowWildcard: { user: true, session: true, command: true }, isPlatform: true, isSub: true, agent: 'platform' },
+                // platform can listen for browser tab presence (shared subscription)
+                // - ff/v1/browser/tab-presence/<userId>/<sessionId>/<heartbeat|context>
+                // Uses [^/]+ for the message-type segment because the subscription wildcard (+)
+                // is matched as a literal character. The publish-side ACL on teamFrontend
+                // already restricts to heartbeat|context.
+                { topic: /^ff\/v1\/browser\/tab-presence\/[^/]+\/[^/]+\/[^/]+$/, shared: true }
```
Steve-Mcl:
```suggestion
                { topic: /^ff\/v1\/browser\/tab-presence\/[^/]+\/[^/]+\/[^/]+$/ }
```
shared is for subs only

Steve-Mcl (approved):
approved bar outstanding nit

cstns:
The shared flag isn't a vestige. EMQX sends the full topic including the $share/group/ prefix to our auth webhook, so aclManager still needs to strip it and match against the bare topic. The shared: true on individual ACL rules is a deliberate gate: it prevents clients from wrapping any topic they can subscribe to in $share/ to create unintended consumer groups. Re-added it as it also caused unit test failures

PR #8183, opened 2026-08-11
Add TabPresencePublisher and MCP toggle integration in ExpertButton component

### LINKED ISSUE #8159: MCP header button: opt-in browser session exposure for third-party agents

cstns:
## Description

  - Introduce a publisher pattern for the frontend, the counterpart to the existing subscriber pattern. Subscribers receive MQTT messages, publishers push them. Same lifecycle (connect/disconnect), same transport sharing, same singleton factory.
  - Add TabPresencePublisher as the first concrete publisher. It publishes heartbeats (every 45s + visibility changes) and context updates (on route navigation) to ff/v1/tab-presence/... MQTT topics, feeding the backend Redis cache from #8158.                                                                           
  - Transform the Expert header button into a composite split button. Left half opens the Expert drawer (unchanged behavior). Right half is an MCP toggle that starts/stops the presence publisher. Toggle state persists to sessionStorage so it survives page refreshes.
  - Extend the Transport interface with a publish method, making it bidirectional so publishers use the same abstraction as subscribers.

**Publisher pattern** (new frontend/src/publishers/):                                                                                                                                                                                                                                                                           
  - team-publisher.contract.ts - base class mirroring TeamSubscriber: operation serialization, transport attachment sharing, _publish() helper, abstract _onStarted()/_onStopped() hooks
  - publisher.factory.ts - singleton factory (definePublisherSingleton)
  - publisher.registry.ts - empty registry for future auto-connected publishers
  - tab-presence.publisher.ts - concrete publisher: heartbeat interval, router afterEach guard, visibility change listener, beforeunload cleanup

**Transport**:
  - transport.types.ts - added TransportPublishOptions interface and publish() to Transport                                                                                                                                                                                                                                   
  - mqtt.transport.ts - implemented publish() delegating to mqttService.publishMessage()                                                                                                                                                                                                                                      

**Orchestrator**:                                                                                                                                                                                                                                                                                                               
  - app.orchestrator.ts - added $publishers, bootPublishers(), publisher cleanup in dispose()

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/8159

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


frontend/src/stores/account-settings.js
```diff
@@ -50,6 +51,10 @@ export const useAccountSettingsStore = defineStore('account-settings', {
             // adding in PostHog Feature Flags
             checks.isPostHogFeatureFlagsEnabled = !!state.posthogFlags[POSTHOG_FLAGS.FF_FEATURE_FLAGS]
 
+            // MCP third-party: platform + team flag + PostHog gate
+            checks.isMcpThirdPartyEnabled =
```
n-lark, line 55: `isMcpThirdPartyEnabled` doesn't appear to be consumed anywhere. Should the MCP toggle in `ExpertButton.vue` be gated on this?

frontend/src/components/icons/McpIcon.js
```diff
@@ -0,0 +1,22 @@
+const { createVNode: _createVNode, openBlock: _openBlock, createBlock: _createBlock } = require('vue')
```
n-lark, line 1: Do you use the `frontend/src/components/icons/mcp-icon.svg` icon here? Should you murder it?

frontend/src/components/ExpertButton.vue
```diff
@@ -55,6 +132,41 @@ export default {
...
+
+    &:hover {
+        background-color: var(--ff-color-bg-surface);
+    }
+}
+
+/* Right half: MCP toggle */
+.expert-composite__mcp {
+    border: none;
+    border-left: 1px solid rgba(128, 128, 128, 0.2);
```
n-lark: You should use a --ff-color- token for this.

frontend/src/components/ExpertButton.vue
```diff
@@ -1,36 +1,111 @@
...
+        onMcpClick () {
+            if (this.mcpActive) {
+                this.stopMcp()
+            } else {
+                this.startMcp()
+            }
+        },
+        startMcp () {
+            if (!this.team) return
+            const orchestrator = getAppOrchestrator()
```
n-lark: Why do all this here in `ExpertButton` rather than in the publisher? Sure you have a reason just wondering.

frontend/src/components/ExpertButton.vue
```diff
@@ -1,36 +1,111 @@
...
+        if (stored === 'true' && this.team) {
+            this.startMcp()
+        }
+    },
+    beforeUnmount () {
+        if (this.publisher) {
+            destroyTabPresencePublisher()
+            this.publisher = null
+        }
+    },
```
n-lark, line 80: When we switch teams or log out, `setTeam` in `account.js` reconnects/tears down all subscribers (`ensureTeamChannelConnected` / `disconnectTeamSubscribers`). 

We're not doing that for the publisher. It's created in `ExpertButton` and never re-pointed, so after a team switch it's left attached to the old team. Do you think that's going to be an issue?

frontend/src/components/ExpertButton.vue
```diff
@@ -1,36 +1,111 @@
...
+                this.stopMcp()
+            } else {
+                this.startMcp()
+            }
+        },
+        startMcp () {
+            if (!this.team) return
+            const orchestrator = getAppOrchestrator()
+            const { mqtt } = orchestrator.$services
+            const { createMqttTransport } = require('@/transport/mqtt.transport')
```
n-lark: Was this breaking a circular import??

n-lark (dismissed):
Hefty lift 💪  - should you include some specs for this?

frontend/src/transport/mqtt.transport.ts
```diff
@@ -16,6 +16,10 @@ export class MqttTransport implements Transport<TransportConnectOptions> {
         return this.$mqtt.subscribe(key, topics, options)
     }
 
+    publish (key: string, options: TransportPublishOptions): Promise<void> {
+        return this.$mqtt.publishMessage(key, options)
```
n-lark, line 20: You've got a typescript error here: 

```
Argument of type 'TransportPublishOptions' is not assignable to parameter of type 'MqttPublishRequest'.
  Types of property 'payload' are incompatible.
    Type 'unknown' is not assignable to type 'MqttPayload'.
```

frontend/src/components/icons/McpIcon.js
cstns: yeah, you're right! i thought it would be a good idea to keep the icon as a plain svg too

frontend/src/stores/account-settings.js
cstns: i thought of adding the feature flags but forgot to toggle the mcp button visibility based of it. fixed

frontend/src/components/ExpertButton.vue
cstns: fixed, thank you!

frontend/src/components/ExpertButton.vue
cstns: Honestly, no good reason. Since it's a user-initiated action (not always-on like subscribers), I just wired it up inline and didn't think much about it.                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                              
  Two options to clean it up:                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                              
  1. Export a convenience factory from the publisher module that hides the orchestrator/transport wiring. Component just calls something like startTabPresence(team) and gets a publisher back. The UI stuff (sessionStorage, alerts) stays in the component.                                                               
  2. Register it in the orchestrator's publisher registry so it's booted automatically, then the component just calls connect/disconnect on a pre-wired instance. Doesn't fit as naturally since the publisher is opt-in, but it would keep things consistent with how subscribers work.
                                                                                                                                                                                                                                                                                                                              
  Leaning toward (1), but (2) has merit if we want that consistency. What do you think?

frontend/src/components/ExpertButton.vue
cstns: The presence publisher tracks user/session level data (heartbeat, visibility, page context), not team-scoped data. The MQTT credentials happen to come from a team connection, but the information being published has nothing to do with which team is active.

Steve-Mcl: lets switch off MCP enablement when switching teams - kinda a cop-out but also a genuine security thing (i.e. when switching teams did you _really_ want MCP actions to be going ahead without a physical confirmation)?

cstns: I'll be taking the easy way out

frontend/src/components/ExpertButton.vue
cstns: no, just claude being lazy

frontend/src/transport/mqtt.transport.ts
cstns: good catch

frontend/src/components/ExpertButton.vue
n-lark: 1 sounds good to me!

Steve-Mcl (approved):

PR #8195, opened 2026-08-14
Add SendMqttMessageAwaitReply utility

cstns:
## Summary

Adds a generic utility for turning async MQTT publish/subscribe into awaitable calls, using MQTT v5 `correlationData` packet properties for request/response correlation.

The existing `sendCommandAwaitReply` in `DeviceCommsHandler` does this for device commands but is tightly coupled to device topics and message formats. This utility extracts the core correlation pattern into a standalone class that any handler can use (the MCP gateway handler being the first consumer, coming in a follow-up PR).

* `create()` returns a `correlationData` UUID, MQTT v5 publish options, and a promise
* `resolve(correlationData, payload)` matches incoming responses (handles Buffer from MQTT v5 packets)
* No dependency on any MQTT client, topic structure, or message format
* Each consumer instantiates its own instance via `require`, not bound to `app.comms`

## Test plan

* [ ] `npx mocha 'test/unit/forge/comms/sendMqttMessageAwaitReply_spec.js'` (11 tests covering create, resolve, Buffer handling, timeout, clear, concurrent correlation)
* [ ] `npx mocha 'test/unit/forge/comms/*_spec.js'` (no regressions on existing comms tests)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Steve-Mcl (approved):
Looks good and is inert so all good.

PR #6511, opened 2026-01-14
Refactor tour cancel logic and add final step handling for hosted instance tours

cstns:
## Description

starting a new product tour when closing the initial one before completion brought clunky ui artefacts and perturbed the order in which follow-up actions reacted.

I overwrote the tour cancel method and replaced it with the initial onClose hook.
The cancel method now redirects to the 2n'd to last step or last step in the welcome tour. The 2nd to last tour step highlights the newly created instance if it exists, and if it doesn't exist it will redirect to the last step with the 'you're all set` message.

PH events should remain the same, alongside subsequent actions that occurred after the tour completed.

https://github.com/user-attachments/assets/65690ac7-3fad-4e4e-a3cb-7ed29779b111

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/5580

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] Suitable unit/system level tests have been added and they pass 
 - [ ] Documentation has been updated
    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Changes `flowforge.yml`?
    - [ ] Issue/PR raised on `FlowFuse/helm` to update ConfigMap Template
    - [ ] Issue/PR raised on `FlowFuse/CloudProject` to update values for Staging/Production
 - [ ] Link to Changelog Entry PR, or note why one is not needed.

## Labels

 - [ ] Includes a DB migration? -> add the `area:migration` label

Issue #5580, opened 2025-06-03
Highlight new instance at conclusion of product tour


gstout52:
### Description

About half of new users cancel the product tour.

![Image](https://github.com/user-attachments/assets/98234ab1-774f-44a3-bddc-eec7d3f41f60)

To ensure that users reach the final activation step in product, when a user cancels the product tour, start a new single-item "tour" that highlights the Open Editor button on the automatically-created instance, and says: "Get started building with Node-RED here" (or similar).

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
Feel like we are missing a key metric in this issue which is how many people open the editor? Especially then broken down by those that cancel the tour and those that completed it. If there is a difference, then we have merit in doing this, if it's the same, then the highlighting of the "Open Editor" button will be irrelevant.

joepavitt:
We do actually have an issue whereby the product tour currently assumes you're on the Applications page though, which is no longer the case with our new landing page... that does need addressing...

gstout52:
This chart shows the funnel from signup to open editor, broken down by completion step for the product tour. Users who cancel the tour at the start open the editor 7 percentage points less than users who do more than half of the tour or all of it. (Funny enough, users who did some of the tour but less than half of it opened the editor 100% of the time - but there were only 5 such users, polluting the chart.)

<img width="604" height="691" alt="Image" src="https://github.com/user-attachments/assets/6b78dc38-3105-4363-9592-9681c08b4239" />

https://eu.posthog.com/project/2209/insights/67lAxmhU

joepavitt:
Thanks Greg, the other interesting thing looking at the Insight is the time to next step:

- Tour Completed: 3m 16s
- Over 50%: 3m 15s
- Cancelled: 5m 17s

I'd also be interested to know, of those that cancel and do not open the Editor, what instead _are_ they doing?

gstout52:
I've now watched several of the videos of users from [this cohort](https://eu.posthog.com/project/2209/cohorts/64453), and overwhelmingly they: close the tour, then click around frantically.

Other behaviors:

- Create a second instance (likely because they don't know they already have an instance)
- Delete their team
- Try to create a remote instance
- A couple of users were on mobile, where the product tour display is very poor. For a quick win, I've [created an issue](https://github.com/issues/created?issue=FlowFuse%7Cflowfuse%7C5773) to present an alternative display for mobile users.

[...]

PR #8215, opened 2026-08-18
Make closing the welcome tour actually close it

dimitrieh:
## Summary

The close icon, the escape key and the first step's Exit button all route through `tour.cancel`. https://github.com/FlowFuse/flowfuse/pull/6511 overrode that method to jump to the final step, so none of them ended the tour on the first press and a button labelled Exit did not exit.

It also reported `ff-tour-cancel` against the final step rather than the step the user left on, so the drop-off funnel behind https://github.com/FlowFuse/flowfuse/issues/5580 shows everyone abandoning at the end.

Back to listening for the `cancel` event, with an optional cancel hook on `Tours.create`. The welcome tour uses it to pulse the Open Editor button once the overlay is gone, so the nudge survives without holding the user in a tour they closed.

## Test plan

* [ ] `npx vitest run --config ./config/vitest.config.ts test/unit/frontend/tours/` (drives the real shepherd.js in jsdom and clicks the rendered close icon; fails on `main`)
* [ ] `npx vitest run --config ./config/vitest.config.ts` (no regressions)

### DIFF
--- a/frontend/src/stores/ux-tours.js
+++ b/frontend/src/stores/ux-tours.js
@@ -2,7 +2,7 @@ import { defineStore } from 'pinia'
 import { markRaw } from 'vue'
 
 import Tours from '../tours/Tours.js'
-import TourWelcome, { id as WelcomeTourId } from '../tours/tour-welcome.js'
+import TourWelcome, { id as WelcomeTourId, onCancel as onWelcomeTourCancel } from '../tours/tour-welcome.js'
 
 export const useUxToursStore = defineStore('ux-tours', {
     state: () => ({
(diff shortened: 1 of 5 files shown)


dimitrieh:
@cstns could you take a look here when you have time? You were the one created https://github.com/FlowFuse/flowfuse/pull/6511

See clip for context:

https://github.com/user-attachments/assets/b896f00d-2e18-40d2-837f-25e46fa9dd36

cstns:
This was the intended use case, I argued against it at the time but was vetoed. Why are we circling back now, is there an issue to this PR?

dimitrieh:
Closing for now. Adjusting onboarding is up ahead

PR #8216, opened 2026-08-18
docs: name the file each config snippet belongs in

dimitrieh:
## TL;DR

Config snippets in docs now say which file they are, and point at the line you have to change. A reader landing on the Kubernetes install guide from search sees `customization.yml` on the block instead of having to scroll back through the prose to work it out.

Depends on FlowFuse/website#5627. Until that merges the website drops both the file name and the highlights, so this changes nothing on flowfuse.com and breaks nothing either.

## What changed

- **Kubernetes install guide**: every `customization.yml` snippet is labelled. Where the prose tells you to edit a value, that line is highlighted: the cluster issuer, the private CA certs, the external database switch, the SES IAM role, the storage class.
- **Device agent**: the `https`, `httpStatic` and `localAuth` snippets in `register.md`, and `nodeOptions` in `running.md`, are labelled `device.yml`.

Only fence info strings changed. No prose, no snippet contents, no new files.

## Checked

The website's docs sync passes fence info strings through untouched, so the label and line numbers survive into the built site. The rendering itself was verified on the website side in FlowFuse/website#5627, against blocks using this exact syntax.

### DIFF
--- a/docs/device-agent/register.md
+++ b/docs/device-agent/register.md
@@ -304,7 +304,7 @@ The values should be set to the contents of the certificate/key.
 Alternatively, the properties `caPath`, `keyPath` and `certPath` can be used instead
 to provide absolute paths to files containing the certificates/keys.
 
-```yml
+```yml [device.yml]
 https:
    keyPath: /opt/flowfuse-device/certs/key.pem
    certPath: /opt/flowfuse-device/certs/cert.pem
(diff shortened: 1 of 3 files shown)


dimitrieh:
depends on https://github.com/FlowFuse/website/pull/5627

KristopherLeads (approved):

Issue #8230, opened 2026-08-21
Product repackaging: Hub / Edge / Fleet


dimitrieh:
Collection of prod/eng work regarding the product repackaging work.

PR #8239, opened 2026-08-21
Announcements: target specific teams, rich body, and an in-app toast

dimitrieh:
## Description

The Notifications Hub could send a plain text announcement to whole team types, with the whole card as its only affordance. Moving a named set of teams onto a new product needs narrower targeting and a message that can explain itself.

Raised as a proposal, so opening as a draft.

- **Audience is one thing or the other.** An explicit choice between teams of a type and specific teams, with only that mode's controls rendered. Previously team type and billing state appeared twice on the same screen, once as audience selectors and again as filters on the team table.
- **Specific teams come from a table.** Filterable and paginated, with row checkboxes and bulk actions: select every team matching the current filter, select the visible page, clear. Selection is owned outside the table so it survives searching, filtering and paging. New admin route returns the ids for a filter so selecting a thousand teams is one request, sharing its where clause with the teams list so the two cannot drift.
- **Richer body.** Optional markdown, an embedded YouTube video, and a button with its own label and url. The admin form previews through the same component recipients see. Video and button urls are normalised and validated server side, so the embed src is built from a validated id rather than typed text.
- **Bottom right toast.** Unread announcements surface without opening the drawer, and clear when it is opened. A rich announcement's card is not itself a link, since the body owns its links, video and button. A plain announcement keeps the existing card wide click.

Suspended teams are excluded from selection: they are filtered out when an announcement is addressed, so offering them built a selection that could never receive anything.

`ff-data-table` gains an optional `checked` prop. Left unset it owns its selection exactly as before.

Context: https://github.com/FlowFuse/flowfuse/issues/8230

<details><summary>Screenshots</summary>
<p>

pr-1-hub-form-and-preview
pr-2-audience-team-type-mode
pr-3-specific-teams-bulk-select
pr-4-toast-bottom-right
pr-5-toast-beside-expert-drawer
pr-6-drawer-rich-and-plain

</p>
</details> 

### Test plan

<details><summary>Verified against a local platform with 63 teams across three team types</summary>

Backend, unit tested:

- [x] Audience by explicit team list, by team type, and by billing state; recipients deduped when one user owns many selected teams
- [x] Ids endpoint matches the teams list for the same filter, filters by type, name and state, and is admin only
- [x] Video link accepted from watch, share, shorts, live and bare id forms; other hosts and non http schemes rejected
- [x] Card link, `to.url` and button url reject anything that can execute or resolve off platform, including targets hiding a host behind a stripped tab

Browser:

- [x] Select all matching, then narrow the filter: selection holds while the matching count drops, and unchecking one row leaves the rest
- [x] Switching audience mode clears the other side, so a stale selection cannot be sent
- [x] Markdown, video playback and the button all work inside the toast and the drawer card
- [x] Toast clears when the notifications drawer opens; sits beside an open right drawer at several widths, and holds back when there is no room for it
- [x] Clicking a rich card's body neither navigates nor marks it read

</details>

## Related Issue(s)

None. Raised from a question about informing teams in app when they are moved onto a new product.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts
 - [ ] Link to Changelog Entry PR, or note why one is not needed. Draft proposal, so no changelog entry yet.

## Labels

### DIFF
--- a/forge/db/models/Team.js
+++ b/forge/db/models/Team.js
@@ -367,6 +367,44 @@ module.exports = {
                         count,
                         teams: rows
                     }
+                },
+                /**
+                 * The ids of every team matching a search and filter, with no
+                 * pagination. For callers that need the whole matching set
+                 * rather than a page of it, such as selecting every team an
+                 * announcement should go to.
+                 *
+                 * Shares its where-clause construction with getAll so the ids
+                 * cannot drift from the list the admin is looking at.
+                 *
+                 * @param {Object} pagination search options; `query` only, any cursor is ignored
+                 * @param {Object} where additional filters, as built for getAll
+                 * @param {number} idLimit hard cap on how many ids are returned
+                 * @returns {{ids: string[], count: number, truncated: boolean}}
+                 */
+                getAllIds: async (pagination = {}, where = {}, idLimit = 10000) => {
+                    where = buildPaginationSearchClause({ query: pagination.query }, where, ['Team.name'])
+                    if (pagination.query) {
+                        const queryId = M.Team.decodeHashid(pagination.query)
+                        if (queryId && queryId.length === 1) {
+                            // The query term is a valid hashid - go look it up
+                            where = { id: queryId }
+                        }
+                    }
+                    const include = []
+                    if (app.billing) {
+                        // The billing filter references $Subscription.status$
+                        include.push({ model: app.db.models.Subscription, attributes: [] })
+                    }
+                    const [rows, count] = await Promise.all([
+                        this.findAll({ where, include, attributes: ['id'], order: [['id', 'ASC']], limit: idLimit + 1 }),
+                        this.count({ where, include })
+                    ])
+                    return {
+                        ids: rows.slice(0, idLimit).map(row => row.hashid),
+                        count,
+                        truncated: rows.length > idLimit
+                    }
                 }
             },
             instance: {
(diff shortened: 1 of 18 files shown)


dimitrieh:
Two pre-existing issues found while reviewing this branch. Both predate it and are left untouched here, noting them so they are not lost.

**`Generic.vue` renders an announcement body through `v-html` with no sanitiser.** A plain announcement's `message` goes straight into `v-html` in `frontend/src/components/notifications/Generic.vue`, so admin authored markup is rendered unfiltered. The new rich renderer added here does sanitise, which makes the untouched plain path the weaker of the two.

**A Cypress selector will break if a fixture grows.** `test/e2e/frontend/cypress/tests/admin/notifications-hub.spec.js` does `cy.get('[data-el^="audience-teamType-"]').click()`, which matches every team type row and fails on more than one match. It passes only because the CE fixture seeds a single team type.

dimitrieh:
@allthedoll can you take a look here? 

This is a request from sales to help with the migration of pro/starter on FlowFuse Cloud. It builds out the notification mechanism we already have to be able to target specific teams and accept markdown inputs/rich text formatting so we can put in a youtube embed, links to documentation, and/or a button to reach out to sales. Check screenshots in the body of this PR. 

Happy to discuss this proposal next week. Ideally, need a dev to review this. This can happen any time just before the migration.

Issue #7423, opened 2026-06-11
Story 2: OAuth2 for MCP Agents


cstns:
### Summary

Enable external AI agents (Claude Desktop, Cursor, Claude Code) to authenticate with FlowFuse through OAuth2 PKCE, building on the existing OAuth2 scaffolding. This evolves from manual PAT setup (phase 1, covered by #7411) to automated OAuth flows, and eventually to Claude plugin/connector directory submission.

### Context

FlowFuse already has a working OAuth2 Authorization Code Grant with PKCE implementation (`forge/routes/auth/oauth.js`). It supports the `ff-plugin` client for the Node-RED editor, dynamic project/device clients, consent pages, token exchange, and refresh tokens. The infrastructure is sound. This story extends it for MCP agent clients without rearchitecting what exists.

The POC (PR #7414) added MCP-specific OAuth bits (hardcoded `mcp-agent` client, `AccessRequestMCP.vue` consent page, `.well-known` discovery endpoints). These should not be reused directly. Instead, this work builds around the existing OAuth scaffolding following the established patterns.

### Requirements

**Build on existing OAuth2 PKCE scaffolding**
- The existing implementation in `forge/routes/auth/oauth.js` already handles:
  - Authorization Code Grant with PKCE (S256)
  - `/account/authorize` and `/account/token` endpoints
  - Client validation (hardcoded + dynamic via `AuthClient` table)
  - Consent/authorization page workflow with approve/deny
  - Token exchange with refresh token support
  - 5-minute TTL on OAuth sessions, single-use authorization codes
- This is the foundation. The MCP agent auth work extends it, it doesn't replace it.

**MCP client registration**
- Register an MCP agent client, following the same pattern as the existing `ff-plugin` hardcoded client.
- The client needs appropriate scope definitions for MCP platform access.
- Redirect URI validation should support localhost (for local dev tools like Claude Code, Cursor) and, eventually, Anthropic's callback URLs (for Claude plugin).

**Consent page**
- An MCP-specific consent page that clearly communicates what the agent will be able to do.
- Should show the scope/permissions being granted in user-friendly terms.

**Discovery endpoints (`.well-known`)**
- `/.well-known/oauth-authorization-server` (RFC 8414): Standard OAuth2 metadata. This is what enables clients to auto-discover the authorization server without manual configuration.
- `/.well-known/oauth-protected-resource` (RFC 9728): Protected resource metadata pointing to the MCP endpoint.
[...]

PR #8263, opened 2026-08-24
feat(mcp): add .well-known OAuth discovery endpoints

### LINKED ISSUE #7431: .well-known Discovery Endpoints for MCP OAuth
### Summary

Add `.well-known` discovery endpoints that let MCP clients auto-discover the OAuth authorization server and protected resource metadata without manual configuration. This is what bridges the gap between "paste a PAT" and "paste your FlowFuse URL and authenticate."

### Prerequisites

- #7429 (MCP endpoint handler, so the protected resource URL exists)

### Requirements

**`/.well-known/oauth-authorization-server` (RFC 8414)**
- Returns JSON metadata describing the OAuth2 authorization server
- Must include: `issuer`, `authorization_endpoint`, `token_endpoint`, `response_types_supported`, `code_challenge_methods_supported`, `grant_types_supported`
- Values derived from `app.config.base_url` so they're correct for both local dev and production/self-hosted
- No auth required (public endpoint)

**`/.well-known/oauth-protected-resource` (RFC 9728)**
- Returns JSON metadata describing the MCP endpoint as a protected resource
- Must include: `resource` (the MCP endpoint URL), `authorization_servers` (pointing to the authorization server metadata URL)
- No auth required (public endpoint)

**Route placement**
- Register in `forge/routes/wellKnown.js` (new file) or similar, outside the EE routes since discovery should work regardless of license tier
- Mounted at the root level so `.well-known` paths resolve correctly

**Tests**
- Both endpoints return valid JSON with the required fields
- URLs are correctly derived from `app.config.base_url`
- Endpoints are publicly accessible (no auth required)
- Response content types are `application/json`

> **Aside on EE gating:** These endpoints are placed outside EE routes so that discovery works regardless of license tier. If `.well-known` were EE-gated, non-EE instances would 404 on discovery and MCP clients would fail with a confusing error instead of a clean "not available." The actual MCP endpoint and OAuth flow remain EE-gated. Might be worth reconsidering if there's a reason to hide discovery entirely on non-EE instances.

### References
- RFC 8414: OAuth 2.0 Authorization Server Metadata
- RFC 9728: OAuth 2.0 Protected Resource Metadata
- POC reference: PR #7414, `forge/routes/wellKnown.js` (reference only)
- Parent story: #7423

andypalmi:
## Summary

Adds the OAuth discovery endpoints an MCP client reads before it can authenticate.

- `GET /.well-known/oauth-authorization-server` (RFC 8414): advertises the authorize, token and registration endpoints, the supported grant types (`authorization_code`, `refresh_token`), `code_challenge_methods_supported: ["S256"]`, and `token_endpoint_auth_methods_supported: ["none"]`.
- `GET /.well-known/oauth-protected-resource` (RFC 9728): advertises the protected resource (`/mcp`) and its authorization server.

Both endpoints are anonymous and derive their URLs from `base_url`.

First of a stack of three PRs adding OAuth2 for MCP agents.

Part of #7423. Closes #7431.

### DIFF
--- a/forge/ee/routes/mcp/index.js
+++ b/forge/ee/routes/mcp/index.js
@@ -7,6 +7,7 @@
  * @param {import('../../../forge').ForgeApplication} app
  */
 module.exports = async function (app) {
+    await app.register(require('./wellKnown'), { prefix: '/.well-known', logLevel: app.config.logging.http })
     await app.register(require('./registrations'), { prefix: '/api/v1/teams/:teamId/mcp', logLevel: app.config.logging.http })
     await app.register(require('./server'), { prefix: '/mcp', logLevel: app.config.logging.http })
 }
--- a/forge/ee/routes/mcp/server.js
+++ b/forge/ee/routes/mcp/server.js
@@ -31,6 +31,8 @@ module.exports = async function (app) {
     // Resolves the caller's identity and scope, or sends an error reply and returns null.
     async function resolveCaller (request, reply) {
         if (!request.session?.User) {
+            // RFC 9728 §5.1: point unauthenticated callers at the resource metadata.
+            reply.header('WWW-Authenticate', `Bearer resource_metadata="${app.config.base_url}/.well-known/oauth-protected-resource/mcp"`)
             reply.code(401).send({ code: 'unauthorized', error: 'Unauthorized' })
             return null
         }
@@ -55,7 +57,7 @@ module.exports = async function (app) {
     // POST serves the MCP Streamable HTTP protocol in JSON mode: each request is
     // forwarded to the gateway and its response returned. Notifications (no id)
     // are acknowledged; the gateway populates its session on the first request.
-    app.post('/', async (request, reply) => {
+    app.post('/', { config: { allowAnonymous: true } }, async (request, reply) => {
         const caller = await resolveCaller(request, reply)
         if (!caller) {
             return
--- a/forge/ee/routes/mcp/wellKnown.js
+++ b/forge/ee/routes/mcp/wellKnown.js
@@ -1,0 +1,34 @@
+module.exports = async function (app) {
+    // RFC 9728: OAuth 2.0 Protected Resource Metadata for the /mcp resource.
+    // Lives in the EE mcp plugin so it is only advertised where /mcp exists.
+    function protectedResourceMetadata () {
+        const baseUrl = app.config.base_url
+        return {
+            resource: `${baseUrl}/mcp`,
+            authorization_servers: [baseUrl]
+        }
+    }
+
+    const schema = {
+        tags: ['Authentication', 'X-HIDDEN'],
+        response: {
+            200: {
+                type: 'object',
+                properties: {
+                    resource: { type: 'string' },
+                    authorization_servers: { type: 'array', items: { type: 'string' } }
+                }
+            }
+        }
+    }
+
+    // RFC 9728 §3.1: clients derive the metadata URL by inserting the resource
+    // path, so the /mcp resource is served at oauth-protected-resource/mcp. The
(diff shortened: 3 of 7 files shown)


cstns (changes requested):
### `oauth-protected-resource` won't scale past the one resource

  Two things about this one, they're related and I think the same fix sorts out both.

  **1. One document can only describe one resource.** RFC 9728 has no "multiple resources" shape, so the document describes exactly one resource identifier, which
  here is hardcoded to `${base_url}/mcp`. As soon as there's a second protected resource (per team MCP endpoint, a separate REST resource, whatever) there's nowhere
  to put it and the root document has to pick a winner.

  **2. The document is at the wrong path for the resource it describes.** RFC 9728 §3.1 uses path insertion, so metadata for `https://host/mcp` belongs at
  `https://host/.well-known/oauth-protected-resource/mcp`. The bare root path is reserved for a resource whose identifier is exactly `https://host`. Right now a
  client fetching the root gets back `resource: "https://host/mcp"`, which doesn't match what it asked for, and stricter clients reject that mismatch. Current MCP
  clients will probably still work because the spec has them try the path inserted URL first and fall back to the root, so we're leaning on the fallback rather than
  on being correct.

  Might be worth having each protected resource serve its own document at its own path:

  ```
  /.well-known/oauth-protected-resource/mcp     # owned by the mcp plugin
  /.well-known/oauth-protected-resource         # optional alias, if we want the fallback to keep working
  ```

  That also happens to fix a third thing: this plugin is registered unconditionally in `forge/routes/index.js`, but `/mcp` only exists under EE with an active license
  (`forge/ee/routes/mcp/index.js:11`). On OSS or an unlicensed install we're advertising a resource that 404s. If the document moves into the EE mcp plugin it gets
  the right path and the right gating for free. The `oauth-authorization-server` half can stay where it is, one issuer per deployment, so the root path is correct
  there and stays correct.

  One more, since it's the bit clients actually rely on: RFC 9728 §5.1 and the MCP spec have the resource answer an unauthenticated request with

  ```
  401 WWW-Authenticate: Bearer resource_metadata="https://host/.well-known/oauth-protected-resource/mcp"
  ```

  so the client never has to guess a path at all. I couldn't find `WWW-Authenticate` anywhere in the stack. Happy for it to land in a later PR, just flagging it so it
  doesn't fall through the gaps.

  ### So, roughly

  - [ ] Serve the doc at `/.well-known/oauth-protected-resource/mcp`, keep the root path as an alias if we want the fallback
  - [ ] Move it into the EE mcp plugin so it's license gated alongside the resource it describes
  - [ ] Have `/mcp` return `401` + `WWW-Authenticate: Bearer resource_metadata="..."` (later PR is fine)

andypalmi:
Thanks for the thorough review, all three make sense and I have pushed a commit addressing them.

1. Path insertion: the protected resource document is now served at `/.well-known/oauth-protected-resource/mcp` per RFC 9728 section 3.1, with the bare `/.well-known/oauth-protected-resource` kept as an alias.

2. Licensing: I moved the protected resource document out of the root well-known handler into a new plugin under `forge/ee/routes/mcp`, so it is only advertised where the `/mcp` resource actually exists. The authorization server metadata stays at the root since the issuer is per deployment.

3. WWW-Authenticate: I went ahead and added it here rather than deferring. Unauthenticated requests to `/mcp` now return 401 with `WWW-Authenticate: Bearer resource_metadata="..."` pointing at the path inserted document (RFC 9728 section 5.1). This meant marking the route anonymous at the auth layer so the handler emits the challenge instead of the generic session check.

Tests cover the new metadata paths, the alias, and the challenge header. Happy to split point 3 into its own PR if you would prefer.

cstns (approved):

PR #8264, opened 2026-08-24
feat(mcp): OAuth2 PKCE flow with dynamic client registration for MCP agents

### LINKED ISSUE #7432: OAuth2 PKCE Flow Extension for MCP Clients
### Summary

Extend the existing OAuth2 PKCE flow in `forge/routes/auth/oauth.js` to handle MCP agent clients. This adds an `mcp-agent` client path alongside the existing `ff-plugin` path, with localhost redirect URI validation, MCP-specific scope handling, and token issuance that produces scoped PATs (from #7411).

### Prerequisites

- #7431 (`.well-known` discovery endpoints)
- #7411 (PAT scopes) is assumed to be implemented. The OAuth flow issues scoped PATs as the token type for MCP.

### Requirements

**MCP client registration (hardcoded)**
- Add `mcp-agent` as a recognized `client_id` in the OAuth flow, following the `ff-plugin` pattern
- No database record, no `AuthClient` model changes

**`/account/authorize` extension**
- When `client_id=mcp-agent`:
  - Validate `redirect_uri` is a loopback address: `http://localhost` or `http://127.0.0.1`, any port, any path (per RFC 8252 Section 7.3)
  - Accept an MCP-specific scope value (e.g. `mcp-agent`)
  - Redirect to an MCP consent page route (frontend implementation in a separate task). The route pattern should be distinct from the existing `/account/request/:requestId` and `/account/request/:requestId/editor` paths.

**`/account/complete/:code` extension**
- When the request originates from an `mcp-agent` client:
  - Skip the project/device ownership checks (MCP is user-scoped, not resource-scoped)
  - Admin user handling is resolved in #7411: PATs with `adminOptIn = true` are rejected from MCP (#7429). Admin users can use MCP with tokens that have `adminOptIn = false` (the token behaves as a regular user scoped to their team role).
  - Generate the authorization code and redirect to the client's `redirect_uri` as normal

**`/account/token` extension**
- When `client_id=mcp-agent` and `grant_type=authorization_code`:
  - Validate PKCE code_verifier as normal
  - Issue a scoped PAT (from #7411) tied to the user, with the scope (read-only vs full access) and team selection that was approved during consent
  - Return `access_token`, `expires_in`, `refresh_token`
- When `client_id=mcp-agent` and `grant_type=refresh_token`:
  - Refresh the token following the existing refresh pattern

**Tests**
- Full OAuth flow: authorize -> consent -> complete -> token exchange returns a valid scoped PAT
- Redirect URI validation: localhost and 127.0.0.1 on various ports are accepted, non-loopback URIs are rejected
- Invalid scope is rejected at the authorize step
- PKCE verification works correctly for MCP client
- Refresh token flow works for MCP-issued tokens
- Existing `ff-plugin` and dynamic client flows are unaffected

### References
- Existing OAuth2 implementation: `forge/routes/auth/oauth.js`
- `ff-plugin` client pattern: `oauth.js:91-117` (authorize), `oauth.js:179` (complete), `oauth.js:407-439` (token)
- RFC 8252 Section 7.3: Loopback redirect URIs for native apps
- PAT scope model: #7411
- Parent story: #7423

andypalmi:
## Summary

Adds the OAuth2 authorization code flow with PKCE for MCP clients, plus dynamic client registration so third-party agents can obtain a `client_id` without being provisioned ahead of time.

- `POST /account/client` (RFC 7591): registers a public MCP client and returns a generated `client_id` with `token_endpoint_auth_method: none` and no client secret. Redirect URIs are validated (loopback with port flexibility per RFC 8252, otherwise HTTPS).
- The authorize, consent, token and refresh handlers recognise MCP clients by looking up the registered client rather than a fixed id.
- `AuthClient` gains `type`, `name` and `redirectURIs` (migration + model) with a `createMCPClient` controller.
- The authorization code exchange mints a scoped access token for the user, honouring the access level and teams chosen at consent.
- Adds a `platform_get_active_user` tool so an agent can resolve its own user id and, alongside the profile, read back what its token is allowed to do (`token.readOnly`, `token.allTeams`, `token.teams`). This is the introspection counterpart to the scoped token above, letting an agent check up front whether a requested action is within the session's access. With no scope (the first-party Expert path) it reports the user's own unrestricted access, since Expert acts with the same permissions the user has.

The consent UI is added in the next PR in the stack. See the issue for the note on why dynamic registration is used instead of a hardcoded client id.

Stacked on #7431. Part of #7423. Closes #7432.

> [!IMPORTANT]
> This PR introduces a **database change**. It includes a migration that runs on upgrade

### DIFF
--- a/forge/db/controllers/AccessToken.js
+++ b/forge/db/controllers/AccessToken.js
@@ -265,6 +265,37 @@ module.exports = {
         await app.settings.set('platform:stats:token', false)
     },
 
+    createMCPOAuthToken: async function (app, userId, { readOnly = false, teamIds = [] } = {}) {
+        const token = generateToken(32, 'ffpat')
+        const refreshToken = generateToken(32, 'ffpat')
+        const expiresAt = Date.now() + DEFAULT_TOKEN_SESSION_EXPIRY
+
+        await app.db.sequelize.transaction(async (t) => {
+            const tok = await app.db.models.AccessToken.create({
+                name: 'MCP Agent',
+                token,
+                refreshToken,
+                scope: '',
+                expiresAt,
+                readOnly,
+                adminOptIn: false,
+                ownerId: '' + userId,
+                ownerType: 'user'
+            }, { transaction: t })
+
+            if (teamIds.length > 0) {
+                const scopes = teamIds.map(teamId => ({
+                    AccessTokenId: tok.id,
+                    TeamId: app.db.models.Team.decodeHashid(teamId),
+                    UserId: userId
+                }))
+                await app.db.models.AccessTokenTeamScope.bulkCreate(scopes, { transaction: t })
+            }
+        })
+
+        return { token, expiresAt, refreshToken }
+    },
+
     createPersonalAccessToken: async function (app, user, scope, expiresAt, name, { readOnly = false, adminOptIn = false, teamIds = [] } = {}) {
         const userId = typeof user === 'number' ? user : user.id
         const token = generateToken(32, 'ffpat')
--- a/forge/db/controllers/AccessToken.js
+++ b/forge/db/controllers/AccessToken.js
@@ -4,6 +4,15 @@ const { generateToken, generateNumericToken, sha256, randomPhrase } = require('.
 
 const DEFAULT_TOKEN_SESSION_EXPIRY = 1000 * 60 * 30 // 30 mins session - with refresh token support
 
+const DEFAULT_REFRESH_TOKEN_EXPIRY = 1000 * 60 * 60 * 24 * 30 // 30 days - sliding refresh token lifetime
+
+// Concurrent refreshes of the same stable refresh token reuse the most recently
+// minted access token from this shared cache (Valkey in production) rather than
+// each minting a new one and overwriting the row. Re-mint once the cached token
+// is within this window of expiry so a client is never handed a token about to die.
+const MCP_ACCESS_TOKEN_CACHE = 'mcp-oauth-access-token'
+const MCP_ACCESS_TOKEN_REMAINING_LIMIT = 1000 * 60 // 60 seconds
+
 const DEFAULT_DEVICE_OTC_EXPIRY = 1000 * 60 * 60 * 24 // 24 hours
 
 /*
@@ -269,6 +278,7 @@ module.exports = {
         const token = generateToken(32, 'ffpat')
         const refreshToken = generateToken(32, 'ffpat')
         const expiresAt = Date.now() + DEFAULT_TOKEN_SESSION_EXPIRY
+        const refreshTokenExpiresAt = Date.now() + DEFAULT_REFRESH_TOKEN_EXPIRY
 
         await app.db.sequelize.transaction(async (t) => {
             const tok = await app.db.models.AccessToken.create({
@@ -277,6 +287,7 @@ module.exports = {
                 refreshToken,
                 scope: '',
                 expiresAt,
+                refreshTokenExpiresAt,
                 readOnly,
                 adminOptIn: false,
                 ownerId: '' + userId,
@@ -433,8 +444,14 @@ module.exports = {
 
     refreshToken: async function (app, refreshToken) {
         const existingToken = await app.db.models.AccessToken.byRefreshToken(refreshToken)
-        if (existingToken) {
-            const [prefix] = refreshToken.split('_')
+        if (!existingToken) {
+            return null
+        }
+        const [prefix] = refreshToken.split('_')
+
+        // Tokens without their own refresh lifetime (e.g. editor sessions) rotate
+        // the refresh token on each use.
+        if (!existingToken.refreshTokenExpiresAt) {
             const tokenUpdates = {
                 token: generateToken(32, prefix),
                 refreshToken: generateToken(32, prefix),
@@ -443,7 +460,35 @@ module.exports = {
             await app.db.models.AccessToken.update(tokenUpdates, { where: { refreshToken: existingToken.refreshToken } })
             return tokenUpdates
         }
-        return null
+
+        // A refresh token with its own lifetime is no longer valid once that
+        // lifetime has passed - remove it rather than issuing a new access token.
+        if (existingToken.refreshTokenExpiresAt.getTime() < Date.now()) {
+            await existingToken.destroy()
+            return null
+        }
+
+        // These tokens keep a stable refresh token, so concurrent refreshes can
+        // reuse the access token most recently minted for it: the first mint
+        // populates the shared cache and the rest return the same token instead of
+        // overwriting the row. If a truly simultaneous mint slips past the cache,
+        // the stale side simply refreshes again with its unchanged refresh token.
+        const cache = app.caches?.getCache?.(MCP_ACCESS_TOKEN_CACHE, { ttl: DEFAULT_TOKEN_SESSION_EXPIRY, max: 10000 })
+        const cacheKey = sha256(refreshToken)
+        const cached = await cache?.get(cacheKey)
+        if (cached && cached.expiresAt - Date.now() > MCP_ACCESS_TOKEN_REMAINING_LIMIT) {
+            return { token: cached.token, expiresAt: cached.expiresAt, refreshToken }
+        }
+
+        const token = generateToken(32, prefix)
+        const expiresAt = Date.now() + DEFAULT_TOKEN_SESSION_EXPIRY
+        await app.db.models.AccessToken.update(
+            { token, expiresAt, refreshTokenExpiresAt: Date.now() + DEFAULT_REFRESH_TOKEN_EXPIRY },
+            { where: { refreshToken: existingToken.refreshToken } }
+        )
+        await cache?.set(cacheKey, { token, expiresAt })
+        // The refresh token is unchanged, so hand back the one the client presented.
+        return { token, expiresAt, refreshToken }
     },
 
     /**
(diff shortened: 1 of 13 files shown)


cstns (commented):
One thing this makes visible: with dynamic registration, third party agents become the normal case, and right now the audit log can't tell them apart from the Expert.

  Every tool call, Expert or third party, goes through the same platform-automation:forge handler, and platformAutomation.js hardcodes source: 'mcp:expert' on the nonce. So a third party agent's actions show up as "via Expert" in the audit log. The frontend already has a source === 'mcp' branch ("via MCP")
  but nothing ever emits it.

  The client identity is also dropped on the way through. Registration captures client_name onto AuthClient.name, then createMCPOAuthToken hardcodes the token name to 'MCP Agent' and keeps no reference back to the client, so every agent looks identical in both the audit log and the user's token list. The
  nonce carries no tokenId either, and inject() swaps in the Expert's platform token, so there's no indirect signal left. Only the direct REST path records a tokenId (as source: 'api').

  Would be good to land this branch with the three cases clearly separated in the log:
  - mcp:expert for the first party Expert
  - mcp for a third party agent, ideally with the registered client name so you can see which agent
  - api for a plain token calling the REST API directly

  server.js already threads scope down to platformAutomation, so that same seam might be the natural place to carry the caller identity, and storing the client on the token at issuance (or at least using client_name instead of 'MCP Agent') would give the log something to name. Happy to go a different route
  if you have one in mind.

andypalmi:
Good catch, and I agree the three cases should be distinguishable. I have confirmed the gaps: the platform-automation handler hardcodes source 'mcp:expert' for every tool call, the issued token is named 'MCP Agent' with no link back to the AuthClient, and the frontend 'mcp' branch is never emitted.

To keep this PR focused on the OAuth mechanics I have opened a dedicated follow-up, #8271, to land the attribution end to end: link the issued token to its AuthClient and name it after the registered client, then separate the runtime source so a third party agent logs as 'mcp' with the client name, 'mcp:expert' stays for the Expert, and a direct token stays 'api'. The runtime part threads caller identity through the gateway and comms seam alongside scope and overlaps with the audit trail work, so it reads cleaner on its own.

Let me know if you would rather see any of it in this PR instead.

andypalmi:
Good call on this. #8272 now implements the full attribution, stacked on these PRs.

The MCP door records the calling client against its session, and the platform-automation handler stamps audit source `mcp` with the client name for a third-party agent, while the first-party Expert keeps `mcp:expert`. The issued token is also named after its registered client and linked to it, so the two are distinguishable in both the audit log and the user's token list.

#8271 tracks the whole thing and #8272 closes it.

forge/db/controllers/AccessToken.js
```diff
@@ -412,7 +460,35 @@ module.exports = {
             await app.db.models.AccessToken.update(tokenUpdates, { where: { refreshToken: existingToken.refreshToken } })
             return tokenUpdates
         }
-        return null
+
+        // A refresh token with its own lifetime is no longer valid once that
+        // lifetime has passed - remove it rather than issuing a new access token.
+        if (existingToken.refreshTokenExpiresAt.getTime() < Date.now()) {
```
cstns, line 463: Could we fall into a trap here where a malicious MCP client figures out that issuing a refreshToken request on a token without refreshTokenExpiresAt will automatically extend its lifetime? I'm worried about a time escalation attack surface.

forge/db/migrations/20260824-01-add-mcp-authclient-fields.js
```diff
@@ -0,0 +1,39 @@
...
+            defaultValue: null
+        })
+        await context.addColumn('AuthClients', 'redirectURIs', {
+            type: DataTypes.TEXT,
+            allowNull: true,
+            defaultValue: null
+        })
+    },
+    down: async (context) => {
+        await context.removeColumn('AuthClients', 'type')
```
cstns: nice to have but we don't do down anywhere else

forge/db/migrations/20260825-01-add-refreshTokenExpiresAt-to-AccessTokens.js
```diff
@@ -0,0 +1,29 @@
...
+module.exports = {
+    up: async (context) => {
+        await context.addColumn('AccessTokens', 'refreshTokenExpiresAt', {
+            type: DataTypes.DATE,
+            allowNull: true,
+            defaultValue: null
+        })
+    },
+    down: async (context) => {
+        await context.removeColumn('AccessTokens', 'refreshTokenExpiresAt')
```
cstns: we don't do down

forge/ee/lib/mcp/tools/users.js
```diff
@@ -0,0 +1,34 @@
+module.exports = [
+    {
+        name: 'platform_get_active_user',
```
cstns, line 3: i know there's a need for such a tool, I also identified it's absence but why include it in this PR?

andypalmi:
Heads up for reviewers: the refresh-token handling introduced here is superseded by #8273, which replaces the non-rotating refresh token with proper rotation (grace window plus replay detection, per RFC 9700 section 4.14.2). The final refresh behaviour to review lives in #8273 rather than this PR.

cstns (changes requested):

forge/db/migrations/20260824-01-add-mcp-authclient-fields.js
andypalmi: good catch, making the down a no op like the others

forge/db/migrations/20260825-01-add-refreshTokenExpiresAt-to-AccessTokens.js
andypalmi: good catch, making the down a no op like the others

forge/db/controllers/AccessToken.js
andypalmi: The expiry is set server-side at issuance, not by the client. Every MCP token is minted through createMCPOAuthToken, which always sets refreshTokenExpiresAt, so an MCP client can never present a token that lands in the non-expiring branch. That path only serves editor/device tokens, which have never had a refresh lifetime. So there is no way for a client to extend its own lifetime here.

forge/ee/lib/mcp/tools/users.js
andypalmi: It's the introspection counterpart to the scoped token this PR issues, letting an agent read back what its token can actually do.

forge/db/migrations/20260824-01-add-mcp-authclient-fields.js
```diff
@@ -0,0 +1,35 @@
...
+ *   type          - 'mcp' for dynamically registered MCP clients, null otherwise
+ *   name          - the client_name supplied at registration
+ *   redirectURIs  - JSON array of redirect URIs the client may use
+ */
+
+const { DataTypes } = require('sequelize')
+
+module.exports = {
+    up: async (context) => {
+        await context.addColumn('AuthClients', 'type', {
```
cstns: we could ditch this column in favor of the existing ownerType column which serves the same purpose and is already freeform text; we can re-map the object when presenting it to 3rd party clients involved in the oauth flow

forge/db/controllers/AccessToken.js
```diff
@@ -4,6 +4,15 @@ const { generateToken, generateNumericToken, sha256, randomPhrase } = require('.
...
 const DEFAULT_TOKEN_SESSION_EXPIRY = 1000 * 60 * 30 // 30 mins session - with refresh token support
 
+const DEFAULT_REFRESH_TOKEN_EXPIRY = 1000 * 60 * 60 * 24 * 30 // 30 days - sliding refresh token lifetime
+
+// Concurrent refreshes of the same stable refresh token reuse the most recently
+// minted access token from this shared cache (Valkey in production) rather than
+// each minting a new one and overwriting the row. Re-mint once the cached token
+// is within this window of expiry so a client is never handed a token about to die.
+const MCP_ACCESS_TOKEN_CACHE = 'mcp-oauth-access-token'
+const MCP_ACCESS_TOKEN_REMAINING_LIMIT = 1000 * 60 // 60 seconds
```
cstns: we should increase this limit to at least 60 minutes

cstns (approved):
tested e2e with my local claude code, works as expected; we should sync with @knolleary to validate the ui for the oauth consent page

cstns:
my bad, the consent page was on the next stack up

andypalmi:
Verified the refresh flow end to end against a real MCP client, on this PR's code (without the later rotation work).

Steps: authorized the client to obtain a token pair, expired the access token in the database while leaving the refresh token valid, then made another tool call. The door rejected the expired access token without destroying the row, the client refreshed via `grant_type=refresh_token`, and the call succeeded transparently. The row came back with a new access token, the same refresh token, and a fresh 30 minute expiry, which is the non-rotating behaviour this PR implements.

Also confirmed non-MCP OAuth is unaffected: the refresh handling is gated on `refreshTokenExpiresAt`, which only MCP tokens set, so editor, device and ff-plugin sessions take the same code path as before.

forge/db/migrations/20260824-01-add-mcp-authclient-fields.js
andypalmi: Should I?

forge/db/controllers/AccessToken.js
andypalmi: Can't go to 60min: this is the re-mint margin against a 30min token lifetime, so at 60min the cache is never reused and every refresh overwrites the row (the race the cache prevents). Set it to 5min, which keeps coalescing and still re-mints before expiry.

forge/db/migrations/20260824-01-add-mcp-authclient-fields.js
andypalmi: Done, dropped `type` and switched MCP clients to `ownerType='mcp'`.

PR #8272, opened 2026-08-25
feat(mcp): attribute third-party MCP audit actions

### LINKED ISSUE #8271: Attribute MCP audit-log actions to the calling agent
## Why

With dynamic client registration, third-party agents become the normal caller of the platform MCP tools, but the audit log cannot tell them apart from the first-party Expert. Every tool call, Expert or third-party, goes through the same platform-automation handler, which hardcodes `source: 'mcp:expert'` on the nonce, so a third-party agent's actions show up as "via Expert". The client identity is dropped too: registration captures `client_name` onto `AuthClient.name`, but the issued token is hardcoded to the name `MCP Agent` with no reference back to the client, so every agent looks identical in both the audit log and the user's token list.

The frontend already renders the cases (`mcp:expert` gives "via Expert", `mcp` gives "via MCP"), but nothing ever emits `mcp`.

## Goal

Separate the three cases clearly in the audit log:

- `mcp:expert` for the first-party Expert.
- `mcp` for a third-party agent, with the registered client name so it is clear which agent acted.
- `api` for a plain token calling the REST API directly (already the case).

## Proposal

- Link the issued MCP OAuth token to its `AuthClient` and name it after the registered client instead of `MCP Agent`. The authorization request already carries `client_id` at the point the token is minted, so the client name and id are available. This alone fixes the user's token list and gives any token-id based attribution something to name.
- Carry the caller identity through the same runtime seam that already threads `scope` (the platform MCP door through to the platform-automation handler), so the nonce source is `mcp` for a third-party agent and `mcp:expert` for the Expert, and the client name reaches the audit log.
- Surface the client name in the audit entry for the `mcp` case.

## Acceptance criteria

- A third-party agent's platform tool call is logged with source `mcp` and the registered client name, not "via Expert".
- The first-party Expert's tool calls remain `mcp:expert`.
- A direct REST token call remains `api`.
- The issued MCP token is named after its registered client and is linked to that client.

## Notes

- Follows #8264 (OAuth2 PKCE and dynamic client registration). Part of #7423.
- The token to client link and naming is self-contained. The runtime source separation depends on the caller identity propagating across the gateway and comms seam, so it can land as a second step if that plumbing is not yet in place.

Part of #7423

andypalmi:
## Summary

Distinguishes third-party MCP agent platform tool calls from the first-party Expert in the audit log, and runs a third-party call under the caller's own Personal Access Token rather than a freshly minted platform token.

- The MCP door stashes the caller's PAT against its `mcpSessionId`. On a cache hit the platform-automation handler injects that token for the tool call, so the action runs under the caller's own credential, and stamps the audit source as `mcp`.
- The first-party Expert path never opens that door, so it has no cached token: it mints a scope-capped platform token as before and keeps the `mcp:expert` source.
- The audit log renders `via MCP` for a third-party source and `via Expert` for the first-party path.

Distinguishing individual third-party clients by name in the audit log is deferred to a follow-up.

Part of #7423. Closes #8271.


forge/ee/routes/mcp/server.js
```diff
@@ -82,6 +89,10 @@ module.exports = async function (app) {
         }
 
         const mcpSessionId = request.headers['mcp-session-id'] || randomUUID()
+        if (caller.clientName) {
+            const cache = app.caches?.getCache?.(MCP_SESSION_SOURCE_CACHE, { ttl: MCP_SESSION_SOURCE_CACHE_TTL, max: 10000 })
+            await cache?.set(mcpSessionId, { clientName: caller.clientName })
```
cstns: you need to store the pat here and then retrieve it for later usage in inject calls

forge/comms/platformAutomation.js
```diff
@@ -154,10 +167,11 @@ class PlatformAutomationHandler {
                 if (user) {
                     const { token } = await this.app.expert.mcp.getOrCreatePlatformToken(user)
```
cstns: you don't need to create a token for 3d party, you can use the stored pat from the cache

forge/comms/platformAutomation.js
```diff
@@ -109,6 +115,13 @@ class PlatformAutomationHandler {
             let result = {}
             this.app.log.info(`platform-automation request: userId=${userId} mcpSessionId=${mcpSessionId} command=${command} tool=${data?.name || 'n/a'}`)
 
+            // A cache hit means a third-party agent opened this session at the MCP
+            // door; a miss keeps the 'mcp:expert' default for the first-party Expert
+            // path, which never goes through that door.
+            const sessionSourceCache = this.app.caches?.getCache?.(MCP_SESSION_SOURCE_CACHE)
+            const sessionSource = mcpSessionId ? await sessionSourceCache?.get(mcpSessionId) : null
```
cstns: this should be the pat

forge/db/migrations/20260825-02-add-authclient-to-access-token.js
```diff
@@ -0,0 +1,27 @@
+/**
```
cstns: we don't need this

forge/db/controllers/AccessToken.js
```diff
@@ -274,15 +274,16 @@ module.exports = {
...
         const token = generateToken(32, 'ffpat')
         const refreshToken = generateToken(32, 'ffpat')
         const expiresAt = Date.now() + DEFAULT_TOKEN_SESSION_EXPIRY
         const refreshTokenExpiresAt = Date.now() + DEFAULT_REFRESH_TOKEN_EXPIRY
 
         await app.db.sequelize.transaction(async (t) => {
             const tok = await app.db.models.AccessToken.create({
-                name: 'MCP Agent',
+                name: client?.name || 'MCP Agent',
+                AuthClientId: client?.clientID || null,
```
cstns: we don't need this

forge/routes/auth/index.js
```diff
@@ -156,11 +156,24 @@ async function init (app, opts) {
                                 }))
                             }
 
+                            // Only tokens issued to a registered MCP client carry an
+                            // AuthClientId, so the client name lookup is skipped for
+                            // every other token rather than joined on each request.
+                            let clientName = null
+                            if (accessToken.AuthClientId) {
```
cstns: we don't need this

frontend/src/components/audit-log/AuditEntry.vue
```diff
@@ -93,7 +93,9 @@ export default {
                 return toolName ? `via Expert (Tool name: ${toolName})` : 'via Expert'
             }
             if (this.entry.source === 'mcp') {
-                return toolName ? `via MCP (Tool name: ${toolName})` : 'via MCP'
+                const clientName = this.entry.body?.sourceContext?.clientName
```
cstns: we don't need to distinguish between clients, only expert/third party/api

cstns (changes requested):
a lot of overhead for something that had it's grown work laid out

cstns (approved):
looking amazing, tested out perfectly!

PR #8273, opened 2026-08-25
feat(mcp): rotate MCP OAuth refresh tokens with replay detection

### LINKED ISSUE #8270: Rotate MCP OAuth refresh tokens with replay detection
## Why

The MCP OAuth flow issues each client two credentials: a short-lived access token (sent on every /mcp call) and a longer-lived refresh token (used only to mint a new access token when the old one expires). Today the refresh token is stable: every refresh mints a new access token but returns the same refresh token unchanged.

A stable refresh token is a long-lived bearer credential. If it leaks (a log, disk, a stolen laptop) an attacker can mint access tokens for its full 30 day lifetime, and nothing signals the theft, because the real client and the attacker present the identical token and both succeed. RFC 9700 (OAuth Security BCP) section 4.14.2 recommends that public clients rotate refresh tokens with replay detection instead:

- Rotation: each refresh returns a new refresh token and invalidates the old one, so a refresh token is single use.
- Replay detection: if an already rotated-out refresh token is presented again, two copies exist (a likely theft), so the whole token family (the original grant and everything derived from it) is revoked and the client must re-authorize. A leaked token is then usable only until the next legitimate refresh, and the collision surfaces the breach.

## The race this has to handle

Naive rotation breaks legitimate clients, because two refreshes with the same token happen normally:

- Concurrent: a client fires several tool calls at once; if the access token just expired they all try to refresh together with the current refresh token. The first rotates it, the rest now hold a used token and get falsely flagged as replay.
- Retried: a refresh response is lost to a timeout and the client retries with the same token; the server already rotated on the first attempt, so the retry looks like replay.

Either false positive would revoke the family and eject a well-behaved user. The stable design shipped first precisely to avoid this. Rotation must therefore keep a short grace window: the immediately previous refresh token stays valid for a few seconds and returns the current tokens, and only a token used after that window counts as a genuine replay.

## Proposal

- Rotate the refresh token on each successful refresh and mark the presented one as rotated.
- Accept the immediately previous refresh token within a short grace window, returning the current tokens, so concurrent and retried refreshes still succeed.
- Treat a refresh token presented after its grace window as a replay: revoke the entire token family and require re-authorization.

## Acceptance criteria

- A successful refresh returns a new refresh token; the previous one stops working once its grace window elapses.
- Two near-simultaneous refreshes with the same token both succeed and neither triggers a revocation.
- A rotated-out token presented after the grace window revokes the family and returns an OAuth error, and that family's access tokens stop authenticating.
- Token types without their own refresh lifetime (for example editor sessions) are unchanged.
- Rotation and grace-window state lives in the shared cache, so it works on a single replica (docker compose) and multiple replicas (Helm).

## Notes

- Fifth PR in the epic, stacked on the consent-page PR; depends on the refresh-lifetime and coalescing changes already merged.
- If rotation, grace window, and family revocation are too large for one review, split into a rotation PR and a replay-detection PR under this issue.

Part of #7423

andypalmi:
## Summary

Rotates the MCP OAuth refresh token on every refresh, with a grace window and replay detection, replacing the stable (non-rotating) refresh token.

## How it works

- Each refresh issues a new refresh token and retires the presented one into a dedicated `AccessTokenRefreshRotations` table, keyed by a unique hash. The full lineage is kept, so a token retired several rotations ago is still recognised as belonging to the grant.
- A rotated-out token presented within a short grace window re-mints only a fresh access token and returns no `refresh_token`, so the client keeps its current one (RFC 6749 §6). A concurrent or retried refresh still succeeds.
- Presented after the window, a rotated-out token is treated as a replay: the grant is revoked (its lineage cascades away), an audit-log entry records the client and grant, and the client must re-authorize (RFC 9700 §4.14.2).
- Concurrent refreshes are serialised with a compare-and-swap on the current refresh token: the update matches only while the token is still current, so exactly one refresh rotates and the loser re-mints deterministically from the database.
- The refresh grant is rate limited and the presented-token lookup is indexed, so probing the endpoint with guessed tokens is throttled and does not scan the AccessTokens table.

## Scope

Only MCP OAuth grants are affected. Editor, ff-plugin, and device-editor sessions have no refresh lifetime and keep their existing rotate-each-use behavior with no replay handling.

## Tests

Unit and route specs cover rotation, the grace-window re-mint, replay after grace including a token retired several rotations back, and a concurrent refresh that neither returns null nor revokes the grant.

Part of #7423. Closes #8270.


cstns:
it's unclear to me what we're fixing here

cstns:
It looks to me that we're treating a symptom before it presented it's effects; could this be postponed to 3.1?

andypalmi:
Yes, let's postpone this. Agreed it is security hardening rather than something the OAuth flow needs to function, so I'll pull it out of the stack into its own PR and target it separately (3.1 works for me).

For context on what it does: the base refresh flow (exchanging a valid refresh token for a new access token) already lands in #8264 and is complete on its own. This PR adds hardening on top: rotating the refresh token on each use, a compare-and-swap so two concurrent refreshes converge on one result, a short grace window for retries, and replay detection that revokes the grant if an already-rotated refresh token is presented again (RFC 9700 section 4.14.2). None of that is required for refresh to work, which is why it can land later without blocking the rest of the stack.

cstns:
Took a careful read through the rotation logic and ran the Sequelize behavior locally to confirm the core mechanism holds up: static `update()` does run the hashing setters, so rotated tokens store hashed and resolve through `byRefreshToken`, and the compare-and-swap correctly lets exactly one of two racing refreshes win. The algorithm itself is solid. Most of what I want to raise is about the machinery around it, where the cache has quietly become a correctness dependency. A few of these I'd consider addressing before this merges.

**Plaintext refresh token in the shared cache.** On each refresh the cache stores `{ token, expiresAt, refreshToken }` with the access token and the new (up to 30 day) refresh token both in plaintext, TTL 30 minutes, in Valkey. The whole reason `AccessTokens` hashes tokens at rest is defeated for anything that can read that cache instance (co-tenant services, a `MONITOR` session, RDB/AOF dumps, backups). Pre-PR only a 30 minute access token was exposed this way, so this widens the blast radius to a long lived credential.

**`previousRefreshToken` has no index, and the route no longer rejects unknown MCP tokens up front.** So every MCP refresh with a non current token runs a `findOne` on an unindexed column, over a table holding every device, session, HTTP and PAT row. Someone looping random `refresh_token` values gets a sequential scan per request against the shared token endpoint. Might be worth adding `addIndex('AccessTokens', ['previousRefreshToken'])` in the migration (a partial index `WHERE previousRefreshToken IS NOT NULL` keeps it small on Postgres), plus a rate limit on the refresh grant.

**The grace window and the losing refresh both depend on the cache, and a miss can escalate to grant loss.** The rotation result's plaintext lives only in the cache, so on a cache miss (multi replica with the default in process memory cache, or a Valkey restart/eviction) a legitimate retry inside the grace window finds the row via `previousRefreshToken` but no cache entry, returns null, and once 60s pass the same retry hits `rotated.destroy()` and revokes the whole grant, killing an access token the client never received. Relatedly, the compare-and-swap loser reads the cache in the gap between the winner's DB commit and its `cache.set`, so a valid concurrent refresh can come back as a 400.

On that cluster, one option worth considering is to stop treating the cache as the source of truth for the grace window and let the DB be authoritative. Concretely: within the grace window, let the presented (rotated out) token re-mint a fresh access token for the same grant and return it, rather than trying to hand back the identical cached token. The old and new refresh tokens both stay valid until the grace window closes, then the old one is done. That removes the need to cache any plaintext at all, makes the losing refresh just re-read and re-mint deterministically, and means a slow or retried refresh succeeds without depending on Valkey being up. The tradeoff is that single use becomes single use with a short overlap window, which is a pretty standard rotation pattern and still satisfies the replay intent. If keeping strict single use matters, the alternative is to encrypt the pending tokens at rest (same app secret used elsewhere) so the hashing-at-rest property isn't broken, but the overlap approach is simpler and dissolves more of these at once.

A few smaller things while you're in here:

* Replay tracking only keeps the immediately previous token, so an attacker who refreshes twice (A to B to C) overwrites `previousRefreshToken`, and the victim's still held A then resolves to neither a grace recovery nor a revocation, just a plain 400. That leaves the attacker with a live grant and defeats the detection this PR is adding. Keying replay off the grant lineage (a rotation counter, or a small pruned set of retired hashes per grant) rather than only the last hash would close it.
* Legacy cache entries from the current code survive a rolling deploy under the same cache name and key, and they have no `refreshToken` field, so the first post deploy refresh on a still current token returns `refresh_token: undefined` and skips rotation for up to ~25 minutes. Versioning the cache key (or only using a hit when `cached.refreshToken` is present) avoids it.
* When a grant is revoked as a replay it happens silently. An audit log entry with the client and grant id would make a forced re-consent explainable to support later.
* `forge/ee/routes/mcp/server.js` caches the raw bearer per session, and rotation overwrites the access token column every refresh, so a proactive refresh between MCP requests can 401 in-flight tool calls until the next request rewrites the session cache. Only re-minting the access token when it's actually near expiry (while still rotating the refresh token) would cut that churn.
* Minor cleanups: the route's `isMcpClient` escape hatch duplicates the not-found handling the controller already owns and re-runs the same query, and the cache-hit result is rebuilt field by field at three sites while `MCP_ACCESS_TOKEN_REMAINING_LIMIT` is now dead under rotation.

Nice test coverage on the rotate/grace/replay paths overall, the concurrency test in particular is a good idea, though it currently tolerates a null result (`filter(Boolean)` then `>= 1`), which is masking the losing-refresh 400 above.

cstns (changes requested):
see comment above

andypalmi:
Thanks, this was a really useful review. I reworked it along the lines you suggested:

- Grace is now DB-authoritative. A within-grace refresh re-mints only an access token and omits `refresh_token` (RFC 6749 §6), so no token plaintext is cached anywhere. That removes the plaintext-in-cache exposure and the cache-miss to grant-loss path.
- Retired tokens now live in an indexed `AccessTokenRefreshRotations` table keyed by hash, keeping the full lineage, so the A to B to C replay is caught and resolving a presented token is an indexed lookup rather than a scan.
- Also rate-limited the refresh grant, added an audit-log entry on replay revoke, dropped the duplicate route lookup, and tightened the concurrency test so it no longer tolerates a null.

The one I didn't do is the access-token churn: returning the existing access token would mean storing its plaintext, so "only re-mint near expiry" isn't clean without multi-token support. Impact is low for reactive clients, so happy to track it as a follow-up unless you feel strongly.

forge/db/migrations/20260910-01-add-refresh-token-rotation-tracking.js
```diff
@@ -0,0 +1,54 @@
...
+        })
+
+        await context.addIndex('AccessTokenRefreshRotations', {
+            name: 'access_token_refresh_rotation_hash',
+            fields: ['tokenHash'],
+            unique: true
+        })
+    },
+    down: async (context) => {
+        await context.dropTable('AccessTokenRefreshRotations')
```
cstns: please remove, the repo convention is an empty down

forge/routes/auth/oauth.js
```diff
@@ -608,13 +612,13 @@ module.exports = async function (app) {
             if (!refresh_token) {
                 return badRequest(reply, 'invalid_request', 'Invalid refresh_token')
             }
-            const existingToken = await app.db.models.AccessToken.byRefreshToken(refresh_token)
-            if (!existingToken) {
-                badRequest(reply, 'invalid_request', 'Invalid refresh_token')
+            const rateCache = app.caches?.getCache?.('mcp-refresh-rate', { ttl: MCP_REFRESH_RATE_WINDOW, max: 100000 })
```
cstns, line 615: It's not actually a 60-second window, the lru-cache resets the entry TTL on every set(), so the counter only expires after 60 seconds of silence. Any IP with sustained traffic of one refresh per minute or more accumulates to 30 and then gets 429'd until a quiet gap occurs.

cstns (changes requested):
Rotation rows accumulate forever, one per refresh, only cleared when the grant dies. A grant refreshing every 30 minutes gathers ~17k rows/year. A housekeeper prune of rows older than the refresh-token lifetime (30 days) bounds growth while keeping replay detection meaningful.

cstns:
Also adding @knolleary as reviewer to get his blessing as this touches auth

cstns:
for more context, this turned into a prerequisite for https://github.com/FlowFuse/engineering/issues/272 via https://github.com/FlowFuse/engineering/issues/334

andypalmi:
All three addressed: the rate limit now anchors on a stored window start so a busy IP resets after the window instead of staying blocked, rotation rows are pruned in the expireTokens housekeeper past the refresh lifetime, and the migration down is now empty per convention.

cstns (approved):
approved from my end, waiting on @knolleary's pass

andypalmi:
@cstns I am going to update the migration date since main has a newer migration in and this migration needs to be last when we merge it

andypalmi:
Ok migration date updated https://github.com/FlowFuse/flowfuse/pull/8273/commits/db0e16ce17aeb25574f8659517b30adb6be6ade1

knolleary:
I asked claude to pick through it due to the complexity.

### Blocking
1. **The grace-window re-mint destroys the access token the winning refresh just issued** —`forge/db/controllers/AccessToken.js:502`

`AccessToken.update({ token, expiresAt }, { where: { id: grant.id } })` overwrites the row's single token column. Probed directly:

So with two concurrent refreshes, the client that won the CAS holds an access token that is dead the moment the loser's grace call lands. This is exactly the failure the cache being removed here existed to prevent (`// Concurrent refreshes of the same refresh token reuse the cached access token rather than each minting a new one and overwriting the row`). The new concurrency test masks it — `live.length.should.be.aboveOrEqual(1)` passes with one of the two callers holding a dead token; it should assert both are live.

2. **The grace window doesn't rescue the stranded client, it just delays revoking its grant by 60s** — same function

A client only lands in the grace path because it is still holding the retired token (lost response, retry, or parallel process). Returning no `refresh_token` leaves it holding that retired token, and its next refresh — 30 minutes later, far outside the 60s grace — is classified as a replay. 
A benign concurrent refresh therefore ends in the whole grant being revoked and the user forced to re-authorize. Both findings have one fix: during grace, return the same `{token, refreshToken}` pair the winner got. Caching the winner's response keyed by `sha256(retired token)` with TTL = the grace window does it, reuses the cache infrastructure this PR deletes, and keeps the security property (a replay past grace still revokes).

### Worth addressing

3. **The rate limit gates every refresh grant, not just MCP** — `forge/routes/auth/oauth.js:615-625`

The gate sits above the `client_id` resolution, so editor, device-editor and ff-plugin refreshes share one 30/min/IP budget. Device agents behind a single site NAT all present the same source IP; a synchronised reconnect of more than 30 of them in a minute gets 429s on a path with no retry semantics. Scope it to MCP clients, or key on `client_id + ip`.

4. **request.ip is attacker-controlled.** `forge/forge.js:159` sets `trustProxy: true`, so `request.ip` is the leftmost `X-Forwarded-For` entry on any request — a prober bypasses the cap by varying the header. It's also per-process under the default memory driver, so the real limit is 30 × instance count. The platform already registers `@fastify/rate-limit` with route-level `config.rateLimit` (used in `forge/routes/auth/index.js:333`); worth at least noting why this hand-rolled one is needed instead (I assume: because `rate_limits.enabled` is opt-in), and that both will apply when it is enabled.

5. **Rotation table growth.** One row per refresh (every 30 min per grant), kept 30 days ≈ 1,440 rows per active MCP grant. Replay attacks are realistically near-term; a shorter retention or a lineage depth cap would bound this.

6. **Two queries where one would do** — `forge/db/models/AccessToken.js:85`.  `byRotatedRefreshToken` is a finder on `AccessToken` that returns an `AccessTokenRefreshRotation`, and the controller then issues a second findOne for the grant. Either move it to the rotation model or include the AccessToken so one call returns both.

### Nits

 - The replay audit body carries the raw numeric `AccessToken.id` as grant; ids are hashid-encoded everywhere else they're user-visible.
 - info: 'MCP refresh token replay detected; grant revoked' is never read — `AuditEntryVerbose.vue` writes its own sentence.
 - No test for the new logger in `test/unit/forge/auditLog/user_spec.js]`, where the convention is one test per event.
 - The grace branch doesn't re-check `grant.refreshTokenExpiresAt`, so a just-expired grant can still mint an access token within the window.
 - The rate-limit test reaches into `rateCache.lru` (memory driver only) and depends on leaving a stale window behind so later tests in the file aren't throttled.

cstns (approved):

PR #8279, opened 2026-08-25
feat(mcp): forward telemetry attribution and self-hosted opt-out context

andypalmi:
## Summary

Forwards the attribution and self-hosted opt-out context the MCP gateway needs to emit one telemetry event per tool call, for both first-party and third-party callers.

- The third-party door (`forge/ee/routes/mcp/server.js`) attaches `telemetryEnabled`, `username`, `deployment`, `patId` and `teamId` as MQTT user properties, since the gateway can't see the database.
- `telemetry:enabled` (effective state, licence-aware) is exposed to every logged-in user, and `username`, `deployment`, `telemetryEnabled` are added to the Expert chat context so the self-hosted opt-out reaches first-party sessions too.
- A missing flag defaults telemetry off: only current emitters stamp it, so legacy self-hosted traffic stays off by default.

No message content or tool arguments are ever sent.

Part of FlowFuse/engineering#286. Closes FlowFuse/engineering#287.


forge/routes/api/settings.js
```diff
@@ -43,6 +43,10 @@ module.exports = async function (app) {
                 'user:offboarding-required': app.settings.get('user:offboarding-required'),
                 'user:offboarding-url': app.settings.get('user:offboarding-url'),
                 'user:team:trial-mode:projectType': app.settings.get('user:team:trial-mode:projectType'),
+                // Effective telemetry state, matching the MCP door: a licence forces it on,
+                // otherwise it follows the yml and admin opt-out.
+                'telemetry:enabled': app.license.active() || (app.config.telemetry?.enabled !== false && app.settings.get('telemetry:enabled') !== false),
+                'telemetry:anonymize': app.config.telemetry?.anonymize !== false,
```
cstns: you should follow the same pattern using `app.settings.get`; that's where the default should be set to true when the setting isn't explicitly set in conf see forge/config/index.js

cstns (approved):

PR #8289, opened 2026-08-26
Improve reliability of browser MCP session pinning

### LINKED ISSUE #8285: Cache driver parity and pin lifecycle
Third-party MCP clients pin their session to an open browser tab, and the platform keeps those pins in the shared cache. It works in local dev and not reliably in production, because the memory and redis/valkey drivers don't behave the same way. The redis driver silently ignores `updateAgeOnGet`, so pin entries expire mid-use. On top of that there's no way to find the pins belonging to a given browser session, so stale pins can survive for up to 30 minutes after a tab closes or after a pin moves to another tab.

The goal is pin state you can trust: both drivers behave identically, a pin lives exactly as long as the tab that owns it, and both directions of a pin expire together.

What's involved:

* `updateAgeOnGet` in the redis driver (`HGETEX`, with a fallback for servers that don't have it)
* a `scan(pattern)` method on both drivers, with glob escaping
* `BrowserSession` fixes: heartbeats refresh the tab's pins, re-pinning and tab close clean up the old entries, and reads keep both directions of a pin aligned
* validate against a real valkey instance, not just the mocked client, since the mock is what hid this in the first place

Steve-Mcl:
### TLDR

Makes browser MCP session pinning reliable. Pins now live exactly as long as the tab that owns them: heartbeats keep them alive, closing a tab or moving a pin cleans up immediately, and a closed tab can no longer be reported as present just because something kept polling it. Getting there required two cache-layer additions (pattern lookup and read-refresh TTLs on the redis driver), which are included here and validated against real Valkey and Redis servers.

### Detail

BrowserSession pinning:

- The presence cache no longer refreshes TTL on read. Only the tab's own heartbeat keeps it alive, so a closed tab expires even while consumers poll it.
- Pins are stored as a forward/reverse entry pair, and the pair now ages together: heartbeats and reads touch both directions, re-pinning to a new tab deletes the old reverse entry, and `removeSession` clears any pins referencing the closed tab (with a guard for pins that already moved elsewhere).
- `getActiveMcpSessions` returns the pinned MCP session ids (it previously returned raw cache keys) and `[]` when there are none.

Cache drivers (in support of the above):

- The redis driver now honours `updateAgeOnGet`. It uses `HGETEX` (one round trip) where the server supports it and falls back to `HGET` + `HPEXPIRE` after a single "unknown command" probe. Valkey 9.1 turns out to support `HGETEX`, so production takes the fast path (assuming valkey 9.1 / later versions of redis).
- Both drivers gain `scan(pattern)` with redis glob semantics (`*`, `?`, `[abc]`, `\` escape). The memory driver matches through a shared `globToRegExp` in `caches/util.js`, so the two drivers agree on what a pattern matches. An `escapeGlob()` helper lets IDs be embedded in patterns safely.
- Fixed `get()` so falsy values (`0`, `false`, `''`) round-trip instead of becoming `undefined`. The heartbeat error counter stores `0`, so this mattered.

### Tests

- New `BrowserSession_spec.js`: pin lifecycle against the real memory driver. Pin/resolve round trips, user scoping, re-pin cleanup, `removeSession` clearing presence and pins (verified against the raw cache keys, not just the API), the moved-pin race guard, glob characters in session and user ids, and heartbeat pin refresh.
- New `util_spec.js`: ~40 extensive regression cases for `globToRegExp` (wildcards, character classes, escaping, anchoring, newlines, regex metacharacters as literals) plus `escapeGlob` round-trip cases.
- `redis-cache_spec.js`: coverage for the `HGETEX` fast path, the fallback, permanent fallback after "unknown command", error propagation, falsy round-trips, and `scan` cursor iteration with dedupe.
- `memory-cache_spec.js`: `scan` glob matching against real cache entries.

#### Additional manualt testing
- Live validation (not in CI) was done too. The driver passed the same 22 checks against Valkey 9.1, Redis 8.10, and Redis 7.4. The 7.4 run proves the fallback path against a real server that lacks `HGETEX`.

### Notes

- Pre-existing hard requirement, discovered during live validation and now documented in the driver header: TTL caches need Redis >= 7.4 or Valkey >= 9.0, because `set()` has always used `HPEXPIRE` and throws "unknown command" on older servers (Redis 7.2 fails exactly this way). Nothing in this PR raises that floor. Added as a note in the redis-cache file for awareness.
- The heartbeat scan walks the whole pins hash. Fine while MCP access is opt-in per tab. If it ever becomes default-on, add the per-tab "has pins" marker first (this was noted in the BrowserSession file in the header).
- Glob edge cases where it deliberately differs from redis is an unterminated `[` is treated as a literal rather than an open class. Pinned by tests.

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


Steve-Mcl:
Adding @hardillb as a reviewer for the drivers since it touches the redis/valkey stuff.

Ben, for reference, there is a MCP enablement button on the UI. When pressed, it permits an external MCP server to pin itself against this tab (mcp session id gets mapped to the browsers internal UUID session id). The cache (memory and redis) needed a means of updating life on get and a means of searching (known as scan in redis world) - thats what was implemented to support Serbans needs here.


@cstns for the overarching solution to the pinning issue.

forge/caches/redis-cache.js
```diff
@@ -1,13 +1,23 @@
+/**
+ * Redis/Valkey cache driver. Caches with a ttl use per-field hash TTLs
+ * (HPEXPIRE), which requires Redis >= 7.4 or Valkey >= 9.0 — set() throws
+ * "unknown command" on older servers. HGETEX (Redis >= 8.0, Valkey >= 9.1)
+ * is used opportunistically for updateAgeOnGet, with a fallback for servers
+ * without it.
+ */
```
hardillb, line 7: Helm chart installs Valkey 9.0.2 which matches the first test but not the >= 9.1

hardillb:
```suggestion
/**
 * Redis/Valkey cache driver. Caches with a ttl use per-field hash TTLs
 * (HPEXPIRE), which requires Redis >= 7.4 or Valkey >= 9.0 — set() throws
 * "unknown command" on older servers. HGETEX (Redis >= 8.0, Valkey >= 9.0)
 * is used opportunistically for updateAgeOnGet, with a fallback for servers
 * without it.
 */
```

forge/caches/util.js
```diff
@@ -11,6 +11,70 @@ function assertStringKey (key) {
...
+/**
+ * Scan patterns must be strings for the same driver-parity reasons as keys.
+ * @param {*} pattern
+ */
+function assertStringPattern (pattern) {
+    if (typeof pattern !== 'string') {
+        throw new TypeError(`Cache scan pattern must be a string, got ${typeof pattern}`)
+    }
+}
+
```
hardillb, line 23: This function just appears to test the pattern is a string and do nothing else? Given the name this might be valid but just double checking

hardillb (approved):
Looks OK, just the fix for the comment on the redis driver and a minor quibble about the assertStringPattern function

forge/caches/util.js
Steve-Mcl: I wrote that then couldn't think of anything meaningful to actually add (turns out redis's matcher (stringmatchlen, which backs HSCAN MATCH) never rejects a pattern) - I did consider dropping it but the separate message earns its keep! 

You ok with that @hardillb?

hardillb: yeah, just wanted to make sure it wasn't left in by mistake

PR #8292, opened 2026-08-26
feat(mcp): gate the third-party MCP surface per team on the AI feature

andypalmi:
## Summary

Gates the third-party MCP surface on the per-team AI feature, not just the platform-level flag, so a token can no longer reach platform tools for a team that has opted out of AI.

- The front door narrows a token's forwarded team reach to the teams with AI enabled, resolved per request in `resolveCaller`.
- An empty team list already means all teams, so when filtering leaves no teams a sentinel that matches no team is substituted (deny) rather than collapsing back to all teams.
- The gate is backed by two hashid-keyed caches, the team AI value and the user's team memberships, which cache inputs rather than the filtered result so each is invalidated with a single delete. Invalidation is wired to the routes that change the effective AI value: per-team and admin properties edits, team-type feature changes, and team membership changes. A five minute TTL is the backstop, and the shared cache driver propagates deletes across instances.
- The user-teams cache is fail closed: downstream RBAC re-checks live membership on every call, so a stale list can only make an all-teams token temporarily more restrictive, never grant a team the user has left.
- Tests cover the resolution cases (all-teams, team-scoped, multi-team, team-type inherited, no-teams user), the caches, and invalidation through the team and team-type APIs.

### DIFF
--- a/forge/db/controllers/Team.js
+++ b/forge/db/controllers/Team.js
@@ -1,6 +1,64 @@
 const { Roles, RoleNames } = require('../../lib/roles')
 
+// Per-team AI gate caches for the MCP door, keyed by hashid. The TTL backstops
+// the memory driver, whose del only clears the process that handled the write.
+const TEAM_AI_CACHE = 'mcp-team-ai-enabled'
+const USER_TEAMS_CACHE = 'mcp-user-teams'
+const AI_CACHE_TTL = 1000 * 60 * 5 // 5 minutes
+const AI_CACHE_MAX = 10000
+
 module.exports = {
+
+    init (app) {
+        app.caches.createCache(TEAM_AI_CACHE, { ttl: AI_CACHE_TTL, max: AI_CACHE_MAX })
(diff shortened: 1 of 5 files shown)


forge/db/controllers/Team.js
```diff
@@ -1,7 +1,65 @@
...
+    // Cached per-team AI feature value for the door's gate; a missing team
+    // resolves to false (fail closed).
+    isTeamAiEnabled: async function (app, teamHashId) {
+        const cache = app.caches.getCache(TEAM_AI_CACHE)
+        const cached = await cache.get(teamHashId)
+        if (cached !== undefined) {
+            return cached
+        }
+        const team = await app.db.models.Team.byId(teamHashId)
+        const enabled = team ? !!team.getFeatureProperty('ai', true) : false
```
cstns, line 26: i don't think we should store anything in cache if there's no team

cstns (approved):

andypalmi:
Closing this as we are going another route with this

PR #8293, opened 2026-08-26
feat(expert): fetch flow-building tool catalog over MQTT

### LINKED ISSUE #8611: Fetch the flow-building tools catalog over the message bus
**What it is**
The tool-permissions catalog fetched the flow-building tools with a direct HTTP call to the Expert service, authenticated with the Expert service token.

**Done when**
- The catalog loads in the permissions UI over the message bus, with no service token on that path.
- It works on deployments that haven't enabled third-party MCP.
- If the gateway doesn't answer, the UI degrades to the platform tools instead of erroring.
- Older self-hosted deployments keep working.

andypalmi:
Closes #8611

## Description

The `/api/v4/expert/mcp/tools` route (tool-permissions catalog, added in #7639) builds the list of tools shown in the permissions UI. It previously fetched the flow-building tools with a direct `axios.get` to the Expert instance's `/mcp/flow-tools` endpoint, authenticated with the `expert.service.token`.

This replaces that HTTP fetch with a request over MQTT to the gateway's `list_flow_catalog` tool, reusing the existing third-party MCP request/response channel:

- request: `ff/v1/mcp/<platformId>/flow-building-tool-catalog/<uuid>/request`
- response: `ff/v1/mcp/<platformId>/flow-building-tool-catalog/<uuid>/response`

The `flow-building-tool-catalog` sentinel sits in the `userId` topic level to mark the fetch as the first-party, session-less catalog read.

Design points:

- **No service token** is needed on this path.
- **No new topic.** The catalog rides the channel the bridge and the central gateway already carry (`ff/v1/mcp/+/+/+/request` / `.../response`), so no new bridge source, rule, or gateway handler is required. `list_flow_catalog` is already a registered tool on the gateway's MCP server, so the request is served by the same dispatch as any other MCP call.
- **Not gated on `mcpThirdParty`.** The ACL recognises the sentinel `userId` and treats it as a first-party read: exempt from the third-party feature gate and from the user lookup. Every other check (platform client, platformId shape, topic-safe session id) still applies, so the sentinel can only perform the designed request/response exchange. Deployments that have not enabled third-party MCP still get the permissions-UI catalog.
- If the gateway is unreachable or errors, the route degrades to the platform tools alone and still returns 200, rather than forwarding an upstream failure to the browser.

## Backward compatibility

Forge no longer sends the service token on this path. The Expert instance keeps its HTTP `/mcp/flow-tools` endpoint for older Forge deployments; new Forge uses MQTT only, with no HTTP fallback.

## Testing

Unit tests cover the sentinel ACL path (including that the catalog channel works with `mcpThirdParty` disabled and without a real user, while the topic-shape checks still apply) and the route fetching the catalog over MQTT. Validated live on a self-hosted deployment with the central broker.

### DIFF
--- a/forge/ee/routes/expert/index.js
+++ b/forge/ee/routes/expert/index.js
@@ -57,8 +57,7 @@ const curatePlatformTool = (def) => {
     }
 }
 
-// Unwraps list_flow_catalog's CallToolResult: nr-mcp-server-nodes wraps the payload as JSON
-// in content[0].text, matching the Expert gateway-client's parseToolResult.
+// Unwraps a CallToolResult: the payload is JSON in content[0].text.
 const parseFlowCatalogResult = (mcpResponse) => {
     const result = mcpResponse?.result
     const text = result?.content?.[0]?.text
@@ -72,8 +71,7 @@ const parseFlowCatalogResult = (mcpResponse) => {
     return result?.structuredContent ?? null
 }
 
-// Maps a gateway flow-building descriptor to the permissions-UI entry, identical to the
-// Expert gateway-client's toUiCatalogEntry so the shape matches the old HTTP path.
+// Maps a flow-building tool descriptor to the permissions-UI entry shape.
 const toUiCatalogEntry = (d) => {
     const ann = d.annotations || {}
     const meta = d._meta || {}
@@ -731,9 +729,9 @@ module.exports = async function (app) {
             return reply.status(404).send({ code: 'not_found', error: 'Not Found' })
         }
 
-        // Fetch the flow-building catalog from the gateway's list_flow_catalog meta-tool over
-        // the MQTT bridge (no Expert service token). It is session-less, so no team/scope is
-        // sent; failure degrades to the platform tools alone.
+        // Flow-building catalog, over MQTT:
+        // - no service token, no team/scope (catalog is global)
+        // - failure degrades to platform tools alone
         let catalog = []
         let hash = null
         const mcpGateway = app.comms?.mcpGateway
@@ -761,8 +759,6 @@ module.exports = async function (app) {
             }
         }
 
-        // Merge in the global platform tools, reusing the app.comms singleton (constructing
-        // one re-registers its MQTT listener).
         const platformHandler = app.comms?.platformAutomation
         if (platformHandler) {
             const platformDefs = platformHandler.getToolDefinitions() || []
--- a/forge/ee/routes/expert/index.js
+++ b/forge/ee/routes/expert/index.js
@@ -57,6 +57,36 @@ const curatePlatformTool = (def) => {
     }
 }
 
+// Unwraps list_flow_catalog's CallToolResult: nr-mcp-server-nodes wraps the payload as JSON
+// in content[0].text, matching the Expert gateway-client's parseToolResult.
+const parseFlowCatalogResult = (mcpResponse) => {
+    const result = mcpResponse?.result
+    const text = result?.content?.[0]?.text
+    if (typeof text === 'string') {
+        try {
+            return JSON.parse(text)
+        } catch (e) {
+            return null
+        }
+    }
+    return result?.structuredContent ?? null
+}
+
+// Maps a gateway flow-building descriptor to the permissions-UI entry, identical to the
+// Expert gateway-client's toUiCatalogEntry so the shape matches the old HTTP path.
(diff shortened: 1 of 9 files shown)


Steve-Mcl:
not a review - just a quick eye over this. This PR introduces yet another topic for the bridge. I suspect we could use an existing topic channel e.g. ff/v1/mcp/+/+/+/request already. Could just set the "userid" and "mcpSessionId" fields of that topic as something arbitrary/known/checked e.g. `ff/v1/mcp/<platformId>/flow-building-tool-catalog/<random-uid-for-separation-of-concerns>/request` (ensuring if element [5] is === "flow-building-tool-catalog" the only operation it can make is the designed request) (or some other topic that fits `ff/v1/mcp/+/+/+/request`)

test/unit/forge/comms/authRoutesV2_spec.js
```diff
@@ -1914,5 +1914,67 @@ describe('Broker Auth v2 API', async function () {
...
+                    topic: `ff/v1/mcp/${app.comms.id}/${CATALOG_USER}/${SESSION}/request`
+                })
+            })
+            it('allows forge_platform to publish a catalog request for another replica\'s platformId', async function () {
+                await allowWrite({
+                    username: 'forge_platform',
+                    topic: `ff/v1/mcp/${OTHER_PLATFORM_ID}/${CATALOG_USER}/${SESSION}/request`
+                })
+            })
+            it('allows a catalog request without a real user (sentinel skips the user lookup)', async function () {
```
Steve-Mcl, line 1948: this test is identical to the one above (differs only in name)

forge/comms/utils/mcpToolResult.js
```diff
@@ -0,0 +1,23 @@
...
+    if (typeof text === 'string') {
+        try {
+            return JSON.parse(text)
+        } catch (e) {
+            return null
+        }
+    }
+    return result?.structuredContent ?? null
+}
+
```
Steve-Mcl, line 22: Silent failures. A JSON-RPC {error}, an isError: true tool result, and unparseable text all collapse to null and then to an empty catalog. The only app.log.warn is in the catch (where its called), which none of these reach. Is it worth logging when parsed is null or mcpResponse.error is set? otherwise "gateway is broken" and "there are no flow tools" are indistinguishable (in the logs and in the UI).

forge/ee/routes/expert/index.js
```diff
@@ -700,40 +711,43 @@ module.exports = async function (app) {
...
+                    {
+                        mcp: {
+                            jsonrpc: '2.0',
+                            id: 1,
+                            method: 'tools/call',
+                            params: { name: 'list_flow_catalog', arguments: {} }
+                        },
+                        toolGroups: ['flow_building']
+                    },
+                    app.expert.requestTimeout
```
Steve-Mcl: `app.expert.requestTimeout` defaults to 60000 (`forge/ee/lib/expert/index.js:21`). That was a
ok for `axios.get`, because an unreachable Expert failed fast on DNS or
ECONNREFUSED and never got near the timeout. An MQTT publish has no equivalent failure. With
nothing listening on the far end of the bridge, this waits the full minute before falling back
to the platform tools.

So any self-hosted deployment with `expert.enabled` but no working bridge (no licence, bridge
not provisioned yet, bridge mid-resync) now blocks `GET /mcp/tools` for 60s

Worth giving the catalog its own shorter budget instead of reusing the HTTP one. 5 or 10
seconds is probs plenty for a round trip!

Skipping the publish entirely when the bridge is known to be down might be a better bet???

Steve-Mcl (commented):
nits

test/unit/forge/comms/authRoutesV2_spec.js
andypalmi: Good catch, folded the two into one.

forge/comms/utils/mcpToolResult.js
andypalmi: Added warn logs for a gateway error, isError result, and null parse, so broken gateway vs no tools is now distinguishable.

forge/ee/routes/expert/index.js
andypalmi: Gave the catalog its own 10s budget instead of reusing the HTTP timeout.

Steve-Mcl (approved):
tested locally!

PR #8295, opened 2026-08-26
Extend visual and interaction feedback for MCP features

### LINKED ISSUE #8286: Tab heartbeat and pin visibility in the UI
Follows on from the cache work. The tab is what makes a pin live, so the client side has to hold up its end.

Two things:

1. Heartbeat while the tab holds pins, so the backend can refresh them and drop them when the tab goes away. Right now the presence heartbeat is the only signal, and it isn't tied to whether anything is actually pinned to this tab.
2. Show the user the pin state. Which MCP sessions are currently attached to this tab, and a signal when one attaches or detaches. Today the MCP toggle only says "exposed", so there's no way to tell that an agent is driving this tab, or which one.

cstns:
## Description

The MCP toggle now shows whether something is actually driving the tab, not just whether the tab is exposed.

Exposed and targeted are two different things, so the store tracks them separately. Targeting isn't exclusive, so it's a set rather than a flag and the button shows a count when more than one client holds the tab. The tab never works that count out itself: a pin can end by just expiring, which produces no event, so the platform restates the full set
on every heartbeat and pin change.

**Backend**: pins are stored both ways round so we can ask who holds a tab, and there's a new per-session topic (`.../mcp/clients`) for the platform to answer on, with the ACL pinning the session segment to the subscriber's own credential. MCP session ids are hashed to opaque refs before they reach the browser. Cache drivers gained `scan`, with glob matching shared so memory and redis agree.

**Frontend**: new subscriber for that topic, and four button states instead of two (off, waiting, connected, interrupted). Interrupted outranks the rest since nothing is confirming the others while we can't hear the platform. Toggle mounts are counted now, because the header renders two and the first to unmount was tearing down comms the other still showed as live.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/8286

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/comms/aclManager.js
+++ b/forge/comms/aclManager.js
@@ -130,8 +130,9 @@ module.exports = function (app) {
                 return false
             }
         },
-        checkUserIsTeamMember: async function (requestParts, usernameParts) {
-            // requestParts = [ fullTopic , <teamHash> [, <userHash> [, <sessionId>]] ]
+        checkTeamUserSession: async function (requestParts, usernameParts) {
+            // requestParts = [ fullTopic , <teamHash> [, <userHash>] ] - v1
+            // requestParts = [ fullTopic , <teamHash>, <userHash>, <sessionId> ] - v2 (MCP)
             // usernameParts = [ 'fe-team', <userHash>, <teamHash>, <sessionId> ]
             const topicTeamHash = requestParts[1]
             const usernameUserHash = usernameParts[1]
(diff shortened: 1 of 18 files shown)


forge/comms/aclManager.js
```diff
@@ -767,7 +767,10 @@ module.exports = function (app) {
                 { topic: /^ff\/v1\/expert\/([^/]+)\/([^/]+)\/platform\/([^/]+)\/response$/, verify: 'checkExpertPlatformTopic', isPlatform: true, isPub: true, agent: 'platform' },
                 // platform can publish third-party MCP requests to the central gateway
                 // - ff/v1/mcp/<platformId>/<userId>/<mcpSessionId>/request
-                { topic: /^ff\/v1\/mcp\/([^/]+)\/([^/]+)\/([^/]+)\/request$/, verify: 'checkMcpTopic', isPlatform: true, isPub: true }
+                { topic: /^ff\/v1\/mcp\/([^/]+)\/([^/]+)\/([^/]+)\/request$/, verify: 'checkMcpTopic', isPlatform: true, isPub: true },
+                // platform can tell one browser tab about its MCP state
+                // - ff/v1/<team>/u/<user>/s/<session>/mcp/<event>
+                { topic: /^ff\/v1\/[^/]+\/u\/[^/]+\/s\/[^/]+\/mcp\/[^/]+$/ }
```
Steve-Mcl: isnt this only one `<event>` right now - ../mcp/event?

Steve-Mcl: wait no, this is heartbeat|stop here right?

Steve-Mcl (approved):
approved. pulled locally, tested with claude, and co-reviewed with Serban.

PR #8301, opened 2026-08-26
fix(auth): enforce PAT team scope on entity-addressed routes

andypalmi:
## Summary

Team-scoped personal access tokens were only checked against their team scope on routes that resolve the team directly (`request.team`). Routes addressed by an entity id set the entity and the team membership but not `request.team`, so the team-scope check was skipped and a token scoped to one team could reach resources in other teams the user belongs to.

- The check runs in `needsPermission`, alongside the existing team-scope enforcement, so it applies to every team-scoped route.
- A new helper resolves the team from whatever the route already loaded (directly, or via the application, instance, or device), falling back to the request's team membership, so a route addressed by an application, instance, device, or snapshot id is covered the same way as one addressed by a team id.
- Both team-scope checks (the admin-bypass branch and the main permission block) use the helper, so the two paths enforce scope identically.
- A blocked call returns 403 `unauthorized`, matching the existing team-scope behaviour.

## Testing

- New unit tests drive real entity-addressed routes with a team-scoped personal access token, covering an out-of-scope resource addressed by entity id (denied), an in-scope resource addressed by entity id (allowed), and an out-of-scope snapshot addressed by id (denied, exercising the team-membership fallback).


andypalmi:
## Live validation

Tested against a running instance with a personal access token scoped to a single team (`<TEAM_A>`). The user is a member of both `<TEAM_A>` and `<TEAM_B>`. Each tool was called for a resource in `<TEAM_A>` (in-scope, expected `200`) and the equivalent resource in `<TEAM_B>` (out-of-scope, expected `unauthorized`).

### Entity-addressed routes (the fix - these bypassed the check before)

| tool | `<TEAM_A>` | `<TEAM_B>` |
|---|---|---|
| get_application | 200 | unauthorized |
| get_application_audit_log | 200 | unauthorized |
| get_application_bill_of_materials | 200 | unauthorized |
| list_application_snapshots | 200 | unauthorized |
| get_hosted_instance | 200 | unauthorized |
| get_hosted_instance_status | 200 | unauthorized |
| get_hosted_instance_logs | 200 | unauthorized |
| get_instance_audit_log | 200 | unauthorized |
| get_instance_history | 200 | unauthorized |
| list_instance_http_tokens | 200 | unauthorized |
| get_remote_instance (device) | 200 | unauthorized |
| list_instance_snapshots | 200 | unauthorized |
| create_instance_snapshot (write) | 200 | unauthorized |
| get_snapshot (snapshotId only) | 200 | unauthorized |
| get_snapshot_full (snapshotId only) | 200 | unauthorized |
| get_pipeline_stage (pipelineId only) | 200 | unauthorized* |

*blocked in the preHandler before the stage is looked up.

### Team-addressed routes (already enforced before the fix - regression check)

| tool | `<TEAM_A>` | `<TEAM_B>` |
|---|---|---|
| get_team | 200 | unauthorized |
| get_team_audit_log | 200 | unauthorized |
| get_team_bill_of_materials | 200 | unauthorized |
| list_team_device_groups | 200 | unauthorized |
| list_pipelines | 200 | unauthorized |
| list_brokers | 200 | unauthorized |
| list_broker_clients | 200 | unauthorized |
| get_broker | 200 | unauthorized |
| list_broker_topics | 200 | unauthorized |
| list_team_databases | 200 | unauthorized |

### No team context (ungated by design - not addressed here)

| tool | result |
|---|---|
| list_teams | returns all of the user's teams (`<TEAM_A>`, `<TEAM_B>`, ...) |
| list_notifications | returns cross-team instance references |
| list_own_invitations | returns inviting teams incl. out-of-scope |
| get_active_user | own profile; defaultTeam may be out-of-scope |

These routes carry no team context, so team-scope enforcement does not apply at the permission layer. Narrowing their responses to the token's scope is a separate concern; handling of these ungated-by-design tools is deferred to a later decision.

forge/routes/auth/permissions.js
```diff
@@ -30,6 +30,44 @@ function patAllowsWrite (pat, scope) {
...
+ * @param {object} app - The forge app instance
+ * @param {object} request - The Fastify request
+ * @returns {string|null} the team hashid, or null when there is no team context
+ */
+function resolveRequestTeamHashid (app, request) {
+    const loadedTeam = request.team || request.application?.Team || request.project?.Team || request.device?.Team
+    if (loadedTeam?.hashid) {
+        return loadedTeam.hashid
+    }
+    const teamId = request.teamMembership?.TeamId || request.teamId
```
knolleary:
```suggestion
    const teamId = request.teamMembership?.TeamId ?? request.owner?.TeamId ?? request.teamId
```

The `/api/v1/snapshot` routes were not covered. The prehandler on those routes sets `request.owner` to either the Project or Device object - both of which will have `TeamId` set.

Also switches to `??` as we're protecting against `null`/`undefined` rather than all falsey values.

knolleary (approved):

PR #8302, opened 2026-08-26
fix(auth): enforce PAT team scope on user-scoped list routes

andypalmi:
## Summary

Team-scoped personal access tokens are enforced on team- and entity-addressed routes, but the user-scoped routes return data spanning every team the user belongs to, with no single team for the `needsPermission` check to gate on. A token scoped to one team could therefore read teams, invitations, and a default team outside its scope.

- `GET /user/teams` now returns only the teams in the token's scope.
- `GET /user/invitations` drops invitations to out-of-scope teams.
- `GET /user` omits `defaultTeam` when it is out of scope.
- A new `patTeamScopeFilter(request)` helper returns a predicate for the token's team scope, or null for callers that are not team-scoped PATs, so cookie sessions and all-teams tokens are unaffected.
- Out-of-scope entries are omitted rather than flagged, so a response never discloses the existence of a team the token cannot see.

## Testing

- New unit tests drive the routes with a team-scoped personal access token: only in-scope teams are returned from `/user/teams`, an out-of-scope invitation is dropped from `/user/invitations` (the cookie session still sees it), and an out-of-scope `defaultTeam` is omitted from `/user`.
- Validated live against a running instance with a token scoped to a single team (see the testing comment).

### DIFF
--- a/forge/routes/api/user.js
+++ b/forge/routes/api/user.js
@@ -41,6 +41,10 @@ module.exports = async function (app) {
         if (app.license.active() && app.billing && app.db.controllers.Subscription.freeTrialCreditEnabled()) {
             response.free_trial_available = await app.db.controllers.Subscription.userEligibleForFreeTrialCredit(user)
         }
+        const allowTeam = app.patTeamScopeFilter(request)
+        if (allowTeam && response.defaultTeam && !allowTeam(response.defaultTeam)) {
+            delete response.defaultTeam
+        }
 
         reply.send(response)
     })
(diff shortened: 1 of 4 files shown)


andypalmi:
Tested a token scoped to a single team (`<TEAM_A>`) against a user who belongs to `<TEAM_A>`, `<TEAM_B>` and `<TEAM_C>`.

platform_list_teams (`/user/teams`):

| Token scope | Result |
| --- | --- |
| all teams | `<TEAM_A>`, `<TEAM_B>`, `<TEAM_C>` (3) |
| `<TEAM_A>` only | `<TEAM_A>` (1) |

platform_list_own_invitations (`/user/invitations`), two invitations seeded (one to `<TEAM_A>`, one to `<TEAM_B>`):

| Token scope | Result |
| --- | --- |
| `<TEAM_A>` only | `<TEAM_A>` invitation only (1); `<TEAM_B>` hidden |

platform_get_active_user (`/user`), token scoped to `<TEAM_A>`:

| defaultTeam | Result |
| --- | --- |
| `<TEAM_B>` (out of scope) | `defaultTeam` omitted |
| `<TEAM_A>` (in scope) | `defaultTeam` returned |

PR #8307, opened 2026-08-27
Move MCP comms lifecycle out of the Expert button and into the orchestrator

### LINKED ISSUE #8284: Browser session pinning for third party MCP clients
### Description

First party Expert never has to ask which browser it is talking to. It runs inside the tab, so the tab is the session. Third party MCP clients (Claude Desktop, Cursor, an agent on someone's laptop) have no browser at all. They hold a PAT or an OAuth token, and that identifies a user, not a place to act.

  That is fine for platform tools, which execute inside the platform. It breaks for the two groups that only exist in a browser: `platform_ui` (context, routes, navigation) and `flow_building` (the Node-RED canvas, which is an iframe inside our UI). Those calls have to be delivered to a real tab over MQTT, and a user routinely has several tabs open on different teams, instances, and pages. Picking the wrong one is not a no-op, it is an action on the wrong instance.

  So a third party client needs an explicit answer to "which tab", it needs that answer to survive across many tool calls in one conversation, and the user needs to have opted that tab in first. Pinning is the mechanism: the user exposes a tab, the agent lists what is exposed, pins one, and every browser bound call after that targets it without threading a session id through each call.

  Most of this is already built (see the sub-tasks for what is left). This story exists to give the work an issue, because we shipped it without one.

### Epic/Story

_No response_

cstns:
## Description

The MCP comms (tab presence publisher, in-flight subscriber, session subscriber) were started and stopped by the Expert button mounting and unmounting. Moving between a platform page and the immersive editor swaps the layout, so the header is a different component instance. Vue runs the old `beforeUnmount` before the new `mounted`, so the teardown was still awaiting its stops when the remount had already recreated everything, and the stops then landed on the fresh objects.

The result was a tab that still looked exposed (toggle on, still listed, still heartbeating) but answered nothing. Every `ui_*` and `flow_building_*` call just timed out. Refreshing the page fixed it, which is what confirmed it was lifecycle rather than
capability.

The mount counter was an earlier attempt at this. It could not work, because it was arbitrating a lifecycle that should never have been component scoped in the first place.

  ### What changed

The comms now live in the subscriber and publisher registries, so the orchestrator creates them once at boot and disposes them with the app. Nothing in the component tree owns them anymore.

* The store only connects and disconnects. It never creates or destroys.
* The Expert button is purely a control now. It subscribes, unsubscribes, and kicks off  the heartbeat, with no unmount hook at all.
* Registry entries carry an `autoConnect` flag. MCP is `false`, so a team load never  quietly exposes a tab.
* Everything from the previous attempt is gone: `mounts`, `retain`, `release`, `teardown`,
 `activePublisher`, `silenceCloseNotice` and the `announceClose` param.

## Related Issue(s)

closes https://github.com/FlowFuse/flowfuse/issues/8284

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/frontend/src/components/ExpertButton.vue
+++ b/frontend/src/components/ExpertButton.vue
@@ -79,8 +79,13 @@ export default {
         }
     },
     watch: {
-        async team () {
+        async team (team, previousTeam) {
             if (!this.mcpActive) {
+                return
+            }
+            // The first team resolving is not a switch.
+            if (!previousTeam?.id) {
+                this.resumeMcp(team)
                 return
             }
             // Straight to the store action rather than stopMcp(), which announces the close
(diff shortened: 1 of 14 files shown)


knolleary:
Once @Steve-Mcl signs off on the code changes, you have my approval to be merged today prior to release.

Steve-Mcl (approved):
reviewed with @cstns and witnessed in action. 
noting that prior fix has actually been removed in favour of the app mount point (not on the button) solution

PR #8321, opened 2026-08-28
docs: show the agent picker before the written-out steps

dimitrieh:
On [Connect Your Own Agent](https://flowfuse.com/docs/user/expert/third-party-agents/) the picker sat under **Where to add it, per agent**, so a reader met the raw MCP address in step 1 and only reached the per-agent tabs much further down.

The single `::agent-setup-tabs` embed now sits at the top of **Connect your agent**, directly after the "any MCP client" line. Nothing is dropped: the three numbered steps, the Cloud and self-hosted addresses, and every per-agent section stay in text below it. The embed is not duplicated, since two instances on one page would emit duplicate tab and panel ids.

### DIFF
--- a/docs/user/expert/third-party-agents.md
+++ b/docs/user/expert/third-party-agents.md
@@ -14,6 +14,13 @@ Because the agent is yours, so is the model it runs on.
 
 Any MCP client that supports the HTTP transport can connect. That is the only requirement.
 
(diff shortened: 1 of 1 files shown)


dimitrieh:
up for review

cstns (approved):

PR #8354, opened 2026-09-01
docs: add networking requirements page for self-hosted installs

dimitrieh:
Adds an outbound egress list for self-hosted installs. Comes from a self-hosted enterprise renewal on a narrow-egress OT network that needs to file one change-controlled firewall request: https://github.com/FlowFuse/accounts/issues/128. Sales reports it is a common question.

Today there is no self-hosted egress list in the docs. The only list is on the Device Agent overview page, which covers agent to Cloud traffic, so a self-hoster behind a firewall reads Cloud-only endpoints as authoritative for their install.

## Where it lives

`docs/install/networking-requirements.md`, in the Installing FlowFuse nav group alongside DNS Setup and Configuring FlowFuse. It is reference material (what is required), not a how-to, so it is a flat table with conditions rather than a walkthrough. Linked from:

- install overview
- Docker install, Requirements
- Kubernetes install, Prerequisites

## Open questions for review

1. Wording of the restricted-networks caveat: an allowlist covers most nodes, but some packages fetch native builds from arbitrary hosts at install time and cannot be enumerated.
2. Should the Device Agent overview list be trimmed of Cloud-only rows and cross-linked here, or left as is? It also lists `registry.npmjs.com`, which should be `.org`.
3. Kubernetes side: any image-pull or operator hosts missing (EMQX operator, Traefik, cert-manager charts)?

### DIFF
--- a/docs/install/introduction.md
+++ b/docs/install/introduction.md
@@ -31,6 +31,12 @@ We also provide one-click installs of the Docker version:
  - [Digital Ocean Docker Install Guide](/docs/install/docker/digital-ocean.md)
  - [AWS Docker Install Guide](/docs/install/docker/aws-marketplace.md)
 
+## Networking requirements
+
+If the platform runs behind a firewall or proxy, see
+[Networking requirements](/docs/install/networking-requirements.md) for the outbound
+hostnames and ports FlowFuse needs.
+
 ## Upgrading FlowFuse
 
 If you are upgrading FlowFuse, please refer to the [Upgrade Guide](/docs/upgrade/README.md)
--- a/docs/install/introduction.md
+++ b/docs/install/introduction.md
@@ -31,11 +31,9 @@ We also provide one-click installs of the Docker version:
  - [Digital Ocean Docker Install Guide](/docs/install/docker/digital-ocean.md)
  - [AWS Docker Install Guide](/docs/install/docker/aws-marketplace.md)
 
-## Networking requirements
-
-If the platform runs behind a firewall or proxy, see
-[Networking requirements](/docs/install/networking-requirements.md) for the outbound
-hostnames and ports FlowFuse needs.
+Whichever installation you choose, the platform needs outbound network access to a set of
+hostnames. If it runs behind a firewall or proxy, see
+[Networking requirements](/docs/install/networking-requirements.md).
 
 ## Upgrading FlowFuse
 
--- a/docs/install/networking-requirements.md
+++ b/docs/install/networking-requirements.md
@@ -1,0 +1,69 @@
+---
+navTitle: Networking requirements
+meta:
+   description: The outbound network access a self-hosted FlowFuse platform requires, listed by hostname, port and purpose, so a firewall change can be requested in one go.
+   tags:
+      - installation
+      - networking
+      - firewall
+      - proxy
+      - self-hosted
+      - flowfuse
+---
+
+# Networking requirements
+
+This page lists the outbound access a self-hosted FlowFuse platform needs. It applies to
+both the Docker and Kubernetes installations, and covers the FlowFuse application, the
+Node-RED instances it hosts, and the browsers that open the editor.
+
+Two related lists live elsewhere:
+
+- Inbound access to the platform is ports `80` and `443`. See
+  [Docker install](./docker/README.md#requirements) and [DNS setup](./dns-setup.md).
+- The Device Agent has its own list, which includes FlowFuse Cloud endpoints. See
+  [Device Agent networking requirements](../device-agent/install/overview.md#networking-requirements).
+
+All destinations below are outbound TCP.
+
+## Always required
+
+| Destination | Port | Purpose |
+| --- | --- | --- |
+| `registry.npmjs.org` | 443 | Installing Node-RED and node packages |
+| `catalogue.nodered.org` | 443 | Default node catalogue used by the editor palette |
+| `registry-1.docker.io`, `auth.docker.io`, `production.cloudflare.docker.com` | 443 | Pulling FlowFuse and Node-RED container images, at install and at every upgrade |
+| `ping.flowforge.com` | 443 | Anonymous usage telemetry |
+
+Telemetry can be turned off on open source installations, but
+[licensed installations cannot disable it](../admin/telemetry.md#configuring-telemetry).
+
+## Required for specific features
+
+Only open these if the feature is in use.
+
+| Destination | Port | Feature | Turn off with |
+| --- | --- | --- | --- |
+| `registry.flowfuse.com`, `ff-certified-nodes.flowfuse.cloud` | 443 | FlowFuse Certified Nodes, from the server, the editor browsers and the hosted instances | Leave the certified nodes token unset |
+| `app.flowfuse.com` | 443 | Daily import of the public blueprint library | `blueprintImport.enabled: false` |
+| `expert.flowfuse.com` | 443 | FlowFuse Expert and the in-editor assistant | `ai.enabled: false` |
+| `expert-broker.flowfuse.com` | 8883 | FlowFuse Expert MQTT broker, a Helm chart default | `ai.enabled: false` |
+| `api.github.com`, `github.com` | 443 | GitOps pipelines | Do not configure a GitOps integration |
+| `www.googleapis.com` | 443 | Google SSO | Do not configure Google SSO |
+| `acme-v02.api.letsencrypt.org` | 443 | Automatic TLS certificates | Use a certificate you supply yourself |
+| Your SMTP relay | 587 or 465 | Invitations, password resets and notifications | Leave `email` unconfigured |
+
+## Restricted networks
+
+A hostname allowlist covers the large majority of Node-RED nodes, but not all of them. Some
+packages download native builds at install time from arbitrary hosts, such as GitHub releases
+or S3 buckets. Those cannot be listed up front.
+
+If that is not acceptable, host the packages yourself and point FlowFuse at them:
+
+- An internal npm registry and your own node catalogue. See
+  [3rd party npm registries](../user/custom-npm-packages.md#npm-registries).
+- A mirror of the FlowFuse and Node-RED container images in your own registry.
+
+On Kubernetes, also see
+[Network Policies](./kubernetes/README.md#i-use-kubernetes-network-policies%2C-how-can-i-configure-them%3F).
--- a/docs/install/networking-requirements.md
+++ b/docs/install/networking-requirements.md
@@ -39,10 +39,6 @@ All destinations below are outbound TCP.
 | `registry.npmjs.org` | 443 | Instances | Installing Node-RED and node packages |
 | `catalogue.nodered.org` | 443 | Editor browsers | The node catalogue listed in the editor palette. The editor fetches it directly, the platform does not |
 | `registry-1.docker.io`, `auth.docker.io`, `production.cloudflare.docker.com` | 443 | Platform | Pulling the FlowFuse and Node-RED container images, at install and at every upgrade. Not needed if you pull the images from an internal registry or a pull-through proxy instead |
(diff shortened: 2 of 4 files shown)


dimitrieh:
This follows a conversation with Nick, and solves a sales need from https://github.com/FlowFuse/accounts/issues/128#issuecomment-5452823013

Up fro changes and considerations as eng sees fit. This is just the initial proposal to kickstart it

docs/install/networking-requirements.md
```diff
@@ -0,0 +1,69 @@
...
+  [Device Agent networking requirements](../device-agent/install/overview.md#networking-requirements).
+
+All destinations below are outbound TCP.
+
+## Always required
+
+| Destination | Port | Purpose |
+| --- | --- | --- |
+| `registry.npmjs.org` | 443 | Installing Node-RED and node packages |
+| `catalogue.nodered.org` | 443 | Default node catalogue used by the editor palette |
```
hardillb: This is only required for the machines running the browser accessing the NR editor (for both hosted and remote), not from the FF install it's self.

docs/install/networking-requirements.md
```diff
@@ -0,0 +1,69 @@
...
+| `registry-1.docker.io`, `auth.docker.io`, `production.cloudflare.docker.com` | 443 | Pulling FlowFuse and Node-RED container images, at install and at every upgrade |
```
hardillb: Unless using a local Container registry/proxy

docs/install/networking-requirements.md
```diff
@@ -0,0 +1,69 @@
...
+| `ping.flowforge.com` | 443 | Anonymous usage telemetry |
+
+Telemetry can be turned off on open source installations, but
+[licensed installations cannot disable it](../admin/telemetry.md#configuring-telemetry).
+
+## Required for specific features
+
+Only open these if the feature is in use.
+
+| Destination | Port | Feature | Turn off with |
```
hardillb: > Turn off with

This doesn't feel the right way round

docs/install/networking-requirements.md
```diff
@@ -0,0 +1,69 @@
...
+
+A hostname allowlist covers the large majority of Node-RED nodes, but not all of them. Some
+packages download native builds at install time from arbitrary hosts, such as GitHub releases
+or S3 buckets. Those cannot be listed up front.
+
+If that is not acceptable, host the packages yourself and point FlowFuse at them:
+
+- An internal npm registry and your own node catalogue. See
+  [3rd party npm registries](../user/custom-npm-packages.md#npm-registries).
+- A mirror of the FlowFuse and Node-RED container images in your own registry.
```
hardillb: That's a bit vague and hard to do in practice as you only find out they are pull native components when it fails and the only way to host them yourself is to either do proxy interception or build and host the nodes and their components yourself locally, this is a VERY big task.

hardillb (commented):
Comments added before I am OOO

docs/install/networking-requirements.md
dimitrieh: Fixed. Added a "Needed by" column and this row is now Editor browsers only, with a note that the editor fetches the catalogue directly and the platform does not. Confirmed in the editor client, palette-editor.js does the $.getJSON on the catalogue URL from the browser.

docs/install/networking-requirements.md
dimitrieh: Fixed. The row now ends with "Not needed if you pull the images from an internal registry or a pull-through proxy instead", and the restricted networks section lists an internal registry or pull-through proxy as an option.

docs/install/networking-requirements.md
dimitrieh: Changed the column to "Required when", so each row states the condition that makes the destination necessary instead of how to switch it off. Config flags stay inside the cell where they help, for example blueprintImport.enabled.

docs/install/networking-requirements.md
dimitrieh: Rewrote that section. It now says a minority of packages fetch native components at install time from hosts that cannot be listed, that you only find out when an install fails, and that the ways round it are an intercepting proxy or building and hosting the packages and their components yourself, both a significant piece of work. The advice is to plan for a small number of nodes being unavailable rather than for full coverage.

dimitrieh:
Note on the red check. "Test Documentation links" fails in `npm ci`, before the link test runs:

```
npm error `npm ci` can only install packages when your package.json and package-lock.json are in sync
npm error Invalid: lock file's globals@17.11.0 does not satisfy globals@17.12.0
```

It is not caused by this branch, which only adds and edits markdown under `docs/`:

- The same job passed on the first commit of this PR at 10:43 today, same lockfile, same npm (11.17.0).
- Every `docs.yml` run since about 10:55 today fails the same way, including unrelated branches (`docs/packaging-subflows`, `docs-website-preview`).
- `main` has not moved since 31 August, and its `package.json` (`globals: ^17.7.0`) and `package-lock.json` (17.11.0) agree with each other.
- There is precedent for the same package and symptom in https://github.com/FlowFuse/flowfuse/pull/8027, where the fix was a lockfile resync on `main`.

So this needs a `package-lock.json` resync on `main` rather than a change here. Happy to open that separately if it is not already in hand.

The job that validates the content of this PR, "Test Documentation with website" (site build plus `hyperlink --check-anchors`), passes. Links and anchors were also checked locally.

dimitrieh:
@knolleary or @ppawlowski can you take a look here now that Ben is OOO

knolleary (approved):

knolleary:
Approved on the basis there could be finer details to add when we discover them - but lets get something merged so we can iterate as needed.

docs/install/introduction.md
```diff
@@ -31,6 +31,12 @@ We also provide one-click installs of the Docker version:
  - [Digital Ocean Docker Install Guide](/docs/install/docker/digital-ocean.md)
  - [AWS Docker Install Guide](/docs/install/docker/aws-marketplace.md)
 
+## Networking requirements
+
+If the platform runs behind a firewall or proxy, see
+[Networking requirements](/docs/install/networking-requirements.md) for the outbound
+hostnames and ports FlowFuse needs.
+
```
ppawlowski: I don't think the `Introduction` page is a proper one for such detailed information as networking requirements.

docs/install/networking-requirements.md
```diff
@@ -0,0 +1,78 @@
...
+All destinations below are outbound TCP.
+
+## Always required
+
+| Destination | Port | Needed by | Purpose |
+| --- | --- | --- | --- |
+| `registry.npmjs.org` | 443 | Instances | Installing Node-RED and node packages |
+| `catalogue.nodered.org` | 443 | Editor browsers | The node catalogue listed in the editor palette. The editor fetches it directly, the platform does not |
+| `registry-1.docker.io`, `auth.docker.io`, `production.cloudflare.docker.com` | 443 | Platform | Pulling the FlowFuse and Node-RED container images, at install and at every upgrade. Not needed if you pull the images from an internal registry or a pull-through proxy instead |
+| `ping.flowforge.com` | 443 | Platform | Anonymous usage telemetry |
```
ppawlowski: Are we sure about this? Will a FlowFuse Platform on a self-hosted environment crash if access to the `ping.flowforge.com` is not possible?
Additionally, it should be `ping.flowfuse.com`. The old domain is used for legacy purposes and shouldn't be used in documents.

docs/install/networking-requirements.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Required for specific features
+
+| Destination | Port | Needed by | Required when |
+| --- | --- | --- | --- |
+| `registry.flowfuse.com`, `ff-certified-nodes.flowfuse.cloud` | 443 | Editor browsers and instances | FlowFuse Certified Nodes are in use, which needs a certified nodes token set on the platform |
+| `app.flowfuse.com` | 443 | Platform | The public blueprint library is imported daily, which is the default on a licensed install (`blueprintImport.enabled`) |
+| `expert.flowfuse.com` | 443 | Platform | FlowFuse Expert or the in-editor assistant is enabled (`ai.enabled`). Both are proxied by the platform, so instances and browsers do not need this |
+| `expert-broker.flowfuse.com` | 8883 | Platform | FlowFuse Expert is enabled on Kubernetes, where the Helm chart sets this broker as the default. On Docker there is no default, the broker has to be set explicitly |
+| `api.github.com`, `github.com` | 443 | Platform | A GitOps pipeline is configured |
```
ppawlowski: If we list GitHub, we should add other git providers we support. Or drop this line.

ppawlowski (commented):
See my comments; none of them are blockers, though.

docs/install/introduction.md
dimitrieh: Removed the section. The overview page now carries a single sentence pointing at the dedicated page, with no heading of its own, so none of the detail lives there.

The Docker and Kubernetes guides also link it from their own Requirements sections, which is where an installer actually hits it.

docs/install/networking-requirements.md
dimitrieh: Two separate things here.

**Will it crash:** no. The ping is a housekeeper task (`forge/housekeeper/tasks/telemetryMetrics.js`), scheduled once a day at a random time plus once ten seconds after startup, and the `axios.post` sits in a try/catch that logs `Failed to send telemetry report`. Blocking it costs you the report and nothing else. I have moved the row out of "Always required" into the conditional table and added a note saying exactly that, so nobody reads it as load-bearing.

**The hostname:** changed to `ping.flowfuse.com` as you asked, but flagging a mismatch with the code. The shipped default is still the old domain: `telemetryMetrics.js:60` is `app.config.telemetry.url || 'https://ping.flowforge.com/ping'`, and a code search across the org finds `ping.flowfuse.com` in no files and `ping.flowforge.com` in three, including `docs/admin/telemetry.md:108`. Both hostnames answer 200, but they resolve to different API Gateway deployments, so they are not aliases.

As the page now stands, an admin who allowlists only `ping.flowfuse.com` gets telemetry failing silently on every current release. Options as I see them:

1. List both hostnames in the row, so the allowlist survives the switch.
2. Move the default in code first, then the docs follow.

Which would you prefer? Happy to list both here and open a separate issue for the code default.

docs/install/networking-requirements.md
dimitrieh: Split into three rows:

| `github.com`, `api.github.com` | GitHub |
| `dev.azure.com` | Azure DevOps |
| Your own Git server | Any other HTTPS Git server, for example GitLab, Bitbucket, Gitea or self-hosted |

That matches the three backends in `forge/ee/lib/gitops/backends`: `github.js` and `azure.js` each pin their host with a regex, and `generic.js` accepts any `https://` URL, so the third row cannot name a hostname.

dimitrieh:
@dgatti0213 can you have a look here as well. See if it aligns from your side as well

docs/install/networking-requirements.md
dimitrieh: @ppawlowski this needs your view again :)

ppawlowski: Any `flowforge.com` reference should not be used in publicly available documents.

dgatti0213 (approved):

docs/install/networking-requirements.md
dimitrieh: @ppawlowski did you see this:

>As the page now stands, an admin who allowlists only ping.flowfuse.com gets telemetry failing silently on every current release. Options as I see them:

ppawlowski: My statement about avoiding the use of the old company name/domain name in the documentation remains.

dimitrieh: @ppawlowski Got it, merging. In that case, "elemetry failing silently", I'll take is a non-problem to be solved. If it actually is an issue, please open up an issue that we can solve separately.

PR #8356, opened 2026-09-01
[327] Dashboard adoption tracking in PostHog KPI  

n-lark:
## Description

See https://github.com/FlowFuse/engineering/issues/327

## Related Issue(s)

Resolves https://github.com/FlowFuse/engineering/issues/327

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

### DIFF
--- a/forge/housekeeper/tasks/dashboardAdoption.js
+++ b/forge/housekeeper/tasks/dashboardAdoption.js
@@ -1,0 +1,59 @@
+const { randomInt } = require('../utils')
+
+const DASHBOARD_MODULE = '@flowfuse/node-red-dashboard'
+
+const QUERIES = {
+    postgres: `
+        SELECT
+            COUNT(*) FILTER (
+                WHERE (s.settings::jsonb -> 'nodes' -> :module ->> 'local')::boolean IS TRUE
+            ) AS with_dashboard,
+            COUNT(*) AS total
+        FROM "StorageSettings" s
+        JOIN "Projects" p ON p.id = s."ProjectId"
+        WHERE p.state <> 'suspended'`,
+    sqlite: `
+        SELECT
+            SUM(CASE
+                WHEN json_extract(s.settings, '$.nodes."' || :module || '".local') = 1 THEN 1
+                ELSE 0
+            END) AS with_dashboard,
+            COUNT(*) AS total
+        FROM "StorageSettings" s
+        JOIN "Projects" p ON p.id = s."ProjectId"
+        WHERE p.state <> 'suspended'`
+}
+
+module.exports = {
+    name: 'dashboardAdoption',
+    schedule: `${randomInt(0, 59)} ${randomInt(0, 23)} * * *`,
+    run: async function (app) {
+        if (process.env.FF_TELEMETRY_DISABLED) {
+            return
+        }
+        const isLicensed = app.license.active()
+        const isTelemetryEnabled = (app.config.telemetry.enabled !== false && app.settings.get('telemetry:enabled'))
+        if (!app.settings.get('setup:initialised') || !(isLicensed || isTelemetryEnabled)) {
+            return
+        }
+        if (!app.config.telemetry?.frontend?.posthog?.apikey) {
+            return
+        }
+        const query = QUERIES[app.db.sequelize.getDialect()]
+        if (!query) {
+            return
+        }
+        const [result] = await app.db.sequelize.query(query, {
+            replacements: { module: DASHBOARD_MODULE },
+            type: app.db.sequelize.QueryTypes.SELECT
+        })
+        const withDashboard = Number(result.with_dashboard) || 0
+        const total = Number(result.total) || 0
+
+        app.product.capture(app.settings.get('instanceId'), '$ff-dashboard-adoption', {
+            with_dashboard: withDashboard,
+            total,
+            pct: total > 0 ? Math.round(1000 * withDashboard / total) / 10 : 0
+        })
+    }
+}
--- a/test/unit/forge/housekeeper/tasks/dashboardAdoption_spec.js
+++ b/test/unit/forge/housekeeper/tasks/dashboardAdoption_spec.js
@@ -1,0 +1,133 @@
+const should = require('should') // eslint-disable-line
+const sinon = require('sinon')
+
+const dashboardAdoptionTask = require('../../../../../forge/housekeeper/tasks/dashboardAdoption')
+
+const setup = require('../setup')
+
+const DASHBOARD_MODULE = '@flowfuse/node-red-dashboard'
+
+function runtimeSettings (modules = {}) {
+    return JSON.stringify({
+        instanceId: 'abc123',
+        nodes: {
+            'node-red': { name: 'node-red', version: '4.1.6', local: false },
+            ...modules
+        }
+    })
+}
+
+describe('Dashboard Adoption Task', function () {
+    /** @type {import('../setup').TestApplication} */
+    let app
+    let instances
+    let telemetryDisabled
+
+    async function createInstanceWithSettings (name, settings) {
+        const instance = await app.factory.createInstance(
+            { name },
+            app.application,
+            app.stack,
+            app.template,
+            app.projectType,
+            { start: false }
+        )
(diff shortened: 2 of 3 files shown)


forge/housekeeper/tasks/dashboardAdoption.js
```diff
@@ -0,0 +1,59 @@
+const { randomInt } = require('../utils')
+
+const DASHBOARD_MODULE = '@flowfuse/node-red-dashboard'
+
+const QUERIES = {
```
cstns, line 5: I think the raw SQL here is going to bite us in a few ways at once:

* one malformed `settings` row aborts the whole aggregate on both dialects (sqlite's `json_extract` and postgres's `::jsonb` cast both throw), and since the housekeeper just logs and moves on, the task would fail silently every day until someone notices in Sentry. All the other readers of this column wrap `JSON.parse` in try/catch for that reason.
* the two dialects don't quite agree on the answer. Postgres's `::boolean` accepts `'t'`, `'1'`, `'yes'` etc while the sqlite `= 1` check doesn't, so self-hosted and cloud numbers wouldn't be comparable. And a non boolean value like `"local": "user"` blows up the postgres query entirely.
* the postgres branch never runs in CI (tests are sqlite), so it can break without anything going red.

Might be worth considering skipping the raw SQL and doing this through the ORM instead: count via a `LIKE '%@flowfuse/node-red-dashboard%'` prefilter, fetch just the candidate rows, and do the exact `nodes[module].local === true` check in JS with a try/catch around the parse. That's dialect agnostic, tolerant of a bad row, and it's the pattern `certifiedNodes.js` already uses for this kind of data. There's also `StorageSettings.getProjectModules` and the `hasDashboard2UI` predicate in `views/Project.js` which already own this logic, so reusing one of those would keep this metric from drifting away from what the UI considers "has a dashboard".

forge/housekeeper/tasks/dashboardAdoption.js
```diff
@@ -0,0 +1,59 @@
...
+    postgres: `
+        SELECT
+            COUNT(*) FILTER (
+                WHERE (s.settings::jsonb -> 'nodes' -> :module ->> 'local')::boolean IS TRUE
+            ) AS with_dashboard,
+            COUNT(*) AS total
+        FROM "StorageSettings" s
+        JOIN "Projects" p ON p.id = s."ProjectId"
```
cstns, line 13: The inner join means `total` counts StorageSettings rows, not instances. Those rows are only created the first time Node-RED boots and posts its settings, so instances that were created but never started don't count towards the denominator, which inflates the percentage. It also makes the number hard to relate to `platform.counts.projects` from the existing telemetry ping.

Maybe count non suspended projects as the denominator instead, and treat "no settings row" as "no dashboard"? As a minor aside, there's no unique index on `StorageSettings.ProjectId`, so a project that somehow ends up with two rows would be counted twice in the total. Counting projects sidesteps that too.

forge/housekeeper/tasks/dashboardAdoption.js
```diff
@@ -0,0 +1,59 @@
...
+            END) AS with_dashboard,
+            COUNT(*) AS total
+        FROM "StorageSettings" s
+        JOIN "Projects" p ON p.id = s."ProjectId"
+        WHERE p.state <> 'suspended'`
+}
+
+module.exports = {
+    name: 'dashboardAdoption',
+    schedule: `${randomInt(0, 59)} ${randomInt(0, 23)} * * *`,
```
cstns, line 29: This wants a `startup` property alongside `schedule`. The random cron time gets re-rolled on every restart, so a platform that restarts more than once a day can keep missing its slot and never report. `telemetryMetrics.js` pairs the same random daily cron with `startup: 10000` for exactly this reason.

forge/housekeeper/tasks/dashboardAdoption.js
```diff
@@ -0,0 +1,59 @@
...
+    run: async function (app) {
+        if (process.env.FF_TELEMETRY_DISABLED) {
```
cstns, line 31: This is now the fourth copy of the telemetry opt-in check (telemetryMetrics, the settings route, and the EE MCP server have the others). Fine for now, but might be worth pulling into a shared helper so a future change to the opt-out logic doesn't miss one of them. Two bits here are also already covered elsewhere: `telemetry:enabled` is forced to true when licensed, so the `isLicensed` branch is dead, and `product.capture` is already a no-op when there's no apikey.

test/unit/forge/housekeeper/tasks/dashboardAdoption_spec.js
```diff
@@ -0,0 +1,133 @@
...
+            })),
+            notLocal: await createInstanceWithSettings('dashboard-not-local', runtimeSettings({
+                [DASHBOARD_MODULE]: { version: '1.2.3', local: false }
+            })),
+            withoutDashboard: await createInstanceWithSettings('no-dashboard', runtimeSettings())
+        }
+    })
+
+    after(async function () {
+        process.env.FF_TELEMETRY_DISABLED = telemetryDisabled
```
cstns, line 58: Small one: if `FF_TELEMETRY_DISABLED` was unset when the suite started, this assignment writes the literal string `"undefined"` back, which is truthy and would suppress telemetry behaviour in any spec file that runs after this one in the same process. Probably wants a `delete process.env.FF_TELEMETRY_DISABLED` when the saved value is undefined.

test/unit/forge/housekeeper/tasks/dashboardAdoption_spec.js
```diff
@@ -0,0 +1,133 @@
...
+
+        app.product.capture.calledOnce.should.be.true()
+        const [distinctId, event, properties] = app.product.capture.firstCall.args
+        distinctId.should.equal(app.settings.get('instanceId'))
+        event.should.equal('$ff-dashboard-adoption')
+        properties.should.match({ with_dashboard: 1, total: 3, pct: 33.3 })
+    })
+
+    it('excludes suspended instances', async function () {
+        instances.withDashboard.state = 'suspended'
```
cstns, line 84: If the assertion fails here, the instance never gets un-suspended and the remaining tests fail with confusing counts. The other tests in this file use try/finally for their cleanup, might as well do the same here.

cstns (changes requested):
`StorageSettings.getProjectModules` and the `hasDashboard2UI` predicate in `views/Project.js` already own this logic

test/unit/forge/housekeeper/tasks/dashboardAdoption_spec.js
n-lark: fixed

test/unit/forge/housekeeper/tasks/dashboardAdoption_spec.js
n-lark: fixed

forge/housekeeper/tasks/dashboardAdoption.js
n-lark: ty dropped those two checks

forge/housekeeper/tasks/dashboardAdoption.js
n-lark: added good catch

forge/housekeeper/tasks/dashboardAdoption.js
n-lark: Switched the denominator to a non-suspended `Project.count()` so never-started instances count as no-dashboard, and the numerator dedupes by `ProjectId` so duplicate settings rows can't double-count either.

forge/housekeeper/tasks/dashboardAdoption.js
n-lark: Yeah, moved it all to the ORM. `LIKE` prefilter, then parse each row in a `try/catch` so one bad row can't take out the whole count. No dialect branches left either. Also pulled the parse bit out of `getProjectModules` into `getLocalModules(settings)` so both go through the same check. It's truthy local rather than === `true`, since that's what that helper already did.

Skipped hasDashboard2UI though, it goes off `palette.modules`, which can list Dashboard on instances that never actually installed it.

cstns (approved):

Issue #8416, opened 2026-09-04
MCP OAuth refresh token flow is broken


cstns:
### Description

Third party MCP agents authenticate with a 30 minute access token and a 30 day refresh token stored on the same AccessToken row. A few things around refreshing them need attention.

**Nightly cleanup deletes active grants (witnessed in real use)**
The daily `expireTokens` housekeeper task deletes every token whose `expiresAt` has passed, without checking `refreshTokenExpiresAt`. MCP access tokens expire every 30 minutes, so the sweep removes grants that are still refreshable. The agent's next refresh fails with "Invalid refresh_token" and the token vanishes from the UI.
Fix: only delete rows where the refresh token is also expired or missing, the same guard `getOrExpire` already has.

**Missing refresh_token returns a 500**
A token request with `grant_type=refresh_token` and no `refresh_token` hits `sha256(undefined)` and throws.
Fix: validate the field and return `400 invalid_request`, like the authorization_code branch does.

**Refresh cache is per process**
Concurrent refresh dedup uses an in-memory cache. With multiple replicas each one mints its own access token and overwrites the same row, so a token handed out by another replica stops working until the next refresh.
Fix: might be worth a shared cache, or dropping the dedup entirely.

**Cached refreshes don't extend the window**
A refresh served from the cache skips the `refreshTokenExpiresAt` bump, so the sliding window only extends on some refreshes.
Fix: bump it on every refresh.

**MCP tokens look like PATs in the UI**
They're stored with `ownerType: 'user'` and a name, so they show up in the Personal Access Tokens list with the 30 minute access expiry. That already misled a user into thinking his MCP access dies every half hour. Editing one there rewrites `expiresAt` but never `refreshTokenExpiresAt`, and clearing the expiry makes the token never expire.
Fix: show the real grant lifetime for MCP tokens or move them out of the PAT list, and keep the PAT edit path away from them.

cstns:
**Re: "Refresh cache is per process", parking this one for now.**

The failure only shows up on multi-replica deployments when two refreshes for the same grant race each other. In practice OAuth clients serialize their token refreshes, and even when it does happen the damage is one 401 which the agent recovers from on its next refresh. So it's a self-healing edge case, not something users are hitting.

[...]

PR #8420, opened 2026-09-04
Keep refreshable tokens out of the nightly expireTokens sweep

### LINKED ISSUE #8421: Nightly expireTokens sweep deletes MCP grants that are still refreshable
### Description

The daily `expireTokens` housekeeper task deletes every AccessToken whose `expiresAt` has passed, without checking `refreshTokenExpiresAt`. MCP OAuth grants keep a 30 minute access token and a 30 day refresh token on the same row, so the sweep removes grants that are still refreshable. The agent's next refresh fails with "Invalid refresh_token" and the token vanishes from the user's token list. Witnessed in real use.

Fix: only delete rows where the refresh token is also expired or missing, the same guard `getOrExpire` already applies in the request path (RFC 6749 §1.5).

### Epic/Story

#8416

### Does this need to be reflected in the FlowFuse Expert?

No Expert impact

cstns:
Closes #8421 (sub-issue of #8416, the lockout we witnessed in real use).

The daily `expireTokens` housekeeper task deleted every AccessToken whose `expiresAt` had passed, without looking at `refreshTokenExpiresAt`. MCP OAuth grants keep a 30 minute access token and a 30 day refresh token on the same row, so unless the agent had refreshed within the last half hour before the task fired, the sweep deleted the whole grant. The agent's next refresh failed with "Invalid refresh_token" and the token vanished from the user's token list.

The request path already handles this correctly (`getOrExpire` keeps the row while the refresh token is valid, per RFC 6749 §1.5). This applies the same guard to the sweep: only delete rows where the refresh token is also expired or absent.

Tokens without a `refreshTokenExpiresAt` (plain PATs, editor sessions, device tokens) are swept exactly as before. Added unit tests for the task covering all four combinations.

### DIFF
--- a/forge/housekeeper/tasks/expireTokens.js
+++ b/forge/housekeeper/tasks/expireTokens.js
@@ -10,7 +10,18 @@ module.exports = {
     schedule: `${randomInt(0, 59)} ${randomInt(0, 23)} * * *`,
     run: async function (app) {
         await app.db.models.Session.destroy({ where: { expiresAt: { [Op.lt]: Date.now() } } })
-        await app.db.models.AccessToken.destroy({ where: { expiresAt: { [Op.lt]: Date.now() } } })
+        // A token row can outlive its access token expiry: MCP OAuth grants keep a
+        // longer-lived refresh token on the same row (see getOrExpire, RFC 6749 §1.5).
+        // Only remove rows once the refresh token is also expired or absent.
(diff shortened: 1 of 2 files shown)


knolleary (approved):

knolleary:
Approved - failing postgres test appears unrelated. Re-running

Issue #8367, opened 2026-09-03
Make the signup auto-provisioning reusable and skippable


cstns:
`completeUserSignup` in `forge/lib/userTeam.js` currently does two things in one go: creates the personal team, and then, if `user:team:auto-create:instanceType` is set, creates an application and a hosted instance. It runs on email verification.

Two changes needed.

**Skip the application and instance when the AI onboarding flag is on.** Team creation still happens, just not the workspace inside it. The Expert does that part later based on the conversation.

**Make that provisioning callable on its own.** Someone who bails out of the onboarding conversation needs the same result a classic signup would have given them, and so does the control cohort of the A/B. Neither of those happens at email verification, so the logic needs to be reachable from elsewhere.

### Done when

* A non-flagged signup behaves exactly as it does today, verified by tests.
* A flagged signup ends up with a team and nothing in it.
* The provisioning logic can be called with a user and a team from outside the signup flow.

### Worth knowing

There are three partial copies of "make a default application and instance" in the codebase right now:

* the block inside `completeUserSignup` (`forge/lib/userTeam.js`)
* `createTeamApplication` in `forge/routes/api/team.js`
* the `defaultTeamCreated` block further down `forge/routes/api/team.js`

Since we need a callable version anyway, it might be worth folding them into one and pointing all three callers at it rather than leaving a fourth variant behind. Worth a look at whether they actually behave identically first, because if they've drifted then consolidating changes behaviour somewhere and that needs to be a deliberate call rather than a side effect.

The existing guard only provisions if the user has no instances anywhere (`Project.byUser(user)`). That guard still makes sense when the logic is called later, so it's worth keeping rather than moving the check to the caller.

This is the one code path every user goes through, so the test bar is higher than usual here.

Issue #8365, opened 2026-09-03
Gate onboarding behind a platform flag and stop provisioning at signup


cstns:
Today `completeUserSignup` in `forge/lib/userTeam.js` runs when someone verifies their email, and depending on the `user:team:auto-create*` platform settings it creates their team, an application called `<name>'s Application`, and a hosted instance with a machine-generated name. All of that happens before the user has told us anything about what they want.

For AI-led onboarding we need the opposite: land them in a team with nothing in it, so the Expert can set things up around what they actually say.

But the classic path still needs the old behaviour, and so does anyone who bails out of the conversation. So this is three pieces:

* **Add a platform flag for AI-led onboarding.** Hidden, config-only, off by default. The frontend reads it through `FeatureChecks.ts` like the other feature checks.
* **Make the signup auto-provisioning reusable and skippable.** Skip it when the flag is on, and pull it out into something we can call later.
* **Add an endpoint that provisions the default workspace on demand.** So "I'd rather set it up myself" produces exactly what a classic signup would have.

### Worth knowing

There are currently three partial copies of "make a default application and instance": the block in `userTeam.js`, `createTeamApplication` in `forge/routes/api/team.js`, and the `defaultTeamCreated` block further down the same file. We need a callable version for the third piece anyway, so it might be worth consolidating them while we're in there rather than adding a fourth.

This touches the classic signup path, which is the one path every single user goes through, so it wants proper test cover. The bar is that a non-flagged signup behaves exactly as it does today.

The on-demand endpoint gets used twice: once by the escape hatch in the onboarding surface, and once by the control cohort when we split the A/B. Worth knowing so it doesn't get built as an onboarding-only thing.

PR #8431, opened 2026-09-07
Add hidden aiOnboarding platform flag

### LINKED ISSUE #8366: Add a platform flag for AI-led onboarding
We need a way to turn AI-led onboarding on for a deployment without exposing it as a setting anyone can toggle.

Config-only, off unless explicitly set. No admin UI. Cloud first, and whether it ever gets exposed to self-hosted installs is a separate conversation we don't need to have yet.

The frontend reads it through `FeatureChecks.ts`, same as the other feature checks, so it comes out on the `featuresCheck` getter of the `account-settings` store and the onboarding path can just ask for it.

### Done when

* The flag exists in config and defaults to off.
* The frontend can read it via `featuresCheck`.
* Nothing changes for a deployment that hasn't set it.

### Worth knowing

The AI onboarding path only works if the Expert is available and the browser can reach a broker, since everything runs over MQTT rather than HTTP. `shouldUseMqtt` in the `product-expert` store already resolves that as `isExternalMqttBrokerFeatureEnabled && isSupportAgent`, so this flag probably wants to depend on that rather than restate it. `FeatureChecks.ts` has `dependsOnPlatform` and `dependsOnTeam` for exactly this, so it may be worth wiring the dependency there instead of checking two things at every call site.

cstns:
Closes #8366

Part of #8365. This adds the config-only platform flag that AI-led onboarding will sit behind. Nothing uses it yet, the signup changes and the on-demand provisioning endpoint come in follow-up PRs.

The backend registers an `aiOnboarding` flag from `expert.onboarding.enabled` in the config, ANDed with the existing ai and expert gates, same pattern as `expertInsights`. Off unless a deployment explicitly sets it, and there is no admin UI for it.

On the frontend it comes out as `isAiOnboardingFeatureEnabled` on `featuresCheck`. The entry depends on `isExpertAssistantFeatureEnabled` plus the `externalBroker`/`teamBroker` chain, since the onboarding conversation runs over MQTT and this saves call sites from restating what `shouldUseMqtt` checks. `isSupportAgent` stays a runtime store check as it is session state rather than a feature.

Tests cover the new feature check and each dependency gate in `FeatureChecks.spec.js`. The one-line backend registration is not unit tested directly, matching how the sibling flags are handled; the signup work in #8367 will boot with the flag set and cover that wiring.

### DIFF
--- a/forge/ee/lib/index.js
+++ b/forge/ee/lib/index.js
@@ -52,6 +52,11 @@ async function commonFeatures (app, opts) {
     // explicitly disabled in the config.
     const isMcpThirdPartyEnabled = isAiEnabled && (app.config?.expert?.enabled ?? false)
     app.config.features.register('mcpThirdParty', isMcpThirdPartyEnabled, true)
+
+    // Set the AI-led onboarding flag. Hidden and config-only: off unless
+    // expert.onboarding.enabled is explicitly set in the config.
+    const isAiOnboardingEnabled = isAiEnabled && (app.config?.expert?.enabled ?? false) && app.config?.expert?.onboarding?.enabled === true
+    app.config.features.register('aiOnboarding', isAiOnboardingEnabled, true)
 }
 
 module.exports = fp(async function (app, opts) {
--- a/frontend/src/composables/FeatureChecks.ts
+++ b/frontend/src/composables/FeatureChecks.ts
@@ -131,8 +131,6 @@ export const FEATURE_CONFIGS: FeatureConfig[] = [
         output: 'isAiOnboardingFeatureEnabled',
         platformKey: 'aiOnboarding',
         dependsOn: 'isExpertAssistantFeatureEnabled',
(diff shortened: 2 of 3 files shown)


andypalmi (approved):
Looks good, approving.

One small optional thing: the inline comment on the `aiOnboarding` entry says it mirrors `shouldUseMqtt`, but that getter only checks `externalBroker` (plus `isSupportAgent`), not `teamBroker`. The `teamBroker` dependency is actually right, since on self-hosted the team broker is the local broker that bridges up to the central broker, so it is a real connection prerequisite. Might be worth having the comment say that directly instead of referencing `shouldUseMqtt`.

Separately, I know #8366 says self-hosted exposure is a later conversation, so nothing to change in this PR. But would it be worth filing follow-up issues on the helm and docker-compose repos now to track adding this flag there, so it does not get lost when that conversation happens?

Issue #8368, opened 2026-09-03
Add an endpoint that provisions the default workspace on demand


cstns:
Once signup stops provisioning automatically, something has to be able to do it later. Two callers need it:

* someone who clicks out of the onboarding conversation and would rather set things up themselves
* the control cohort of the A/B, which calls it on arrival so those users get the classic experience

Both want the same outcome: the application and instance a classic signup would have created, machine-generated names and all.

So this is an endpoint over the callable provisioning logic, with the usual schema, permission check and tests.

### Done when

* A team with nothing in it can be brought to the same state as a classic signup with one call.
* It refuses politely if the team already has instances, rather than making a second set.
* Permissions are right: a member of the team can call it, nobody else can.

### Worth knowing

Deciding factor for the shape of this: the A/B assignment has to happen in the browser, not on the server. `forge/product/index.js` only decorates `capture`, so there's no server-side PostHog flag evaluation available, and the auto-provision decision used to happen at email verification when no browser is running at all. Routing both callers through one endpoint is what lets the assignment stay client-side without adding server-side flag evaluation.

Worth thinking about whether this returns the created application and instance, or just a success. The onboarding surface probably wants the instance back so it can navigate the user to it, but that's a guess rather than a requirement.

One thing to settle: whether this is idempotent or whether a second call is an error. Idempotent is friendlier if the browser retries, but it hides a bug where something calls it twice for different reasons.

PR #8432, opened 2026-09-07
Consolidate default workspace provisioning and skip it for AI onboarding

### LINKED ISSUE #8367: Make the signup auto-provisioning reusable and skippable

cstns:
Closes #8367

Part of #8365, stacked on #8431.

`completeUserSignup` used to create the default application and instance inline at email verification, and `forge/routes/api/team.js` had two more partial copies of the same logic (the trial block and the `user:team:auto-create:application` path). This folds all of it into two Team controller methods: `createDefaultApplication` and `provisionDefaultWorkspace`. The latter is what the on demand endpoint in #8368 will sit on top of. It guards with typed errors (`team_not_empty`, `invalid_instance_type`, `invalid_stack`, `invalid_template`) and returns the created application and instance so callers can do their own auditing and navigation.

When the `aiOnboarding` flag from #8431 is on, signup now stops after invite handling and team creation, so the user lands in an empty team and the Expert can set things up from the conversation. This covers all four signup callers (email verify, admin create, both SSO paths).

Two deliberate behavior changes: the application auto created at signup now also shows up in the team and application audit logs (it previously only emitted the user scoped `verify.autoCreateApplication` event), and the trial path gained the team empty guard and application reuse, both of which are no ops for a brand new team.

The trial block and the auto create application path had no test coverage at all, so this adds pin tests for them before the refactor, plus tests for the new controller methods and the flagged signup. The existing signup suite passes unchanged, which was the bar here.

### DIFF
--- a/forge/routes/api/team.js
+++ b/forge/routes/api/team.js
@@ -1,5 +1,3 @@
-const crypto = require('crypto')
-
 const { Op } = require('sequelize')
 
 const { Roles } = require('../../lib/roles')
@@ -515,17 +513,6 @@ module.exports = async function (app) {
         }
     })
 
-    async function createTeamApplication (user, team) {
-        const applicationName = `${user.name}'s Application`
-        const application = await app.db.models.Application.create({
-            name: applicationName.charAt(0).toUpperCase() + applicationName.slice(1),
-            TeamId: team.id
-        })
-        await app.auditLog.Team.application.created(user, null, team, application)
-        await app.auditLog.Application.application.created(user, null, application)
-        return application
-    }
-
     /**
      * Create a new team
      * /api/v1/teams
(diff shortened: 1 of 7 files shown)


forge/routes/api/team.js
```diff
@@ -625,25 +612,11 @@ module.exports = async function (app) {
...
-                            app.log.warn(`Unable to create Trial Instance in team ${team.hashid}: Instance type with id ${instanceTypeId} from 'user:team:auto-create:instanceType' not found`)
-                        } else if (!instanceStack) {
-                            app.log.warn(`Unable to create Trial Instance in team ${team.hashid}: Unable to find a stack for use with instance type ${instanceTypeId}`)
-                        } else if (!instanceTemplate) {
-                            app.log.warn(`Unable to create Trial Instance in team ${team.hashid}: Unable to find the default instance template`)
-                        } else {
-                            const safeTeamName = team.name.toLowerCase().replace(/[\W_]/g, '-')
-                            const safeUserName = request.session.User.username.toLowerCase().replace(/[\W_]/g, '-')
-                            const application = await createTeamApplication(request.session.User, team)
+                        try {
```
andypalmi, line 615: Previously a provisioning failure here bubbled up to the route's outer catch and failed the whole team-create request with `unexpected_error`; now it's caught and logged, so the request returns 200 whether or not the instance was created. 

Failing the request would be worse since the team already exists, so swallowing seems right. 

But right now the catch discards both the result and the error, so the response is identical whether the instance was created or not. 

Could we instead carry the outcome into the reply, for example add a 
`defaultInstance: { provisioned: true }` on success 
and `defaultInstance: { provisioned: false, error: err.code }` (`invalid_instance_type`, `invalid_stack`, `invalid_template`, `team_not_empty`) in the catch, so the frontend can tell the instance is missing and show a "your workspace didn't finish setting up, retry" prompt rather than sending the user into a team that looks complete but has no instance?

andypalmi (approved):
I am going to approve it and suggest we follow up on this

https://github.com/FlowFuse/flowfuse/pull/8432#discussion_r3959563598

PR #8435, opened 2026-09-07
Run only forge unit and system tests in the postgres CI job

### LINKED ISSUE #8434: postgres-tests CI job re-runs lint and frontend unit tests for nothing
The `postgres-tests` job in `tests.yml` runs `npm run test`, which expands to lint, forge unit tests, frontend unit tests and system tests.

Lint and the frontend vitest suite never touch the database, and the `ui-unit-tests` job already runs both. So the postgres job spends time repeating work that gains nothing from running against Postgres. It can even end up running the frontend tests when the UI jobs were skipped by the changed files filter, for example on a migration only change.

The parts that actually benefit from a Postgres run are the forge unit tests and the system tests, which is also exactly what `backend-tests` runs against SQLite. Might be worth trimming the postgres job down to just those two suites, without coverage instrumentation since Codecov already gets its report from the SQLite run.

cstns:
Closes #8434

The `postgres-tests` job was running `npm run test`, which includes lint and the frontend vitest suite. Neither touches the database and both already run in the `ui-unit-tests` job, so the postgres run was just repeating them.

The job now runs `npm run test:unit:forge && npm run test:system`, matching what `backend-tests` runs against SQLite. Coverage instrumentation is skipped too, Codecov already gets its report from the SQLite run.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -131,7 +131,7 @@ jobs:
           FF_TEST_DB_POSTGRES_PASSWORD: secret
           FF_TEST_DB_POSTGRES_DATABASE: flowforge
           NODE_OPTIONS: "--max-old-space-size=6144"
-        run: npm run test
+        run: npm run test:unit:forge && npm run test:system
 
   ui-unit-tests:
     if: ${{ needs.check-changes.outputs.run_ui_tests == 'true' }}


.github/workflows/tests.yml
```diff
@@ -131,7 +131,7 @@ jobs:
           FF_TEST_DB_POSTGRES_PASSWORD: secret
           FF_TEST_DB_POSTGRES_DATABASE: flowforge
           NODE_OPTIONS: "--max-old-space-size=6144"
-        run: npm run test
+        run: npm run test:unit:forge && npm run test:system
```
ppawlowski, line 134: How will this step differ from the `backend-test` job, which executes exactly the same tests:
https://github.com/FlowFuse/flowfuse/pull/8435/changes#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fL82-L85 ?

cstns: It binds a pg container and the test step is configured via env variables to use pg as it's db. The `backend-tests` job is running the exact suite of tests but running with mysqli

ppawlowski: Cool. wasn't aware of this. Thanks for the clarification!

ppawlowski (approved):

Issue #8384, opened 2026-09-03
Carry the conversation into the editor when the Expert navigates


cstns:
When the user approves the plan, the Expert navigates them into the immersive editor so it can build. The onboarding surface can't go with them, because it's rendered on a team page. The Expert drawer can, because `RightDrawer` is mounted above the router views and survives navigation.

So at that moment the surface needs to close and the drawer needs to open, with the conversation intact.

The good news is that there's nothing to transfer. The messages live in the store, not in either renderer, so this is unmounting one and mounting the other. `openAssistantDrawer` already exists in `product-expert.js`.

### Done when

* Approving a plan takes the user into the editor with the conversation visible in the drawer.
* Nothing is lost across the transition: transcript, pending state, whatever the Expert was mid-way through.
* The surface doesn't linger behind the editor or reopen afterwards.

### Worth knowing

Timing is the risk. The navigation and the surface teardown and the drawer opening all happen around the same moment, and the editor takes a while to load. Getting the order wrong gives you either a flash of nothing or a surface briefly on top of a loading editor.

Worth deciding what triggers it. The Expert makes a `ui_navigate` call, so the handoff could hang off that, or off the route change itself. Hanging it off the route change is probably more robust, since it also covers the user navigating away by other means.

flowfuse#8323 is directly relevant: the Expert can start writing flows before the immersive editor has finished loading. Whatever this task does about ordering may overlap with that fix, so worth reading it first.

The drawer has a pinned and an unpinned state. Coming out of a full-attention surface into a collapsed drawer would feel like the conversation vanished, so it probably wants to arrive pinned regardless of what the user's previous preference was. Worth checking whether that stomps a stored preference they'd expect to keep.

andypalmi:
Approving a plan now carries the conversation into the editor drawer with the transcript intact, verified across all three live runs. Closing as done.

Issue #8382, opened 2026-09-03
Send new signups into onboarding, and give them a way out


cstns:
Two halves of the same thing: getting people into the conversation, and letting them leave it.

**In.** When the flag is on, a new signup lands in onboarding after team creation. No question about whether they want it, no choice screen first. It just starts, with the Expert already talking.

**Out.** A control that's always available and says something like "I'd rather set it up myself". Pressing it calls the on-demand provisioning endpoint and drops them on the classic team home with the application and instance a classic signup would have given them.

The out control stays visible the whole way through, but goes quieter once the user engages with the conversation. Present, not competing.

### Done when

* A flagged signup ends up in onboarding without being asked.
* A non-flagged signup goes exactly where it goes today.
* The way out produces the same result as a classic signup and never leaves a team with nothing in it.
* The control is reachable at any point in the conversation, and recedes visually after the user engages.

### Worth knowing

Where onboarding sits in the routing is worth thinking about rather than assuming. A dedicated route gives it a URL, which makes it returnable and makes the PostHog funnel cleaner to instrument. A modal over the team home is less structural but means the user is nowhere in particular. Either works, and the surface direction being chosen on the parent might decide it.

Billing won't get in the way. `requiresBilling` is false while `isTrialAccount` is true and the trial hasn't ended, so `team/index.vue` won't bounce a fresh trial signup to the billing page before onboarding can render.

The current path after email verification sets a flag that triggers the welcome tour, which is being suppressed on this path in a sibling task. Worth coordinating so the two don't fight over what happens on arrival.

"Recedes after engagement" needs defining. Probably first answer submitted, or first message sent, rather than something time-based. Worth picking something observable rather than a timer.

Issue #8380, opened 2026-09-03
Build the onboarding UI


cstns:
Where the onboarding conversation actually happens. Four pieces underneath: the surface itself, getting new signups into it and giving them a way out, stopping the existing welcome tour and education modal from firing on this path, and handing the conversation over to the Expert drawer when the user gets taken into the editor.

The surface embeds the existing expert components and reads the same store, so it's the same conversation rendered differently rather than a second chat implementation. Restyled chrome, shared innards.

### Direction still to be agreed

We have four sketched treatments. This needs a team decision rather than being settled here, so the mockups are below and the tradeoffs are summarised so we can argue about it properly.

**Standalone conversation.** Full window, one column, the transcript grows above you. Progress is implied rather than shown, carried by what the Expert says.

<img width="1129" height="1106" alt="Image" src="https://github.com/user-attachments/assets/9d446a0a-d890-4147-89c1-4569ef5786d2" />

---

**One question, full bleed.** Each question owns the screen with large option tiles. A thin unlabelled thread is the only progress signal.

<img width="1122" height="690" alt="Image" src="https://github.com/user-attachments/assets/0310f236-6dbb-4308-aafc-b23748df00c1" />

---

**Conversation beside the build.** Left half conversation, right half an integral panel showing what exists so far, so provisioning is something you watch rather than something you're told about.

<img width="1130" height="788" alt="Image" src="https://github.com/user-attachments/assets/db6bde78-7274-4534-b0be-45eb719df47e" />

---

**Collapsing transcript.** Answered questions fold into one quiet line each, with the picks kept as editable chips. The stack of folded answers is the progress indicator.

<img width="1134" height="919" alt="Image" src="https://github.com/user-attachments/assets/22d7fc0f-a99d-48d6-a6c0-35bb0818b9be" />

--- 

### Recommended approach

For what it's worth, my reading is that the collapsing transcript is the strongest fit for 3.1, with the right-hand build panel added later. Two reasons, both practical rather than aesthetic:

* The Expert decides at runtime how many questions to ask, one to four per turn. Anything that commits to a fixed number of steps up front will eventually be wrong, and a progress indicator that lies is worse than none. A stack that grows with what actually happened can't be wrong.
* Answers staying visible and editable means a wrong pick is cheap to fix, which matters when the options are being read by someone unfamiliar with the domain language.

[...]

Issue #8369, opened 2026-09-03
Open the onboarding conversation without waiting for the user


cstns:
The Expert is purely reactive today. It answers when spoken to. `Verify input` in the expert flows rejects any request that doesn't carry a `payload.query` or an abort signal, so a chat session can't exist until the user has typed something.

Onboarding needs the other way round. The user lands and the Expert is already talking, already asking the first thing it needs to know. Waiting for a brand-new user to open with a question is asking them to know what to ask, which is the whole problem we're trying to get away from.

Two pieces:

* **Send the onboarding flag to the Expert with each turn.** So it knows to behave differently. One field on the context object.
* **Let the Expert open a chat session before the user types.** Relax the input check for onboarding, and have it greet with the first question set.

### Worth knowing

We're keeping this on the existing `support` agent channel with a flag, rather than adding a fourth channel alongside `support`, `insights` and `mcp`. A new channel would mean editing the two hardcoded `support` chat rules in `emxq-bridge/templates.js`, re-provisioning EMQX, adding ACL entries and new IO wiring, all to carry a boolean. The tradeoff is that onboarding behaviour then lives inside the same system prompt as normal support, so it's conditional branching rather than isolation. Worth revisiting if that gets unwieldy, but not worth paying for up front.

The second piece is a change to the flows on `expert-api-dev`, not to this repo. Filed here to keep the epic in one place.

There's a related thing we'll want but haven't specified yet: what the actual opening questions are. That's content design rather than plumbing and it's getting its own issue.

PR #8438, opened 2026-09-07
Build the onboarding surface shell

### LINKED ISSUE #8381: Build the onboarding surface shell
The surface the onboarding conversation renders in. It embeds the existing expert components and reads the same store, so it's the shipped chat with different chrome around it rather than a parallel implementation.

That's the important constraint. `product-expert-support-agent.js` holds the messages, so any renderer over that store shows the same conversation. If this ends up with its own message state, the handoff to the drawer stops working and we've got two chat implementations to keep in step.

### Done when

* The conversation renders in the onboarding surface identically to how it renders in the drawer.
* Messages, question cards, plan cards and tool calls all work without surface-specific handling.
* No message state lives in the surface.

### Worth knowing

Layout direction is a team decision that hasn't been made yet. Four treatments are sketched on the parent issue. This task is the shell and the wiring, so most of it is the same whichever way that lands, but it's worth waiting for the decision before finalising the chrome rather than building one and reworking it.

Components to reuse rather than reimplement: `ExpertMessages`, `AiMessage`, `HumanMessage`, `AnswerWrapper` and everything it dispatches to (`QuestionsList`, `PlanCard`, `ToolCalls`, the resource cards), and `ExpertChatInput` or a restyled equivalent.

If the chosen direction has a step indicator of any kind, `MultiStepForm.vue` and `StepSlider.vue` already exist and the create-instance and create-application flows both use them, so borrowing that keeps the visual language consistent with the rest of the platform. Worth noting the recommendation on the parent argues against a fixed step count, so this may not apply.

The onboarding path is feature-gated, so this surface should never render for a deployment that hasn't turned the flag on, and the classic signup path has to be completely unaffected.

cstns:
Closes #8381

Part of #8380, stacked on #8436. The surface the onboarding conversation renders in, using the collapsing transcript treatment we settled on.

The page lives at `/team/:team_slug/onboarding` on the plain layout, deliberately outside the `team` route tree so there is no RightDrawer (the conversation can never render twice), no platform chrome and no billing bounce. It renders a 404 in place whenever onboarding does not apply: flag off, or the team already has instances, mirroring the backend guards on the provisioning endpoint.

The conversation is the existing `Expert.vue` panel mounted whole, reading the same store as the drawer. No message state lives in the surface, which is what keeps the future drawer handoff (#8384) free. An injected `expert-surface` variant (defaulting to `drawer`, so the drawer is untouched) switches the treatment: answered question turns fold into quiet lines with the picks as editable pill chips (a new pure `buildCollapsedTranscript` helper plus a `CollapsedQuestionTurn` component), the info banner, canned welcome message, Start over, Plan mode and settings controls are hidden, and the composer reads "Or just tell me in your own words". The rest of the treatment (flattened bubbles, the active question card with full width option tiles) is scoped deep CSS in the page, so deleting the page deletes the look.

The header gets a teleport target in the plain layout, and the page teleports a "Set it up myself" escape hatch into it, wired to the on-demand provisioning endpoint from #8436: provision, then land on the classic team home, with a 409 `team_not_empty` treated as already provisioned. The receding-after-engagement behaviour stays with #8382.

Since the Expert cannot open a conversation on its own yet (#8369), the page seeds a hardcoded placeholder transcript when the conversation is empty, clearly marked temporary with a pointer to that issue.

Two unrelated fixes are split into their own commits: `hydrateMessages` silently dropped every user message (its switch(true) matched a truthy string against `true`), which also affects the editor to platform handoff, and the horizontal resize bar painted a stray 1px tick inherited from the vertical variant's edge divider.

Verified against the mockup in a live browser in light and dark themes. 31 new unit tests across the fold helper, the folded turn component, the page and the hydrate regression.

### DIFF
--- a/frontend/src/api/team.ts
+++ b/frontend/src/api/team.ts
@@ -573,6 +573,15 @@ const deleteGitToken = async (teamId: string, tokenId: string) => {
     return client.delete(`/api/v1/teams/${teamId}/git/tokens/${tokenId}`)
 }
 
+/**
+ * Provision the default workspace (application + instance) in an empty team.
+ * Used by AI-led onboarding's "set it up myself" escape hatch. 409s with
+ * code 'team_not_empty' if the team already has instances.
+ */
+const provisionDefaultWorkspace = async (teamId: string) => {
+    return client.post(`/api/v1/teams/${teamId}/default-workspace`).then(res => res.data)
+}
+
 const getTeamInstanceCounts = async (teamId: string, states: string[], type: string, applicationId: string | null = null) => {
     const params = new URLSearchParams()
     states.forEach(state => params.append('state', state))
@@ -627,5 +636,6 @@ export default {
     getTeamDeviceGroups,
     getGitTokens,
     createGitToken,
-    deleteGitToken
+    deleteGitToken,
+    provisionDefaultWorkspace
 }
--- a/frontend/src/components/ResizeBar.vue
+++ b/frontend/src/components/ResizeBar.vue
@@ -58,6 +58,10 @@ export default {
     &.horizontal {
         width: 100%;
         height: 6px;
+        // the base border-right is the drawer-edge divider for the vertical
+        // variant; in horizontal mode it painted a stray 1px tick at the
+        // right end of the bar
+        border-right: none;
 
         &::before {
             writing-mode: horizontal-tb;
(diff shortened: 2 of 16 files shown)


cstns:
Splitting this into four smaller PRs: the two bug fixes as standalone PRs off main, the expert surface variant machinery, and the onboarding page itself.

PR #8439, opened 2026-09-07
Add the onboarding surface variant to the expert components

### LINKED ISSUE #8385: Let people skip a question or answer in their own words
The Expert's question cards are currently all-or-nothing. `QuestionsList.vue` gates its Send button on every question having an answer, and it only renders radios and checkboxes, so there's no way to say something that isn't on the list.

Both of those are wrong for onboarding, where the whole point is that we're guessing at what someone's situation is. If none of four options fits, the user is stuck choosing a wrong one or abandoning the card.

Two changes:

**Skippable.** Drop the requirement that every question is answered. The Expert should get on with it using what it has, and treat an unanswered question as "they didn't want to say" rather than waiting.

**An "Other" option with free text.** Picking it reveals a text input, and whatever they type goes back as that question's answer.

### Done when

* Send works with one or more questions left blank.
* An Other option accepts free text and the Expert receives it as the answer to that question.
* The Expert handles a missing answer without re-asking the same thing.

### Worth knowing

There's a related problem in the same file worth fixing while it's open. `compose()` currently flattens every answer into one text string, `"question answer1, answer2"` per line, and sends it as an ordinary chat message. Which option belonged to which question isn't retained anywhere, on either side. That's fine for the Expert reading prose, but it means nothing downstream can use the structured answers, and a couple of things want to: the traits the Expert deduces about how technical the user is, and any surface treatment that shows answered questions as editable chips.

There's an existing escape valve that partly covers custom input already: the Edit button emits the composed answers into the message box so the user can rewrite before sending. Worth keeping, since it handles "your question is wrong" rather than just "your options are wrong".

The schema side needs a small change too, so the Expert knows an answer may be absent rather than assuming every question comes back filled. That part lands in the expert flows.

Use the existing `ff-radio-group` and `ff-checkbox` components for the Other option rather than raw inputs, so it stays consistent with the rest of the card.

cstns:
Part of #8381 / #8380, stacked on #8436. First of two PRs for the onboarding surface; the page that activates all of this comes in the next one.

This teaches the shared expert chat components an `expert-surface` variant, injected with a `'drawer'` default so nothing changes anywhere until a surface provides `'onboarding'`. Under that variant:

* `ExpertMessages` renders through a new pure `buildCollapsedTranscript` helper: past question turns (the AI questions message plus the composed human reply) fold into `folded-turn` entries, the collapsing transcript treatment from the mockups on #8380. The drawer path maps messages through unchanged.
* A new `CollapsedQuestionTurn` renders the fold: one quiet line per question with the picks as pill chips (one chip per pick for multi selects), a Skipped marker for unanswered stale turns, expand/collapse to the original messages, and chip clicks loading the "question answer" line into the composer, since correcting a sent answer means sending a new message.
* `Expert.vue` hides the info banner and skips the canned welcome message (the Expert opens the onboarding conversation itself, #8369).
* `ExpertChatInput` hides Start over, Plan mode and the settings cog, and reads "Or just tell me in your own words".

Everything is inert without a provider, so the reviewable claim here is that the drawer rendering is byte identical. Covered by 12 unit tests over the fold helper and the folded turn component.

Closes https://github.com/FlowFuse/flowfuse/issues/8385


andypalmi:
A few things we spotted in the onboarding surface variant:

- No free-text option on question cards, and an Edit button rather than letting users type their own answer.
- A composer reply that matches no question is repeated as the answer under every folded question.
- The folded chip uses a `×`, which reads as delete, but clicking it loads the answer back into the composer to edit.
- Each folded question has its own left bar, so one turn reads as several separate items when collapsed while it is a single bubble when expanded.
- Re-expanding a folded turn re-streams the card, so the title appears before its options, and collapsing is instant with no animation.

Two open questions:

1. A reply typed in the composer might read better as Skipped plus the raw reply, rather than attributed back to the questions.
2. A multi-select answer is split into one pill per selected option (for example A, B, E become three pills), but the pills are not individually actionable, clicking any of them loads the whole answer. A single pill per question may be clearer.

andypalmi:
Fixes for the above:

- #8475: free-text answer on every question card; Edit button replaced by a single Send.
- #8477 (stacked on #8475): unmatched composer reply now shown once per turn; chip `×` swapped for an edit icon; folded questions share one left bar; re-expanding renders the card at once and the expand and collapse animate.

andypalmi (approved):
Approving since we merged #8475 and #8477 in this branch

PR #8440, opened 2026-09-07
Add the AI-led onboarding page

### LINKED ISSUE #8381: Build the onboarding surface shell

cstns:
Closes #8381

Part of #8380, stacked on #8439, which carries the expert component changes this page activates.

The page lives at `/team/:team_slug/onboarding` on the plain layout, deliberately outside the `team` route tree: no RightDrawer (the conversation can never render twice), no platform chrome, no billing bounce. It provides `expert-surface: 'onboarding'` and mounts the existing `Expert.vue` panel whole, so the conversation state stays in the store and the future drawer handoff (#8384) is free. It renders a 404 in place whenever onboarding does not apply (flag off, or the team already has instances), mirroring the backend guards on the provisioning endpoint.

The plain layout header gains a teleport target, and the page teleports a "Set it up myself" escape hatch into it, wired to the on-demand provisioning endpoint from #8436 through a new `teamApi.provisionDefaultWorkspace` wrapper: provision, then land on the classic team home, with a 409 `team_not_empty` treated as already provisioned. The receding-after-engagement behaviour stays with #8382.

The rest of the collapsing-transcript look (flattened bubbles, the active question card with full width option tiles, the quiet composer) is scoped deep CSS inside the page, so deleting the page deletes the treatment.

Since the Expert cannot open a conversation on its own yet (#8369), the page seeds a hardcoded placeholder transcript when the conversation is empty or holds only canned messages, clearly marked temporary with a pointer to that issue. Note the placeholder's user replies render fully only once #8442 (the hydrateMessages fix) is merged.

Verified against the mockup in a live browser in light and dark themes. 14 unit tests over the guard, the teleported escape hatch and the fixture seeding.


frontend/src/pages/team/Onboarding.vue
```diff
@@ -0,0 +1,253 @@
...
+        <template v-if="team && !notAvailable">
+            <!-- The layout's header (the teleport target) finishes mounting
+                 after this page does, so the teleport waits a tick -->
+            <Teleport v-if="teleportReady" to="#plain-layout-actions">
+                <a
+                    href="#"
+                    class="skip-onboarding"
+                    data-action="skip-onboarding"
+                    @click.prevent="skipOnboarding"
+                >Set it up myself</a>
```
andypalmi: Since this provisions the workspace on click, it is an action rather than navigation. Could we make it a real button so it gets proper keyboard and screen-reader behaviour, and reuse `ff-button` while we are here? The scoped `.skip-onboarding` styles were mostly mimicking a link, so they can largely drop once this is a tertiary button (keep any spacing the header needs).

```suggestion
                <ff-button
                    kind="tertiary"
                    class="skip-onboarding"
                    data-action="skip-onboarding"
                    :disabled="provisioning"
                    @click="skipOnboarding"
                >Set it up myself</ff-button>
```

frontend/src/stores/ux.js
```diff
@@ -20,12 +41,22 @@ export const useUxStore = defineStore('ux', {
...
+            }
         },
+        startOnboardingBuild () { this.onboardingStage = ONBOARDING_STAGES.BUILDING },
+        endOnboarding () { this.onboardingStage = ONBOARDING_STAGES.DONE },
         openOverlay () { this.overlay = true },
         closeOverlay () { this.overlay = false }
     },
     persist: {
-        pick: ['isNewlyCreatedUser', 'userActions'],
+        pick: ['isNewlyCreatedUser', 'onboardingStage', 'userActions'],
```
andypalmi, line 59: `onboardingStage` is persisted only to `localStorage`, so a user who finishes or skips onboarding on one browser and then opens another (or clears site data) within their first week gets reseeded to `intake` by `checkIfIsNewlyCreatedUser` and dropped back into onboarding. Since `user.createdAt` is already durable, would it be worth holding the completed/stage state server-side per user so it survives across devices? There is no existing preferences column to reuse, so it would mean a small new `User.onboardingStage` column and migration threaded through the existing `PUT /api/v1/user` endpoint (`tcs_accepted` is a close precedent). It might sit more naturally alongside the enter and leave logic in the follow-up PR than here. Flagging for a decision rather than asking for it in this PR.

frontend/src/stores/ux.js
```diff
@@ -20,12 +41,22 @@ export const useUxStore = defineStore('ux', {
...
+            // Only seeds while undecided. isNewlyCreatedUser is recomputed on
+            // every boot for a week, so seeding unconditionally would restart
+            // onboarding for anyone who had already finished or skipped it.
+            if (this.onboardingStage === null) {
+                this.onboardingStage = this.isNewlyCreatedUser
+                    ? ONBOARDING_STAGES.INTAKE
+                    : ONBOARDING_STAGES.DONE
+            }
         },
+        startOnboardingBuild () { this.onboardingStage = ONBOARDING_STAGES.BUILDING },
```
andypalmi, line 53: I could not find a caller for `startOnboardingBuild()` anywhere in the stack, so the stage only ever moves `intake` to `done` and never into `building`. Could you say a bit about what the `building` stage is meant to track: when does something flip it to `building`, when does it leave, and what reads it? Mainly want to understand the intended lifecycle. If it is groundwork for a later PR that is completely fine. (No question on `isOnboarding`, that one clearly feeds the onboarding flag into the assistant context.)

andypalmi (commented):
This is looking really good, the stage model and the folded-transcript surface read cleanly. A few inline notes below: one small suggested change on the escape-hatch button, and two questions (onboarding-state persistence, and the `building` stage).

frontend/src/stores/ux.js
cstns: It's a sensible request, up to now we've been trying to stay away from these types of additions on the backend; In my mind a UserProperties tbale would scale better, we'd also be able to store theme (light dark), i18n preference when we get there too. We can stick to localStorage for now, but lets raise an issue in the engineering meeting for it next Wednesday

frontend/src/stores/ux.js
cstns: the building stage is meant for the stage where the plan having been approved the agent starts implementing it; it is then switched to completed when done

cstns: not in use today though

andypalmi (approved):
Approving. The replies on all three threads land well: localStorage for now with the UserProperties table to raise separately, and the building stage plus startOnboardingBuild() as intended groundwork that nothing calls yet. Nothing blocking here.

PR #8450, opened 2026-09-07
Add comms heartbeat to detect lack of message flow

### LINKED ISSUE #8362: CommsClient heartbeat health check
We have seen a few cases when the MQTT messages between the forge app and broker appear to stop flowing. No errors are logged on either side - but the MQTT messages just stop.

This is most apparent with Device Checkin messages. We have ops alerting in place for when Device status update log messages drop as a stop-gap solution. A restart of the forge app 'fixes' the connection.

We haven't been able to identify the root cause as it happens very rarely with no event to correlate it against.

The purpose of this task is to introduce an application-level heartbeat to the Comms connection.

It should:
 - publish to a dedicated heartbeat topic every 30seconds with an incrementing payload.
 - subscribe to the same topic and validate the heartbeats arrive
 - if 2 heartbeats are missed, trigger a comms client reconnect.

The code should log these events clearly so we can monitor their occurrences.

knolleary:
Closes #8362 

This adds a heartbeat to the CommsClient MQTT connection.

- It publishes to `ff/v1/platform/${this.platformId}/heartbeat` every 15 seconds with a payload of the current time in ms.
- It subscribes to the same topic.
- A watchdog thread checks how long ago the last heartbeat message was received.
   - <30s - that's fine
   - 30-60s - log a warning - `No comms heartbeat received for ${delta} ms`
   - \>60s - disconnect and reconnect the mqtt client - `No comms heartbeat received for ${delta} ms - reconnecting`

For the issue we're trying to handle, messages appear to stop flowing after a client reconnect. The underlying mqtt client handles resubscribing - the theory here is something goes wrong with those subscriptions.

We cannot simply call `client.reconnect` in case it is the internal state of the local client that has gone wrong. Instead we disconnect the existing client and reinitialise a new one. Some care has gone into this PR to ensure the old client is properly cleaned up and doesn't interfere with the new one.

One key assumptions about the issue we're trying to fix here: That *all* subscriptions are similarly impacted - so adding one more here works as a detection of a problem elsewhere.

Given we cannot reproduce the issue at will, there is definitely some speculation about whether this fix will help detect it.

The most notable symptom is the Device Status checkin messages not getting through. They are on a shared subscription. *If* the underlying issue is broker-side related to the shared sub, then this new non-shared subscription isn't going to necessarily spot it.

An alternative would be to send a heartbeat on the device checkin topic - with placeholder team/device ids that can be passed through the ACL check. Going to think about that separately - but we can get this watchdog in place first.

### DIFF
--- a/forge/comms/aclManager.js
+++ b/forge/comms/aclManager.js
@@ -706,6 +706,8 @@ module.exports = function (app) {
     const ACLS = {
         forge_platform: {
             sub: [
+                // Receive cluster heartbeat from the platform
+                { topic: /^ff\/v1\/platform\/[^/]+\/heartbeat$/ },
                 // Receive status events from project launchers
                 // - ff/v1/<team>/l/<instance>/status
                 { topic: /^ff\/v1\/[^/]+\/l\/[^/]+\/status$/, shared: true },
@@ -735,6 +737,8 @@ module.exports = function (app) {
                 { topic: /^ff\/v1\/[^/]+\/u\/[^/]+\/s\/[^/]+\/[^/]+$/, shared: true }
             ],
             pub: [
+                // Heartbeat topic - platform can publish to its heartbeat topic to ensure broker is alive and we're connected
+                { topic: /^ff\/v1\/platform\/[^/]+\/heartbeat$/ },
                 // Send commands to project launchers
                 // - ff/v1/+/l/+/command
                 { topic: /^ff\/v1\/[^/]+\/l\/[^/]+\/command$/ },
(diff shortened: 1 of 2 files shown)


hardillb:
Just double checking, because each instance (in the case of FFC where we have 2) has it's own `platformId` so they should each work and reset independently of each other, but to get the symptoms we are seeing they both must be failing at the same time? (assuming shared subscriptions group should be sending half of device check in to each instance)

hardillb (approved):
Approved with the comments about if this is a shared subscription problem this won't catch it from our chat this morning.

cstns (approved):

knolleary:
Going to revisit this before merging. Following discussion with Ben, the shared subscription aspect is actually fairly critical here. A heartbeat running on the device checkin topic is likely more valuable.

PR #8452, opened 2026-09-08
Fix form validation of azure repo urls

### LINKED ISSUE #8451: Azure Git repository URLs rejected by pipeline stage form (regression from #8058)
### Current Behavior

No Azure DevOps repository URL can be saved on a pipeline git stage. The validation in `frontend/src/pages/application/PipelineStage/form.vue:729` reads:

```
/^https:\/\/dev\.azure\.com\/[^/]+\/[^/]\/_git\/[^/]+$/
```

The project segment is `[^/]`, matching exactly one character, so only a one character project name passes. Both valid Azure forms are rejected:

- `https://dev.azure.com/org/project/_git/repo`
- `https://dev.azure.com/org/_git/repo` (accepted before #8058)

### Expected Behavior

Both forms validate, matching what `forge/ee/lib/gitops/backends/azure.js` and `frontend/src/components/pipelines/Stage.vue` already accept:

```
/^https:\/\/dev\.azure\.com\/[^/]+(\/[^/]+)?\/_git\/[^/]+$/
```

### Steps To Reproduce

1. Team settings, add an Azure git token.
2. Application, pipeline, add a stage of type Git Repository.
3. Enter `https://dev.azure.com/myorg/myproject/_git/myrepo`.
4. "Please enter a valid GitHub or Azure DevOps repository URL", and the stage cannot be saved.

### Environment

- FlowFuse version: v2.33.2, v3.0.0, v3.0.1 and main. Confirmed present in the bundle FlowFuse Cloud is currently serving.
- Platform/OS: all
- Browser: all

### Notes

Regression from #8058, which fixed #8055 on the backend but left the form validation wrong, so the July workaround of dropping the project name no longer works either. Blocking an Enterprise Cloud customer: https://app.hubspot.com/contacts/26586079/record/0-2/210431677688

knolleary:
Fixes #8451 

Fixes the regex for url validation of azure git repos.

For clarity, I've separated the github/azure error reporting - was odd to say 'enter a github or azure url' when we know which it should be based on the token type.

### DIFF
--- a/frontend/src/pages/application/PipelineStage/form.vue
+++ b/frontend/src/pages/application/PipelineStage/form.vue
@@ -725,10 +725,19 @@ export default {
             const type = this.selectedGitTokenType
             if (url === '') {
                 this.errors.url = ''
-            } else if (type === 'github' || type === 'azure') {
-                this.errors.url = (/^https:\/\/github\.com\/[^/]+\/[^/]+$/.test(url) || /^https:\/\/dev\.azure\.com\/[^/]+\/[^/]\/_git\/[^/]+$/.test(url))
+            } else if (type === 'github') {
+                // github url can be in the form of:
+                //  - https://github.com/org/repo
+                this.errors.url = /^https:\/\/github\.com\/[^/]+\/[^/]+$/.test(url)
                     ? ''
-                    : 'Please enter a valid GitHub or Azure DevOps repository URL'
+                    : 'Please enter a valid GitHub repository URL'
+            } else if (type === 'azure') {
+                // azure devops url can be in the form of:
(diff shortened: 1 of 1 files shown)


cstns (approved):

knolleary:
@cstns I've just pushed an update.

Azure provides a clone-url button that provides the url with a username in the url:

```
https://ORGNAME@dev.azure.com/org/project/_git/project
```

Previously we'd reject that - which is inconvenient to the user. Updated the form validation to accept that pattern, but strips off the ORGNAME before saving so the backend doesn't need to worry about it.

cstns (approved):

PR #8456, opened 2026-09-08
Allow admin to modify type of team in manual billing mode

### LINKED ISSUE #8454: Allow admin to modify team type of team already in manual billing mode
When a billing team is placed into manual billing mode by an admin, we allow the team type to be updated at the same time.

If the team is already in manual billing mode, an admin cannot currently modify the team type.

Purpose of this task is to enable that specific scenario.

 - Update Change Team Type form to not block changing type if user is admin and team in manual billing mode
 - Verify backend applies the update
 - Verify any limit checks in the UI and whether they need to be applied in this scenario

knolleary:
Fixes #8454 

Previously, if a team was in manual billing mode, we block changing its type in the front end.

This PR updates the rules so that an admin user *is* able to modify the team type for a team in manual billing mode.

It adds a clear message so the admin realises what they are doing:

image

The existing limit checks still apply; if the team's resource usage exceeds any limits of the target type, the change will be blocked. We *may* need to find a way around that, but it'll be a separate piece of work if needed.

### DIFF
--- a/frontend/src/pages/team/changeType.vue
+++ b/frontend/src/pages/team/changeType.vue
@@ -26,7 +26,7 @@
                             :price="billingEnabled ? (!isAnnualBilling ? teamType.billingPrice : teamType.annualBillingPrice) : ''"
                             :price-interval="billingEnabled ? (!isAnnualBilling ? teamType.billingInterval : teamType.annualBillingInterval) : ''"
                             :value="teamType.id"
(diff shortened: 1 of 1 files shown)


knolleary:
Sticking back in draft - found an edge case I want to squash

knolleary:
Edge case was a false alarm. This is good to go.

n-lark (approved):

PR #8458, opened 2026-09-09
Keep a browser session's MCP pins alive across a reload

### LINKED ISSUE #8459: MCP browser session pin is dropped when the tab reconnects
An MCP client pins a browser tab as its active session (`platform_set_active_browser_session`). If that tab's connection then briefly drops and reconnects (most simply, a page reload), the platform treated the disconnect, the MQTT last will, as the tab going away and purged the session and every pin on it. The client silently lost its pin mid-session, and subsequent `flow_building` / `platform_ui` calls could no longer route to the tab.

A reload is indistinguishable from a real disconnect at that point: the same session reconnects a moment later and resumes heartbeating. The pin should survive that brief gap.

Expected: a pin stays alive across a reconnect; it is only dropped when the tab is genuinely gone (no reconnecting heartbeat).

andypalmi:
Closes #8459

When a pinned tab reloads, its MQTT connection closes ungracefully and the broker would fire the connection's last will, which purges the browser session and every MCP pin on it, so the tab lost its pin even though the same session reconnects a moment later.

The presence connection now sets an MQTT v5 will delay interval (20s), with a matching session expiry and a non-clean session. A reload reconnects with the same clientId and resumes the session, so the broker cancels the pending will and the tab keeps its pins; a genuine close still fires the will once the delay passes. No application-side state is needed.

Only the fe-team connection carries a will (from `createClientForTeamFrontend`), so the session-expiry and will-delay options apply to it alone; every other broker connection is unchanged.

### DIFF
--- a/forge/comms/browserSessionLifecycle.js
+++ b/forge/comms/browserSessionLifecycle.js
@@ -4,10 +4,9 @@
  * Every event a browser tab emits arrives on one topic shape:
  *   ff/v1/<teamHash>/u/<userHash>/s/<sessionId>/<event>
  *
- * `disconnected` is the connection's last will. The broker holds it for a short delay (the
- * will delay interval) and publishes it only if the tab has not reconnected, so a reload does
- * not fire it. New consumers of that signal belong in handleDisconnected - they should not
- * need a topic of their own.
+ * `disconnected` is the connection's last will. The broker publishes it when a
+ * tab goes away without a clean disconnect. New consumers of that signal belong in
+ * handleDisconnected - they should not need a topic of their own.
  *
  * Traffic also goes the other way: notifyMcp publishes to one tab on
  * ff/v1/<teamHash>/u/<userHash>/s/<sessionId>/mcp/<event>. The ACL pins the session
--- a/forge/comms/browserSessionLifecycle.js
+++ b/forge/comms/browserSessionLifecycle.js
@@ -4,9 +4,10 @@
  * Every event a browser tab emits arrives on one topic shape:
  *   ff/v1/<teamHash>/u/<userHash>/s/<sessionId>/<event>
  *
- * `disconnected` is the connection's last will. The broker publishes it when a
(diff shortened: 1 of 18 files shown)


forge/comms/browserSessionLifecycle.js
```diff
@@ -4,9 +4,9 @@
  * Every event a browser tab emits arrives on one topic shape:
  *   ff/v1/<teamHash>/u/<userHash>/s/<sessionId>/<event>
  *
- * `disconnected` is the connection's last will. The broker publishes it when a
- * tab goes away without a clean disconnect. New consumers of that signal belong in
- * handleDisconnected - they should not need a topic of their own.
+ * `disconnected` is the connection's last will. A reload fires it too, so it is flagged for a
```
cstns: no need to update the comment; mentioning the grace period should be more than enough

forge/comms/browserSessionLifecycle.js
```diff
@@ -45,11 +45,11 @@ class BrowserSessionLifecycleHandler {
     }
 
     /**
-     * The tab's connection is gone. Fans out to every subsystem that keeps
-     * per-session state.
+     * The tab's connection is gone, but a reload looks the same from here. Flag it for the
+     * grace window instead of dropping it, so a reconnecting tab keeps its pins.
```
cstns: same here

cstns (changes requested):
Everything needed is already in place:

The UI client connects with protocolVersion: 5 (frontend/src/services/mqtt.service.ts:547), and MQTT 5 wills accept a willDelayInterval property. 
So the fix belongs in the connect options in mqtt.service.ts, roughly:

```js
  protocolVersion: 5,
  clean: false,
  properties: { sessionExpiryInterval: 30 },
  will: {
      topic: credentials.will.topic,
      payload: credentials.will.payload,
      qos: 1,
      retain: false,
      properties: { willDelayInterval: 30 }
  }
```

cstns (approved):

PR #8463, opened 2026-09-10
Take a newly registered user into onboarding, and let them out

### LINKED ISSUE #8382: Send new signups into onboarding, and give them a way out

cstns:
Gets a newly registered user into the onboarding conversation, and lets them out of it.

Registration sets a one-shot flag, persisted because email verification reloads the app before anything can act on it. `checkState` consumes it on the next arrival and routes to the onboarding page.

Deliberately one-shot rather than "redirect whenever the user is mid-onboarding". The latter resumes people who come back cold, but it also pins them to the page: anyone who clicks through to their account settings gets pulled straight back. The flag is consumed whether or not it is acted on, so it never fires twice, and the account-age check does not raise it, so an account under a week old is not sent to onboarding on every boot.

An explicit redirect target wins over onboarding. Invitations are auto-accepted during signup, so an invited user also looks freshly registered, and they are headed somewhere specific on a team that is not theirs to set up.

The escape control now recedes once the user engages, where engaging means contributing a turn rather than a timer elapsing. The count of their turns when the page opens is the baseline, so a seeded or resumed transcript does not count. It drops to 45% opacity and returns to full on hover or keyboard focus, so it stays reachable rather than becoming decorative.

Closes #8382


frontend/src/pages/team/Onboarding.vue
```diff
@@ -74,6 +84,19 @@ export default {
...
             }
         },
+        userTurns: {
+            immediate: true,
+            handler (turns) {
+                // Engagement is the user answering or saying something, not a
+                // timer, so the control only recedes once they have committed
+                // to the conversation
+                if (this.initialUserTurns === null) {
+                    this.initialUserTurns = turns
```
andypalmi, line 94: Small timing question. The baseline here is captured by an immediate watcher, but the fixture is seeded from the `team` watcher, which no-ops on its first run when the team context is not resolved yet (a direct load or refresh of this URL rather than the in-app redirect from Home). In that case `initialUserTurns` is captured as 0 before the seed, then the seeded human turns push `userTurns` above it and `hasEngaged` flips on with no user action. The effect is only cosmetic since the control stays reachable, but it defeats the intent. Deferring the capture until the team, and so the seed, has resolved avoids it, since `hydrateMessages` runs synchronously:

```suggestion
                if (this.initialUserTurns === null) {
                    // Wait until the team, and so the seeded transcript, has
                    // resolved. A direct page load would otherwise capture the
                    // baseline before the seed and mark the user engaged.
                    if (!this.team) {
                        return
                    }
                    this.initialUserTurns = turns
```

frontend/src/pages/Home.vue
```diff
@@ -88,13 +89,20 @@ export default {
...
-                if (this.team || this.defaultUserTeam) {
+                const teamSlug = this.team?.slug || this.defaultUserTeam?.slug
+                if (teamSlug) {
+                    // A newly registered user is taken to AI-led onboarding once
+                    if (this.featuresCheck?.isAiOnboardingFeatureEnabled && useUxStore().consumeOnboardingEntry()) {
+                        return this.$router.push({
+                            name: 'team-onboarding',
+                            params: { team_slug: teamSlug }
+                        })
+                    }
```
andypalmi, line 102: Because of the `&&` short-circuit, `consumeOnboardingEntry()` only runs when the feature is enabled, so on a platform where AI onboarding is off a new user keeps `shouldEnterOnboarding: true` in storage indefinitely. If the feature is later turned on, that stale flag fires on their next visit here, and if their stage has moved past intake the onboarding page bounces them to page-not-found. Consuming the one-shot unconditionally and gating only the redirect on the feature keeps it from lingering, and avoids dragging a no-longer-new user in when the feature flips on:

```suggestion
                    // A newly registered user is taken to AI-led onboarding
                    // once. Consume the one-shot regardless so it cannot linger
                    // and fire on a later visit if the feature is toggled on.
                    const enterOnboarding = useUxStore().consumeOnboardingEntry()
                    if (enterOnboarding && this.featuresCheck?.isAiOnboardingFeatureEnabled) {
                        return this.$router.push({
                            name: 'team-onboarding',
                            params: { team_slug: teamSlug }
                        })
                    }
```

frontend/src/stores/ux.js
```diff
@@ -48,15 +58,23 @@ export const useUxStore = defineStore('ux', {
...
+        endOnboarding () {
+            this.onboardingStage = ONBOARDING_STAGES.DONE
+            this.shouldEnterOnboarding = false
+        },
         openOverlay () { this.overlay = true },
         closeOverlay () { this.overlay = false }
     },
     persist: {
-        pick: ['isNewlyCreatedUser', 'onboardingStage', 'userActions'],
+        pick: ['isNewlyCreatedUser', 'onboardingStage', 'shouldEnterOnboarding', 'userActions'],
```
andypalmi, line 77: Following the persistence thread on the onboarding-page PR (https://github.com/FlowFuse/flowfuse/pull/8440#discussion_r3989330188), where the call was to keep localStorage for now and raise a UserProperties table in the engineering meeting. Just noting that this PR sharpens the symptom worth capturing in that issue: `shouldEnterOnboarding` is localStorage only, and `checkIfIsNewlyCreatedUser` re-raises it whenever the stage resolves from null, which is any device with empty storage inside the seven day window. So a user who finishes or skips onboarding on one browser and opens another within their first week is now actively redirected back into onboarding, where before it was only a reachable page. No change needed here, flagging for that future backend issue.

andypalmi (commented):
Building on the stage model from the PR below this in the stack. The one-shot entry flag reads cleanly and the store tests are thorough. Three inline notes: a small suggestion on the engagement baseline, one on consuming the entry flag unconditionally, and a persistence note carried over from the onboarding-page PR.

frontend/src/pages/team/Onboarding.vue
cstns: Good catch, the race is real on this branch: on a direct load the baseline gets captured as 0 before the team resolves, then the seeded human turns flip `hasEngaged` with no user action.

I'm not taking the guard as suggested though, because it trips over itself one PR up the stack. #8467 replaces the seed with `openConversation()`, which sends an empty query and adds no human turns. With nothing changing `userTurns` when the team resolves, the watcher never refires, so the baseline would only get captured on the user's first message. That message gets swallowed as the baseline and the control only recedes on their second one.

Instead I've moved the baseline capture into the `team` watcher, right after the seed call. The seed hydrates synchronously, so by the time the capture runs the seeded turns are already counted in, and on the follow-up branch it just captures 0 once the team is there. The `userTurns` watcher is left with only the engagement check.

andypalmi (approved):
Approving. The one-shot in Home.vue is now consumed unconditionally so it cannot linger and fire after a later feature toggle, and moving the engagement baseline capture into the team watcher resolves the direct-load race without the guard that would have tripped up the follow-up in #8467. Reads cleanly.

PR #8464, opened 2026-09-10
Stop the welcome tour firing on the AI onboarding path

### LINKED ISSUE #8383: Stop the welcome tour and education modal firing on the onboarding path
Two things currently fire right after email verification, and both land before the user has anything to look at.

The **Shepherd welcome tour** gets flagged from `UnverifiedEmail.vue` on successful verification, then picked up in `pages/team/index.vue` once resources have loaded. It walks abstract concepts: what a hosted instance is, what a remote instance is, where the audit log lives.

The **education modal** fires as the tour's completion callback from the same place. It's "Where would you like to get started?" with a list of external doc links that open in new tabs.

Neither makes sense on the onboarding path. The Expert is having a conversation with the user, and a tour talking over it about concepts they haven't met yet is actively in the way.

Both stay exactly as they are for classic signups.

### Done when

* Neither the tour nor the education modal fires when the onboarding flag is on.
* Both behave exactly as they do today when it's off.
* Both remain reachable from the user menu, which is where `PageHeader.vue` already offers them.

### Worth knowing

The tour flag persists to localStorage via the `ux-tours` store, so suppressing it needs care about whether the flag is set and ignored, or never set in the first place. Never setting it is cleaner, since a flag that's set and ignored will fire later when the user hits a page that checks it.

Not replacing them with anything. Explaining the platform is out of scope for this epic, and the Expert covering platform concepts in context is a separate idea for another time. What the Expert does explain is how to use the thing it just built, which is a different task.

Worth checking whether anything else keys off tour completion. `pages/team/index.vue` runs the education modal from the tour's close hook, and there's also a branch that wakes the Expert instead, so the interaction between tour, modal and assistant wake-up on arrival needs untangling rather than just switching one off.

cstns:
The welcome tour explains hosted instances, remote instances and the audit log to someone who has just arrived and has none of them. On the AI onboarding path the Expert covers the same ground by actually building something, so the tour ends up talking over it.

The education modal goes with it. It is not triggered independently anywhere: it opens as the tour's completion callback in `team/index.vue`, so suppressing the tour suppresses both.

The flag is never raised rather than raised and ignored. Onboarding lives outside the team route tree, so nothing checks the flag while the user is in the conversation, and leaving it set would fire the tour at them later when they finally reached the team page.

Both stay reachable from the user menu, and both behave exactly as before when AI onboarding is off. `UnverifiedEmail.vue` had no spec, so this adds one covering the classic path, the onboarding path, and the failed-verification case.

One thing worth settling in review: this gates on the feature flag rather than on whether this particular user is heading into onboarding. So on a deployment with onboarding enabled, an invited user joining an existing team also loses the tour. That is the same gap already open on #8382 around invited users, and it probably wants one answer covering both rather than two separate patches.

Closes #8383


frontend/src/pages/UnverifiedEmail.vue
```diff
@@ -60,7 +62,9 @@ export default {
             try {
                 await userApi.verifyEmailToken(this.token)
                 clearTimeout(this.resendTimeout)
-                this.presentTour()
+                if (!this.featuresCheck?.isAiOnboardingFeatureEnabled) {
+                    this.presentTour()
+                }
```
andypalmi: `shouldPresentTour` is persisted in localStorage, so if it was set true on an earlier visit (standard flow, or before this feature was on), skipping the write here still leaves it queued and `team/index.vue` would dispatch the tour on the next team load. Worth actively withdrawing it instead? `withdrawTour()` already exists on the store with no callers:

```suggestion
                if (this.featuresCheck?.isAiOnboardingFeatureEnabled) {
                    // clear any tour queued under the standard flow so it cannot fire later
                    this.withdrawTour()
                } else {
                    this.presentTour()
                }
```

Needs `withdrawTour` adding to the `mapActions(useUxToursStore, ...)` list above. Happy to leave it if you'd rather stay minimal, since a brand-new user has no stale flag anyway.

andypalmi (commented):
Small, focused change and the new spec covers both paths nicely. One suggestion inline on the tour-suppression path.

frontend/src/pages/UnverifiedEmail.vue
andypalmi:
```suggestion
                if (this.featuresCheck?.isAiOnboardingFeatureEnabled) {
                    // clear any tour queued under the standard flow so it cannot fire later
                    this.withdrawTour()
                } else {
                    this.presentTour()
                }
```

andypalmi (approved):

PR #8466, opened 2026-09-10
Add housekeeping leadership subscription

hardillb:
## Description

It appears the leadership subscription was missed when implemented.


hardillb:
Noticed while reviewing #8450 that we might not have a clean shutdown publish to remove the leader vote. Will fix before removing draft status

hardillb:
Nevermind, found it

https://github.com/FlowFuse/flowfuse/blob/37fcff25572be6ef6a97801833b2a8dc08c89ac9/forge/comms/index.js#L124

knolleary (approved):

PR #8467, opened 2026-09-10
Have the Expert open the onboarding conversation

### LINKED ISSUE #8369: Open the onboarding conversation without waiting for the user

### LINKED ISSUE #8370: Send the onboarding flag to the Expert with each turn
The Expert needs to know when it's running an onboarding conversation rather than a normal support one, because the behaviour differs quite a bit: no platform vocabulary, picks defaults silently instead of asking, provisions before it plans.

The context object in `frontend/src/stores/context.js` already ships about twenty five fields to the agent on every turn, including `planMode`, `questionCadence`, `isTrialAccount` and the various `supports*` capability flags. This is one more field alongside those.

### Done when

* The agent receives the flag on every turn of an onboarding conversation.
* It's absent or false everywhere else.

### Worth knowing

Both branches of `context.js` need it. There's an early-return path used before a team has loaded and a main path used after, and they build the object separately, so a field added to one and not the other goes missing depending on timing.

Worth deciding whether this is driven by the platform flag directly or by the onboarding surface being open. They're different: the platform flag says the deployment does AI onboarding, the surface being open says this particular conversation is one. A user who finishes onboarding and later opens the Expert drawer normally shouldn't still be getting onboarding behaviour, so it probably wants to track the conversation rather than the deployment.

### LINKED ISSUE #8371: Let the Expert open a chat session before the user types
`Verify input` in the expert flows rejects any request that doesn't carry a `payload.query` or an abort signal:

> `data must contain a 'query' or abort signal`

Which means there's no way to open a chat session without the user having typed something first. Today the session gets created lazily on the first message the user sends.

For onboarding the Expert has to speak first, so it needs to accept a turn that has no query and respond with its opening greeting and first question set.

### Done when

* An onboarding session can be opened with no user query.
* The Expert responds with its opening turn.
* A non-onboarding request with no query is still rejected exactly as it is now.

### Worth knowing

The work is in the flows on `expert-api-dev`, not in this repo. Filed here to keep the epic together.

Worth being careful about the shape of the relaxation. The check currently guards against malformed requests generally, so opening it up unconditionally would let genuine bad input through and produce a confusing failure further along. Narrowing it to "no query is fine when the onboarding flag is set" keeps the existing protection for everything else.

There's a decision about what the opening turn actually is. It could be a plain greeting that then waits, or a greeting plus the first question set in one turn. The second is fewer round trips and gets the user doing something immediately, which is probably what we want, but it means the first question set is fixed rather than derived from anything. The specific questions are getting their own issue.

Also worth checking what happens on the frontend when a session exists with no user message in it. `product-expert-support-agent.js` tracks `sessionStartTime` and there's a session expiry warning built on it, so a session that opens and then sits there while the user reads could start warning about expiry before they've said a word.

cstns:
The Expert speaks first on the onboarding page, so the frontend needs a way to start a turn the user has not typed.

## Heads up for whoever reviews this

**This is the first PR in the stack whose behaviour cannot be verified from the frontend side.** It sends the opening turn and renders whatever comes back. Whether the Expert actually produces a sensible first turn for an empty query is the other half, and that lives in the expert flows (#8371), not here.

If the agent is not ready for it, nothing errors. `Verify input` accepts an empty string, the fast-reply switch does not match it, and the turn goes to the LLM as an empty user message. So the page opens, the request goes out, a reply comes back and renders. **It reads as working right up until you look at what the Expert actually said.** A blank or confused first message is the failure mode, not a stack trace.

So testing this means reading the opening turn, not just checking that one appears.

## What it does

`openConversation` sends a turn with an empty query. That is not a new protocol: `resumeToolApprovals` already does exactly this, described in its own comment as "an ordinary chat request with no query". Worth noting the expert-side `Verify input` check turns out to be looser than #8371 assumed, it requires `query` to be a string rather than to be non-empty, so an empty string already passes it today.

Two things it deliberately does not do:

* **No user message is added.** The user has not said anything, so nothing of theirs belongs in the transcript.
* **The session clock is left unstarted.** This is the concern raised in #8371. The expiry warning fires at 25 minutes and the session is declared expired at 28, tracking the agent's own 30 minute history GC. If that clock started when the conversation opened, someone reading the first question could be told their history had expired without ever having typed. `handleQuery` already starts the clock on the first turn where it is unset, so the user's first reply picks it up naturally.

The `onboarding` flag is added to both branches of the context getter. They build their objects separately, so a field added to one goes missing depending on load timing. It is sourced from `ux.isOnboarding`, which tracks the conversation rather than the deployment: it stays true after the Expert moves the user into the editor, and goes false once onboarding is finished or skipped, so a later ordinary chat is not treated as one. That settles the open question on #8370.

Also drops `onboardingFixture.js`, the hardcoded placeholder transcript that stood in while this was missing, along with its seeding call and the stale TEMPORARY comment. The drawer's canned welcome was already suppressed on this surface, so nothing else was hardcoded. The old guard is kept: a transcript holding only canned messages counts as empty and gets cleared, so arriving from the drawer does not leave its greeting in the way, while a real conversation is left alone, which is what makes the page resumable.

Closes #8369
Closes #8370
Closes #8371


frontend/src/stores/product-expert.js
```diff
@@ -219,6 +219,35 @@ export const useProductExpertStore = defineStore('product-expert', {
...
+                }
+                return result
+            } catch (error) {
+                if (error.name === 'AbortError' || error.name === 'CanceledError') {
+                    return undefined
+                }
+                if (!this.shouldUseMqtt) {
+                    console.error('Expert API error:', error)
+                }
+                this.addPredefinedAiMessage('Sorry, I could not get started. Please refresh to try again.', { isError: true })
```
andypalmi, line 246: Over MQTT `_onMqttError` already adds an error bubble on a failed publish, so this line adds a second one on top of it. `handleQuery` keeps its message inside the `!shouldUseMqtt` guard for that reason. Moving it in matches that and avoids the double bubble:

```suggestion
                if (!this.shouldUseMqtt) {
                    console.error('Expert API error:', error)
                    this.addPredefinedAiMessage('Sorry, I could not get started. Please refresh to try again.', { isError: true })
                }
```

andypalmi (commented):
One inline note on the opening-turn error path.

andypalmi:
Minor: the `skip-onboarding` `ff-button`'s label sits inline with its tags, tripping the `singleline-html-element-content-newline` warning. Putting it on its own line clears it (whitespace-only, non-blocking):
```html
                >
                    Set it up myself
                </ff-button>
```

frontend/src/stores/product-expert.js
andypalmi:
```suggestion
                if (!this.shouldUseMqtt) {
                    console.error('Expert API error:', error)
                    this.addPredefinedAiMessage('Sorry, I could not get started. Please refresh to try again.', { isError: true })
                }
```

andypalmi (approved):

PR #8468, opened 2026-09-10
feat(expert): pinned task list in the Expert chat

andypalmi:
Closes FlowFuse/engineering#344

## Summary

Adds a pinned task list to the Expert chat that shows the current set of tasks. Tasks are not always bound to a plan: when they belong to a plan the header shows the plan name, otherwise it simply reads `Tasks`. The list sits just above the composer so progress stays visible while you scroll the conversation, and it survives a chat refresh.

The agent side sends a task surface over MQTT whenever the tasks or their states change. The panel renders each task with its state, a done/total counter, and a collapsible header. The whole feature is gated behind a `supportsPlansAndTasks` capability flag so an older instance simply never receives the surface.

## Changes

- New `TaskList.vue`: renders the current tasks with native disc bullets, a per-task state label, and a done/total count in the header.
- New `CollapsibleSection.vue`: a small reusable collapsible used by the task list header (and reused by the plan card in the stacked PR).
- `product-expert.js`: new `activeTaskList` state, an `expert:tasks` inflight handler that stores the surface and acks it, clears the list on a new chat, and persists `activeTaskList` in `sessionStorage` so it comes back after a refresh.
- `context.js`: advertises the new `supportsPlansAndTasks` capability flag alongside the existing ones.
- `Expert.vue`: mounts the task list between the messages and the composer.

## Task states

Every task carries a state and is labelled in place, so the list reads clearly without a legend:

- To Do: plain text, `(To Do)`
- In Progress: italic with an accent-coloured `(In Progress)`
- Done: struck through and subtle, `(Done)`

The header shows the done/total count and auto-opens while any task is still pending or in progress.

**All three states together (light):**

Task list with To Do, In Progress, and Done states in light mode

**Same, in dark mode:**

Task list with all three states in dark mode

**Every task done:** the counter reads 5/5 and each row is struck through.

Task list with every task done, counter 5/5

**Collapsed:** the header collapses to a single line and keeps the counter (here 4/4).

Collapsed task list header showing 4/4

## Persistence

The active task list is persisted to `sessionStorage`, so refreshing the chat brings the list back in the same way the messages already return. Starting a new chat clears it.

## Backward compatibility

The agent keeps planning and tracking task state internally either way. The only thing the `supportsPlansAndTasks` flag gates is the emission: the task surface is published over MQTT only when the client advertises the flag. On an older frontend the surface is never sent, so the agent behaves exactly as before and there is simply no panel to receive it.

## Testing

- Started a plan-driven conversation and confirmed the task list appears above the composer with the correct states and counter.
- Toggled the collapsible header open and closed.
- Refreshed the chat and confirmed the list is restored.
- Started a new chat and confirmed the list is cleared.
- Verified light and dark rendering.

### DIFF
--- a/frontend/src/components/expert/components/messages/components/CollapsibleSection.vue
+++ b/frontend/src/components/expert/components/messages/components/CollapsibleSection.vue
@@ -1,0 +1,137 @@
+<template>
+    <div class="ff-collapsible" :class="{ 'is-up': openUp }">
+        <div class="ff-collapsible--header" @click="toggle">
+            <ChevronRightIcon class="ff-collapsible--chevron" :class="{ rotated: expanded }" />
+            <slot name="header" />
+        </div>
+        <transition :css="false" @enter="onEnter" @after-enter="onAfterEnter" @leave="onLeave">
+            <div v-if="expanded" ref="content" class="ff-collapsible--content">
+                <slot />
+            </div>
+        </transition>
+    </div>
+</template>
+
+<script>
+import { ChevronRightIcon } from '@heroicons/vue/20/solid'
+
+const DURATION = 220
+
+export default {
+    name: 'CollapsibleSection',
+    components: { ChevronRightIcon },
+    props: {
+        autoOpen: {
+            type: Boolean,
+            default: false
+        },
+        forceOpen: {
+            type: Boolean,
+            default: false
+        },
+        openUp: {
+            type: Boolean,
+            default: false
+        }
+    },
+    data () {
+        return {
+            manualExpanded: null
+        }
+    },
+    computed: {
+        expanded () {
+            if (this.forceOpen) return true
+            return this.manualExpanded !== null ? this.manualExpanded : this.autoOpen
+        }
+    },
+    methods: {
+        toggle () {
+            if (this.forceOpen) return
+            this.manualExpanded = !this.expanded
+        },
+        reducedMotion () {
+            return window.matchMedia?.('(prefers-reduced-motion: reduce)').matches
+        },
+        onEnter (el, done) {
+            if (this.reducedMotion()) return done()
+            const target = el.scrollHeight
+            el.style.overflow = 'hidden'
+            el.style.height = '0'
+            el.style.opacity = '0'
+            requestAnimationFrame(() => {
+                el.style.transition = `height ${DURATION}ms ease, opacity ${DURATION}ms ease`
+                el.style.height = `${target}px`
+                el.style.opacity = '1'
+            })
+            this.onEnd(el, 'height', done)
+        },
+        onAfterEnter (el) {
+            el.style.height = ''
+            el.style.opacity = ''
+            el.style.overflow = ''
+            el.style.transition = ''
+        },
+        onLeave (el, done) {
+            if (this.reducedMotion()) return done()
+            el.style.overflow = 'hidden'
+            el.style.height = `${el.scrollHeight}px`
+            el.style.opacity = '1'
+            // Force a reflow so the browser registers the start height before collapsing.
+            this.forceReflow(el)
+            el.style.transition = `height ${DURATION}ms ease, opacity ${DURATION}ms ease`
+            el.style.height = '0'
+            el.style.opacity = '0'
+            this.onEnd(el, 'height', done)
+        },
+        forceReflow (el) {
+            // Reading a layout property flushes pending style writes so the next one animates.
+            return el.offsetHeight
+        },
+        onEnd (el, prop, done) {
+            const handler = (event) => {
+                if (event.target !== el || event.propertyName !== prop) return
+                el.removeEventListener('transitionend', handler)
+                done()
+            }
+            el.addEventListener('transitionend', handler)
+        }
+    }
+}
+</script>
+
+<style scoped lang="scss">
+.ff-collapsible {
+    display: flex;
+    flex-direction: column;
+
+    &.is-up {
+        flex-direction: column-reverse;
+    }
+}
+
+.ff-collapsible--header {
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
+    padding: 0.25rem 0;
+    cursor: pointer;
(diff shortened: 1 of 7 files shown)


cstns:
The feature works and the overall shape is good, but there are two things I'd want addressed before merging. Both are cheap to fix relative to their impact.

**1. The first auto-open of the task list locks the panel at a fixed height and clips everything added after.**

In `CollapsibleSection.vue`, `onEnter` reads `scrollHeight`, sets the start height to `0`, and then sets the transition and target height inside a single `requestAnimationFrame`. rAF callbacks run before the frame's style recalc, so the browser never snapshots the `height: 0` start state. Its first recalc sees `auto -> Npx`, which isn't interpolable, so no transition is generated, `transitionend` never fires, and `onAfterEnter` never runs to clean up. The content is left with a permanent inline `height: Npx; overflow: hidden`.

This hits the primary path: the panel auto-opens on the first `expert:tasks` message, locks at the height of that first list, and every task the agent adds afterwards is clipped and unreachable. Your own `onLeave` already has the `forceReflow(el)` call with a comment explaining exactly why it's needed, it just didn't make it into `onEnter`.

While you're in there, it might be worth adding a timeout fallback next to the `transitionend` listener. With `:css="false"`, Vue only removes the leaving element when `done()` is called, so any collapse that produces no transition (hidden tab, `display: none` ancestor, zero-height content) currently orphans the old node in the DOM forever, and the next expand renders alongside it.

**2. `activeTaskList` has one exit door and lives on the wrong store.**

The list is only ever cleared by Start Over. That leaves stale, frozen task lists pinned above the chat in several everyday flows:

* Hitting Stop mid-plan: `stopInflightChat` clears `_inFlightRequests` but not the task list, so the panel keeps showing "(In Progress)" indefinitely. Same via `handleMqttError` on a broker drop.
* Refresh: `activeTaskList` is in the sessionStorage persist pick, so a reload restores a live-looking task list above a chat whose MQTT session is gone and will never update it.
* Mode switching: the field sits on the shared store while `messages`/`sessionId` are per-agent, and `setAgentMode` doesn't touch it. Flip from Support to Insights and the Support tasks stay pinned over the Insights transcript. Inversely, Start Over while in Insights wipes the Support agent's still-valid list.

All three have the same root cause, so one structural change might cover it: move `activeTaskList` into the per-agent stores the same way `messages` and `sessionId` work, exposed through a getter, and clear it in the stop/error paths. That fixes mode switching, reset, and the persistence scoping in one place rather than patching each path individually.

A few smaller things noticed along the way, none blocking: a final `expert:tasks` racing the chat reply on its own topic gets dropped by the in-flight guards (the reply branch deletes the inflight entry before the trailing update arrives, so the last statuses stay stale and the agent's ack times out), the new ack publish isn't wrapped in try/catch like the neighbouring automation cases so a publish failure is an unhandled rejection, and the completed list auto-collapses the instant the last task turns done, which is exactly when I'd want to read it. Also worth a thought: `supportsPlansAndTasks` is advertised on every transport including HTTP and Insights turns, where the `expert:tasks` channel that delivers the list doesn't exist. If the agent changes behaviour based on that flag it might be worth gating it on the MQTT transport.

cstns (changes requested):
see comment above

andypalmi:
Thanks, great review. Both blocking issues fixed, plus two of the smaller ones.

- **Collapsible never settling:** the reflow now sits between the start and target heights (mirroring the collapse path), with a timeout fallback beside `transitionend` so it always settles.
- **activeTaskList scope:** moved to the per-agent stores, exposed via a getter on the umbrella store, and cleared in `stopInflightChat` (covers Stop and broker-drop). Start Over already cleared it.
- **Ack try/catch:** now matches the neighbouring automation cases.
- **List auto-collapsing on the last task:** stays open once everything's done; a manual collapse still sticks.
- **Trailing `expert:tasks` race:** kept separate in #8484 (based on this branch, retargets to main on merge).
- **`supportsPlansAndTasks` on every transport:** left as is, to match `supportsPlanMode`/`supportsHITL`/`supportsQuestions` next to it. Transport safety already lives on the agent side, which won't emit a surface on HTTP. Happy to revisit if you'd rather gate all of them together.

frontend/src/components/expert/components/messages/components/CollapsibleSection.vue
```diff
@@ -0,0 +1,144 @@
...
+            el.style.overflow = 'hidden'
+            el.style.height = `${el.scrollHeight}px`
+            el.style.opacity = '1'
+            // Force a reflow so the browser registers the start height before collapsing.
+            this.forceReflow(el)
+            el.style.transition = `height ${DURATION}ms ease, opacity ${DURATION}ms ease`
+            el.style.height = '0'
+            el.style.opacity = '0'
+            this.onEnd(el, 'height', done)
+        },
```
cstns: The fixes work, but the amount of scaffolding needed here (reflow, transitionend listener, timeout fallback, reduced motion branch) suggests the approach is fighting the platform. Animating height from script is the old workaround, and both bugs so far came from it.

CSS handles this natively now with the grid row trick:

```css
.wrap { display: grid; grid-template-rows: 0fr; transition: grid-template-rows 220ms ease; }
.wrap.expanded { grid-template-rows: 1fr; }
.wrap > .inner { min-height: 0; overflow: hidden; }
```

Toggle a class, drop the `<transition>` and all the JS hooks. Bonus: content growing mid-animation just works, which matters since tasks stream in. Only catch is the content stays mounted when collapsed, so it needs `inert` on the collapsed state.

Worth doing now while #8484 is still stacked on this branch.

cstns (changes requested):

andypalmi:
@cstns good call, done. Dropped the `<transition>` and all the JS hooks for the `grid-template-rows: 0fr/1fr` trick, with `inert` on the collapsed content. 58 lines of scaffolding gone, and streaming tasks growing mid-animation now just work. Propagated up to #8469 too.

cstns (approved):

PR #8469, opened 2026-09-10
feat(expert): structured plan card with active-plan pointer

andypalmi:
> Stacked on top of the task list PR #8468. Review and merge that one first.

Closes FlowFuse/engineering#345

## Summary

Turns the plan reply into a structured card with a name, a description, and a collapsible body, and introduces an active-plan pointer so the chat always shows which plan is currently in effect.

When a plan is waiting for approval the card is expanded with its action buttons. Once approved it collapses to a single line, and the plan that owns the currently pinned task list is marked with an `Active` pill. Approving another plan, or reactivating an earlier one, moves the pill and swaps the pinned task list to match.

## Changes

- `PlanCard.vue`: structured layout with a `Plan: <name>` heading, a description line, and the plan body inside a collapsible section; an `Active` pill in the collapsed header; awaiting-approval vs approved states; heading sized like a top-level heading and the pill given a solid accent fill for legibility.
- `AnswerWrapper.vue`: passes `name`, `description`, `active`, and `awaiting-approval` into the card, deriving `isActivePlan` from the plan id and the active-plan pointer.
- `product-expert.js`: adds the `activePlanId` getter, derived from the pinned task list's `planId`, so the active plan and its task list stay in sync.

## Anatomy of the card

While a plan awaits approval the card is fully expanded. The heading is the plan name, followed by the plan description, then the plan content, with the action buttons at the bottom. The card also falls back gracefully: a plan without structured fields renders as before, so this is backward compatible.

Plan card awaiting approval, annotated with plan name, description, and content

## Collapsed states

Once approved the card collapses to a single line. The active plan carries the `Active` pill; other plans do not.

**Active:**

Collapsed plan card with the Active pill

**Not active:**

Collapsed plan card without the Active pill

## Active-plan pointer

Only one plan is active at a time. Approving a new plan makes the previous one no longer active, so the `Active` pill moves to the newest approved plan.

Two plans in the chat, the Active pill on the newest approved plan

## Reactivating an earlier plan

You can ask the Expert to reactivate a previous plan. Doing so moves the `Active` pill back to that plan and swaps the pinned task list to that plan's tasks.

**Reactivating the first plan:** the pinned task list at the bottom now shows the first plan's tasks.

Reactivating the first plan, pinned task list shows the first plan's tasks

**Reactivating the second plan again:** the pinned task list swaps to the second plan's tasks, matching the currently active plan.

Reactivating the second plan, pinned task list shows the second plan's tasks

## Backward compatibility

The whole plan and task experience is gated behind the `supportsPlansAndTasks` capability flag, which is introduced in the stacked task list PR #8468. When the flag is absent the agent runs in its previous mode and does not emit the plan/task surface.

A plan reply without the structured name/description fields also renders exactly as it did before, so older agent responses are unaffected.

## Testing

- Approved a plan and confirmed the card collapses and gains the `Active` pill.
- Approved a second plan and confirmed the pill moves off the first.
- Reactivated the first plan and confirmed the pill and the pinned task list both switch back to it.
- Reactivated the second plan and confirmed both switch to it.
- Confirmed an unstructured plan reply still renders correctly.
- Verified light rendering; card uses shared theme tokens for both themes.

### DIFF
--- a/frontend/src/components/expert/components/messages/components/AnswerWrapper.vue
+++ b/frontend/src/components/expert/components/messages/components/AnswerWrapper.vue
@@ -83,6 +83,11 @@
             :plan="answer.content"
             :message-uuid="messageUuid"
             :answer-uuid="answer._uuid"
+            :name="answer.name || ''"
+            :description="answer.description || ''"
+            :status="answer.status || 'proposed'"
+            :active="isActivePlan"
+            :awaiting-approval="isAwaitingPlanApproval"
             :disabled="interactionDisabled"
             :should-stream="shouldStream"
             class="mb-3"
@@ -171,7 +176,7 @@ export default {
     },
     computed: {
         ...mapState(useProductAssistantStore, ['supportedActions', 'toolApprovalStatuses']),
-        ...mapState(useProductExpertStore, ['agentMode', 'isWaitingForResponse', 'messages']),
+        ...mapState(useProductExpertStore, ['agentMode', 'isWaitingForResponse', 'messages', 'activePlanId']),
         isLatestMessage () {
             const msgs = this.messages || []
             return msgs.length > 0 && msgs[msgs.length - 1]?._uuid === this.messageUuid
@@ -234,6 +239,12 @@ export default {
         },
         isPlanAnswer () {
             return this.answer.kind === 'plan'
+        },
+        isAwaitingPlanApproval () {
+            return (this.answer.status || 'proposed') === 'proposed' && !this.interactionDisabled
+        },
+        isActivePlan () {
+            return !!this.answer.planId && this.answer.planId === this.activePlanId
         },
         isEditorContext () {
             // In editor context, the route name includes 'editor'
--- a/frontend/src/components/expert/components/messages/components/AnswerWrapper.vue
+++ b/frontend/src/components/expert/components/messages/components/AnswerWrapper.vue
@@ -85,7 +85,6 @@
             :answer-uuid="answer._uuid"
             :name="answer.name || ''"
             :description="answer.description || ''"
-            :status="answer.status || 'proposed'"
             :active="isActivePlan"
             :awaiting-approval="isAwaitingPlanApproval"
             :disabled="interactionDisabled"
@@ -241,7 +240,7 @@ export default {
             return this.answer.kind === 'plan'
         },
         isAwaitingPlanApproval () {
-            return (this.answer.status || 'proposed') === 'proposed' && !this.interactionDisabled
+            return !this.isActivePlan && !this.interactionDisabled
         },
         isActivePlan () {
             return !!this.answer.planId && this.answer.planId === this.activePlanId
--- a/frontend/src/components/expert/components/messages/components/resources/PlanCard.vue
+++ b/frontend/src/components/expert/components/messages/components/resources/PlanCard.vue
@@ -1,5 +1,5 @@
 <template>
-    <div class="expert-plan">
+    <div v-if="!hasStructure" class="expert-plan">
         <rich-content
             :content="plan"
             :message-uuid="messageUuid"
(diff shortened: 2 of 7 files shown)


frontend/src/components/expert/components/messages/components/AnswerWrapper.vue
```diff
@@ -239,6 +243,12 @@ export default {
         isPlanAnswer () {
             return this.answer.kind === 'plan'
         },
+        isAwaitingPlanApproval () {
+            return !this.isActivePlan && !this.interactionDisabled
+        },
+        isActivePlan () {
+            return !!this.answer.planId && this.answer.planId === this.activePlanId
```
cstns, line 250: This gates the approval UI on planId equality alone. If the agent ever reuses a planId on a revised plan (Request changes flow), or emits `expert:tasks` at proposal time, the fresh unapproved plan renders collapsed and Active with no way to approve or reject it. Is planId reuse ruled out by the agent contract?

frontend/src/components/expert/components/messages/components/resources/PlanCard.vue
```diff
@@ -75,6 +159,11 @@ export default {
         }
     },
     emits: ['approve', 'edit-manual', 'request-changes', 'reject', 'streaming-complete'],
+    computed: {
+        hasStructure () {
```
cstns, line 163: Keying the whole structured/legacy split on the display name means a plan with a planId but a blank name silently drops the description and can never show the Active pill. `answer.planId` feels like the sturdier discriminator.

frontend/src/components/expert/components/messages/components/resources/PlanCard.vue
```diff
@@ -44,14 +44,82 @@
             </ff-button>
         </div>
     </div>
+
+    <div v-else class="expert-plan expert-plan--structured">
+        <template v-if="awaitingApproval">
+            <div class="plan-head">
+                <span class="plan-name">Plan: {{ name }}</span>
```
cstns, line 51: The comment in AnswerWrapper says plan answers carry their heading inside the Markdown content. If that's still true, structured plans render the title twice (once here, once in the body). If the contract changed, that comment needs updating instead.

frontend/src/components/expert/components/messages/components/resources/PlanCard.vue
```diff
@@ -44,14 +44,82 @@
...
+            </div>
+            <p v-if="description" class="plan-desc">{{ description }}</p>
+            <rich-content
+                :content="plan"
+                :message-uuid="messageUuid"
+                :answer-uuid="answerUuid"
+                :should-stream="false"
+                class="plan-body"
+            />
+            <div class="plan-actions">
```
cstns, line 61: The four button block is now duplicated byte for byte between the legacy and structured branches, and the rich-content invocation appears three times. Rendering the actions once behind `v-if="!hasStructure || awaitingApproval"` would collapse most of this.

frontend/src/components/expert/components/messages/components/resources/PlanCard.vue
```diff
@@ -65,6 +133,22 @@ export default {
...
+        },
+        description: {
+            type: String,
+            default: ''
+        },
+        active: {
+            type: Boolean,
+            default: false
+        },
+        awaitingApproval: {
```
cstns, line 148: This prop is fully derivable from `active` and `disabled` at its only call site, and the three props can contradict each other (the defaults already allow awaiting approval and active at once). A local computed would remove the drift risk.

frontend/src/components/expert/components/messages/components/resources/PlanCard.vue
```diff
@@ -84,20 +173,51 @@ export default {
...
+    gap: 0.5rem;
+}
+
+.plan-name {
+    font-size: 1.125rem;
+    font-weight: 600;
+    color: var(--ff-color-text);
+}
+
+.plan-badge {
```
cstns, line 194: There's a `.forge-badge` primitive in index.css already doing the pill shape, see UserDetails.vue for bare span usage. Needs one local color rule for the accent fill, but the shape/size shouldn't be re-rolled here.

cstns (changes requested):
The card looks good and the active-plan pointer idea is right, but the plan's lifecycle state is inferred rather than tracked, and that one root cause produces three real bugs:

1. **A plan that was never approved loses its buttons forever.** The Approve/Edit/Request changes/Reject block only exists inside the `awaitingApproval` branch, and that flips false the moment the plan stops being the latest message. Any later message does it, including store generated ones like the session expiry warning or an MQTT disconnect banner. The pending plan then collapses and is indistinguishable from an approved one, with no way left to approve it. Before this PR the buttons stayed visible and just went disabled, so this is a regression.

2. **The Active pill can never appear on HTTP.** The only thing that ever writes `activeTaskList` is the MQTT inflight handler, but plan mode is fully reachable over HTTP (the toggle is gated only on the agent, and `sendHttpQuery` carries `planMode` and `supportsPlansAndTasks` in the context), so the pill silently never shows there. This is the concrete version of my transport note on #8468: the frontend now renders transport dependent UI, so "the agent side handles transport safety" no longer covers it. Related question: what does the agent do with `supportsPlansAndTasks: true` on an HTTP turn? If it changes planning behaviour based on a flag it can never fulfil, the degradation is worse than a missing pill.

3. **Stop permanently erases the pill.** `stopInflightChat` nulls `activeTaskList`, so after stopping mid execution the approved plan collapses anonymous for the rest of the session. Same if the agent ever sends an empty task list.

All three would go away if each plan answer carried its own approval state (or the store tracked approved plan ids) instead of deriving it from `isLatestMessage`, `isWaitingForResponse` and the task list pointer. Worth fixing at that level rather than patching the three symptoms.

Smaller things inline.

andypalmi:
Thanks, the root-cause framing was right. Active state now lives in its own agent-owned store field (activePlanId) instead of being inferred from the task-list pointer, which was the shared cause.

1. Fixed. Collapse is gated on active, not on being the latest message. An unapproved plan always keeps its buttons, they just disable off-latest like before.

2. Intended, not a regression we want to close here (happy to be pushed on it): new surfaces (task panel, Active pill, live plan state) are MQTT-only by design, HTTP keeps prior behavior. On your question, planning is driven by a separate flag (supportsPlanMode) that is honored on any transport, so plan mode and the plan card still work over HTTP. supportsPlansAndTasks only gates the live surfaces, which are additionally MQTT-gated, so it is inert over HTTP. No behavior changes on a flag it cannot fulfill.

3. Fixed. Stop and an empty task list no longer clear the active pointer, so the pill survives both.

Inline: discriminator is answer.planId not the name; awaitingApproval is now a local computed off active; actions render once; the pill reuses .forge-badge. The double title is agent-side (the card header owns it now), tracked in FlowFuse/engineering#361.

cstns (approved):

PR #8471, opened 2026-09-10
Allow admin to force team type change

knolleary:
Follow up to #8456 

This allows an admin to change team type of an manual billing team even if the team usage exceeds the new team type usage limits. A warning is shown to the admin making the change in this scenario - telling them to apply the necessary limit overrides after making the team type change.

The second commit fixes the handling of a `0` device limit in the frontend. That was interpreted to mean 'unlimited' - whereas the backend uses `-1` to mean unlimited. With the new Hub plan, a limit of `0` is valid but wasn't being honoured in the UI.

With this fix it now shows the team limit banner on the Remote instance page, and disables the 'add instance' button on the team home page if the limit has been reached.

### DIFF
--- a/forge/routes/api/team.js
+++ b/forge/routes/api/team.js
@@ -801,6 +801,9 @@ module.exports = async function (app) {
             if (request.body.type) {
                 auditLogFunc = app.auditLog.Team.team.type.changed
                 let billingIntervalUpgrade = false
+                // If the request is from an admin, and the team has unmanaged billing, we allow the admin
+                // to override the limit checks and change the team type.
+                let isAdminOverridingLimits = false
                 const bodyOptions = { ...request.body }
                 const targetTypeId = bodyOptions.type
                 delete bodyOptions.type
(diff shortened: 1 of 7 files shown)


hardillb (approved):
Logically looks good, I don't have enough local billing stuff setup to test the all the UI tweaks (should really find time to set it all up again)

cstns:
The new `teamDeviceLimitReached` gate on the Home page reads `totalDevices`, which is summed from `deviceStateCounts`. That count comes from `GET /teams/:teamId/instance-counts`, and it has a few problems as a source for a limit check:

1. It's RBAC-filtered per user (`Device.countByState` skips devices in applications the user lacks `device:read` on). A user with restricted per-application roles sees an undercount, so their Add Instance button stays enabled on a team that's actually at its limit, and the create then fails server-side with `device_limit_reached`. That's the exact flow this PR is trying to get rid of.

2. It's fetched async in `mounted()` and not awaited, so the gate treats the count as 0 on first render. And since `getDeviceStateCounts()` ends with `.catch(e => e)`, a failed fetch leaves it at 0 permanently, so the gate never engages for that visit.

3. It's never refreshed after creating a device. `deviceCreated()` only opens the credentials dialog, so a team one below its limit can create a device and the button stays enabled until a reload (instance deletion on this page does re-fetch, so the create path looks like an oversight).

The same undercount also flows into `RecentlyModifiedDevices` via the `totalDevices` prop and into `TeamDeviceCreateDialog` as `teamDeviceCount`.

Might be worth having the gate read `team.deviceCount` instead, which is the authoritative count the other copies of this computed already use (`DevicesBrowser.vue`, `DeviceStep.vue`), and adding a re-fetch (or team refresh) in `deviceCreated()`. That would cover all three points, and `deviceStateCounts` can stay for the state breakdown widgets it was built for.

cstns (changes requested):
see comment above

knolleary:
@cstns I've updated the limit checks to use `team.deviceCount` and verified locally that a user with rbac applied behaves correctly.

The only piece I haven't done is the refresh on create. That's an edge case I'm not worrying about - need to get this through to unblock Esme.

cstns (approved):

Issue #8122, opened 2026-08-07
Migrate dialogs to the universal store-based modal system


cstns:
We have a growing problem with how dialogs are rendered across the frontend. The current dominant pattern (component-ref based dialogs using `ff-dialog`) always renders the full dialog DOM tree on mount, hidden via CSS class toggling. Every page that uses dialogs has them sitting in the DOM at all times, teleported to `<body>`, just invisible.

This gets worse when dialog-bearing components are rendered inside loops. We have confirmed cases where this results in tens or potentially hundreds of hidden dialog instances in the DOM:

* `BlueprintTile` contains an `AssetDetailDialog` and is rendered in nested v-for loops across 4 pages (blueprint library, blueprint selection, admin blueprints, instance creation flow). 50 blueprints = 50 hidden dialogs.
* `PipelineStage` contains a `DeployStageDialog` and is rendered per-stage per-pipeline. 5 pipelines with 4 stages = 20 hidden dialogs.

Beyond the DOM bloat, the ref-based pattern also creates tight coupling. Every parent that needs a dialog has to import it, mount it in the template, and manage a ref. The same dialog used in two places means duplicated imports, template entries, and refs.

## The universal system

We already have a store-based universal dialog system that solves all of this:

* **Store** (`stores/ux-dialog.js`): Manages dialog state, callbacks, and dynamic component config
* **Service** (`services/dialog.js`): Exposes `Dialog.show()` and `Dialog.showAsync()` for triggering from anywhere
* **Renderer** (`components/dialogs/PlatformDialog.vue`): Mounted once in the `Platform.vue` layout, dynamically renders whatever component the store holds via `<component :is="dialog.is.component">`

This means one dialog container in the DOM, ever. Content renders on demand. No iteration multiplication. No coupling between parent and dialog. The right drawer system (`ux-drawers.js` + `RightDrawer.vue`) follows the exact same pattern and is already widely adopted.

Only ~4 places currently use the universal system vs ~54 using the old ref pattern.

## Scope

The migration covers roughly 49 standalone dialog component files and ~9 inline `ff-dialog` usages across the frontend:

* **Team section**: 15 dialog files
* **Admin section**: 7 dialog files
* **Instance section**: 6 dialog files
* **Device section**: 5 dialog files
* **Application section**: 2 dialog files
* **Account section**: 2 dialog files
* **Shared components**: 12 dialog files
* **Inline dialogs**: ~9 across various pages

## Technical considerations

[...]

PR #8483, opened 2026-09-11
Add an AI connector modal reachable from the nav bar

### LINKED ISSUE #8481: Add an AI connector modal, reachable from the nav bar
## Summary

Surface the AI page CTA inside the platform as a modal, so users can connect a third-party agent without leaving the app. It shows once automatically, and stays reachable afterwards from a button in the nav bar.

## Behaviour

The modal auto-shows once on load. After it has been shown, it does not show again for 10 days. The shown timestamp is kept in localStorage, so the suppression is per browser (a new browser or a cleared cache will show it again, which is acceptable here).

A button in the lower-left region of the nav bar opens the modal at any time, so users who dismissed it can find it again.

## Content

The modal reuses the CTA from the [/ai page](https://flowfuse.com/ai/):

- The short intro at the top.
- The four agent tabs below it (Microsoft Copilot, ChatGPT, Claude, Local and Custom Agents), each with the copy-the-connector-URL step and the open-your-agent step.
- Links out to the [documentation](https://flowfuse.com/docs/user/expert/third-party-agents/) and the /ai page.

## Notes

Reuse existing FlowFuse components (`ff-button` for the nav trigger) rather than raw markup.

andypalmi:
Closes #8481

## What

Surfaces the connector CTA from [flowfuse.com/ai](https://flowfuse.com/ai/) inside the platform as a modal, so users can point their own AI agent at FlowFuse without leaving the app.

- Auto-shows once, then holds off for 10 days. The last-shown timestamp is kept in localStorage, so the suppression is per browser.
- A "Connect an AI agent" button in the lower-left of the nav bar opens the modal at any time, so a user who dismissed it can find it again.
- The modal reuses the /ai content: a short intro, then the four agent tabs (Microsoft Copilot, ChatGPT, Claude, Local and Custom Agents), each with the copy-the-connector-URL step and the open-your-agent step, plus links to the documentation and the /ai page.
- The connector URL is derived from the platform's own origin, so it is correct on Cloud and self-hosted.
- Everything is gated on the AI and MCP third-party features.

## Testing

- [ ] With AI + MCP third-party enabled, the modal auto-shows once and does not return for 10 days.
- [x] The nav button opens it at any time.
- [x] Copying the connector URL works; tabs switch; doc and /ai links open.
- [ ] With the features disabled, neither the button nor the modal appears.

### Modal
Screenshot 2026-09-14 at 11 16 16

### Button in bottom of left nav
Screenshot 2026-09-11 at 18 39 43


ZJvandeWeg:
Love it, and the speed of implementation @andypalmi <3

frontend/src/components/dialogs/AiConnectorModal.vue
```diff
@@ -0,0 +1,357 @@
...
+        step2Url: 'https://chatgpt.com/'
+    },
+    {
+        id: 'local',
+        icon: true,
+        name: 'Local and Custom Agents',
+        step2Title: "Your MCP client's config",
+        step2Body: 'Any MCP client, pointed at your own model.',
+        step2Label: 'See the documentation',
+        step2Url: 'https://flowfuse.com/docs/user/expert/'
```
ZJvandeWeg: @andypalmi Meta remark, but I think that should be updated too. 

Should we have a path `/docs/user/third-party-agents`?

frontend/src/components/icons/ai-agents/microsoft-copilot.svg
cstns, line 1: `frontend/src/components/icons` is where we store icons or svg's

frontend/src/App.vue
```diff
@@ -43,6 +43,7 @@
                 </ff-layout-plain>
             </template>
             <EducationModal />
+            <AiConnectorModal />
```
cstns: please read up on https://github.com/FlowFuse/flowfuse/issues/8122 and use the the universal modal pattern moving forward.

cstns (changes requested):
Looks good overall, please address the img file paths and the modal pattern used. Also there are some tests failing

frontend/src/components/dialogs/AiConnectorModal.vue
andypalmi: The "Local and Custom Agents" step now links to the third-party-agents docs (https://flowfuse.com/docs/user/expert/third-party-agents/)

andypalmi:
@cstns Thanks, all three addressed in the latest push:

- The agent logos now live in `frontend/src/components/icons/ai-agents/` instead of `public/images`, and are imported into the component so webpack bundles them.
- Read up on #8122 and moved the modal onto the universal dialog pattern. `AiConnectorModal.vue` is now a content-only component with no `ff-dialog` wrapper; it is opened through `Dialog.show()` via a small `useAiConnectorModal()` composable, so `PlatformDialog` supplies the shell and it only lives in the DOM while open. That also removes the global mount from `App.vue`.
- The failing tests were a side effect of that global mount: the always-teleported dialog box left five primary buttons in the DOM on every page, so the admin stacks/instance-types specs matched more buttons than they expected. With the modal no longer permanently mounted, those specs pass again.

cstns (approved):

PR #8484, opened 2026-09-11
Deliver a task surface that trails the chat reply

andypalmi:
> Based on #8468 and targeting its branch, so the diff shows only this fix. It will retarget to `main` automatically when #8468 merges.

Closes FlowFuse/engineering#349

## Summary

A task surface that arrives right after the chat reply was being dropped, so the pinned list could keep its pre-final states and the agent's send of that surface would wait out its timeout with no ack. The earlier guards (`_onMqttMessage` bailing on an empty in-flight map, and `handleInFlightRequest` dismissing a request with no matching entry) fired as soon as the reply emptied the map, before a trailing surface could be applied or acked.

## Change

Correlate every inflight message to its originating chat turn instead of inferring intent from the map being empty.

- On the final reply, the finished turn's id is kept briefly in a bounded per-agent recently-completed set.
- An inflight message is applied or run only when its chat transaction matches a live or recently-completed turn, so a surface trailing a normal reply is still applied and acked.
- Anything that matches neither, a request left on the wire after a Stop or a stale surface from an earlier turn, is dropped: not applied, not run against the editor, and not acked, so a stopped turn cannot execute an automation or repaint the list.
- The loading line updates only while the turn is live, so a status trailing the reply cannot repaint the next turn, and Stop clears it.
- Every branch answers through a single response helper.

Browser-session (MCP) requests carry no chat transaction, so they bypass correlation and are handled as before. With the empty-map guard gone they can now dispatch UI tools with no chat open, which matches the MCP contract.

## Testing

- `_onMqttMessage` is now driven end to end: a live surface applies and acks, a surface trailing the reply still applies through recently-completed, a surface from an unseen turn is dropped, and a task surface plus an automation left on the wire after Stop are both dropped, with the automation never dispatching.
- Full store spec passes.

### DIFF
--- a/frontend/src/stores/product-expert.js
+++ b/frontend/src/stores/product-expert.js
@@ -39,6 +39,8 @@ export const useProductExpertStore = defineStore('product-expert', {
         // 'request-plan-change' focuses an empty composer for the plan card's "Request
         // changes"; 'reset' clears a plan loaded via "Edit manually" but not sent.
         composerCommand: null,
+        // Set on stop/drop, cleared on the next send: a surface after a stop acks but does not apply.
+        _chatStopped: false,
         _seenTransactionIds: new Map(),
         // Open human-in-the-loop approval batch (#421). When a turn defers a tool batch
         // for approval the agent ends the turn and returns the card(s); we hold the
@@ -305,6 +307,7 @@ export const useProductExpertStore = defineStore('product-expert', {
             if (!mqttService.hasClient(mqttConnectionKey)) await this.establishMqttComms()
 
             // add the query as an inFlight request
+            this._chatStopped = false
             this._inFlightRequests.set(transactionId, { query, transactionId })
 
             const { entityId, entityType } = mqttTopicHelper.getEntityTopicPaths()
@@ -377,19 +380,17 @@ export const useProductExpertStore = defineStore('product-expert', {
             }
         },
         async handleInFlightRequest ({ topic, message, payload: parsedPayload, transactionId, sessionId, chatTransactionId } = {}) {
-            // Match the originating chat request explicitly (not just the first entry) so a
-            // concurrent in-flight request — e.g. an open tool approval — can't shadow it and
-            // cause us to drop a valid in-flight request.
-            const inFlightRequest = Array.from(this._inFlightRequests.values())
-                .find(r => r.transactionId === chatTransactionId)
-
             // A third-party MCP request is addressed to this tab's browser session rather
             // than a chat session, and has no originating chat request to correlate with,
-            // so it bypasses both checks below. Everything else keeps today's behaviour.
+            // so it bypasses the check below. Everything else keeps today's behaviour.
             const isBrowserSession = !!sessionId && sessionId === useAccountAuthStore().getSessionId()
 
-            // dismiss inFlight requests that don't match the existing sessionId or the inFlight message transactionId
-            if (!isBrowserSession && (sessionId !== this.sessionId || !inFlightRequest)) return
+            // dismiss messages from a different chat session
+            if (!isBrowserSession && sessionId !== this.sessionId) return
+
+            // A chat surface can trail the reply that emptied the in-flight map; it is still
+            // acked and applied. One that trails a stop is acked but not applied.
+            const suppressChatState = !isBrowserSession && this._chatStopped
 
             const servicesOrchestrator = getAppOrchestrator()
             const assistantStore = useProductAssistantStore()
@@ -420,7 +421,7 @@ export const useProductExpertStore = defineStore('product-expert', {
 
             // expert:tasks has its own panel; its status rides expert:status-message.
             // Every other inflight type feeds the loading line.
-            if (parsedTopic.inflightType !== 'expert:tasks') {
+            if (parsedTopic.inflightType !== 'expert:tasks' && !suppressChatState) {
                 this._addInFlightUpdate(payload.status || payload.toolname || 'Processing request...')
             }
 
@@ -451,10 +452,12 @@ export const useProductExpertStore = defineStore('product-expert', {
                 })
                 break
             case parsedTopic.inflightType === 'expert:tasks': {
-                const items = Array.isArray(payload.items) ? payload.items : []
-                this._agentStore.activeTaskList = items.length
-                    ? { planId: payload.planId ?? null, title: payload.title || 'Tasks', items }
-                    : null
+                if (!suppressChatState) {
+                    const items = Array.isArray(payload.items) ? payload.items : []
+                    this._agentStore.activeTaskList = items.length
+                        ? { planId: payload.planId ?? null, title: payload.title || 'Tasks', items }
+                        : null
+                }
                 try {
                     await mqttService.publishMessage(connectionKey, {
                         qos: 2,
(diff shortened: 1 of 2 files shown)


frontend/src/stores/product-expert.js
```diff
@@ -377,19 +380,17 @@ export const useProductExpertStore = defineStore('product-expert', {
...
-            // concurrent in-flight request — e.g. an open tool approval — can't shadow it and
-            // cause us to drop a valid in-flight request.
-            const inFlightRequest = Array.from(this._inFlightRequests.values())
-                .find(r => r.transactionId === chatTransactionId)
-
             // A third-party MCP request is addressed to this tab's browser session rather
             // than a chat session, and has no originating chat request to correlate with,
-            // so it bypasses both checks below. Everything else keeps today's behaviour.
+            // so it bypasses the check below. Everything else keeps today's behaviour.
             const isBrowserSession = !!sessionId && sessionId === useAccountAuthStore().getSessionId()
```
cstns, line 426: Removing the top-level guard makes this bypass reachable on the expert client's support-channel path, so browser-session requests now dispatch UI tools with the chat completely idle. Maybe intended (the MCP subscriber docs say MCP must work with no chat open), but the PR says browser-session handling is unchanged and for this path that's no longer true. Worth stating either way.

frontend/src/stores/product-expert.js
```diff
@@ -420,7 +421,7 @@ export const useProductExpertStore = defineStore('product-expert', {
 
             // expert:tasks has its own panel; its status rides expert:status-message.
             // Every other inflight type feeds the loading line.
-            if (parsedTopic.inflightType !== 'expert:tasks') {
+            if (parsedTopic.inflightType !== 'expert:tasks' && !suppressChatState) {
```
cstns: A status message trailing the reply now pushes onto `inFlightUpdates` after the reply cleared it, and nothing clears it again until the next reply, so the stale text shows up as the *next* turn's loading line. `stopInflightChat` never calls `_clearInFlightUpdates` either, so stopped-turn statuses leak the same way.

frontend/src/stores/product-expert.js
```diff
@@ -39,6 +39,8 @@ export const useProductExpertStore = defineStore('product-expert', {
         // 'request-plan-change' focuses an empty composer for the plan card's "Request
         // changes"; 'reset' clears a plan loaded via "Edit manually" but not sent.
         composerCommand: null,
+        // Set on stop/drop, cleared on the next send: a surface after a stop acks but does not apply.
+        _chatStopped: false,
```
cstns: This is the one piece of per-conversation state on the umbrella store while everything else (`sessionId`, `_inFlightRequests`, `activeTaskList`) is per-agent, and it's toggled by hand in four places with no owning helper. The next teardown or send path added will forget one side of the toggle and silently un-stop or permanently suppress. Goes away entirely with the correlation approach.

frontend/src/stores/product-expert.js
```diff
@@ -451,10 +452,12 @@ export const useProductExpertStore = defineStore('product-expert', {
...
-                this._agentStore.activeTaskList = items.length
-                    ? { planId: payload.planId ?? null, title: payload.title || 'Tasks', items }
-                    : null
+                if (!suppressChatState) {
+                    const items = Array.isArray(payload.items) ? payload.items : []
+                    this._agentStore.activeTaskList = items.length
+                        ? { planId: payload.planId ?? null, title: payload.title || 'Tasks', items }
+                        : null
+                }
                 try {
```
cstns, line 507: This ack and the `expert:status-message` ack above are now byte-identical 12-line publishes, and the same envelope appears five times in this function. A local `respond(body)` next to `responseTopic` collapses each case to a line, and ends the drift where this one got a try/catch and the status one didn't.

test/unit/frontend/stores/product-expert.spec.js
```diff
@@ -318,4 +337,54 @@ describe('product-expert store', () => {
             expect(store.loadingVariant).toBe(SUPPORT_AGENT)
         })
     })
+
+    describe('stopInflightChat', () => {
```
cstns, line 433: Both cases set `_chatStopped` directly and hand-build the payload, and the file-wide `parseTopic` mock hardcodes `expert:tasks`, so the suite verifies "a boolean suppresses two assignments" rather than the reported scenario. Notably the automation-branch gap ships green: nothing drives `_onMqttMessage` end to end or any other inflight type.

cstns (changes requested):
The race is real and worth fixing, but the fix swaps one inferred signal (map emptiness) for another (a mood flag), and in the process drops the `chatTransactionId` correlation that was also our guard against rogue and stale messages. Two concrete regressions fall out:

1. **Stop no longer stops the agent's actions.** `suppressChatState` only gates the loading line and `expert:tasks`; the `automation:*` and `automation-ui:mcp-call-tool` branches run unguarded. An automation already on the wire when the user hits Stop still executes against their editor, and we publish a success result so the agent keeps going. The old top-of-`_onMqttMessage` guard dropped exactly these.

2. **A stale surface from a stopped turn lands on the next turn.** `_chatStopped` is cleared by the very next send and `sessionId` is stable across the whole session, so a delayed turn-A `expert:tasks` (or `automation:*`) passes every remaining check and repaints turn B. `_seenTransactionIds` can't catch it, it keys on the message's own id, not the chat transaction. The removed `find(r => r.transactionId === chatTransactionId)` was the thing dropping these.

I'd rather fix the race by widening the correlation than by removing it: when the reply branch deletes an in-flight entry, remember its id in a small recently-completed set, and accept inflight messages whose `chatTransactionId` matches a live *or* recently completed transaction (ack-without-apply otherwise). That fixes trailing messages of every type, keeps uncorrelated messages out, needs no flag to clear in the right four places, and answers stop-vs-trailing per transaction instead of per global mood. As a bonus, dead-turn surfaces stop getting acked, so the agent isn't encouraged to keep producing after a Stop.

Smaller things inline.

andypalmi:
You are right, thanks for the push. I went with the correlation approach: on the final reply we drop the finished turn's id into a small bounded recently-completed set, and an inflight message is only applied or run when its chatTransactionId matches a live or recently-completed turn. Anything else, a request left on the wire after Stop or a stale surface from an earlier turn, is dropped: not applied, not run against the editor, and not acked. That removes the global flag and answers stop-versus-trailing per transaction instead of per session.

frontend/src/stores/product-expert.js
andypalmi: Good catch on the description. Browser-session (MCP) requests still bypass correlation by design, and yes they now dispatch UI tools with no chat open, which matches the MCP contract. I updated the PR body so it no longer claims that path is unchanged.

frontend/src/stores/product-expert.js
andypalmi: Fixed. The loading line now updates only while the turn is live, so a status trailing the reply cannot repaint the next turn, and stopInflightChat clears the loading line alongside the task list.

frontend/src/stores/product-expert.js
andypalmi: Gone. The correlation state lives per agent next to inFlightRequests, as a bounded recentlyCompletedTransactions map, so there is no hand-toggled flag on the umbrella store to forget in one path.

frontend/src/stores/product-expert.js
andypalmi: Done. There is a local respond(body) next to responseTopic now, and all five branches answer through it, so the envelope exists once and the ack and status paths cannot drift.

test/unit/frontend/stores/product-expert.spec.js
andypalmi: Reworked. The tests drive _onMqttMessage end to end with a configurable parseTopic: a live surface applies and acks, a surface trailing the reply still applies through recently-completed, a surface from an unseen turn is dropped, and a task surface plus an automation left on the wire after Stop are both dropped, with the automation never dispatching.

PR #8492, opened 2026-09-14
Add Gemini to the AI connector modal and the docs

dimitrieh:
## What

Adds Gemini to the AI connector modal from #8483, and the matching section to the third-party-agents docs.

The modal shipped with Copilot, ChatGPT, Claude and Local, matching the widget on flowfuse.com/ai at the time. Neither offered Gemini.

- Gemini reaches a custom MCP server through **Settings & help → Connected Apps → Add a custom app**. The platform supports Dynamic Client Registration, so there is no client ID or secret to enter.
- Google limits custom apps to Gemini Spark on personal accounts in the US. Both the tab and the docs say so, the way the ChatGPT tab says developer mode needs a paid plan.
- `gemini.svg` is the same mark the website carries, copied in because the modal has its own icon directory.

The docs page embeds the website's `agent-setup-tabs` component, so its Gemini tab appears as soon as https://github.com/FlowFuse/website/pull/5799 lands. The prose section keeps "Where to add it, per agent" from listing fewer agents than the tabs above it offer.

## Testing

Checked statically, not by running the app:

- [x] `CLIENTS` parses and every entry has a name, a mark, a step title and a URL.
- [x] The SVG is well formed and its gradient reference resolves inside the file, so it renders standalone as an ``.
- [x] The new import sits in path-alphabetical order with the existing three.
- [ ] **Not verified:** the modal rendered in a running platform. Worth a click-through on the preview: five tabs, Gemini between Claude and Local, the sparkle at tab size, and the copy-URL button still working from that tab.


dimitrieh:
Updated. This now builds on #8499, because the two collided in two places and merging them in either order went wrong on its own.

1. The modal client list auto-merged with **no conflict** but put Gemini after the two coding agents, since both branches insert immediately before the Local tab. Gemini is anchored on the ChatGPT entry now: `claude copilot chatgpt gemini claude-code codex local`.
2. The docs **did** conflict: this branch adds a Gemini section directly above `Command-line and editor agents`, and #8499 renames that heading to `Coding agents`. Resolved keeping both, in reading order.

Worth noting GitHub reported both PRs as MERGEABLE throughout, because it only tests each against main, never against its sibling.

**Merge #8499 first.**

dimitrieh:
converting this to a draft for now until decision has been reached on supporting gemini officially as only gemini spark users could use this.

PR #8494, opened 2026-09-14
Leave onboarding before ending the intake stage

### LINKED ISSUE #8493: Skipping onboarding lands on a 404 instead of the team home
Clicking the "Set it up myself" link on the AI onboarding page takes you to the 404 page instead of the team home.

What happens: `skipOnboarding` ends the onboarding stage before navigating away. Ending the stage flips the page's own `notAvailable` guard, and the guard's watcher replaces the route with the 404 page while the navigation to the team home is still in flight. The page effectively 404s itself on the way out.

The unit tests didn't catch it because the ux store mock never actually flipped `isOnboardingIntake`, so the guard never reacted in tests.

Might be worth navigating away first and only ending the stage once the page is unmounted, so the guard has nothing left to react on.

cstns:
Closes #8493

`skipOnboarding` ended the onboarding stage while the page was still mounted. That flips `notAvailable`, whose watcher replaces the route with the 404 page, and the replace beats the push to the team home. Now the navigation is awaited first and the stage ends after, so the page is unmounted before the stage change can trip the guard.

The spec's ux store mock is now reactive and `endOnboarding` really ends the intake stage, which is what lets the new regression test reproduce the race. Making the store reactive also surfaced that mounted pages leaked between tests and reacted to later tests' store changes, so the spec now auto unmounts after each test.


andypalmi (approved):

andypalmi:
I see this CI warning at line 15 in `frontend/src/pages/team/Onboarding.vue`, should we fix it here already while we are at it?

Expected indentation of 20 spaces but found 19 spaces

frontend/src/pages/team/Onboarding.vue
andypalmi:
```suggestion
                    Set it up myself
```

PR #8496, opened 2026-09-14
Add MCP tool to update an application

### LINKED ISSUE #7691: 5.3-b Write tools, non-destructive (phase 2)
**Parent:** #7672 (5.3 Application management)
**Tool file:** `forge/ee/lib/mcp/tools/applications.js` (extend)

|---|---|---|---|
| `platform_update_application` | `PUT /applications/:applicationId` | `project:edit` | write |

**Design notes:**
- Application CRUD reuses `project:*` scopes (`project:create`/`project:edit`/`project:delete`, `project:read`), not `application:*`.
- `create_application` already ships and evaluates `project:create` against `body.teamId`'s team.

```js
platform_update_application: {
  description: 'Update an application name and/or description.',
  inputSchema: z.object({
    applicationId: z.string().describe('Application id (opaque hashid) to update'),
    name: z.string().optional().describe('New application name'),
    description: z.string().optional().describe('New application description')
  })
}
```

**Tests:**
- Update rejected for read-only PAT.

---

cstns:
Closes #7691

Adds `platform_update_application`, wrapping `PUT /api/v1/applications/:applicationId`.

While verifying the route's actual behavior I found two quirks that the tool description now spells out for agents:

* `name` is effectively required: the route assigns it unconditionally and a missing name fails the model's notNull validation with a 500.
* Omitting `description` leaves the stored value unchanged, but the response body reports `description: null` in that case. The description tells agents not to trust that null and to prefer passing both fields.

The handler only forwards `description` when it's provided, and an empty string goes through so the description can be cleared.


andypalmi (commented):
Verified the route behavior and tests, all good. One cleanup to keep the tool lean: the description restates argument logic that already lives (or should live) in the args. Could we:

- move the name/description quirks (overwrite, omit-returns-null, empty-clears) into the args that own them
- drop the "required" note, since `z.string().min(1)` already enforces it
- keep the description tool-level (what it does + the read-first tip)

I suggest :

```js
    Renames an application and/or updates its description.
    When changing only one field, read the current values with platform_get_application first and pass both.`,
inputSchema: {
    applicationId: applicationId.describe('The hashid of the application to update'),
    name: z.string().min(1).describe('Name for the application. It is always overwritten, so to change only the description pass the current name here to keep it unchanged'),
    description: z.string().optional().describe('New description. Omit to keep the stored value (the response then reports it as null); pass an empty string to clear it')
}
```

Avoids stating each behavior twice and matches the shape of `platform_create_application`.

forge/ee/lib/mcp/tools/applications.js
```diff
@@ -94,6 +94,29 @@ module.exports = [
...
+            description is optional. When omitted, the stored description is left unchanged, BUT the response then reports description as null - do not treat that null as the saved value; call platform_get_application to see the real one.
+            To avoid that ambiguity, prefer reading the current values with platform_get_application first and passing both name and description on every update. Pass an empty string as description to clear it.`,
+        annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: true, openWorldHint: false },
+        inputSchema: {
+            applicationId: applicationId.describe('The hashid of the application to update'),
+            name: z.string().min(1).describe('Name for the application. Required even when only changing the description - pass the current name to keep it'),
+            description: z.string().optional().describe('Description for the application. Omit to leave the stored value unchanged (the response will misreport it as null); pass an empty string to clear it')
+        },
```
andypalmi, line 110: Verified the route behavior and tests, all good. One cleanup to keep the tool lean: the description restates argument logic that already lives (or should live) in the args. Could we:

- move the name/description quirks (overwrite, omit-returns-null, empty-clears) into the args that own them
- drop the "required" note, since `z.string().min(1)` already enforces it
- keep the description tool-level (what it does + the read-first tip)

I suggest :

```suggestion
        description: `FlowFuse platform automation tool:
            Renames an application and/or updates its description.
            When changing only one field, read the current values with platform_get_application first and pass both.`,
        inputSchema: {
            applicationId: applicationId.describe('The hashid of the application to update'),
            name: z.string().min(1).describe('Name for the application. It is always overwritten, so to change only the description pass the current name here to keep it unchanged'),
            description: z.string().optional().describe('New description. Omit to keep the stored value (the response then reports it as null); pass an empty string to clear it')
        },
```

Avoids stating each behavior twice and matches the shape of `platform_create_application`.

cstns: sensible, yes

andypalmi (approved):

forge/ee/lib/mcp/tools/applications.js
```diff
@@ -94,6 +94,26 @@ module.exports = [
             return response
         }
     },
+    {
+        name: 'platform_update_application',
+        title: 'Update Application',
+        description: `FlowFuse platform automation tool:
+            Renames an application and/or updates its description.
+            When changing only one field, read the current values with platform_get_application first and pass both.`,
+        inputSchema: {
```
andypalmi, line 104:
```suggestion
        annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: true, openWorldHint: false },
        inputSchema: {
```

andypalmi (commented):
One blocker: `platform_update_application` is missing its `annotations` block, so the tool ships with `annotations` undefined. That is what fails the catalog invariant test (`catalog_spec.js:25`, `gives every tool a name, title, description, annotations and handler`) - the loop hits `tool.annotations.should` on an undefined value. Every other tool in the file declares one. Suggestion inline. `idempotentHint: true` since the PUT does a full replace, so repeating the same call lands the same state.

forge/ee/lib/mcp/tools/applications.js
andypalmi: Already applied by @cstns , gg

PR #8497, opened 2026-09-14
Add MCP tools for snapshot export, import, update and device target

### LINKED ISSUE #7688: 5.2-b Write tools, non-destructive (phase 2)
**Parent:** #7671 (5.2 Snapshot lifecycle)
**Tool file:** `forge/ee/lib/mcp/tools/snapshots.js` (extend)

|---|---|---|---|
| `platform_export_snapshot` | `POST /snapshots/:id/export` | `snapshot:export` | write |
| `platform_import_snapshot` | `POST /snapshots/import` | `snapshot:import` | write |
| `platform_update_snapshot` | `PUT /snapshots/:id` | `snapshot:edit` | write |
| `platform_set_instance_device_target` | `POST /projects/:id/devices/settings` | `project:snapshot:set-target` | write |

**Consolidation notes:**
- Export is a single tool: `platform_export_snapshot` takes a snapshot id and the generic route resolves the owner (instance or device) from the snapshot, so a separate instance-scoped export tool is not needed. The generic route also accepts an optional `components` selection.

**Design notes:**
- Export requires a `credentialSecret` in the payload (**400** if missing). Import requires `credentialSecret` when the snapshot has encrypted credentials. Model `credentialSecret` as a required input where the route demands it, and surface the validation error descriptively.
- Export is treated as a write even though it "reads" data (it decrypts/exports credentials), so it is annotated `readOnlyHint: false`.
- The generic `POST /snapshots/:id/export` accepts an optional `components` object (`{flows, credentials, envVars}`).

```js
platform_export_snapshot: {
  description: 'Export a snapshot by snapshot id; the owner (instance or device) is resolved from the snapshot. credentialSecret is mandatory; the export fails with 400 if it is missing.',
  inputSchema: z.object({
    id: z.string().describe('ProjectSnapshot id (opaque hashid), not a project UUID'),
    credentialSecret: z.string().describe('Secret used to re-encrypt exported credentials'),
    components: z.object({
      flows: z.boolean().optional().describe('Include flows in the export (default true)'),
      credentials: z.boolean().optional().describe('Include credentials in the export (default true)'),
      envVars: z.union([z.enum(['all', 'keys']), z.literal(false)]).optional().describe("Env vars to include: 'all' values, 'keys' only, or false to exclude")
    }).optional().describe('Optional selection of which snapshot components to export')
  })
}
platform_import_snapshot: {
  description: 'Import a snapshot into an instance or device. ownerId is an instance (project) UUID when ownerType is instance, or a device hashid when ownerType is device. The snapshot object requires name, flows and settings. credentialSecret is required only when the snapshot includes encrypted credentials.',
  inputSchema: z.object({
    ownerId: z.string().describe('Target owner id: instance (project) UUID or device hashid, matching ownerType'),
    ownerType: z.enum(['device', 'instance']).describe('Type of resource that will own the imported snapshot'),
    snapshot: z.object({
      name: z.string().describe('Snapshot name'),
      description: z.string().optional().describe('Snapshot description'),
      flows: z.object({
        flows: z.array(z.any()).describe('Node-RED flows array'),
        credentials: z.record(z.any()).optional().describe('Encrypted credentials object')
      }).describe('Flows payload'),
      settings: z.object({
        settings: z.record(z.any()).optional().describe('Runtime settings'),
        env: z.record(z.any()).optional().describe('Environment variables'),
        modules: z.record(z.any()).optional().describe('Installed module versions')
      }).describe('Settings payload')
    }).describe('Snapshot to import'),
    credentialSecret: z.string().optional().describe('Secret to decrypt the snapshot credentials'),
    components: z.object({
      flows: z.boolean().optional().describe('Import flows (default true)'),
      credentials: z.boolean().optional().describe('Import credentials (default true)'),
      envVars: z.union([z.enum(['all', 'keys']), z.literal(false)]).optional().describe("Env vars to import: 'all' values, 'keys' only, or false to exclude")
    }).optional().describe('Optional selection of which snapshot components to import')
  })
}
platform_update_snapshot: {
  description: 'Update a snapshot name and/or description by snapshot id.',
  inputSchema: z.object({
    id: z.string().describe('ProjectSnapshot id (opaque hashid), not a project UUID'),
    name: z.string().optional().describe('New snapshot name'),
    description: z.string().optional().describe('New snapshot description')
  })
}
platform_set_instance_device_target: {
  description: 'Set the device target snapshot for a hosted instance. The target snapshot is validated to belong to the instance.',
  inputSchema: z.object({
    instanceId: z.string().uuid().describe('Hosted instance (project) UUID whose device target to set'),
    targetSnapshot: z.string().describe('Snapshot id (opaque hashid) to set as the device target')
  })
}
```

**Tests:**
- Export/import/edit and set-target rejected for read-only PAT.
- Export without `credentialSecret` returns the descriptive validation error.
- Export resolves the owner from the snapshot for both instance- and device-owned snapshots.
- Set-target validates the snapshot belongs to the instance.

---

cstns:
Closes #7688

Adds four snapshot write tools:

* `platform_update_snapshot` wraps `PUT /api/v1/snapshots/:id`. The controller does proper partial updates, so the tool only sends the fields it was given. A blank name is not cleanly rejected: the controller throws a sequelize `ValidationError` and nothing maps that to a status code, so the route answers 500. The tool catches it first (whitespace only counts as blank, since the controller trims) and returns a 400. An update with no fields at all gets a 200 back with the snapshot unchanged, which reads as a successful edit that never happened, so the tool rejects that too.
* `platform_export_snapshot` wraps `POST /api/v1/snapshots/:id/export`. `credentialSecret` is unconditionally required by the route (400 without it), and the description warns that the default export includes hidden env var values, plus a reminder that the same secret is needed at import time.
* `platform_import_snapshot` wraps `POST /api/v1/snapshots/import`. One real quirk surfaced while reading the controller: `uploadSnapshot` calls `Object.keys(snapshot.settings.env)` unguarded, so a snapshot without `settings.env` 500s. The handler normalises an omitted env to `{}` so agents don't hit that. Same story for hidden env values, which get decrypted before the component filtering runs, so a keys-only or env-excluded import would 500 without a secret it does not actually need; the tool reduces env up front in both cases and the result is identical to what the route produces on its happy path. One rough edge left: the export response carries six extra fields (`id`, `createdAt`, `updatedAt`, `ownerType`, `user`, `exportedBy`) that this tool's `snapshot` argument does not accept, so pasting an export straight back in fails validation with `unrecognized key`. The description spells out which four fields to copy, but it might be worth letting the schema ignore the extras so the obvious export then import flow just works.
* `platform_set_instance_device_target` wraps `POST /api/v1/projects/:id/devices/settings`. The description carries a caution that setting the target deploys immediately to every assigned device, and notes the route can only set a target, not clear one. The tool makes `snapshotId` required because the route's only `reply.send` sits inside the `if (request.body.targetSnapshot)` block: omit it and you get a 200 with an empty body and nothing changed, a silent no-op rather than an error. It is also annotated `destructiveHint: true`, since it overwrites what every assigned device is running rather than adding to it, which puts it behind destructive tool access instead of plain write.

Descriptions were written from the actual route/controller behavior rather than assumptions, including the owner resolution (instance or device) happening from the snapshot itself. The behaviors above were checked against a running platform, calling each route directly and invoking the matching tool with the same arguments.


andypalmi (commented):
All four tools line up with the routes and controllers, and the tricky bits are handled well: the partial update with the empty-name guard, the `settings.env` normalization that dodges the unguarded `Object.keys` in `uploadSnapshot`, and making the device target snapshot required so the route (which never replies when it is missing) can't hang. Tests cover them nicely.

One optional cleanup, same theme as the pipeline stage tools: `platform_export_snapshot` and `platform_import_snapshot` carry near-identical `components` schemas. The differences are just the direction wording and the "exposes hidden values" caution, and that caution already lives in the export tool's description, so repeating it in the arg is the kind of duplication worth dropping. Could this be a single shared schema in `schemas.js` (next to `snapshotId`/`hostedInstanceId`), spread into both?

```js
// schemas.js
const snapshotComponents = z.object({
    flows: z.boolean().optional().describe('Include flows (default true). Excluding flows also excludes credentials'),
    credentials: z.boolean().optional().describe('Include the flow credentials (default true)'),
    envVars: z.union([z.literal('all'), z.literal('keys'), z.literal(false)]).optional().describe('Environment variables: "all" keeps keys and values (default), "keys" keeps only the names, false removes them')
}).optional()
```

And a smaller one: a couple of descriptions restate mechanics that already live in the args, for example the non-empty name and empty-string-to-clear notes on `platform_update_snapshot`. Could those stay only in the arg `.describe()` that owns them, keeping the description tool-level? The owner-resolution, partial-update, and immediate-deploy caution are genuinely tool-level and read well where they are.

None of this is blocking, happy for it to be a follow-up if you would rather keep this PR focused.

cstns:
Pushed a follow-up to the import tool after a closer read of the controller.

Hidden env vars are also exported encrypted (the env entry gets a `$`), and `uploadSnapshot` decrypts them *before* the flow-credentials guard, iterating the original snapshot rather than the component-filtered copy. Two consequences the tool didn't account for:

* Importing a snapshot with hidden env vars and no `credentialSecret` blows up with a raw 500, even with `components.envVars: false`, because the decrypt runs before the filtering and the catch regex doesn't match the crypto error.
* A wrong secret only produces the documented 400 when flow credentials are present. Without them there's nothing to validate against (env decryption is unauthenticated aes-256-ctr), so hidden values import as silent garbage with a 200.

Since both live in the route/controller, the tool handles them client-side for now:

* Env vars are stripped up front when `envVars: false`. The route ends up emptying them anyway, it just does the decrypt first, so this is the same request without the 500.
* A snapshot carrying encrypted material (hidden env `$`, or `flows.credentials.$` when credentials aren't excluded) with no `credentialSecret` is rejected with a clear 400 before the route is called.
* The description now says when the secret is genuinely needed, and warns that a wrong one can't be detected without flow credentials.

Tests cover the strip, both rejection paths, and the two cases that should still go through.

Guarding the decrypt loop in `uploadSnapshot` itself (and running it on the filtered copy) is the real fix, might be worth a separate issue.

cstns:
Raised the endpoint-side findings from testing these tools as separate issues, since the tools here only work around them rather than fix them:

* #8568 import 500s when `settings.env` is missing entirely, when an env value is null, and when a snapshot carries encrypted hidden env values without a secret. The raw Node error text goes back to the caller too. `platform_import_snapshot` sidesteps all three.
* #8569 import stores unencrypted flow credentials verbatim. Confirmed by importing a plaintext credentials object, then re-exporting and decrypting it back out intact. The tool rejects those up front now.
* #8570 import asks for a credential secret it never uses when `components.flows` is false, since the controller drops the credentials anyway. The tool strips them first so the call goes through.
* #8571 export always emits an encrypted credentials block, even for a snapshot with no flows and no credentials, which is why a secret is in practice always needed on import. Both tool descriptions say so now.

Also #8572 for the gateway dropping the status, `code` and `isError` off tool results. Relevant here because it's what makes the careful 400s in these tools indistinguishable from a 500 on the agent's side.

cstns:
Two more from testing `platform_export_snapshot`, neither fixable from this PR:

* #8574 `snapshot:export` only needs Member, while `snapshot:edit`, `snapshot:delete` and `snapshot:import` all need Owner. Export is the one that hands back every flow credential and every hidden env value in a form the caller can decrypt, so the ordering looks worth a second opinion. Possibly deliberate.
* #8575 the gateway doesn't enforce the whole published input schema. `components: {envVars: true}` is illegal per the schema but gets through and comes back from the route as a bare `Bad Request` naming no field, while an unknown key is caught cleanly. Means a tool can't lean on its own schema for validation.

The tool-side findings from the same pass are fixed in 4a587ea:

* the line claiming an export always carries a credentials block was only half right, with `credentials: false` or `flows: false` it carries `{}` and the import needs no secret at all
* a blank `credentialSecret` now fails in the tool instead of costing a round trip to the route, matching what the update tool already does for a blank name
* `envVars: "keys"` drops the hidden flag, so a secret variable comes back as an ordinary empty one. Noted on the shared components schema, so it covers the import tool too
* carried over the payload-size caution from `platform_get_snapshot_full`, since an export is always a superset of it

Everything else in the export tool checked out: the unconditional `credentialSecret`, every components combination, the re-encryption round-trip for both credentials and hidden env values, owner resolution for device-owned snapshots, and no leak of the snapshot's own stored secret.

cstns:
`platform_update_snapshot` pass. Both claims in the PR description check out against the raw route: a blank name really does 500 (`Snapshot name is required`, the unmapped sequelize `ValidationError`), whitespace-only counts as blank because the controller trims before validating, and an empty body really does answer 200 with the snapshot untouched. Partial updates, clearing a description with `""`, owner resolution for device-owned snapshots and the audit diffs all behave as described.

One new thing, fixed in 2666b0a: a name over 255 characters was going through to a 500 carrying the raw database error, since the column is `DataTypes.STRING`. The tool already shielded the blank end of that check, so this is just the other end. Raised the endpoint side as #8576.

Only bites on postgres, sqlite ignores the declared width, so the guard sits in the handler as well as the schema rather than leaning on a test that would pass on the default dev database regardless.

Also corrected #8575 while I was there. I'd guessed the gateway might be skipping other schema keywords too, but `minLength` turns out to be enforced properly (`name: ""` here is caught with a precise message), so the gap really does look specific to `const`.

cstns:
`platform_set_instance_device_target` pass, which completes all four tools in this PR.

Tested against a throwaway application and instance with two devices assigned to it, so the deploy path ran for real. Everything in the description checks out:

* the destructive classification is wired correctly, `invoke_write_tool` refuses it with `is a delete-class tool; call invoke_delete_tool instead`
* the CAUTION is accurate rather than defensive. One call and both assigned devices carried the new `targetSnapshot`; switching to another snapshot propagated to both again
* the silent no-op is real. The route with `{}` answers 200 with a zero-length body and changes nothing, so making `snapshotId` required is the right call
* snapshot ownership is enforced (`Invalid snapshot` for a snapshot belonging to another instance, and for one that doesn't exist), an unknown instance gives `Not Found`, and a non-UUID instance id is caught at the gateway

Two description fixes in 9bbd6e0:

* "This tool can only set a target, not clear one" read as a limitation of the tool, so an agent would go hunting for another way to clear it. There isn't one. The route ignores `{"targetSnapshot": null}` and answers 200 without changing anything. The only thing that actually clears a target is deleting the snapshot it points at, which I confirmed drops it from the instance and from every assigned device. Worth spelling out, since "delete the snapshot" is a surprising answer to "how do I undo this"
* the description told the caller to confirm before deploying to every assigned device but gave them nothing to confirm against. Setting a target on an instance with no devices returns the same `{"status": "okay"}` as on one with fifty. Now points at `platform_list_remote_instances` scoped by `hostedInstanceId`, which answers exactly which devices a call will hit

Didn't raise an endpoint issue for the missing clear, it looks deliberate rather than broken. Happy to raise it as a question if anyone disagrees.

Also added a third data point to #8575: `pattern` is enforced by the gateway too, so `const` really is the odd one out rather than part of a general gap.

andypalmi (approved):
The latest commits cover everything from the last pass: shared `components` schema, the blank-secret and name-length guards on both ends, and the device-target description now spelling out what it can and cannot undo. Splitting the endpoint and gateway findings into their own issues is the right call, since the tools only work around them here.

On #8574, that role ordering is pre-existing rather than something this PR introduces, so it should not hold this up. Fine to leave to triage on that issue and change the behaviour later if we decide to.

Approving.

PR #8498, opened 2026-09-14
Add MCP tools for team creation, updates, roles and invitations

### LINKED ISSUE #7694: 5.4-b Write tools, non-destructive (phase 2)
**Parent:** #7673 (5.4 Teams and membership)
**Tool file:** `forge/ee/lib/mcp/tools/teams.js` (extend), new `forge/ee/lib/mcp/tools/members.js`

|---|---|---|---|
| `platform_create_team` | `POST /teams` | `team:create` | write |
| `platform_update_team` | `PUT /teams/:teamId` (name/slug only) | `team:edit` | write |
| `platform_change_member_role` | `PUT /teams/:teamId/members/:userId` (role only) | `team:user:change-role` | write |
| `platform_invite_team_member` | `POST /teams/:teamId/invitations` | `team:user:invite` | write |
| `platform_resend_team_invitation` | `POST /teams/:teamId/invitations/:invitationId` | `team:user:invite` | write |

**Design notes:**
- **`platform_update_team` must expose only `name`/`slug`.** `PUT /teams/:teamId` has mutually-exclusive branches: `type` and `suspended` are lifecycle concerns, `features` is owner-restricted to `['ai']`, and **`properties` is admin-only** (tracked in 5.4-d). Do not expose those.
- **`platform_change_member_role` takes only `role`.** The route accepts `role` XOR `permissions`; the `permissions` (granular RBAC) path needs the `rbacApplication` feature. Leave granular permissions out of v1. Handle the last-owner-demotion 403 and the SSO-managed-user block as descriptive errors.
- `invite` and `resend` are kept as separate tools: invite creates a new invitation (`POST /invitations`), resend acts on an existing one (`POST /invitations/:invitationId`). They are not consolidated.

**Tool definitions (description + zod inputSchema; roles are numeric: 5=Dashboard, 10=Viewer, 30=Member, 50=Owner):**
```js
platform_create_team: {
  description: 'Create a new team. A slug is generated from the name when omitted.',
  inputSchema: z.object({
    name: z.string().describe('Display name for the new team'),
    type: z.string().describe('Team type hashid selecting the tier/plan for the team'),
    slug: z.string().regex(/^[a-z0-9-_]+$/i).optional().describe('Optional team slug (URL identifier; lowercase letters, digits, hyphen and underscore)'),
    trial: z.boolean().optional().describe('Request trial-mode setup (only honoured when billing is active and eligibility rules pass)'),
    billingInterval: z.enum(['month', 'year']).optional().describe('Billing cycle for the subscription checkout session')
  })
}
platform_update_team: {
  description: 'Update a team. Only name and slug are editable through this tool; type, suspended, features and properties are not exposed (properties is admin-only).',
  inputSchema: z.object({
    teamId: z.string().describe('Team hashid'),
    name: z.string().optional().describe('New display name for the team'),
    slug: z.string().regex(/^[a-z0-9-_]+$/i).optional().describe('New team slug (URL identifier; lowercase letters, digits, hyphen and underscore)')
  })
}
platform_change_member_role: {
  description: 'Change a team member role. Role only; the granular-permissions path is left out of v1. Demoting the last owner returns a descriptive 403, and editing an SSO-managed member returns a descriptive block.',
  inputSchema: z.object({
    teamId: z.string().describe('Team hashid'),
    userId: z.string().describe('User hashid of the member whose role is changing'),
    role: z.union([z.literal(5), z.literal(10), z.literal(30), z.literal(50)]).describe('New team role: 5=Dashboard, 10=Viewer, 30=Member, 50=Owner')
  })
}
platform_invite_team_member: {
  description: 'Invite one or more people to a team. The invited role is granted when the invitation is accepted, defaulting to 30=Member when omitted.',
  inputSchema: z.object({
    teamId: z.string().describe('Team hashid'),
    user: z.string().describe('Comma-separated list of usernames and/or email addresses to invite (maximum 5 entries per call after de-duplication)'),
    role: z.union([z.literal(5), z.literal(10), z.literal(30), z.literal(50)]).optional().describe('Team role: 5=Dashboard, 10=Viewer, 30=Member, 50=Owner')
  })
}
platform_resend_team_invitation: {
  description: 'Resend an existing team invitation. No request body.',
  inputSchema: z.object({
    teamId: z.string().describe('Team hashid'),
    invitationId: z.string().describe('Invitation hashid to resend')
  })
}
```

**Tests:**
- `update_team` payload cannot include `properties`/`type`/`suspended` (schema-restricted).
- `change_member_role` last-owner demotion returns the descriptive 403.
- SSO-managed member edit returns the descriptive block.

---

cstns:
Closes #7694

Adds five team and membership write tools:

* `platform_create_team` wraps `POST /api/v1/teams`. The description covers the things an agent can't guess: the team type hashid comes from `platform_list_team_types`, slug rules (`create` is reserved, uniqueness), the self-service team creation setting for non-admins, the strict trial eligibility rules, and the fact that billing platforms return a `billingURL` checkout link the user has to visit.
* `platform_update_team` wraps the name/slug branch of `PUT /api/v1/teams/:teamId` and only exposes those two fields. Type, suspended, features and properties stay out on purpose (the route treats them as mutually exclusive branches, and properties is admin-only).
* `platform_change_member_role` wraps `PUT /api/v1/teams/:teamId/members/:userId` with role only (the permissions branch needs the RBAC feature and stays out of v1). One reality worth noting from the route: failures like demoting the only owner come back as a generic 403 `invalid_request` with no detail, so the description explains the usual causes instead of promising a descriptive error. The SSO-managed-membership block does have a descriptive message and is quoted verbatim.
* `platform_invite_team_member` wraps `POST /api/v1/teams/:teamId/invitations`. The route reports per-person failures as HTTP 200 with `code: invitation_failed` and an error map, so the description tells agents to read the body rather than trust the status code. Max 5 per call (429 beyond that), role defaults to Member.
* `platform_resend_team_invitation` wraps `POST /api/v1/teams/:teamId/invitations/:invitationId` (no body, extends expiry, 404 for foreign/unknown invitations).


cstns:
Tested all five tools against a running platform. Most of the PR description holds up well, in particular the two calls that were clearly made from reading the routes rather than guessing:

* the `invitation_failed` warning is real. Per-person failures genuinely come back as HTTP 200 with an error map (`{"nobody@example.com": "External invites not permitted"}`, `{"definitely-not-a-user": "Not an existing user"}`, `{"cstns": "Already a member of the team"}`)
* the "generic 403 without detail" note on role changes is accurate. A non-member and a last-owner demotion both return an identical contentless `403 invalid_request` / `"Invalid request"`, so describing it honestly rather than promising a useful error was right

Also confirmed: reserved and duplicate slugs, `invalid_team_type`, trial rejection for an established user, a real Stripe checkout `billingURL` on create, `>5` invitees giving `429 too_many_invites`, resend extending the expiry (verified in mailpit, real mail goes out), and 404 for unknown or foreign invitations.

Four things changed in df928b7:

**Role changes are now delete-class.** Demoting a team's only owner is blocked for Member but not for Viewer or Dashboard, and those succeed and leave the team with zero owners. Proved it on a throwaway team: `{"role": 30}` gives 403, `{"role": 10}` gives 200, and afterwards I couldn't promote myself back, delete the team, rename it, or invite anyone. No non-admin recovery at all. Raised as #8580. Since one call can do that irreversibly, `destructiveHint: true` felt more honest than leaving it alongside ordinary writes, and the caution now says which demotions actually bite.

**`invitation_failed` has two shapes, not one.** Alongside the documented HTTP 200 with an error map, there's an HTTP 400 with `error` as a plain string when the whole call is rejected, which is what the team user limit returns. An agent following the old wording would iterate a string.

**The invite route has its own rate limit** of 5 calls per 30 seconds, which is a second 429 that means the opposite of `too_many_invites`: one clears by waiting, the other never will. Worth noting `platform_resend_team_invitation` already documented its limit, so the two are consistent now.

**An empty team update was a silent no-op**, answering 200 with the team unchanged. `platform_update_snapshot` in #8497 already rejects exactly this, so this one now does too.

Also noted on `platform_create_team`: with a team-scoped token the new team falls outside the scope, so the create succeeds and then reading, updating or inviting to it all fail, and it doesn't show up in `platform_list_teams`. Not a defect, but it's a trap worth naming since scoping an agent's token is the natural way to sandbox it.

One more from reading rather than running: #8581, `POST /api/v1/teams` replies 403 for a non-admin when self-service creation is off and then keeps executing.

andypalmi (approved):
Checked all five tools against the routes they wrap and everything holds up:

- required vs optional fields on each tool
- the reserved/duplicate slug and `invalid_team_type` handling on create
- the trial rules and `billingURL` on create
- the only-owner demotion asymmetry sitting behind `destructiveHint`
- both `invitation_failed` shapes, plus the separate rate-limit `429`
- the empty-update guard matching `platform_update_snapshot`

Descriptions are accurate, handlers forward what the routes expect, and the tests cover it.

The two route issues this surfaced (#8580 for the owner-orphaning demotion, #8581 for the create route continuing after a `403`) are real but correctly left as their own follow-ups rather than holding up this PR.

PR #8499, opened 2026-09-14
Add Claude Code and Codex to the AI connector modal and the docs

dimitrieh:
## What

Adds a **Claude Code** tab and a **Codex** tab to the AI connector modal, and reworks the matching docs section.

A coding agent installs the connector into itself, so each tab is one prompt and nothing else:

> Add the FlowFuse MCP tool at `<this platform>/mcp`. Then ask me to complete the sign-in in the browser that opens.

The prompt is built from the platform's own origin, the same source the address step already used, so it is correct on Cloud and self-hosted with nothing for the reader to edit.

Nothing here goes stale when a client changes how remote servers are added, which they do at different times and in different ways: Claude Code documents an `mcp add` subcommand, Codex documents only its config file and its UI.

### Modal changes

- Step one is overridable per client, so a tab can show a prompt instead of the bare address. Both go through the same copy path and the same PostHog event.
- The tab glyph is chosen per client through `<component :is>` rather than one hardcoded icon. `ServerIcon` and `CommandLineIcon` are values now, not registered components.

### Docs

`Command-line and editor agents` splits into a coding-agent section leading with the prompt, and a section for the rest. The hand-written `claude mcp add` command stays for anyone who wants it.

The widget on the docs page comes from the website, so its two new tabs arrive with https://github.com/FlowFuse/website/pull/5800.

**Conflicts with #8492** (the Gemini tab), which inserts into the same array. Whichever merges second needs a one-line rebase.

## Testing

Checked statically, not by running the app:

- [x] `CLIENTS` parses; every entry has a name, a glyph or logo, a step-two label and URL.
- [x] Both prompt builders return the passed endpoint verbatim, so a self-hosted origin survives into the copied text.
- [x] The doc's `#clients-without-a-sign-in-flow` anchor still resolves after the heading split.
- [ ] **Not verified:** the modal rendered in a running platform. Worth a click-through: six tabs, the two coding-agent tabs showing the prompt rather than the address, and Copy putting the whole sentence on the clipboard.


dimitrieh:
Updated: both new tabs now carry the vendor mark instead of the command-line glyph. Codex takes the OpenAI mark already on the ChatGPT tab, Claude Code takes the Anthropic mark already on the Claude tab.

With no glyph left but ServerIcon, the dynamic `<component :is>` and its registration change were no longer needed, so they are reverted. The diff is now the two new clients plus the overridable step one.

Rebased onto your merge of main.

dimitrieh:
@andypalmi @cstns https://github.com/FlowFuse/website/pull/5800 has been merged, can you look into this one?

cstns (approved):

PR #8500, opened 2026-09-14
Add MCP tools for pipeline creation, stages and deploys

### LINKED ISSUE #7697: 5.5-b Write tools, non-destructive (phase 2)
**Parent:** #7674 (5.5 Pipelines (DevOps))
**Tool file:** new `forge/ee/lib/mcp/tools/pipelines.js`

`readOnlyHint: false`, `destructiveHint: false`. Deploy is the CI/CD action; it changes state but does not overwrite or destroy existing content, so it lives with the ordinary writes.

|---|---|---|---|
| `platform_create_pipeline` | `POST /pipelines` | `pipeline:create` | write |
| `platform_update_pipeline` | `PUT /pipelines/:pipelineId` | `pipeline:edit` | write |
| `platform_add_pipeline_stage` | `POST /pipelines/:pipelineId/stages` | `pipeline:edit` | write |
| `platform_update_pipeline_stage` | `PUT /pipelines/:pipelineId/stages/:stageId` | `pipeline:edit` | write |
| `platform_deploy_pipeline_stage` | `PUT /pipelines/:pipelineId/stages/:stageId/deploy` | `pipeline:edit` | write |

**Design notes:**
- **`deploy` has a state-dependent Owner gate:** if the target instance has `KEY_PROTECTED` enabled and the caller is not Owner, the route returns 403 `protected_instance`. Surface this as an actionable message ("only a team Owner can deploy to a protected instance").
- Deploy is a no-op 200 when the stage action is `NONE` and there is no git repo; deploy behaviour branches by target type (instance/device/deviceGroup/gitRepo) with no further permission differences.
- Stage creation accepts action/git/device stage types under one `pipeline:edit` scope. The stage `action` enum values are `create_snapshot`, `use_active_snapshot`, `use_latest_snapshot`, `prompt`, `none` (from `PipelineStage.SNAPSHOT_ACTIONS`).
- **`update_pipeline` body shaping:** the tool exposes a flat `name`; its handler wraps it into the `{ pipeline: { name } }` shape the route actually reads. The route's declared schema is a flat `{ name }`, but the handler reads `request.body.pipeline.name`, so the nested shape is what works in practice (source inconsistency worth a follow-up fix on the route). Keep this translation in the handler so the agent-facing input stays flat.

```js
platform_create_pipeline: {
  description: 'Create a new pipeline in an application.',
  inputSchema: z.object({
    applicationId: z.string().describe('Application hashid the pipeline is created in'),
    name: z.string().describe('Name for the new pipeline (required)')
  })
}
platform_update_pipeline: {
  description: 'Rename a pipeline.',
  inputSchema: z.object({
    pipelineId: z.string().describe('Pipeline hashid to update'),
    name: z.string().describe('New pipeline name')
  })
}
platform_add_pipeline_stage: {
  description: 'Add a stage to a pipeline. Accepts action, git, and device stage types; supply only the fields relevant to the chosen stage type.',
  inputSchema: z.object({
    pipelineId: z.string().describe('Pipeline hashid to add the stage to'),
    name: z.string().optional().describe('Stage name'),
    instanceId: z.string().optional().describe('Hosted instance hashid bound to this stage'),
    deviceId: z.string().optional().describe('Remote instance/device hashid bound to this stage'),
    deviceGroupId: z.string().optional().describe('Device group hashid bound to this stage'),
    deployToDevices: z.boolean().optional().describe('Whether deploying the stage also pushes to its bound devices'),
    action: z.enum(['create_snapshot','use_active_snapshot','use_latest_snapshot','prompt','none']).optional().describe('Snapshot action for the stage (from PipelineStage.SNAPSHOT_ACTIONS)'),
    gitTokenId: z.string().optional().describe('Git token hashid for a git-repo stage'),
    url: z.string().optional().describe('Git repository URL for a git-repo stage'),
    branch: z.string().optional().describe('Git branch to push to'),
    pullBranch: z.string().optional().describe('Git branch to pull from'),
    pushPath: z.string().optional().describe('Repository path to push to'),
    pullPath: z.string().optional().describe('Repository path to pull from'),
    credentialSecret: z.string().optional().describe('Secret used to encrypt stored git credentials'),
    source: z.string().optional().describe('Stage hashid this stage sources its snapshot from')
  })
}
platform_update_pipeline_stage: {
  description: 'Update a pipeline stage. Accepts action, git, and device stage types; supply only the fields relevant to the chosen stage type.',
  inputSchema: z.object({
    pipelineId: z.string().describe('Pipeline hashid the stage belongs to'),
    stageId: z.string().describe('Pipeline stage hashid to update'),
    name: z.string().optional().describe('Stage name'),
    instanceId: z.string().optional().describe('Hosted instance hashid bound to this stage'),
    deviceId: z.string().optional().describe('Remote instance/device hashid bound to this stage'),
    deviceGroupId: z.string().optional().describe('Device group hashid bound to this stage'),
    deployToDevices: z.boolean().optional().describe('Whether deploying the stage also pushes to its bound devices'),
    action: z.enum(['create_snapshot','use_active_snapshot','use_latest_snapshot','prompt','none']).optional().describe('Snapshot action for the stage (from PipelineStage.SNAPSHOT_ACTIONS)'),
    gitTokenId: z.string().optional().describe('Git token hashid for a git-repo stage'),
    url: z.string().optional().describe('Git repository URL for a git-repo stage'),
    branch: z.string().optional().describe('Git branch to push to'),
    pullBranch: z.string().optional().describe('Git branch to pull from'),
    pushPath: z.string().optional().describe('Repository path to push to'),
    pullPath: z.string().optional().describe('Repository path to pull from'),
    credentialSecret: z.string().optional().describe('Secret used to encrypt stored git credentials'),
    source: z.string().optional().describe('Stage hashid this stage sources its snapshot from')
  })
}
platform_deploy_pipeline_stage: {
  description: 'Deploy a source stage to the next stage in the pipeline. If the target instance has KEY_PROTECTED enabled and the caller is not a team Owner, the deploy is rejected with 403 protected_instance (surface as "only a team Owner can deploy to a protected instance"). Deploy is a no-op 200 when the stage action is NONE and there is no git repo; behaviour otherwise branches by target type (instance/device/deviceGroup/gitRepo) with no further permission differences. The request body is optional/nullable; only send sourceSnapshotId when the stage action is prompt.',
  inputSchema: z.object({
    pipelineId: z.string().describe('Pipeline hashid the stage belongs to'),
    stageId: z.string().describe('Source pipeline stage hashid to deploy to the next stage'),
    sourceSnapshotId: z.string().optional().describe('Snapshot hashid to deploy for a prompt-action stage')
  })
}
```

**Tests:**
- Write and deploy tools rejected for read-only PAT.
- Deploy to a `KEY_PROTECTED` instance as a Member returns the descriptive 403.
- Deploy no-op path returns 200 without error.

---

cstns:
Closes #7697

Adds the five pipeline write tools: `platform_create_pipeline`, `platform_update_pipeline`, `platform_add_pipeline_stage`, `platform_update_pipeline_stage` and `platform_deploy_pipeline_stage`. With this, agents can finally run the deploy step of a pipeline.

A few route realities shaped the handlers and descriptions:

* `PUT /pipelines/:id` declares a flat `{ name }` body but its handler reads `request.body.pipeline.name`, so the tool keeps a flat agent-facing input and wraps it into the nested shape (the route's own tests use the nested shape too). Might be worth a follow-up fix on the route itself.
* Stage updates only apply git settings inside the rebinding branch, which is triggered by `instanceId`/`deviceId`/`deviceGroupId` being present in the body. The frontend gets this by always sending a null `deviceGroupId` (AJV coerces it to `''`); the tool handler mirrors that by adding a blank `deviceGroupId` when only git fields are sent, so a git-only update actually lands.
* `addGitRepo` resets omitted git fields to empty strings on update, except `credentialSecret` which is kept. The field descriptions tell agents to resend the full git settings on every git update.
* Deploy replies `{ status: "importing" }` as soon as the deploy has started and finishes in the background, so the description tells agents to poll the target status instead of treating the 200 as completion. The protected-instance Owner gate (403 `protected_instance`), the prompt-action `sourceSnapshotId`, and the action-none no-op are also spelled out.
* Stage ordering is a linked list: `source` is how a new stage is attached after an existing one, and the API's ordering rules (no device group first, no instance/device after a device group) are in the description since the errors are generic `invalid_input`.


andypalmi (commented):
Route behavior all checks out, and the tricky bits are handled well: the nested `pipeline.name` wrap, the blank `deviceGroupId` to reach the git rebind branch, and the deploy status semantics. Tests cover them nicely.

One optional cleanup, mostly on the two stage tools since they overlap a lot:

- The descriptions restate logic that already lives in the args (`action`, the "resend git settings as a set / reset to empty" note, `source`/linked-list). Could those move fully into the args that own them, keeping the description tool-level? Same idea we landed on for the update-application tool. I'm talking about this comment https://github.com/FlowFuse/flowfuse/pull/8496#discussion_r4007530503
- `platform_add_pipeline_stage` and `platform_update_pipeline_stage` duplicate the whole git field block, where the only real difference is the "reset to empty" note on update. Could it be a shared fragment in `schemas.js` (next to `teamId`/`applicationId`), spread into both? Something like:

```js
// schemas.js
function gitStageFields ({ resetNote = false } = {}) {
    const reset = resetNote ? '. Resent on every git update; reset to empty when omitted' : ''
    return {
        url: z.string().optional().describe('Git repository URL' + reset),
        branch: z.string().optional().describe('Git branch to push to' + reset),
        pullBranch: z.string().optional().describe('Git branch to pull from' + reset),
        pushPath: z.string().optional().describe('Repository path to push to' + reset),
        pullPath: z.string().optional().describe('Repository path to pull from' + reset),
        credentialSecret: z.string().optional().describe('Secret used to encrypt flow credentials pushed to the repository' + (resetNote ? '. Keeps its stored value when omitted' : ''))
    }
}
```

- The handler payload key list is duplicated between the two as well, so it could be a shared const.

None of this is blocking, happy for it to be a follow-up if you would rather keep this PR focused.

cstns:
Tested against a running platform with a throwaway application, two hosted instances, a device group and a git token.

The route reading behind this PR holds up well, and both handler workarounds turned out to be load-bearing rather than defensive:

* the nested-body wrap is required. The flat `{name}` body the route's own schema declares gives `500 TypeError: Cannot read properties of undefined (reading 'name')`; only `{pipeline:{name}}` works (#8583)
* the blank `deviceGroupId` on git-only updates is required. I ran identical payloads differing only by that key: without it the update returns 200 and changes nothing, with it the settings land (#8584)
* "omitted git fields are reset to empty, credentialSecret is the exception" is exactly right. A partial update left `url` changed and `branch`, `pullBranch`, `pushPath`, `pullPath` all `''`, credential secret intact

Deploy behaves as described throughout: `{"status":"importing"}` with genuine background completion (a Deploy Snapshot appeared on the target), `404 not_found` from the last stage, `action: none` giving `{"status":"okay"}` and doing nothing, and `action: prompt` giving `400 invalid_source_snapshot` without a `sourceSnapshotId` and deploying with one. Ordering and target rules all enforced as documented.

Two changes in 9415a58:

**Deploy is now delete-class.** It replaces the flows running on the next stage's target rather than adding to them, which is the same reasoning that made `platform_set_instance_device_target` destructive in #8497. Leaving it as a plain write next to renaming a pipeline felt inconsistent.

**The `source` warning was too soft.** "A stage added without source is not linked into the chain" undersells what actually happens. On a pipeline already containing a stage, adding one without `source` leaves two unlinked heads and the listing then returns only the new stage:

```
before: stages: ['GitStage']
after:  stages: ['NoSource']
```

`GitStage` still exists and fetches fine by id, it just stops appearing. So one missing argument reads as if it wiped the pipeline. Raised the listing behaviour as #8582.

One small correction to the PR description: it says the ordering rules are spelled out "since the errors are generic `invalid_input`". The `code` is generic but the `error` text is specific ("A Device Group cannot be the first stage", "An instance or device cannot be added after a device group"), and since the gateway drops `code` and surfaces `error` (#8572), an agent sees the clear message. Documenting the rules is still worth it, the reasoning just isn't quite right.

Everything above was exercised at the route level with payloads matching exactly what each handler sends; the tool-level pass is pending a gateway catalog refresh.

andypalmi:
The destructive reclassification and the sharper source warning both read well, and the ordering-error correction is fair.

Coming back to the two cleanups from the last pass, since this is the same shape as what we landed on #8497:

1. The stage descriptions still restate logic the args already carry (the `action` modes, the git reset-to-empty note, the `source` linked-list rule). Could those move fully into the arg `.describe()` that owns them, keeping the description tool-level, the way we did for the update-application tool?

2. `platform_add_pipeline_stage` and `platform_update_pipeline_stage` still duplicate the whole git field block, the only real difference being the reset-to-empty note on update. Could it become a shared fragment in `schemas.js` (next to `teamId`/`applicationId`), spread into both, with the handler payload key list shared too?

Can you fold these in before this one merges, the same way the shared `components` schema went into #8497?

cstns:
Both folded in, 44304c5e.

**Shared git block.** `gitStageFields` now lives in `schemas.js` next to the other shared pieces, spread into both stage tools. It takes a `forUpdate` flag because that is the only genuine difference between the two: on update the git settings are applied as a set so omitted fields empty out, with `credentialSecret` as the exception. Putting the flag in the fragment keeps that caveat on the arguments that own it rather than pushing it back into prose, which seemed truer to the first point than leaving a neutral fragment and re-stating the difference in the description.

`action` and `deployToDevices` were duplicated too, so they are declared once as well, and both handlers now build their payload from a shared `stagePayloadKeys` list (`[...stagePayloadKeys, 'source']` for add).

**Descriptions.** Moved onto the arguments that own them: the `action` modes, the git resend rule, the `source` linked-list rule, and the partial-update note. What is left is what each tool does plus the ordering rules, which I kept tool-level since they are cross-field invariants that no single argument owns, and you did not list them among the things to move. Shout if you would rather they hung off `deviceGroupId`.

One thing worth confirming: the `source` warning was deliberately in both places, because the failure mode is nasty (a missing `source` leaves the pipeline listing showing only the new stage, #8582). It now appears once, on the `source` argument, which is consistent with the principle but does drop the redundancy that was there on purpose. Happy either way, just flagging it rather than quietly halving it.

No behaviour change. I diffed the generated JSON Schema for both tools before and after: identical keys in the same order, same types, same `required` sets, same `additionalProperties`, with only the descriptions differing. 241 MCP tests still pass and lint is clean.

andypalmi (approved):
Cleanups look good. `gitStageFields` in `schemas.js` with the `forUpdate` flag is the right shape, and the shared `action` / `deployToDevices` / `stagePayloadKeys` deduplication matches what we did on #8497. Confirmed the generated schema is unchanged bar descriptions.

On the `source` warning: keep it in both places. The redundancy is deliberate given how the failure reads (a missing `source` looks like it wiped the pipeline, #8582), so I'd rather it stay slightly over-stated than have someone miss it.

Only thing left is the merge conflict with `main` in `schemas.js` (both this and #8497's `snapshotComponents` fragment landed in the same spot, purely additive). Resolve that against main and this is good to merge.

PR #8501, opened 2026-09-14
Add MCP tools for device group creation, updates, membership and settings

### LINKED ISSUE #7700: 5.6-b Write tools, non-destructive (phase 2)
**Parent:** #7675 (5.6 Device groups (fleet))
**Tool file:** new `forge/ee/lib/mcp/tools/deviceGroups.js`

`readOnlyHint: false`, `destructiveHint: false`. Creation, edits, membership changes, and settings are distinct endpoints and scopes, so they stay separate tools.

|---|---|---|---|
| `platform_create_device_group` | `POST /applications/:applicationId/device-groups` | `application:device-group:create` | write |
| `platform_update_device_group` | `PUT /applications/:applicationId/device-groups/:groupId` | `application:device-group:update` | write |
| `platform_update_device_group_membership` | `PATCH /applications/:applicationId/device-groups/:groupId` | `application:device-group:membership:update` | write |
| `platform_update_device_group_settings` | `PUT /applications/:applicationId/device-groups/:groupId/settings` | `application:device-group:update` | write |

**Design notes:**
- All routes require the `deviceGroups` feature (404 otherwise). Surface as a descriptive "device groups not enabled for this team" error.
- `update_device_group` accepts `name`, `description`, and `targetSnapshotId` (string or null) - setting `targetSnapshotId` is how a group's deployed snapshot is pinned/cleared.
- `PUT /:groupId/settings` carries a Member-only `{ env }` field restriction in the handler, but its scope is `application:device-group:update` (Owner), so under standard team RBAC only Owners reach it and they may send the full settings body (the declared schema only names `env`). The Member-only branch is reachable solely via app-specific granted permissions, not ordinary roles.

```js
platform_create_device_group: {
  description: 'Create a device group in an application.',
  inputSchema: z.object({
    applicationId: z.string().describe('Application hashid to create the group in'),
    name: z.string().describe('Name for the new device group (required)'),
    description: z.string().optional().describe('Optional description for the group')
  })
}
platform_update_device_group: {
  description: 'Update a device group. Setting targetSnapshotId pins the group deployed snapshot; passing null clears it.',
  inputSchema: z.object({
    applicationId: z.string().describe('Application hashid the group belongs to'),
    groupId: z.string().describe('Device group hashid to update'),
    name: z.string().optional().describe('New group name'),
    description: z.string().optional().describe('New group description'),
    targetSnapshotId: z.string().nullable().optional().describe('Snapshot hashid to pin as the group target, or null to clear it')
  })
}
platform_update_device_group_membership: {
  description: 'Change the membership of a device group. Combine add/remove for incremental changes, or use set to replace the entire membership.',
  inputSchema: z.object({
    applicationId: z.string().describe('Application hashid the group belongs to'),
    groupId: z.string().describe('Device group hashid whose membership to change'),
    add: z.array(z.string()).optional().describe('Device hashids to add to the group'),
    remove: z.array(z.string()).optional().describe('Device hashids to remove from the group'),
    set: z.array(z.string()).optional().describe('Device hashids to set as the exact group membership')
  })
}
platform_update_device_group_settings: {
  description: 'Update a device group settings. The declared schema names only env; under standard team RBAC only Owners can update device group settings, and only Owners may send settings fields other than env.',
  inputSchema: z.object({
    applicationId: z.string().describe('Application hashid the group belongs to'),
    groupId: z.string().describe('Device group hashid whose settings to update'),
    env: z.array(z.record(z.any())).optional().describe('Environment variable objects for the group')
  })
}
```

**Tests:**
- Feature-disabled team returns the descriptive gate error.
- `update_device_group_settings` write is accepted for an Owner PAT.

---

cstns:
Closes #7700

Adds the four device group write tools: `platform_create_device_group`, `platform_update_device_group`, `platform_update_device_group_membership` and `platform_update_device_group_settings`.

The descriptions lean on what the routes and `DeviceGroup` controller actually do, because most of these writes have immediate fleet-wide effects an agent must warn the user about:

* Pinning `targetSnapshotId` on a group applies it to every member device and tells them to deploy right away; `null` clears the pin from the group and its devices. The snapshot must belong to the same application (controller check, 400 otherwise).
* Membership changes are transactional (one bad device id fails the whole call) and also deploy: added devices pick up the group's pinned snapshot, removed devices have it cleared. `set` is atomic replace and overrides `add`/`remove`.
* The settings route replaces the full env list and 500s if `env` is missing, so the tool makes `env` required and documents the full-replacement semantics, the name rules, and the hidden-value convention (resend `hidden: true` with an empty value to keep a secret you can't read).
* All routes 404 when the team's `deviceGroups` feature is off, indistinguishable from a missing application, so each description says so.


cstns:
Tested against a running platform with a throwaway application, an instance, a snapshot, two devices and several groups.

The descriptions in this one are the most accurate of the set so far. Everything I could check held up:

* `env` really does 500 when missing (`Cannot read properties of undefined (reading 'reduce')`), so making it required is right
* the hidden-value convention works exactly as written. I checked at the storage layer, not just the API view: resending `{"name":"SECRET","value":"","hidden":true}` kept `s3cret` in the database while the API masks it as `""`, and an omitted variable was dropped in the same call
* name validation and duplicate rejection are descriptive: `Invalid Env Var name '9BAD'`, `Duplicate Env Var names provided`
* membership really is transactional: a device from another application and a nonexistent id both give a 400 with nothing changed
* `set` overrides `add`/`remove`, `targetSnapshotId: null` clears the pin, a foreign snapshot gives `Snapshot does not belong to the same application`, and updates return `{}`

Four changes in e6020e5:

**`description` now defaults to `''` instead of being omitted.** This is the significant one. Creating a group without a description stores a null, which fails the route's own response schema: the create returns 500 having committed the group anyway, and from then on both `GET /applications/:id/device-groups` and `GET /teams/:id/device-groups` 500 on that row. Two creates during testing took out the device group listings for a whole team, with an error that just names an array index. Raised as #8592; the tool now can't produce such a row.

**Three writes are now delete-class.** Pinning `targetSnapshotId` deploys to every member device (the same action as `platform_set_instance_device_target`), `set` replaces the whole membership and clears the pinned snapshot from dropped devices, and the settings `env` list is a full replacement that deletes anything omitted.

**Corrected the add/remove wording.** "A device in both add and remove ends up removed" is only half true, it toggles against current membership:

| device state | in both lists | result |
|---|---|---|
| already a member | | removed |
| not a member | | added |

So using both lists to force a device out does the opposite when it wasn't there.

**Empty bodies rejected.** `{}` on both the update and membership routes answers 200 with nothing changed, same as the snapshot and team update tools now guard against.

Also worth knowing, though I left it alone: device group names aren't unique within an application.

All of the above was exercised at the route level with payloads matching what each handler sends; the tool-level pass is pending a gateway catalog refresh.

forge/ee/lib/mcp/tools/deviceGroups.js
```diff
@@ -65,5 +65,123 @@ module.exports = [
...
+            applicationId: applicationId.describe('The hashid of the application to create the device group in'),
+            name: z.string().min(1).describe('Name for the new device group'),
+            description: z.string().optional().describe('Description for the group. Defaults to an empty string rather than being left unset, because a group stored without one breaks the device group listings')
+        },
+        handler: async (args, { inject }) => {
+            // A group created without a description is stored with a null one, which
+            // fails the route's own response schema: the create answers 500 having
+            // created the group anyway, and every later listing of the application's
+            // and the team's groups then 500s on that row too. Default it instead.
+            const payload = { name: args.name, description: args.description ?? '' }
```
andypalmi, line 88: Optional simplification: fold the empty-string default into the schema and drop the handler fallback, so the default lives in one place and clients see it advertised. `applyInputDefaults` fills the schema default into the args before the handler runs, and the route rejects a wrong-type description on its own, so nothing else needs the guard.

```suggestion
            description: z.string().default('').optional().describe('Description for the group; defaults to an empty string, because a group stored without one breaks the device group listings')
        },
        handler: async (args, { inject }) => {
            // A group created without a description stores null and 500s the listings (#8592).
            const payload = { name: args.name, description: args.description }
```

andypalmi (approved):
Checked the four tools against the routes and the `DeviceGroup` controller: required vs optional fields, the empty-body `400` guards on update and membership, `targetSnapshotId` deploy-on-pin and clear-on-`null` with the same-application check, the transactional membership semantics (`set` overriding add/remove, one foreign device failing the whole call), and the settings full-`env`-replacement rules all match. Descriptions accurate, handlers forward correctly, tests cover it.

#8592 (a null or unset description `500`ing the group listings) is correctly left as a follow-up. The create tool defaults the description so it can't produce that row.

PR #8502, opened 2026-09-14
Add MCP tools for notification read state, invitations and profile

### LINKED ISSUE #7703: 5.9-b Write tools, non-destructive (phase 2)

cstns:
Closes #7703

Adds the three self-service write tools:

* `platform_set_notification_read_state` consolidates the single (`PUT /user/notifications/:id`) and bulk (`PUT /user/notifications`) routes behind one tool: pass `notificationId` for one, `ids` for several, exactly one of the two. The description points out there is no mark-all shortcut, that unknown ids in a bulk call are silently skipped (the model filters them), and that the two routes reply with different shapes (`{status: okay}` vs the refreshed list).
* `platform_respond_to_team_invitation` maps `accept` to `PATCH /user/invitations/:id` and `reject` to `DELETE`, and explains it only works on invitations addressed to the caller (404 otherwise).
* `platform_update_profile` exposes only `name` and `defaultTeam` from `PUT /user` (username/email/tcs stay out on purpose). Route realities in the description: an empty name resets the display name to the username, and `defaultTeam` must be a team the user belongs to or the route replies `invalid_team`.


cstns:
Tested against a running platform, on my own account, restoring everything afterwards.

Every claim in the description checks out:

* **notifications, single**: `{status: "okay"}`, unknown id gives a 404
* **notifications, bulk**: replies with the refreshed list, genuinely a different shape from the single call. Unknown ids really are silently skipped, and a real id in the same call still applied. All-unknown ids return 200 having changed nothing
* **profile**: an empty `name` really does reset the display name to the username (`'Costin Serban'` became `'cstns'`), and an unknown `defaultTeam` gives `invalid_team`
* **invitations**: an invitation addressed to somebody else returns 404 for both accept and reject, and is not consumed, it was still pending afterwards. Nonexistent ids likewise

Three changes in 0f3c20e:

**Empty profile update rejected.** `PUT /user` with `{}` answers 200 with the profile unchanged, the same silent no-op the snapshot, team and device group update tools already guard against.

**Empty `ids` array rejected.** `ids` had no `.min(1)`, so `ids: []` passed the exactly-one check and reached the bulk route, which answers 200 having done nothing.

**Rejecting an invitation is now delete-class.** It deletes the invitation outright and only the original sender can issue another.

Two things I could not cover:

* accepting or rejecting an invitation **addressed to me**. That needs a second user's credentials so somebody can invite me; the 404 paths are covered but the happy path is not
* profile updates through a PAT, for reasons that are out of scope for this PR

Note the branch had main merged into it while I was working, so my commit is rebased on top of 144d996c rather than the original base. 267 MCP tests pass and lint is clean on the rebased tree.

andypalmi (approved):

Issue #7513, opened 2026-06-15
Scoped PATs - Block Broker Credential Issuance for PAT-Authenticated Requests


cstns:
### Summary

  Prevent PAT-authenticated requests from obtaining MQTT broker credentials. This is a security hardening measure that closes a privilege escalation path.

  ### Why

  Several API endpoints issue MQTT broker credentials that let a client connect to the internal message broker and interact with platform resources (device logs, team notifications, expert assistant chat). Once issued, these MQTT connections operate entirely outside the HTTP request lifecycle. All subsequent authorization is handled by the MQTT ACL layer (`forge/comms/aclManager.js`), which loads team membership fresh from the database and calls `hasPermission`.

  The problem is that the MQTT ACL layer runs outside any Fastify request context. `@fastify/request-context` has no store, so PAT restrictions (team scope, read-only, admin opt-out) are completely invisible to it. A user with a scoped PAT (e.g. read-only, Team A only) could call a credential endpoint, obtain unrestricted (by PAT scopes) MQTT credentials, and then use the broker to access resources in Team B or execute write operations via the expert assistant, fully bypassing the token's intended restrictions.

  This is a blanket block on all PAT-authenticated requests, not just scoped ones. Even an unscoped PAT should not be able to obtain broker credentials, because the MQTT layer has no mechanism to enforce any token-level restrictions and the credentials outlive the originating HTTP request. These endpoints exist to support browser-based UI features (device log streaming, team notifications, expert chat) that use cookie sessions. MCP agents use HTTP endpoints and do not need MQTT credentials.

  ### What to do

  Add a check to each credential-issuing endpoint that rejects the request if `request.session.pat` is present. Consider extracting this as a shared preHandler to avoid repetition.

  **Endpoints to gate:**

  | Endpoint | File |
  |---|---|
  | `POST /api/v1/user/expert-creds` | `forge/routes/api/user.js` |
  | `POST /api/v1/devices/:deviceId/logs` | `forge/routes/api/device.js` |
  | `POST /api/v1/devices/:deviceId/resources` | `forge/routes/api/device.js` |
  | `POST /api/v1/teams/:teamId/comms-credentials` | `forge/routes/api/team.js` |
  | `POST /api/v1/admin/expert-agent-creds` | `forge/routes/api/admin.js` |

  ### Tests

  * Each credential endpoint returns 403 when called with a PAT
  * Each credential endpoint continues to work when called with a cookie session
[...]

PR #8503, opened 2026-09-14
Add MCP tools for broker agent lifecycle and topic writes

### LINKED ISSUE #7712: 5.13-b Write tools, non-destructive (phase 2)
**Parent:** #7682 (5.13 Broker and MQTT (special review))
**Tool file:** new `forge/ee/lib/mcp/tools/broker.js`

`readOnlyHint: false`, `destructiveHint: false`. **Contingent on a product decision to allow broker mutation over MCP.** None of these mint credentials, so #7513 does not block them. Sibling broker lifecycle transitions are consolidated into a single action tool; topic create/update touch broker state but not credentials.

|---|---|---|---|
| `platform_broker_lifecycle_action` (`action`: `start`\|`stop`\|`suspend`) | broker lifecycle routes | `broker:credentials:edit` | write |
| `platform_create_broker_topic` | `POST /teams/:teamId/brokers/:brokerId/topics` | `broker:topics:write` | write |
| `platform_update_broker_topic` | `PUT /teams/:teamId/brokers/:brokerId/topics/:topicId` | `broker:topics:write` | write |

**Design notes:**
- Only the topic **create** route (`POST /topics`) is dual-gated (agent/broker token scope OR user `broker:topics:write`). The **update** route (`PUT /topics/:topicId`) uses a plain `needsPermission('broker:topics:write')` and is not dual-gated; its path is a single `:topicId` param, not a wildcard. The handler only updates `metadata`.
- Broker lifecycle (`start`/`stop`/`suspend`) maps to `PUT /brokers/:brokerId/start|stop|suspend`; it touches broker state but does not mint credentials, so it is not in the #7513 class; it remains contingent on the broker-mutation decision.
- The whole surface is gated by the enterprise license tier and the per-team `teamBroker` feature flag.

```js
platform_broker_lifecycle_action: {
  description: 'Applies a lifecycle transition (start, stop, or suspend) to the broker agent.',
  inputSchema: z.object({
    teamId: z.string().describe('team hashid'),
    brokerId: z.string().describe("broker id: either the literal 'team-broker' or a 3rd-party broker hashid"),
    action: z.enum(['start', 'stop', 'suspend']).describe('lifecycle transition to apply to the broker agent')
  })
},
platform_create_broker_topic: {
  description: 'Creates one or more topics for a broker. Accepts a single topic object instead of an array.',
  inputSchema: z.object({
    teamId: z.string().describe('team hashid'),
    brokerId: z.string().describe("broker id: either the literal 'team-broker' or a 3rd-party broker hashid"),
    topics: z.array(z.object({
      topic: z.string().describe('MQTT topic string'),
      type: z.any().optional().describe('inferred payload schema/type; stored as inferredSchema'),
      metadata: z.record(z.any()).optional().describe('arbitrary topic metadata (e.g. description, schema)')
    })).describe('Topics to store')
  })
},
platform_update_broker_topic: {
  description: 'Updates a broker topic. Only the topic metadata is updated.',
  inputSchema: z.object({
    teamId: z.string().describe('team hashid'),
    brokerId: z.string().describe("broker id: either the literal 'team-broker' or a 3rd-party broker hashid"),
    topicId: z.string().describe('topic hashid'),
    metadata: z.record(z.any()).describe('replacement topic metadata')
  })
}
```

**Tests (only if writes approved):**
- The action tool routes each enum value to the correct lifecycle route.
- Read-only PAT rejected for every tool here.
- Feature-disabled team returns the descriptive gate error.

---

cstns:
Closes #7712

Adds `platform_broker_lifecycle_action`, `platform_create_broker_topic` and `platform_update_broker_topic`. No credential-minting surface is touched, matching the exclusions in #7513.

The lifecycle routes turned out to be subtler than the issue sketch, and the tool descriptions reflect what the routes actually do:

* start/stop/suspend control the broker's topic-collection agent, not the MQTT broker service, so the description says message traffic is unaffected. They're `POST` routes (the sketch said PUT).
* Two team-broker paths never reply at all: starting an agent that's already running, and stopping one that isn't running. Rather than have agents hang on that, the description tells them to check `platform_get_broker` first. Might be worth fixing those no-reply branches on the route.
* Suspending the team broker destroys its agent record entirely, after which `platform_get_broker` reports just `{ state: "suspended" }`, which the read tool already documents.
* Topic creation is fire-and-forget on the route: always an empty 201, per-entry errors swallowed, and a server-side LRU cache can silently skip re-writes of a recently written topic. The description tells agents to verify with the list tool and to use the update tool for metadata changes instead of re-creating.
* Topic update only touches `metadata` and replaces the stored object wholesale, so that's spelled out too.


andypalmi (commented):
All three tools line up with the routes, and the descriptions are honest about what the routes actually do (POST-not-PUT verbs, suspend destroying the team-broker agent, the fire-and-forget 201 with the LRU cache, wholesale metadata replacement). Tests and annotations look right.

One thing to track separately: the `team-broker` no-reply branches you flagged (starting an already-running agent, stopping one that is not) are reachable directly through the tool, and `inject` just hangs with no timeout. The "check `platform_get_broker` first" note mitigates it, but the real fix is a couple of `reply.send` calls on the route. Worth a follow-up issue rather than this PR. Not blocking, the tools themselves are good to go.

Should we hold off on these tools until we implement the fix?

andypalmi:
The no-reply route branches are fixed in #8514 (with test coverage for the previously-hanging transitions), so this can stay focused on the tools.

andypalmi:
Fix from #8514 merged

forge/ee/lib/mcp/tools/broker.js
```diff
@@ -119,5 +119,65 @@ module.exports = [
             const response = await inject({ method: 'GET', url: `/api/v1/teams/${args.teamId}/broker/${args.brokerId}/schema` })
             return response
         }
+    },
+    {
+        name: 'platform_broker_lifecycle_action',
+        title: 'Broker Lifecycle Action',
+        description: `FlowFuse platform automation tool:
```
ZJvandeWeg, line 126: I wonder why we always start this way?

cstns: It’s important for agents to understand which category tools belongs to (platform, UI, or flow-building), as each category is executed and sourced differently

cstns:
Tested against a running platform, driving the team broker through start, stop and suspend cycles with real `@flowfuse/mqtt-schema-agent` processes, and exercising the topic tools through the gateway.

Both topic tools are accurate in every detail I could check:

* create takes a bare array body, always returns an empty `{}` 201, silently skips an entry with no `topic` string, and **ignores** a re-create of an existing topic. I resent `mcp/test/alpha` with `description: "CHANGED"` and the stored metadata stayed `"first"`, exactly as the description warns. Pointing agents at the update tool instead is the right call
* update returns the updated topic, replaces `metadata` wholesale (two fields down to one dropped the other), clears it on `{}`, and 404s on an unknown id
* suspend really does tear the team-broker agent down: the row disappears and `platform_get_broker` reverts to `{ state: "suspended" }`

Two corrections to the lifecycle description, in 7be4a15:

**"Gets no response from the API" is not what happens.** Those branches answer **200 with an empty body**, having done nothing. That is worse for an agent than a hang, because it looks like success. Raised as #8617, which also covers `stop` never persisting a stopped state.

**The recommended pre-check cannot work.** The description said to check `platform_get_broker` first, but for the team broker that route hardcodes `status = { connected: true, error: null }` (`3rdPartyBroker.js:354`) while the 3rd-party branch thirteen lines earlier calls `getBrokerAgentState` properly. With a real agent running I got identical output before and after a stop: `state: "running"`, `connected: true`. So the check always says running. Raised as #8618. The description now says only the presence of the agent is observable there, and that start/stop results should be treated as unverified.

Also flipped `platform_broker_lifecycle_action` to `destructiveHint: true`, since suspend removes the agent record entirely.

One more found along the way, #8619: after a suspend I was left with an orphaned `mqtt-schema-agent` process while `TeamBrokerAgents` was empty. `stopBrokerAgent` kills by `settings.pid` behind an `if`, and the route destroys the row 1.5s later, so a missing pid leaks the process with nothing left to find it by.

Correction to something I said earlier in this PR's testing: I initially wrote that the localfs driver has no broker agent support. That was wrong, it implements the full set in `localfs.js:899-947` and spawns the agent as a child process. The findings above do not depend on the driver.

andypalmi (approved):
Checked the three tools against the broker routes in `3rdPartyBroker.js`:

- Lifecycle start/stop/suspend map correctly. The "no-op transitions answer 200" claim holds now that #8514 landed and gave the already-running-start and not-running-stop branches their `200 {}` reply, so nothing hangs.
- `destructiveHint: true` is right, since suspend destroys the team-broker agent.
- The create tool wraps the topics array and always 201s; the `topics` min length and the per-entry topic-string skip both match the route.
- The update tool does a full `metadata` replace and 404s on a missing topic, which the description spells out.
- Annotations and tests match.

Approving.

PR #8504, opened 2026-09-14
Add MCP tools for instance lifecycle, settings, config surfaces and files

### LINKED ISSUE #7685: 5.1-b Write tools, non-destructive (phase 2)
**Parent:** #7670 (5.1 Instance lifecycle and actions)
**Tool file:** `forge/ee/lib/mcp/tools/instances.js` (extend), `forge/ee/lib/mcp/tools/devices.js` (extend)

`readOnlyHint: false`, `destructiveHint: false`. Sibling actions that share one concept are consolidated into a single tool with an `action` enum, so one tool covers several state transitions.

Instances:

|---|---|---|---|
| `platform_instance_action` (`instanceType`: `hosted`\|`remote`; `action`: `start`\|`stop`\|`restart`\|`suspend`\|`restartStack`) | hosted: `POST /projects/:id/actions/{action}`; remote: `POST /devices/:id/actions/restart` | `project:change-status` (hosted), `device:change-status` (remote) | write |

Hosted instances:

|---|---|---|---|
| `platform_update_hosted_instance_env` | `PUT /projects/:id` (body `{settings:{env}}` only) | `project:edit-env` | write |
| `platform_update_hosted_instance_settings` | `PUT /projects/:id` (full settings) | `project:edit` | write |
| `platform_import_hosted_instance_flows` | `POST /projects/:id/import` | `project:edit` | write |

Remote instances (devices):

|---|---|---|---|
| `platform_update_remote_instance_settings` | `PUT /devices/:id/settings` | `device:edit-env` | write |
| `platform_set_remote_instance_mode` | `PUT /devices/:id/mode` | `device:editor` | write |

**Consolidation notes:**
- `platform_instance_action` covers both hosted and remote instances through an `instanceType` discriminator, mirroring the read tools. Hosted instances accept the full action set; remote instances (devices) expose only `restart`, and the handler rejects any other action for `instanceType: 'remote'` with a descriptive error. It routes to `POST /projects/:id/actions/{action}` (`project:change-status`) or `POST /devices/:id/actions/restart` (`device:change-status`) based on `instanceType`.

**Design notes:**
- **Split env vs full settings** for hosted instances: `PUT /projects/:id` requires `project:edit-env` only when the body is exactly `{settings:{env}}`, else `project:edit` (and the handler narrows env-only writes). Two distinct tools keep annotations and scopes unambiguous. The env tool exposes a flat `env` list; its handler wraps it into exactly `{settings:{env}}` so the narrower edit-env path is selected, and the agent never sees the wrapper.
- **State guards:** start/stop/restart return 400 when the instance is suspended. Surface these as descriptive errors.
- `platform_set_remote_instance_mode` is license-gated (returns 400 `not_licensed` without an active license) and only accepts `autonomous`/`developer`.
- `update_remote_instance_settings` is owner-vs-member field-restricted: Members may write only `env`/`autoSnapshot`. Document this.

```js
platform_instance_action: {
  description: 'Applies a lifecycle action to an instance. instanceType selects the target: hosted instances accept start/stop/restart/suspend/restartStack (start/stop/restart return 400 when the instance is suspended); remote instances (devices) accept only restart. Takes no request body.',
  inputSchema: z.object({
    instanceId: z.string().describe('Instance id: hosted instance UUID, or remote instance (device) hashid'),
    instanceType: z.enum(['hosted', 'remote']).describe('Whether instanceId refers to a hosted instance or a remote instance (device)'),
    action: z.enum(['start', 'stop', 'restart', 'suspend', 'restartStack']).describe('Lifecycle action to apply; remote instances accept only restart')
  })
}

platform_update_hosted_instance_env: {
  description: 'Replaces the full set of environment variables on a hosted instance.',
  inputSchema: z.object({
    instanceId: z.string().uuid().describe('UUID of the hosted instance whose env is updated'),
    env: z.array(z.object({
      name: z.string().describe('Environment variable name'),
      value: z.string().describe('Environment variable value'),
      hidden: z.boolean().optional().describe('Whether the value is masked in the UI')
    })).describe('Full replacement list of instance environment variables')
  })
}

platform_update_hosted_instance_settings: {
  description: 'Updates a hosted instance: display name, hostname, general settings, launcher settings, instance type or stack, or copies configuration and flows from a source instance.',
  inputSchema: z.object({
    instanceId: z.string().uuid().describe('UUID of the hosted instance to update'),
    name: z.string().optional().describe('New display name for the instance'),
    hostname: z.string().optional().describe('New hostname for the instance'),
    settings: z.record(z.any()).optional().describe('Instance settings object (env, palette, editor, and so on)'),
    launcherSettings: z.object({
      healthCheckInterval: z.number().optional().describe('Launcher health-check interval in milliseconds'),
      disableAutoSafeMode: z.boolean().optional().describe('Whether to disable automatic safe mode after repeated crashes')
    }).optional().describe('Node-RED launcher settings'),
    projectType: z.string().optional().describe('Opaque hashid of the instance type to switch to'),
    stack: z.string().optional().describe('Opaque hashid of the stack to switch to'),
    sourceProject: z.object({
      id: z.string().uuid().describe('UUID of the source instance to copy from'),
      options: z.record(z.any()).optional().describe('Flags selecting which parts to copy (flows, credentials, env, and so on)')
    }).optional().describe('Source instance to copy configuration and flows from')
  })
}

platform_import_hosted_instance_flows: {
  description: 'Imports flows into a hosted instance, optionally with credentials and the secret needed to decrypt them.',
  inputSchema: z.object({
    instanceId: z.string().uuid().describe('UUID of the hosted instance to import flows into'),
    flows: z.string().optional().describe('Flows JSON serialized as a string'),
    credentials: z.string().optional().describe('Flow credentials JSON serialized as a string'),
    credsSecret: z.string().optional().describe('Secret used to decrypt the supplied credentials')
  })
}

platform_update_remote_instance_settings: {
  description: 'Updates a remote instance (device) settings. Field-restricted by role: Members may write only env and autoSnapshot, while palette, editor, and security require owner role.',
  inputSchema: z.object({
    deviceId: z.string().describe('Opaque hashid of the remote instance (device) to update'),
    env: z.array(z.record(z.any())).optional().describe('Environment variables as {name, value, hidden?} entries'),
    autoSnapshot: z.boolean().optional().describe('Whether to auto-create a snapshot on each deploy'),
    palette: z.record(z.any()).optional().describe('Palette settings'),
    editor: z.record(z.any()).optional().describe('Editor settings'),
    security: z.record(z.any()).optional().describe('Security settings such as httpNodeAuth or localAuth')
  })
}

platform_set_remote_instance_mode: {
  description: 'Sets the operating mode of a remote instance (device). License-gated: returns 400 not_licensed without an active license. When mode is omitted or null it defaults to autonomous.',
  inputSchema: z.object({
    deviceId: z.string().describe('Opaque hashid of the remote instance (device)'),
    mode: z.enum(['autonomous', 'developer']).optional().describe('Device operating mode')
  })
}
```

**Tests:**
- `platform_instance_action` routes each hosted enum value to the correct `POST /projects/:id/actions/{action}` route, and `instanceType: 'remote'` to `POST /devices/:id/actions/restart`.
- A non-restart action with `instanceType: 'remote'` returns the descriptive error.
- Read-only PAT is rejected for every tool here.
- Suspended-instance start/stop returns the descriptive 400.
- `set_remote_instance_mode` without a license returns the descriptive gate error.

---

### LINKED ISSUE #7706: 5.11-b Write tools, non-destructive (phase 2)
**Parent:** #7680 (5.11 Instance configuration)
**Tool file:** new `forge/ee/lib/mcp/tools/instanceConfig.js`

`readOnlyHint: false`, `destructiveHint: false`. Sibling actions that share one concept are consolidated into a single tool with a discriminator, so one tool covers several transitions of a config surface.

|---|---|---|---|
| `platform_set_instance_config` (`surface`: `ha`\|`customHostname`\|`protection`\|`autoUpdateStack`; `action`: `enable`\|`disable`\|`set`\|`clear`) | `PUT`/`DELETE /projects/:id/{ha,customHostname,protectInstance,autoUpdateStack}` | `project:edit` | write |
| `platform_update_instance_file` (rename/move a file, or set directory sharing) | `PUT /projects/:id/files/_/:path` | `project:files:edit` | write |
| `platform_upload_instance_file` (upload a file, or create a directory) | `POST /projects/:id/files/_/:path` | `project:files:create` | write |
| `platform_create_instance_http_token` (`instanceType`: `hosted`\|`remote`) | hosted: `POST /projects/:id/httpTokens`; remote: `POST /devices/:id/httpTokens` | `project:edit` (hosted), `device:edit` (remote) | write |
| `platform_update_instance_http_token` (`instanceType`: `hosted`\|`remote`) | hosted: `PUT /projects/:id/httpTokens/:tid`; remote: `PUT /devices/:id/httpTokens/:tid` | `project:edit` (hosted), `device:edit` (remote) | write |

**Consolidation notes:**
- `platform_set_instance_config` folds the four config surfaces (`ha`, `customHostname`, `protection`, `autoUpdateStack`) into one tool. `surface` selects the config and `action` selects the transition: `ha`/`protection` use `enable`/`disable`, `customHostname`/`autoUpdateStack` use `set`/`clear`. The handler maps `enable`/`set` to `PUT` and `disable`/`clear` to `DELETE` on the corresponding route, and validates the companion field required by the surface (`replicas` for `ha`, `hostname` for `customHostname`, `schedule` for `autoUpdateStack`).
- The http-token create/update tools cover both hosted and remote instances through an `instanceType` discriminator, mirroring the read tools. They route to `/projects/:id/httpTokens` (`project:edit`) or `/devices/:id/httpTokens` (`device:edit`) based on `instanceType`.

**Design notes:**
- The `ha`/`customHostname`/`protection` config DELETE routes disable a feature rather than destroy data, so they are the `disable`/`clear` action of a non-destructive tool alongside their PUT counterpart.
- **auto-update-stack is set/clear, not enable/disable:** `PUT /autoUpdateStack` sets a weekly restart schedule (an array of `{hour, day, restart}` entries), and `DELETE /autoUpdateStack` clears all schedule entries. There is no boolean toggle, hence the `set`/`clear` action names. These routes have no feature-flag gate.
- **The two file tools are distinct routes, not a rename:** `PUT /files/_/:path` (`project:files:edit`) updates file *properties* (rename/move via body `path`, or set directory sharing via body `share`); it does not upload content. `POST /files/_/:path` (`project:files:create`) is the actual upload path (multipart body) and also creates a directory (body `path`). Do not model the PUT as an upload.
- Plan-gated (enabled per team type/plan, not per user role): HA, custom hostnames, protected instance, files. If the team's plan does not include the feature the route returns a 404, independent of the PAT's permissions; surface this as a clear "feature not enabled for this team" message rather than a bare "not found". auto-update-stack has no feature gate.
- httpTokens PUT uses the `needsPermission('project:edit', true)` two-arg variant; the route blocks modifying Expert-MCP tokens. The tool need not replicate this, only rely on the route.

```js
platform_set_instance_config: {
  description: 'Enables, disables, sets, or clears a configuration surface on a hosted instance. surface selects the config; action semantics and the required companion field depend on surface: ha uses enable/disable (replicas required when enabling, must be 2); customHostname uses set/clear (hostname required when setting); protection uses enable/disable; autoUpdateStack uses set/clear (schedule required when setting). ha, customHostname, and protection are plan-gated and return a descriptive "feature not enabled for this team" error when the plan does not include them; autoUpdateStack has no plan gate.',
  inputSchema: z.object({
    instanceId: z.string().uuid().describe('UUID of the hosted instance'),
    surface: z.enum(['ha', 'customHostname', 'protection', 'autoUpdateStack']).describe('Configuration surface to change'),
    action: z.enum(['enable', 'disable', 'set', 'clear']).describe('Transition to apply: ha/protection use enable/disable, customHostname/autoUpdateStack use set/clear'),
    replicas: z.literal(2).optional().describe('HA replica count; required when enabling ha, must be 2'),
    hostname: z.string().optional().describe('Custom hostname; required when setting customHostname'),
    schedule: z.array(z.object({
      hour: z.number().describe('Hour of the day for the allowed restart window'),
      day: z.number().describe('Day-of-week index (0-6) the entry applies to'),
      restart: z.boolean().describe('Whether an automatic stack restart is permitted in this window')
    })).optional().describe('Weekly restart schedule as {hour, day, restart} entries; required when setting autoUpdateStack')
  })
}

platform_update_instance_file: {
  description: 'Updates properties of an existing file or directory on a hosted instance: renames/moves it (newPath, sent as body.path) or sets directory sharing (share). It does not upload content. newPath and share are mutually exclusive.',
  inputSchema: z.object({
    instanceId: z.string().uuid().describe('UUID of the hosted instance'),
    path: z.string().describe('Path of the existing file or directory to update'),
    newPath: z.string().optional().describe('New path for a rename or move'),
    share: z.record(z.any()).optional().describe('Directory sharing config')
  })
}

platform_upload_instance_file: {
  description: 'Uploads a file (multipart body) or creates a directory (directoryName, sent as body.path) at the target path on a hosted instance.',
  inputSchema: z.object({
    instanceId: z.string().uuid().describe('UUID of the hosted instance'),
    path: z.string().describe('Target path where the file or directory is created'),
    directoryName: z.string().optional().describe('Name of the directory to create')
  })
}

platform_create_instance_http_token: {
  description: 'Creates an HTTP bearer token for an instance. instanceType selects whether instanceId is a hosted instance UUID or a remote instance (device) hashid.',
  inputSchema: z.object({
    instanceId: z.string().describe('Instance id: hosted instance UUID, or remote instance (device) hashid'),
    instanceType: z.enum(['hosted', 'remote']).describe('Whether instanceId refers to a hosted instance or a remote instance (device)'),
    name: z.string().describe('Human-readable name for the HTTP bearer token'),
    expiresAt: z.string().optional().describe('Token expiry as an ISO 8601 timestamp; omit for no expiry'),
    scope: z.string().optional().describe('Scope string limiting what the token can access')
  })
}

platform_update_instance_http_token: {
  description: 'Updates the expiry or scope of an HTTP bearer token on an instance. instanceType selects whether instanceId is a hosted instance UUID or a remote instance (device) hashid.',
  inputSchema: z.object({
    instanceId: z.string().describe('Instance id: hosted instance UUID, or remote instance (device) hashid'),
    instanceType: z.enum(['hosted', 'remote']).describe('Whether instanceId refers to a hosted instance or a remote instance (device)'),
    tokenId: z.string().describe('Opaque hashid of the HTTP bearer token to update'),
    expiresAt: z.string().optional().describe('New expiry as an ISO 8601 timestamp'),
    scope: z.string().optional().describe('New scope string for the token')
  })
}
```

**Tests:**
- `platform_set_instance_config` routes each (surface, action) pair to the correct PUT or DELETE route and enforces the required companion field.
- The http-token tools route by `instanceType` to the project or device route.
- Read-only PAT is rejected for every tool here.
- Feature-disabled instance returns the descriptive gate error (not applicable to autoUpdateStack, which has no feature gate).

---

cstns:
Closes #7685
Closes #7706

The remaining instance write tools, 11 in total, covering hosted and remote instances.

**Lifecycle and cross-type tools** (in `shared-instances-devices.js`, next to the existing `instanceType`-discriminated reads):

* `platform_instance_action`: start/stop/restart/suspend/restartStack for hosted instances, restart only for devices (anything else is rejected tool-side with a clear message). The description covers the suspended-state 400s and that several hosted actions reply before the transition finishes.
* `platform_create_instance_http_token` / `platform_update_instance_http_token`: two route realities worth flagging. The routes never apply `scope` (it's hardcoded to `['']` in the controller), so the spec's `scope` input was dropped rather than shipped as a no-op. And on update, omitting `expiresAt` clears the expiry rather than leaving it unchanged, which the description shouts about. Create returns the raw token value once, so the description says to relay it immediately and treat it as a secret.

**Hosted instance tools** (in `instances.js`):

* `platform_update_hosted_instance_env` sends a body of exactly `{settings: {env}}` so the route's narrower `project:edit-env` path applies (Members can use it). Documents the full-replacement semantics and the hidden-value convention.
* `platform_update_hosted_instance_settings`: name/stack/type changes reply early and restart the instance in the background, `sourceProject` overwrites the target's flows, both flagged as confirm-with-user.
* `platform_import_hosted_instance_flows`: replaces flows and deploys immediately when running; wrong `credsSecret` is the 403 `invalid_credentials_secret`.
* `platform_set_instance_config` consolidates ha/customHostname/protection/autoUpdateStack behind surface+action, mapping enable/set to PUT and disable/clear to DELETE, validating the per-surface companion field tool-side. HA and hostname changes restart the instance; ha/hostname/protection are plan-gated 404s, autoUpdateStack isn't.
* `platform_update_instance_file` / `platform_upload_instance_file`: paths are URL-encoded into the single `:path` segment like the frontend does. The upload tool builds the multipart body by hand (the route only takes multipart for content) and also handles directory creation via the JSON branch; text content only.

**Remote instance tools** (in `devices.js`):

* `platform_update_remote_instance_settings`: the route silently drops palette/editor/security for non-Owners while still replying okay, so the description warns about it instead of pretending the fields always apply. Env hidden-value and password-preservation conventions documented from the handler.
* `platform_set_remote_instance_mode`: autonomous/developer only (the route's "default when omitted" path also writes `undefined` to the device, so the tool just requires `mode`), licensed platforms only.

Also removes a stale eslint-disable in `devices_spec.js` that the linter flagged once the file changed.


andypalmi (commented):
Went through all eleven tools against the routes and the mapping is solid: URLs, methods, permissions, the hosted/remote instanceType split, and the tool-side rejections (remote-only-restart, exactly-one-of on the file tools, the surface/action matrix) all line up. Descriptions are unusually precise about the awkward bits too. Tests and lint are green. A few small notes inline.

forge/ee/lib/mcp/tools/shared-instances-devices.js
```diff
@@ -75,5 +75,79 @@ module.exports = [
...
         }
+    },
+    {
+        name: 'platform_instance_action',
+        title: 'Instance Lifecycle Action',
+        description: `FlowFuse platform automation tool:
+            Applies a lifecycle action to an instance, changing whether and how it runs. Confirm with the user before stopping, suspending, or restarting anything.
+            Hosted instances accept: start (resume a suspended instance, or start the flows of a stopped one), stop (stop the flows, container keeps running), restart (restart the flows), suspend (shut the container down entirely), and restartStack (suspend then relaunch the container, picking up stack changes). stop, restart, and suspend are rejected with a 400 "project_suspended" while the instance is suspended - use start to bring it back first.
+            Remote instances (devices) accept only restart, which asks the device to restart Node-RED; the device must be online and reachable (400 "no_response" on timeout, "device_suspended" while suspended).
+            Several hosted actions reply { status: "okay" } once the transition has STARTED and finish in the background - check platform_get_hosted_instance_status to confirm the final state.`,
```
andypalmi, line 86: stop, restart, and suspend actually await the container operation before replying `{status:"okay"}` (projectActions.js); only start-from-suspended and restartStack reply early, and restartStack sends an empty body rather than `{status:"okay"}`. Could this line name just those two, so it matches the routes?

forge/ee/lib/mcp/tools/instances.js
```diff
@@ -357,6 +357,227 @@ module.exports = [
...
             return response
         }
+    },
+    {
+        name: 'platform_update_hosted_instance_env',
+        title: 'Update Hosted Instance Environment Variables',
+        description: `FlowFuse platform automation tool:
+            Replaces the full set of environment variables on a hosted instance. Variables missing from the list are removed, so read the current set first (platform_get_hosted_instance_config) and resend everything that should stay.
+            To keep an existing hidden (secret) variable's stored value without knowing it, resend it with hidden true and an empty value. A hidden entry with an empty value that does not already exist on the instance is dropped.
+            Values inherited from the instance's template may be locked and are validated against it; violations fail with "settings_validation".
```
andypalmi, line 367: In `validateSettings`, a policy-locked field is silently dropped from the result rather than rejected; `settings_validation` is raised for malformed input instead (bad env var names, duplicate names, invalid palette modules). Worth rewording so it doesn't promise an error on locked values? The settings tool at line 392 has the same wording.

forge/ee/lib/mcp/tools/instances.js
```diff
@@ -357,6 +357,227 @@ module.exports = [
...
+            }
+            const url = `/api/v1/projects/${args.instanceId}/files/_/${encodeURIComponent(args.path)}`
+            if (hasDirectory) {
+                const response = await inject({ method: 'POST', url, payload: { path: args.directoryName } })
+                return response
+            }
+            // The route only accepts file content as multipart/form-data, so build a
+            // single-part body by hand; the target name comes from the URL path, not
+            // the part's filename.
+            const boundary = 'FlowFuseMcpFileUploadBoundary29b18a7f'
```
andypalmi, line 562: This hand-built body reuses a fixed boundary on every call. If uploaded text ever contains that exact literal, the parser treats it as a delimiter and corrupts the upload. Since arbitrary text is the supported case, would generating the body with the `form-data` library (as the route's own tests do) be cleaner? That gives a random per-call boundary and drops the manual assembly.

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -227,5 +227,54 @@ module.exports = [
...
+        title: 'Set Remote Instance Mode',
+        description: `FlowFuse platform automation tool:
+            Switches a remote instance (device) between its two operating modes and notifies the device immediately.
+            "autonomous" is normal fleet operation: the device runs its target snapshot and picks up new target snapshots as they are deployed. "developer" detaches it from fleet updates so its flows can be edited live via the device editor; deploys do not reach it until it returns to autonomous.
+            Requires a licensed platform: unlicensed platforms reject this with 400 "not_licensed".`,
+        annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: true, openWorldHint: false },
+        inputSchema: {
+            remoteInstanceId,
+            mode: z.enum(['autonomous', 'developer']).describe('Operating mode: "autonomous" for normal fleet operation, "developer" for live editing detached from fleet updates')
+        },
```
andypalmi, line 274: The route defaults `mode` to autonomous (`request.body.mode || 'autonomous'`), so could the schema carry that default (`.default('autonomous')`) rather than requiring it, letting the agent see it directly? The route then writes the raw `request.body.mode`, so a caller omitting it persists `undefined`, worth fixing in this PR to write the validated `mode`.

cstns:
Tested against a running platform. Review notes addressed in 8b3ffe2:

1. **Reply timing** — confirmed, reworded to name only start-from-suspended and restartStack (which sends an empty body).
2. **Locked settings** — confirmed (`ProjectTemplate.js:38` drops them), both descriptions reworded.
3. **Multipart boundary** — now `randomUUID()` per call. Kept the hand-built body; `form-data` is only a transitive dep here, so using it means adding a direct dependency. Say if you'd rather have it.
4. **Device mode** — code reading is right, but the effect differs: Sequelize ignores an `undefined` assignment, so the stored mode is untouched. Set a device to developer, sent `{}`, stayed developer. Real issue is a data no-op that still notifies the device and writes an audit entry. Kept `mode` required rather than defaulting, so an agent can't silently pull a device out of developer mode. Route fix is out of scope for these PRs, happy to raise an issue.

Also marked 9 of the 11 writes destructive (overwrite-or-remove rule). Create-token, create-instance, assign-to-application and set-mode stay plain writes.

forge/ee/lib/mcp/tools/instances.js
```diff
@@ -357,6 +359,235 @@ module.exports = [
...
+                options: z.record(z.string(), z.any()).optional().describe('Flags selecting which parts to copy (flows, credentials, envVars, and so on)')
+            }).optional().describe('Copies configuration and flows from another instance onto this one, overwriting current content. Confirm with the user first')
+        },
+        handler: async (args, { inject }) => {
+            const payload = {}
+            for (const key of ['name', 'settings', 'launcherSettings', 'projectType', 'stack', 'sourceProject']) {
+                if (args[key] !== undefined) {
+                    payload[key] = args[key]
+                }
+            }
```
andypalmi, line 420: Guard the empty payload the way the device-group tools do:
```suggestion
            }
            if (Object.keys(payload).length === 0) {
                return toolError(400, 'invalid_request', 'Pass at least one of name, settings, launcherSettings, projectType, stack or sourceProject to update')
            }
```

andypalmi (commented):
Checked the 11 tools against their routes at the branch head. Descriptions, schemas, and annotations all match (scope guards, expiresAt clear-on-omit, hidden-env and password preservation, HA replica 409, hostname 409, not_licensed, the feature-gate 404s, and the early-reply background restarts), and there are no zod ordering issues. A few things to fix before merge, suggested inline:

- `platform_update_hosted_instance_settings`, `platform_update_remote_instance_settings` and `platform_import_hosted_instance_flows` build their payloads from all-optional fields with no empty-body guard, so passing only the instance id sends `{}`. They should get the `Object.keys(payload).length === 0` toolError(400) guard the device-group update and membership tools already use, plus covering tests.
- The remote-settings and import ones are why it matters: an empty hosted-settings PUT is a harmless 200 no-op, but an empty remote-settings PUT still hits `sendDeviceUpdateCommand` and an empty import still calls `importProject`, both of which notify or restart a running instance for a call that changes nothing.
- `platform_set_remote_instance_mode`'s `mode` describe explains what happens when mode is omitted, but mode is required so that path is unreachable, and the explanation is also inaccurate (the route would write `undefined`). Trimmed inline.

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -227,5 +227,55 @@ module.exports = [
...
+            editor: z.record(z.string(), z.any()).optional().describe('Editor settings, e.g. nodeRedVersion (Owner only; silently ignored otherwise)'),
+            security: z.record(z.string(), z.any()).optional().describe('Security settings such as httpNodeAuth or localAuth (Owner only; silently ignored otherwise). Omit passwords to keep the stored ones')
+        },
+        handler: async (args, { inject }) => {
+            const payload = {}
+            for (const key of ['env', 'autoSnapshot', 'palette', 'editor', 'security']) {
+                if (args[key] !== undefined) {
+                    payload[key] = args[key]
+                }
+            }
```
andypalmi, line 259: Same guard here; this is the one that otherwise notifies the device and can restart Node-RED on a no-op:
```suggestion
            }
            if (Object.keys(payload).length === 0) {
                return toolError(400, 'invalid_request', 'Pass at least one of env, autoSnapshot, palette, editor or security to update')
            }
```

forge/ee/lib/mcp/tools/instances.js
```diff
@@ -357,6 +359,235 @@ module.exports = [
...
+            credentials: z.string().optional().describe('Encrypted flow credentials object serialized as a JSON string, as exported from another instance'),
+            credsSecret: z.string().optional().describe('Secret used to decrypt the supplied credentials. Required when credentials are supplied')
+        },
+        handler: async (args, { inject }) => {
+            const payload = {}
+            for (const key of ['flows', 'credentials', 'credsSecret']) {
+                if (args[key] !== undefined) {
+                    payload[key] = args[key]
+                }
+            }
```
andypalmi, line 449: An empty import restarts a running instance for nothing, so guard it too:
```suggestion
            }
            if (Object.keys(payload).length === 0) {
                return toolError(400, 'invalid_request', 'Pass at least one of flows or credentials to import')
            }
```

forge/ee/lib/mcp/tools/devices.js
```diff
@@ -227,5 +227,55 @@ module.exports = [
...
+        title: 'Set Remote Instance Mode',
+        description: `FlowFuse platform automation tool:
+            Switches a remote instance (device) between its two operating modes and notifies the device immediately.
+            "autonomous" is normal fleet operation: the device runs its target snapshot and picks up new target snapshots as they are deployed. "developer" detaches it from fleet updates so its flows can be edited live via the device editor; deploys do not reach it until it returns to autonomous.
+            Requires a licensed platform: unlicensed platforms reject this with 400 "not_licensed".`,
+        annotations: { readOnlyHint: false, destructiveHint: false, idempotentHint: true, openWorldHint: false },
+        inputSchema: {
+            remoteInstanceId,
+            mode: z.enum(['autonomous', 'developer']).describe('Operating mode: "autonomous" for normal fleet operation, "developer" for live editing detached from fleet updates. Required deliberately: the route treats an omitted mode as autonomous for validation but then writes the raw value, so omitting it leaves the stored mode untouched while still notifying the device and writing an audit entry')
```
andypalmi: `mode` is required, so the omitted-mode explanation is both moot and inaccurate (the route would write `undefined`, not leave the mode untouched). Trim it:
```suggestion
            mode: z.enum(['autonomous', 'developer']).describe('Operating mode: "autonomous" for normal fleet operation, "developer" for live editing detached from fleet updates')
```

andypalmi (approved):
Suggestions were added so approving

PR #8511, opened 2026-09-14
user registration: Disallow free email addresses

ZJvandeWeg:
## Description

Disallow registrations from free emails to reduce low-intent sign ups.

## Checklist

 - [X] Suitable unit/system level tests have been added and they pass 


forge/routes/auth/index.js
```diff
@@ -526,6 +527,12 @@ async function init (app, opts) {
             reply.code(400).send(resp)
             return
         }
+        if (app.billing && isFreeEmail(request.body.email.toLowerCase())) {
```
ZJvandeWeg, line 530: I didn't find a better API that basically is just `app.isFlowFuseCloud()`. This seemed the best choice

JoycePlaysFootball:
@n-lark here are a list of emails domains to suppress: https://docs.google.com/spreadsheets/d/1tZe-VxUfwoKEQb2n9yGQ-n1GnbxVlM0lOzmMVGga6VE/edit?usp=sharing 

There should be feedback to a user that wants to use their yahoo.com (for example) but gets blocked; a message like 'business emails only' to let them know.

n-lark (changes requested):
Hey! Noticed a few things we probably want to address before it ships. Lmk if any questions. 

| Issue | What actually happens |
| --- | --- |
| The Google sign-up button on the same page skips the check | `google.js` only inspects `googleUserInfo.hd`, which is set for Workspace accounts only. Personal gmail has no `hd`, so anyone blocked by the form clicks the button below it and gets in. |
| The error message never reaches the user | `Create.vue` has no branch for `invalid_email_domain`, so it shows the generic "An unexpected error occurred, contact support". The "use your company email" copy is thrown away. |
| No client-side validation | Nothing flags the domain until submit, and then the message is the one above. |
| Existing customers can't add teammates who use a personal address | Invited people sign up through this same route, and the check runs after the invite lookup that's meant to let them through. So a customer invites `contractor@gmail.com`, that person can never accept. |
| We should block Joyce's list explicitly instead | Hers is 68 domains, the package blocks 96,649 scraped ones, so it hits people at random: `example.com`, `icloud.com`, `comcast.net`, and universities inconsistently (`stanford.edu` blocked, `mit.edu` not). Hardcoding the 68 targets exactly who we mean to. |

JoycePlaysFootball:
I'm not too sure what to do with any of this @n-lark....do you have any ideas?

n-lark:
Hey @JoycePlaysFootball my comments here are just from my review of this code for @ZJvandeWeg to address. No action needed on your end.

ZJvandeWeg:
> The Google sign-up button on the same page skips the check

Fixed

> The error message never reaches the user

Surprisingly badly engineered frontend, but thanks for the spot. Fixed now.

> Nothing flags the domain until submit, and then the message is the one above.

I don't see an issue here.

> Existing customers can't add teammates who use a personal address

Fixed

> We should block Joyce's list explicitly instead

Why? It seems better to depend on an upstream package then on a self-compiled and maintained list. It might be arbitrary, though so is a self-compiled list.

knolleary:
We can turn off creating new users automatically via the google sign button via the admin settings - 
image

I have raised the appropriate change request - https://github.com/FlowFuse/CloudProject/issues/1577 - and applied it. However, the google button is still shown on the sign-up page - I will raise a separate PR to fix that.

 - https://github.com/FlowFuse/flowfuse/pull/8515

forge/ee/routes/sso/social/google.js
```diff
@@ -1,3 +1,4 @@
+const { isFreeEmail } = require('free-email-domains-list')
 const fp = require('fastify-plugin')
```
n-lark, line 1: You've got a linter error here

n-lark (approved):
Hey approved - just the linter error needs to be resolved. 

> Nothing flags the domain until submit, and then the message is the one above.
> 
> I don't see an issue here.

Generally its nice to have validation as the user types, instead of onsubmit. Not worth holding the PR over. 

> We should block Joyce's list explicitly instead
> 
> Why? It seems better to depend on an upstream package then on a self-compiled and maintained list. It might be arbitrary, though so is a self-compiled list.

Since that package blocks so many emails, I had concerns that we could be blocking users we are not intending to block, like university addresses. The package hasn't been touched in over a year so wonder if its maintained. Fine to go ahead if we're comfortable with that.

forge/ee/routes/sso/social/google.js
ZJvandeWeg: This is fixed now.

ZJvandeWeg:
@n-lark I feel like self-merging, while I can, isn't a great idea. So if you have 10s in the next hour, please 😄

PR #8517, opened 2026-09-15
Git prevent substitution

hardillb:
## Description

Ensure no shell substitution in git commands

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/ee/lib/gitops/backends/github.js
```diff
@@ -86,14 +86,15 @@ module.exports.init = async function (app) {
             await fs.writeFile(snapshotFile, JSON.stringify(snapshotExport, null, 4))
 
             // 6. stage file
-            await execPromised(`git add "${snapshotFile}"`, { cwd: workingDir })
+            await execFilePromised('git', ['add', snapshotFile], { cwd: workingDir })
 
             // 7. commit
-            await execPromised(`git commit -m "Update snapshot\n\nSnapshot updated by FlowFuse Pipeline '${options.pipeline.name.replace(/"/g, '')}', triggered by ${options.user.username.replace(/"/g, '')}" --author="${author}"`, { cwd: workingDir })
+            const commitMessage = `"Update snapshot\n\nSnapshot updated by FlowFuse Pipeline '${options.pipeline.name}', triggered by ${options.user.username}"`
```
knolleary:
```suggestion
            const commitMessage = `Update snapshot\n\nSnapshot updated by FlowFuse Pipeline '${options.pipeline.name}', triggered by ${options.user.username}`
```

Don't need the quotes any more

forge/ee/lib/gitops/backends/github.js
knolleary:
```suggestion
            const author = `${userGitName} <${userGitEmail}>`
```

Do not need to escape quotes anymore

forge/ee/lib/gitops/backends/azure.js
knolleary:
```suggestion
            const author = `${userGitName} <${userGitEmail}>`
```

Do not need to escape quotes manually anymore

knolleary (changes requested):
We need to drop some of the added quote escaping - moving to execFile means escaping is done at the shell level, so the escapes/extra quotes appear as literals in the output.

forge/ee/lib/gitops/backends/azure.js
```diff
@@ -60,17 +60,17 @@ module.exports.init = async function (app) {
...
+            const author = `${userGitName} <${userGitEmail}>`
             workingDir = await fs.mkdtemp(path.join(os.tmpdir(), 'flowfuse-git-repo-'))
 
             // 3. clone repo
             await cloneRepository(url, branch, workingDir)
 
             // 4. set username/email
-            await execPromised('git config user.email "no-reply@flowfuse.com"', { cwd: workingDir })
-            await execPromised('git config user.name "FlowFuse"', { cwd: workingDir })
+            await execFilePromised('git', ['config', 'user.email', 'no-reply@flowfuse.com'], { cwd: workingDir })
```
ZJvandeWeg, line 70: Config writes to the disk, and you'll likely just want to collect config key value pairs and pass them when executing the `git` binary with the `-c` flag.

ZJvandeWeg: Continueing on this thread, I suspect it's better to make a `gitCommand()` which takes a map with config, and a command with their arguments and a worktree location.

knolleary (approved):
I have raised https://github.com/FlowFuse/flowfuse/issues/8648 to capture the feedback on refactoring to a utility function. Approving this change as-is to ensure it gets into 3.1.

PR #8519, opened 2026-09-15
Pre-fill MCP access expiry with default of 1 year minus 1 day

ZJvandeWeg:
The max allowed expiry is one year, so defaulting the date picker to that value avoids the user having to type it in manually.

## Checklist

 - [X] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)


cstns (changes requested):
the changes are sensible but you'll need to update the test expectations to match

cstns (approved):

PR #8536, opened 2026-09-16
Persist HITL approval outcomes across refresh

### LINKED ISSUE #8527: Persist HITL approval outcomes across refresh
On page refresh, already-resolved HITL approval cards show their Allow/Deny buttons again. Question answers survive refresh because they are included in the conversation persistence; approval outcomes never were.

Done when: refreshing mid-conversation shows resolved approval cards in their resolved state.

andypalmi:
Approval card decisions are now kept in a persisted store map (like question answers already were), so resolved cards stay resolved after a page refresh instead of offering Allow/Deny again.

Closes #8527


frontend/src/components/expert/components/messages/components/AnswerWrapper.vue
```diff
@@ -224,11 +224,8 @@ export default {
         hasToolApproval () {
             return this.answer.kind === 'tool-approval' && !!this.answer.id
         },
-        // The answer here is a detached streaming copy, so its own `status` only ever holds
-        // the initial 'pending'. The reactive per-id map in the store carries any later
-        // outcome — including an external one (chat stop / Start Over) the card can't see.
         resolvedToolApprovalStatus () {
```
cstns, line 230: Can we keep the comment that was here? It's the only thing explaining why the status is read from a store map rather than `answer.status`, and that reasoning is still true after the change. Without it the `||` chain reads as arbitrary.

frontend/src/stores/product-expert.js
```diff
@@ -41,6 +41,8 @@ export const useProductExpertStore = defineStore('product-expert', {
         composerCommand: null,
         // question-card answers keyed by answer uuid, so a sent card survives a refresh
         questionAnswers: {},
+        // Terminal outcome per approval id, so a resolved approval card survives a refresh (#8527).
+        approvalOutcomes: {},
```
cstns: Not for this PR, but noting it since we're here: `_approvalBatch` isn't persisted either, so if you refresh while a card is still genuinely pending, the card comes back with Allow/Deny and clicking does nothing, `resolveToolApproval` bails on `if (!batch ...)` and the agent stays parked. The issue only asks about resolved cards so this is out of scope, but might be worth a follow up issue so it doesn't get lost.

cstns (changes requested):
Works, but I think this can be a lot smaller. `product-assistant` already has `toolApprovalStatuses`, same keys, same values, and after this change both maps get written at the same two places and cleared at the same one, so they can never hold different things. The middle operand in `toolApprovalStatuses[id] || approvalOutcomes[id] || answer.status` is only ever reachable in the one tick right after a refresh.

Might be worth just persisting the map that's already there instead. `pinia-plugin-persistedstate` 4.7.1 takes an array of persist configs (`Persist = boolean | PersistenceOptions | PersistenceOptions[]` in its `index.d.ts`), so one store can mix storages:

```js
persist: [
    { pick: ['toolDefaultsByTeam', 'toolPreferencesByTeam'], storage: localStorage },
    // Resolved approval outcomes are chat scoped, so sessionStorage (#8527).
    { pick: ['toolApprovalStatuses'], storage: sessionStorage }
]
```

That would drop the new state field, both mirrored writes, the extra clear in `startOver`, the extra `mapState`, and the `||` chain. The tests here already assert `permState.statuses.u1` so they'd still cover it, the three new `store.approvalOutcomes` assertions just wouldn't be needed.

Two smaller things below.

frontend/src/components/expert/components/messages/components/AnswerWrapper.vue
andypalmi: Restored, the whole file is back to main's version since the fallback operand went away with the rework.

frontend/src/stores/product-expert.js
andypalmi: Nice catch. Rather than a follow-up issue we pulled it into this PR: `_approvalBatch` is now in the sessionStorage persist pick, so a pending card stays answerable after refresh and the resume reuses the persisted sessionId. Covered by a new spec.

andypalmi:
Good call, done in f23b8e4: the store now uses the persist array with a sessionStorage entry for `toolApprovalStatuses`, and the mirror map, double writes, extra clear and the `||` chain are gone. AnswerWrapper and the spec are back to what main has.

cstns (approved):

PR #8537, opened 2026-09-16
Rehydrate an expired Expert conversation instead of disabling chat

### LINKED ISSUE #8534: Rehydrate an expired Expert conversation instead of disabling chat
A user who stays in the Expert chat for more than 30 minutes hits the conversation expiry: "Your conversation history will expire soon" followed by "Your conversation history has expired. Chat is now disabled. Click Start Over to begin a new conversation." During onboarding this is a dead end: the funnel has no navigation, and asking a brand-new user to start over throws away the intake they just went through.

When the history expires, rehydrate the chat with a new conversation id transparently: keep the visible messages in place, swap the id underneath, and let the user simply send their next message and go forward. Never disable the input or ask them to start over.

The same applies to the normal Expert channel: remove the expired-chat Start Over dead end there too and renew silently.

Done when: a chat left open past the expiry keeps accepting messages, in onboarding and in the regular Expert drawer, with no expiry banners and no Start Over prompt.

andypalmi:
When the conversation history expires, silently swap in a fresh conversation id and keep the transcript and input as they are, so the user just carries on. Removes the expiry banners and the Start Over dead end in onboarding and the drawer. Stacked on #8527's PR.

Closes #8534


cstns:
treating as edge case for now, will tackle after we roll out the ai led onboarding

andypalmi:
Currently we only send the query on each Expert message, so this would not work as it is ( we have the `hydrateChat` method which kinda works but it's not ideal already for this type of chat continuation) so a pattern investigation into it would make sense

Currently Expert stores chats for 30 minutes and we would need to send over the full list of messages in this chat for this logic of no expiration of 30 minutes to work

Issue #8520, opened 2026-09-16
Fix onboarding question card styling side-effects


andypalmi:
The onboarding page deliberately restyles question cards into full-width tiles, but the page-scoped `:deep` overrides in `Onboarding.vue` (`.ff-radio-btn` / `.ff-checkbox` set to `display: flex; width: 100%`) also stretch the selection controls themselves. The cards should keep using the same Expert question UI as the chat drawer, just at full chat width; the checkboxes and radios keep their normal size.

Side-effects to fix:

- Single-select options render the description on the same line as the label (the override turns `ff-radio-group`'s label plus `p.ff-description` into a flex row). Multi-select stacks correctly.
- The free-text option's input is crushed to a few px because the control becomes a full-width tile.

Done when: question cards span the full chat width with normally sized controls, single-select descriptions render under their label, the free-text input has full usable width, and the drawer chat rendering is unchanged.

PR #8541, opened 2026-09-16
Let the user sign out of FlowFuse from the onboarding page

### LINKED ISSUE #8522: Always give the user a way out of the onboarding page
During onboarding there is no navigation and no way to sign out of FlowFuse until the flow completes; a user who stalls mid-flow is stuck in the app.

- Set the onboarding flag to false as soon as the agent moves the onboarding phase to done.
- Set it to false as well when the user clicks "Set it up myself".
- Show the user avatar menu on the onboarding page, to the right of "Set it up myself", so the user can sign out of FlowFuse. Sign-out is the only navigation offered there; "Set it up myself" remains the way into the app.

Done when: the user can always sign out of FlowFuse from the onboarding page, and completion (by agent or by skip) restores normal navigation.

andypalmi:
Shows the main navigation's avatar dropdown next to "Set it up myself" with Sign Out as the only option, so a user who stalls mid-onboarding is not stuck in the app. Stacked on #8520's PR.

Closes #8522


frontend/src/pages/team/Onboarding.vue
```diff
@@ -14,6 +14,25 @@
                 >
                     Set it up myself
                 </ff-button>
+                <!-- Sign-out is the only navigation offered on this page -->
+                <ff-dropdown
```
cstns, line 18: lets add the entire ff-dropdown with the full navigation options list the same one we're using for the other layouts

cstns (changes requested):

frontend/src/pages/team/Onboarding.vue
cstns: we'll tackle it in a later iteration

cstns (approved):

PR #8548, opened 2026-09-16
fix: 404 unknown .well-known paths and serve a domain-verification token

andypalmi:
Closes FlowFuse/engineering#335

Unknown paths under `/.well-known/` fell through to the SPA and returned 200 HTML, which fails a domain-verification check.

This adds two routes to the EE mcp plugin: a wildcard handler that returns 404 for unmatched `/.well-known/` paths, and `/.well-known/openai-apps-challenge`, which serves `mcp.domainVerificationToken` from config as `text/plain` when set and 404s otherwise. Serving it from the app lets self-hosted instances enable it through config.


cstns (approved):

knolleary:
@andypalmi do you know if anything was done in the helm chart to support exposing this new setting (`mcp.domainVerificationToken`) in a way that we can set it in production?

andypalmi:
No, it was not. I can file an issue for it and investigate it

@knolleary

knolleary:
@andypalmi thanks - we've got to the point of the openapi registration process that we have the token to use... just need a place to use it :)

PR #8556, opened 2026-09-17
feat: add Support for Agent Deploy

### LINKED ISSUE #8555: Add support for auto deploy by agents
### Description

As described in https://github.com/FlowFuse/flowfuse/issues/8488#issuecomment-5714006620

> * Team-level toggle ("Agent Initiated Deploy") under Team Settings > Danger, gated on the platform ai feature, reusing the same storage/API/audit mechanism as the existing AI opt-out toggle.
> * A live policy check so turning the setting off takes effect immediately, with no instance restart needed.
> * The enablement status is available in the agent's own context, so it can check without triggering a failed deploy first.

### Epic/Story

_No response_

Steve-Mcl:
### Summary

Adds a team-level "Agent Initiated Deploy" setting. When on, an AI agent (first-party Expert or a third-party MCP client) can deploy the flow changes it makes on that team's instances, instead of leaving them staged for a person to click Deploy. Off by default, and gated on the same platform/team `ai` feature flags as the existing AI opt-out toggle.

### Changes

- `forge/routes/api/team.js`: add `agentAutoDeploy` to the `features` allowlist on `PUT /api/v1/teams/:teamId`.
- `forge/routes/api/assistant.js`: new Endpoint `GET /api/v1/assistant/deploy-policy`, a live check combining the platform `ai` flag with the team's `agentAutoDeploy` override. Checked live rather than cached, so turning it off takes effect immediately, no instance restart needed.
- `frontend/src/pages/team/Settings/Danger.vue`: new "AI Flow Deploy" toggle next to AI Features, disabled (with an explanatory hint) when AI Features is off, since the setting has no effect otherwise.
- `frontend/src/stores/context.js`: expose `agentAutoDeployEnabled` on the agent's context object, so an agent can check the setting up front instead of learning it's off from a failed deploy attempt.
- `frontend/src/composables/TeamProperties.js`: `getTeamProperty` now tolerates a team with no `.type` instead of throwing.

### Testing

- `team_spec.js`: owner can toggle `agentAutoDeploy`, member gets 403, feature keys outside the allowlist are ignored.
- `assistant_spec.js`: `/deploy-policy` covers default-off, team-enabled, team-opted-out-of-ai, and platform-ai-disabled.
- `context.spec.js`: `agentAutoDeployEnabled` covers no-team, platform/team ai-off, override-enabled, and TeamType-fallback, plus both branches of the `expert` getter.
- `TeamProperties.spec.js`: new file. No coverage existed for this composable before.

### Notes

The `deploy_flows` tool itself, and the automation that actually triggers the deploy, live in `nr-assistant` and a separate Flow Builder MCP tool surface to be shipped alongside this. This PR is the `flowfuse` half: the setting, its API, and the policy check the agent reads from. 

It should not be merged until nr-assistant and agent work is merged (https://github.com/FlowFuse/nr-assistant/issues/405 and https://github.com/FlowFuse/engineering/issues/370 and https://github.com/FlowFuse/engineering/issues/371)

## Related Issue(s)

closes #8555 

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels


forge/routes/api/assistant.js
```diff
@@ -121,6 +121,24 @@ module.exports = async function (app) {
...
+     * agent-initiated deploys. Deliberately not cached/pushed via settings.js - it
+     * must be checked at the moment a deploy is being considered so that turning the
+     * team setting off takes effect immediately, without an instance restart.
+     * @name /api/v1/assistant/deploy-policy
+     * @static
+     * @memberof forge.routes.api.assistant
+     */
+    app.get('/deploy-policy', {
+        schema: {
+            hide: true // dont show in swagger
```
andypalmi, line 147: Why hide it?

frontend/src/composables/TeamProperties.js
```diff
@@ -21,10 +21,10 @@ function getProperty (properties, key) {
 }
 
 export function getTeamProperty (team, property, defaultValue) {
-    const teamValue = getProperty(team.properties, property)
+    const teamValue = getProperty(team?.properties, property)
```
andypalmi: What requires this change?

frontend/src/composables/TeamProperties.js
```diff
@@ -21,10 +21,10 @@ function getProperty (properties, key) {
...
     if (teamValue === undefined) {
         // No value found in team properties. Check the TeamType properties
-        return getProperty(team.type.properties, property) ?? defaultValue
+        return getProperty(team?.type?.properties, property) ?? defaultValue
```
andypalmi: Same question here

andypalmi:
Some findings from testing this locally:

1. Off by default does not hold on self-hosted. The default TeamType bootstrap (`forge/db/controllers/TeamType.js`) sets `enableAllFeatures: true`, and `TeamType.getFeatureProperty` returns true for any key before reading the `features` map. `Team.getFeatureProperty('agentAutoDeploy', false)` falls through to that for any team that never touched the setting, so on a stock docker-compose/k8s install `/deploy-policy` answers `autoDeploy: true` with nobody opted in. The frontend disagrees: `getTeamProperty` does not honor `enableAllFeatures`, so the new toggle shows off while the policy answers on. The new tests in `assistant_spec.js` pin the flag off, which works around exactly this. Since this setting authorizes unattended deploys it has to be an explicit opt-in: the policy check should read the team's own override directly (`request.team?.properties?.features?.agentAutoDeploy === true`) instead of inheriting through the TeamType fallback. Seeding `features.agentAutoDeploy: false` in the bootstrap does not work, the `enableAllFeatures` short-circuit runs first. Reading the override directly also brings backend and frontend into agreement.
2. The pre-existing `aiEnabled` computed in `Danger.vue` only reads the team type's `features.ai`, not the team-level override, so a team that opted out of AI still sees the new toggle as enabled and the hint never shows. `deploy-policy` returns false there so it is display-only; the disabled state should use `getTeamProperty` like the new `agentAutoDeploy` computed does.
3. Two small ones: `/deploy-policy` has no `rateLimit` config unlike the sibling endpoints in the same file, and the section heading says "AI Flow Deploy" while the confirm dialog says "Agent Initiated Deploy"; aligning the naming would make it read as one setting.

Steve-Mcl:
@andypalmi 
1. see [7a5df7f](https://github.com/FlowFuse/flowfuse/pull/8556/commits/7a5df7fce58376788d3957f99bedd09b6ef1ee4e)
2. altered the existing aiEnabled to use same pettern
3. added rate limiting, updated wording

Please re-check

andypalmi:
Re-checked all three points, all good: explicit team-level opt-in with the enableAllFeatures regression test, rate limit matching the sibling endpoints, consistent naming.

Two follow-ups on our side:
- The deploy_flows tool description still points users at "Agent-Initiated Deploy"; the wording will be aligned to "AI Flow Deploy" in FlowFuse/flowfuse-flow-builder#217.
- The toggle's computed reads through getTeamProperty and so inherits an explicit features.agentAutoDeploy set on a TeamType, while the policy honours only the team's own override; such a type would show the toggle on with the policy answering false. Nothing sets it at type level today; we will raise a follow-up issue.

forge/routes/api/assistant.js
andypalmi: Ok for now, we can come back on it later if somebody wants to query that

frontend/src/pages/team/Settings/Danger.vue
```diff
@@ -175,6 +198,29 @@ export default {
...
             })
+        },
+        showConfirmAgentAutoDeployToggleDialog () {
+            const enabling = this.agentAutoDeploy
+            Dialog.show({
+                header: enabling ? 'Enable AI Flow Deploy' : 'Disable AI Flow Deploy',
+                kind: enabling ? 'danger' : 'primary',
+                text: enabling
+                    ? 'Are you sure you want to allow AI agents to deploy flow changes they make on this team\'s instances, without a person clicking Deploy?'
+                    : 'Are you sure you want to prevent AI agents from deploying flow changes automatically? Changes they make will still need to be deployed manually.',
```
andypalmi:
```suggestion
                    : 'Are you sure you want to disable AI agents deploying flow changes automatically?',
```

frontend/src/pages/team/Settings/Danger.vue
```diff
@@ -175,6 +198,29 @@ export default {
...
                 this.aiEnabledOverride = null
             })
+        },
+        showConfirmAgentAutoDeployToggleDialog () {
+            const enabling = this.agentAutoDeploy
+            Dialog.show({
+                header: enabling ? 'Enable AI Flow Deploy' : 'Disable AI Flow Deploy',
+                kind: enabling ? 'danger' : 'primary',
+                text: enabling
+                    ? 'Are you sure you want to allow AI agents to deploy flow changes they make on this team\'s instances, without a person clicking Deploy?'
```
andypalmi:
```suggestion
                    ? 'Are you sure you want to allow AI agents to deploy flows on this team\'s instances?'
```

andypalmi (approved):

PR #8557, opened 2026-09-17
Await editor readiness in ui_navigate and return context

andypalmi:
Closes FlowFuse/engineering#365

**Problem**

`ui_navigate` resolves as soon as the Vue route change completes. For editor routes the page then loads Node-RED, so the agent has to poll `ui_get_context` before it can safely start working, and when the editor's assistant plugin is missing or outdated the agent only discovers it later through failing tool calls.

**Changes**

- `ui_navigate` now returns the post-navigation context (the same payload `ui_get_context` returns), so a follow-up call is not needed
- for editor routes (`instance-editor*`, `device-editor*`) it waits for the editor's `assistant-ready` signal and returns `editorReady` and `assistantVersion`; the wait deadline is measured from before the route change and capped at 8 seconds so the response always fits the transport's ~10 second window, even when the transition itself is slow
- when the signal does not arrive, or the reported nr-assistant version is below the minimum for agent flow building, the result carries a `notice` telling the agent the plugin needs updating to the latest version; the minimum lives in `frontend/src/mcp/constants.ts`
- a `readyGeneration` counter on the product-assistant store guards against stale readiness: navigating from one editor straight to another reuses the component tree, so the previous editor's version would otherwise read as ready immediately
- tool description updated to match

**Tests**

- new specs for the ready, timeout, below-minimum-version and stale-version paths, plus the context-only non-editor path (fake timers, no real waits)
- full frontend unit suite passing

**Verified live** (local platform, tab driven over MCP, temporary console tracing since removed)

Navigation to a hosted instance editor returned `editorReady: true`, `assistantVersion` and the full context; the fresh signal was detected 600 ms after the route change:

```
[assistant-store] assistant-ready received {version: '0.19.0', readyGeneration: 1}
[ui_navigate] fresh assistant-ready detected {version: '0.19.0', generation: 1, baselineGeneration: 0, elapsedMs: 600}
```

Navigation to a device editor detected its fresh signal at 1.4 s:

```
[assistant-store] assistant-ready received {version: '0.17.0', readyGeneration: 1}
[ui_navigate] fresh assistant-ready detected {version: '0.17.0', generation: 1, baselineGeneration: 0, elapsedMs: 1400}
```

Navigation to a non-editor route (`team-home`) returned the context with no editor fields, and the assistant store was correctly reset on leaving the editor.


frontend/src/mcp/tools/navigation.ts
```diff
@@ -1,7 +1,43 @@
...
+const EDITOR_READY_POLL_INTERVAL_MS = 100
+
+function isEditorRoute (routeName: string): boolean {
+    return routeName === 'instance-editor' || routeName.startsWith('instance-editor-') ||
+        routeName === 'device-editor' || routeName.startsWith('device-editor-')
+}
+
+// Waits for a fresh 'assistant-ready' signal, not just for `version` to be set: editor-to-editor
+// navigation does not unmount the iframe wrapper, so a stale version would otherwise read as ready.
+function waitForFreshAssistantVersion (baselineGeneration: number, deadline: number): Promise<string | null> {
```
cstns: we already have type of helper defined, check frontend/src/composables/TimerHelper.js

cstns (changes requested):

cstns (approved):

Issue #8558, opened 2026-09-17
ChatGPT's MCP client does not return Mcp-Session-Id, which breaks browser tab pinning


cstns:
Browser tab pinning does not work when ChatGPT is the MCP client. `platform_set_active_browser_session` returns successfully, and the next tool call reports that no tab is pinned. Repeating the pin changes nothing.

The cause is session identity. The Streamable HTTP transport in the MCP specification (session management, 2025-03-26 revision onward) states that when a server returns an `Mcp-Session-Id` header, the client MUST include that header on all subsequent HTTP requests. We return it on every response from `forge/ee/routes/mcp/server.js`. ChatGPT's connector never sends it back, so every request presents as a new session. Our pinning design assumed clients would honour that requirement.

Evidence from production on 2026-09-17: 20 requests within a single ChatGPT session produced 17 distinct session ids, all resolving to `activeBrowserSession=null`.

```
14:28:32  840ee83b…  platform_set_active_browser_session   ← pin written
14:28:33  b4b06ae9…  activeBrowserSession=null             ← next call, new session
14:28:45  5bba6c80…  platform_set_active_browser_session   ← pin written again
14:28:47  70391f12…  activeBrowserSession=null             ← gone again
```

The platform compounds the problem. When the header is absent we generate a replacement (`request.headers['mcp-session-id'] || randomUUID()`) rather than rejecting the request. Pins are keyed by that id (`mcp-to-browser::<userId>:<mcpSessionId>`), so each pin is written under an id that is never queried again. The user then sees "no tab is pinned", which points at the pinning logic rather than at session handling, and made this slower to diagnose than it should have been.

Two things that might be worth considering, without a firm view either way: returning a 400 for non-initialize requests that arrive without a session id, which the specification permits and which would surface a non-compliant client immediately; and reviewing whether other parts of the MCP integration rely on spec-compliant client behaviour.

Ruled out during investigation: multi-pod state. Production runs multiple forge replicas and requests do move between them mid-session, but the object cache is redis, so pins are shared correctly.

andypalmi:
Forum post about OpenAI violating MCP https://community.openai.com/t/chatgpt-does-not-echo-back-mcp-session-id-anymore-violating-mcp-transport-rules/1380087/2

cstns:
Interim fix available. OpenAI's clients send a per-conversation id in the request body we can fall back on.

[...]

PR #8561, opened 2026-09-18
Fall back to the openai/session meta when no Mcp-Session-Id is supplied

### LINKED ISSUE #8560: Fall back to _meta["openai/session"] when no Mcp-Session-Id is supplied
Interim fix for #8558.

OpenAI's clients don't return the `Mcp-Session-Id` we hand them, so every request mints a new session and browser tab pins are unreachable by the next call. They do send `_meta["openai/session"]`, an anonymized per-conversation id, which is the scope a pinned tab wants.

Fall back to it in `forge/ee/routes/mcp/server.js` when the header is absent, keeping `randomUUID()` as the final fallback since client `_meta` is a hint and may be absent.

Deliberately narrow: this unblocks ChatGPT web and desktop now. It does not replace the explicit-handle migration tracked in #8558. Codex coverage is unconfirmed, since it uses the Rust `rmcp` client and may send no `openai/*` metadata. To be verified internally by logging `params._meta` and connecting each surface.

cstns:
Closes #8560

OpenAI's MCP clients never return the `Mcp-Session-Id` the server hands them, so every request minted a fresh session id. Anything keyed on it was unreachable by the next call, which is why `platform_set_active_browser_session` reported success and the following tool call reported that no tab was pinned. In one captured production window, 20 requests from a single ChatGPT session produced 17 distinct session ids, all resolving to `activeBrowserSession=null`.

They do send `_meta["openai/session"]`, documented as an "anonymized conversation id for correlating tool calls within the same ChatGPT session". That is per conversation and stable across the calls inside it, which is the scope a pinned tab wants anyway, arguably a better fit than the transport session ever was.

So when the header is absent we now fall back to it, keeping `randomUUID()` as the final fallback because client `_meta` is a hint and may be missing. An explicit header still wins, so compliant clients are unaffected.

Trusting an unauthenticated value here is safe: the pin key is `mcp-to-browser::${userId}:${mcpSessionId}` and `userId` comes from the PAT, so a forged value can only ever collide with that same caller's own state. Worth keeping in mind before reusing this field anywhere that is not already user-scoped.

Tests cover the fallback, stability across two calls in one conversation, the header taking precedence, and two header-less requests still getting distinct ids.

Two things this is not. It is a vendor-specific stopgap rather than the real fix: MCP revision 2026-07-28 removes protocol-level sessions entirely (SEP-2567), and the migration to explicit state handles is tracked in #8558. And Codex coverage is unconfirmed, since it drives MCP through the Rust `rmcp` client and may send no `openai/*` metadata, so this may only cover ChatGPT web and desktop. We will confirm internally by logging `params._meta` and connecting each surface.


forge/ee/routes/mcp/server.js
```diff
@@ -85,7 +85,16 @@ module.exports = async function (app) {
...
+        // OpenAI's clients do not return the Mcp-Session-Id we hand them, so each request
+        // would otherwise present as a new session and anything keyed on it - a pinned
+        // browser tab - would be unreachable by the very next call. They do send a
+        // per-conversation id in _meta, which is the scope a pinned tab wants anyway.
+        // Trusting an unauthenticated value here is safe because every key built from it is
+        // already scoped by the PAT's user, so a forged one can only collide with that same
+        // caller's own state. Interim measure; see #8558 for the explicit-handle migration.
+        const mcpSessionId = request.headers['mcp-session-id'] ||
+            mcpBody.params?._meta?.['openai/session'] ||
+            randomUUID()
```
andypalmi: Can we trim down the comment?

andypalmi (approved):
Trim down the comment and we can get this in.

Also please let's create a follow up issue for us to move our MCP to stateless

Issue #8566, opened 2026-09-18
Expert driven recommendations for hosted and remote instances


cstns:
## Why

Most customers deploy their flows and then leave them alone. They don't know their palette has packages with known advisories, that a flow could be restructured to stop falling over, or that half of what they've built could be exposed through MCP so the Expert's Insights mode can answer questions they're currently answering by hand. Nobody goes looking for this on their own. We should go to them.

We already have most of the pieces. The Expert understands flows, palette and context. We know what's installed on every hosted and remote instance. We have in-app notifications and email. What's missing is something that looks at all of it by itself, tells the customer what it found, and then offers to do it for them.

## What we want

Look at each hosted instance and remote instance on a schedule, and on demand, and produce recommendations. Things worth fixing, upgrading, restructuring, or exposing through MCP. Collect them per instance and per device so there's one place that answers "what's worth doing here", tell people when something new shows up, and let them act on it with a click.

There are really two kinds of findings here, and it might be worth keeping the producers separate:

1. **Plain checks in code.** Outdated or vulnerable packages, deprecated nodes, missing error handling, settings that are obviously costing something. Cheap, repeatable, no surprises.
2. **Expert analysis.** Design patterns, restructuring suggestions, and the MCP angle: spotting flows that hold data worth querying and suggesting they'd be worth wrapping as an MCP server so Insights mode can use them.

Might be worth starting with the first kind while the second is being figured out, since it ships without needing anything new from the Expert service.

## Categories

Each recommendation gets a category, so people can filter and so we can decide what's worth emailing about:

* security
* improvement
* optimization
* qol with mcp

Probably fine to keep this a short fixed list now and grow it once we see what actually comes out.

## Where it shows up

* A list on each instance and each device, showing what was found, why it matters and what to do about it.
* A team level roll-up so someone with twenty devices doesn't have to click through all of them.
* Per recommendation: dismiss, mark as done, or open it in the Expert so the conversation starts with the finding already loaded.
* **Fix this for me.** One click hands the finding to the Expert, which already has the flow building and platform tools to carry it out. The finding needs to carry enough context that the Expert can pick it up cold and knows what "done" looks like.

## Notifications

* Security findings go out as their own email when they land.
* Everything else rolls into a digest, so nobody gets one email per finding.
* In-app notification for all of them.
* Team owners can turn the volume down or off.

## Things we still need to work out

[...]

PR #8595, opened 2026-09-21
Offer prompt suggestions at the start of an Expert conversation

### LINKED ISSUE #8594: Show a few prompt suggestions in the Expert to get people started
## Why

Most people open the Expert and stare at an empty box. They don't know what it can do, and the things it's best at, upgrading Node-RED, checking what's exposed over MCP, telling them which devices have drifted, are exactly the things nobody thinks to ask for.

#8566 wants to go find those things on a schedule and push them at people. This is the same idea pulled forward to the cheapest possible version: put a few of them in front of the user and see which ones get clicked.

## What

Three suggestions, drawn at random from a list in the code, shown under the Expert's opening message.

Clicking one sends it straight away. If the prompt is a half finished sentence, like "Look at what's in my FlowFuse Tables and answer this: ", it goes into the composer with the caret at the end instead, and waits for the user to finish it.

They disappear as soon as the user starts typing, and once one has been used. Start over deals a fresh three. They don't show during onboarding, that conversation already runs itself.

## What it gives us for the rest of #8566

Which suggestions get clicked is the cheapest signal we'll get on where the scheduled analysis should spend its time later. Might be worth firing a PostHog event on click so we actually capture that.

cstns:
Closes #8594

Three prompt suggestions, drawn at random, shown under the Expert's opening message so people who open the drawer and don't know what to ask have somewhere to start.

Clicking one sends it straight away. Two of the prompts are deliberately unfinished ("Look at what's in my FlowFuse Tables and answer this: "), and those land in the composer with the caret at the end instead, waiting for the user to complete the sentence.

They disappear as soon as the user types anything, and once one has been used. Start over deals a fresh three. They don't render in the onboarding conversation, which opens itself.

### Where the prompts live

`frontend/src/components/expert/prompt-suggestions.js` is a plain array of `{ title, prompt, needsInput }`. Adding or removing one is a one line diff, nothing else needs touching. There's no API, no storage, no filtering by what a team actually has, and no memory of what someone has already seen. Everyone gets a random three. Worth being cleverer about that only if people actually click them.

The 37 prompts in there came out of the discussion on #8566 and are grounded in tools that exist: bill of materials, instance resources and logs, instance history, pipelines, device groups and target snapshots, Tables, the team broker, the team node repository, shared library, members and roles, and the MCP registrations behind Insights mode.

### How it's wired

The state and the rules live in `ExpertChatInput.vue`, which already owns the composer text, the transcript and the send path, so the suggestions reuse `handleSend` and keep auto pin on first message. The cards themselves are teleported into a slot under the message list in `Expert.vue`.
Unit tests cover the random draw, the shape of the shipped list, and the component's rendering and click event.

image


andypalmi:
Should we have the samples actually served by the expert? It would allow for easier updates on them

cstns:
This was the simplest way to get something useful out there that showcases what the expert assistant could do

andypalmi:
Nice work, happy with this as the v1. The random draw, hide-on-type, hide-after-use and needsInput caret handling all read cleanly.

A few things before it goes in:

1. Suggestions should only show for the support agent. Right now the composer is shared with Insights and only onboarding is excluded, so they also appear once a resource is selected in Insights.

2. Keying on title is fine to start with, we can add ids later if we need them.

3. pickSuggestions and the card component are tested, but the showSuggestions rules are not. A small test there would lock the behaviour in.

Keeping the hardcoded list for v1 is right. We can serve these from the Expert later, worth a follow-up note.

frontend/src/components/expert/components/ExpertChatInput.vue
```diff
@@ -224,6 +237,17 @@ export default {
...
+            // hasMessages is true from the off, the store seeds a welcome message,
+            // so the transcript counts as started only once the user has said something
+            return this.messages.some(message => message._type === 'human')
+        },
+        showSuggestions () {
+            if (this.expertSurface === 'onboarding') return false
+            if (this.suggestionUsed || this.suggestions.length === 0) return false
+            if (this.inputText.length > 0 || this.hasUserTurns) return false
+            return !this.isInputDisabled
+        },
```
andypalmi, line 251: Gate on the support agent so these don't show up in Insights once a resource is selected:

```suggestion
        showSuggestions () {
            if (this.expertSurface === 'onboarding') return false
            if (this.isInsightsAgent) return false
            if (this.suggestionUsed || this.suggestions.length === 0) return false
            if (this.inputText.length > 0 || this.hasUserTurns) return false
            return !this.isInputDisabled
        },
```

test/unit/frontend/components/expert/PromptSuggestions.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    })
+
+    test('emits the clicked suggestion', async () => {
+        const wrapper = mount(PromptSuggestions, { props: { suggestions } })
+
+        await wrapper.findAll('[data-action="use-prompt-suggestion"]')[1].trigger('click')
+
+        expect(wrapper.emitted('select')[0][0]).toEqual(suggestions[1])
+    })
+})
```
andypalmi, line 55: Could we add a test here for the showSuggestions rules too? The onboarding, insights, hasUserTurns and suggestionUsed gates aren't covered yet, and a small test would lock the behaviour in.

andypalmi (approved):

Issue #8562, opened 2026-09-18
Introduce 'lite' remote instance type


knolleary:
In support of the new Node-RED plugin version of the device agent, we need to be able to distinguish between full Device Agents, and the plugin version. This is because there will be a functional gap between the two, so the platform needs to only offer the appropriate options for the lite instance.

## Steps

Going to split this into two pieces of work:

### Add `agentType` and populate

 - Add `agentType` column to `Devices` table. String/Allow null
    - `null` will represent the full Device Agent
    - `"lite"` for the plugin
- Update the Async Device Registration flow to allow the plugin to self-identify in the registration flow to ensure the value is set.

### Update UX to without capabilities from `lite` remote instances

 This is a big piece of work that can't be iterated. Initially, any functional action in the UI related to devices will need to be blocked for `lite` devices. We can then iteratively re-enable them as they become available in the plugin.

PR #8627, opened 2026-09-22
Add agentType to devices and enable lite type registration

knolleary:
Part of https://github.com/FlowFuse/nr-remote-agent/issues/4

This introduces an `agentType` column to the `Devices` model. It can be:
 - `null`/`"full"` - the default type; a full Device Agent driven remote instance
 - `lite` - the new type driven by the NR Plugin

This will be used to block certain features in the API/UX where they are not possible with the plugin agent.

This PR also updates the async device registration flow so it can be reused by the Plugin. It uses a `type=lite` query parameter to self-identify the type of device being registered. In the case of a plugin, the text shown is different to reflect the different user experience.

This PR does *not* introduce any of the aforementioned feature gates. That will come in a separate PR as it will be chonky to get right.


forge/db/models/Device.js
```diff
@@ -35,6 +35,16 @@ module.exports = {
...
         lastSeenAt: { type: DataTypes.DATE, allowNull: true },
         settingsHash: { type: DataTypes.STRING, allowNull: true },
+        agentType: {
+            // `full` = full agent
+            // `lite` = lite agent
+            type: DataTypes.STRING,
+            allowNull: true,
+            get () {
+                // default null to `full` for backwards compatibility with existing devices
+                return this.getDataValue('agentType') || 'full'
```
cstns, line 45: #8562 has `null` meaning the full agent and `lite` for the plugin, but this getter reports `full` instead. Combined with the `enum: ['lite']` on the create schema, the API ends up emitting a value it won't accept back, so anything that round-trips a device (re-registration, an SDK, the MCP device tooling) hits a 400.

forge/routes/api/device.js
```diff
@@ -248,7 +248,8 @@ module.exports = async function (app) {
                     application: { type: 'string' },
                     setup: { type: 'boolean', enum: [true] }, // enum only permits a value of true for setup
                     agentHost: { type: 'string' }, // future, for audit log
-                    registrationSession: { type: 'string' } // optional, for async device registration flow
+                    registrationSession: { type: 'string' }, // optional, for async device registration flow,
+                    agentType: { type: 'string', enum: ['lite'] }
```
cstns, line 252: Related to the getter: this rejects `full`, which is now what the API returns for every existing device.

forge/routes/api/device.js
```diff
@@ -444,6 +446,13 @@ module.exports = async function (app) {
         },
         schema: {
             summary: 'Start an asynchronous registration for a device',
+            body: {
+                // The body is optional - if provided, it may contain the agent type
+                type: ['object', 'null'],
+                properties: {
+                    type: { type: 'string' }
```
cstns, line 453: No enum here, but line 473 does an exact `=== 'lite'`, so `Lite` or a typo quietly returns a full-agent URL. Might be worth `enum: ['lite']` so it surfaces as a 400 instead.

frontend/src/components/multi-step-forms/device/steps/SuccessStep.vue
```diff
@@ -69,6 +86,9 @@ export default {
         },
         deviceNRRunning () {
             return this.deviceConnected && this.polledDevice?.status === 'running'
+        },
+        agentType () {
+            return this.polledDevice?.agentType || (this.$route.query.type === 'lite' ? 'lite' : null)
```
cstns: This falls back to the URL query while `pollStatus` below branches on the device from the API. The create response already carries `agentType`, so `this.device?.agentType` would work as the fallback and keep both in sync. As it stands, appending `?type=lite` to a normal registration URL hides the OTC block entirely and the Device Agent never gets its code.

frontend/src/components/multi-step-forms/device/steps/SuccessStep.vue
```diff
@@ -2,33 +2,50 @@
...
+                <p>
+                    Starting Node-RED...
+                </p>
+                <p class="flex flex-col gap-2">
+                    <ff-loading scale="small" message=" " />
+                </p>
+            </template>
+            <template v-else>
+                <p>
+                    Node-RED Connected
```
cstns, line 46: Lite branch drops the spinner the full-agent branch has, and the redirect only fires on `status === 'running'`. Until the heartbeat work lands this is where users sit, so it might read as frozen.

frontend/src/components/multi-step-forms/device/steps/SuccessStep.vue
```diff
@@ -91,12 +111,23 @@ export default {
...
-                    await deviceApi.setMode(this.device.id, 'developer')
-                    await deviceApi.enableEditorTunnel(this.device.id)
-                    this.$router.push({
-                        name: 'device-editor',
-                        params: { id: this.device.id }
-                    })
+                    if (device.agentType === 'lite') {
+                        // If this is a Lite Agent:
+                        // - put into developer mode as we don't support fleet mode yet
+                        // - redirect to the device editor page
```
cstns, line 117: Comment says device editor, the push below goes to `device-overview`. Both branches also start with the same `setMode(..., 'developer')`.

forge/db/views/Device.js
cstns, line 1: `agentType` is on the full view but not `DeviceSummary`, which is `additionalProperties: false`, so it's absent from search results, audit log associations, pipeline stage devices and snapshot views. Not a problem for the gating PR since that reads the full view everywhere, just flagging in case those want it later.

frontend/src/components/multi-step-forms/device/steps/SuccessStep.vue
knolleary: This is a temporary arrangement. Unlike the Device Agent, where the user is bouncing between terminal and browser, with the plugin, they are already in a browser environment. The goal will be to self-close this window once registration completes. But that is a later iteration.

frontend/src/components/multi-step-forms/device/steps/SuccessStep.vue
knolleary: Agreed - `this.device.agentType` is already known, and unlike the polled values, it won't change. Will apply.

frontend/src/components/multi-step-forms/device/steps/SuccessStep.vue
knolleary: Fixed up comments.

forge/routes/api/device.js
knolleary: Added enum

forge/db/models/Device.js
knolleary: Updated the enum

forge/db/views/Device.js
knolleary: Added to DeviceSummary

knolleary:
@cstns thanks for the comments. Have applied updates for _most_ of them.

 - ensured the api schema enums include `full`
 - fixed the SuccessStep page to look at `this.device` not `this.pollingDevice` as the value we care about won't be changing between polls

hardillb (approved):

forge/routes/api/device.js
```diff
@@ -437,13 +439,21 @@ module.exports = async function (app) {
     app.post('/_/register', {
         config: {
             allowAnonymous: true,
-            // Rate limit to 5 requests per 30 seconds per IP address. Given
+            // Rate limit to 5 requests per 0 seconds per IP address. Given
```
andypalmi:
```suggestion
            // Rate limit to 5 requests per 30 seconds per IP address. Given
```

I think this was a typo

forge/routes/api/device.js
```diff
@@ -437,13 +439,21 @@ module.exports = async function (app) {
...
             rateLimit: app.config.rate_limits ? { max: 5, timeWindow: 30000 } : false
         },
         schema: {
             summary: 'Start an asynchronous registration for a device',
+            body: {
+                // The body is optional - if provided, it may contain the agent type
+                type: ['object', 'null'],
+                properties: {
+                    // Only include lite in the enum - if not set, the default is full
+                    type: { type: 'string', enum: ['lite'] }
```
andypalmi:
```suggestion
                    // Accept both types; anything other than lite gets the default (full) registration URL
                    type: { type: 'string', enum: ['lite', 'full'] }
```

Would it make sense to also accept `full` here, to match the `agentType` enum on `POST /devices`? Today the device agent sends `{}` so nothing breaks, but a client sending `type: 'full'` explicitly would get a 400 even though it's the default.

Issue #7709, opened 2026-07-04
5.12-b Write tools, non-destructive (phase 2)


andypalmi:
**Parent:** #7681 (5.12 Team data and packages)
**Tool file:** new `forge/ee/lib/mcp/tools/teamData.js`

`readOnlyHint: false`, `destructiveHint: false`. One tool per endpoint; these siblings do not share a single action concept, so they are not consolidated.

| Tool | Endpoint | Scope | Annotation |
|---|---|---|---|
| `platform_create_database_table` | `POST /teams/:teamId/databases/:databaseId/tables` | `team:database:create` | write |
| `platform_create_library_entry` | `POST /storage/library/:libraryId/*` | `library:entry:create` | write |
| `platform_update_git_token` | `PUT /teams/:teamId/git/tokens/:tokenId` | `team:git:tokens:edit` | write |

**Design notes:**
- Plan-gated (enabled per team type/plan, not per user role): `tables`, `npm`, `gitIntegration`, `shared-library`. If the team's plan does not include the feature the route returns a 404 (`Not Found - not available on team`), independent of the PAT's permissions. The tool should surface this as a clear "feature not enabled for this team" message rather than passing through a bare "not found", which an agent would misread as a missing resource.
- `update_git_token` edits an existing token record; it does not mint or return a new credential (the minting POST is excluded, see 5.12-d).

**Scopes to allow-list + access-tag (#7445):**
`team:database:create` (write), `library:entry:create` (write), `team:git:tokens:edit` (write).

**Tool definitions (description + zod inputSchema):**
```js
platform_create_database_table: {
  description: 'Creates a new table in a database. Both name and columns are required to create the table.',
  inputSchema: z.object({
    teamId: z.string().describe('team hashid'),
    databaseId: z.string().describe('database hashid'),
    name: z.string().describe('name for the new table'),
    columns: z.array(z.object({
      name: z.string().describe('column name'),
      type: z.string().describe('column data type (free-form string; the API declares no enum)'),
      nullable: z.boolean().optional().describe('whether the column allows NULL'),
      default: z.string().nullable().optional().describe('default value, or null for none'),
      generated: z.boolean().optional().describe('whether the column value is generated'),
      maxLength: z.number().nullable().optional().describe('maximum length, or null for unbounded')
    })).describe('column definitions for the new table')
  })
},
platform_create_library_entry: {
[...]

PR #8633, opened 2026-09-22
Add MCP tool to create a database table

andypalmi:
Adds the `platform_create_database_table` action tool from #7709, on its own. The two other tools in that issue (`platform_create_library_entry`, `platform_update_git_token`) are not included here.

**What it does**
1. New write tool `platform_create_database_table` backing `POST /api/v1/teams/:teamId/databases/:databaseId/tables` (`team:database:create`), added to the existing `tables.js` alongside the read tools rather than a separate file.
2. `annotations`: `readOnlyHint: false`, `destructiveHint: false`, `idempotentHint: false`, `openWorldHint: false`. Not idempotent because the route returns 409 if a table of that name already exists.

**Input schema**
1. `name` and at least one `column` are both required (`.min(1)`). The route handler only replies when both `name` and `columns` are present, so requiring them keeps a bare call from producing no response.
2. `type` documents the supported set (`bigint, bigserial, boolean, date, timestamptz, real, double precision, text`); the database driver rejects anything else.

**Behaviour notes**
1. Error responses pass through unchanged. The route's own feature-gate 404 already carries a descriptive `error` string ("FlowFuse Tables is not enabled for this team"), and a name clash returns 409 `table_exists`.

**Tests**
1. Posts name and columns to the correct endpoint and returns the response unmodified.
2. Passes through a 409 error response unchanged.


andypalmi:
Checked the tool against `POST /api/v1/teams/:teamId/databases/:databaseId/tables` and both table drivers.

**The tool is correct.** It targets the right route with `team:database:create`, sends `name` and `columns` as the route expects, and passes error responses through unchanged. Two choices are deliberate hardening against route quirks: it requires `name` and at least one `column` (`.min(1)`), which avoids the endpoint's no-reply hang on a partial body, and it documents the supported column types, since anything else is rejected by the driver.

While verifying the route I found some pre-existing backend bugs. They live in the route and drivers, not in this tool, so they do not block the tool, but they affect what a caller can actually do. Filed for follow-up:

1. #8640 create-table hangs on a partial body and accepts an empty `columns` array.
2. #8634 an unsupported column type returns a 500 instead of a 400.
3. #8635 column defaults are wrong for real/double and boolean, and dropped for date/bigserial.
4. #8636 `generated` and `maxLength` are accepted but silently ignored.
5. #8637 the duplicate-table check is name-only and matches on exactly one row.
6. #8638 a successful create returns an empty body and the declared response schema is unused.
7. #8639 deleting a database logs a `created` audit event instead of `deleted`.

andypalmi:
On schema-qualified table creation (e.g. `myschema.mytable`): the tool does not support it because the platform API does not support it yet. The create route accepts only `{ name, columns }` and the driver runs an unqualified `CREATE TABLE`, so a new table always lands in the database's default schema. The UI reflects the same limit: its create form states the table is created in the default schema, and its name validation rejects dotted names. Read, query and delete already take a schema; only create does not.

If we want schema-qualified creation, there are two paths:

1. Extend the API (route and driver) to accept a schema and create the table qualified, then add the field to this tool.
2. Leave creation to the `tables-query` node, which runs arbitrary SQL including `CREATE TABLE myschema.mytable (...)` and other schema changes. This is the documented path for any write or schema change, per the `flowfuse-tables-platform-tools` and `flowfuse-tables-query` skills: the platform tools are read-only inspection, the node is the way to write or alter schema.

@Steve-Mcl

